PR #111, opened 2015-09-03
Remove unnecessary else statements

dullgiulio:
Whenever it makes the code easier to follow, use early return to
avoid else statements.

It's a dirty job, but someone has to do it.

### DIFF
--- a/db/password.go
+++ b/db/password.go
@@ -49,12 +49,11 @@ func (r *passwordInfoRepo) Create(tx repo.Transaction, pw user.PasswordInfo) (er
 	}
 
 	_, err = r.get(tx, pw.UserID)
-	if err != nil {
-		if err != user.ErrorNotFound {
-			return err
-		}
-	} else {
+	if err == nil {
 		return user.ErrorDuplicateID
+	}
+	if err != user.ErrorNotFound {
+		return err
 	}
 
 	err = r.insert(tx, pw)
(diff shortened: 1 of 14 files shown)


db/password.go
```diff
@@ -49,13 +49,12 @@ func (r *passwordInfoRepo) Create(tx repo.Transaction, pw user.PasswordInfo) (er
...
 	_, err = r.get(tx, pw.UserID)
-	if err != nil {
-		if err != user.ErrorNotFound {
-			return err
-		}
-	} else {
+	if err == nil {
 		return user.ErrorDuplicateID
 	}
+	if err != nil && err != user.ErrorNotFound {
```
arnehormann: here and some other places below: `err != nil` can be deleted, it's always true because of the `if err == nil` above

dullgiulio: True, correct this and the others.

bobbyrullo:
Thanks @dullgiulio ! Did you use some automated tool for this, or did you just dig in and read a ton of code?

Minor nit: our commit messages start with subsystem; in your case it touches almost everywhere so it should be like:

>*: Remove unnecessary else branches

bobbyrullo:
otherwise, LGT

dullgiulio:
I used grep and it wasn't that much code to check ;)

bobbyrullo:
Congrats on having the first merged non-CoreOS contribution!

PR #128, opened 2015-09-19
cmd/dex-worker: wait 'til connectors are available

bobbyrullo:
Otherwise, if worker starts without connectors, and then connectors are
added workers have to be restarted to pick up the changes.


aaronlevy:
LGTM

PR #129, opened 2015-09-21
Add smtp support

ecnahc515:
This PR adds support for Dex to send emails via SMTP as an alternative to mailgun. Tested with the mailgun SMTP gateway, and verified the emails were going to mailgun using TLS or STARTTLS (depending on the port used) using wireshark.


bobbyrullo:
almost there: a doc nit, and a question about conifg validation. Thanks Chance!

ecnahc515:
@bobbyrullo Updated in response to your comments, PTAL!

bobbyrullo:
Thank Chancez!

PR #132, opened 2015-09-24
server: use routable url for user creation

joeatwork:
Our httprouter cleans up routes that appear to be missing a trailing slash by redirecting clients. That means that the url `*/users` cannot be hit, since there is a live `*/users/` url - the router will always redirect the url without the slash. This pull request  attempts to avoid the issue by moving the endpoint.


joeatwork:
Tests don't pass. Closing until this works.

PR #133, opened 2015-09-24
server: user management endpoints strictly conform to schema

joeatwork:
This change disables the URL fixing behavior or the router associated
with the user management schema. After this commit, URLS routing
to /api/$VERSION/users must target exactly the specified paths. In
addition, `/api/$VERSION/users/` will serve a 404

This change allows users to hit the user create endpoint, which
would previously serve a redirect rather than actually making the
associated change.


bobbyrullo:
LGTM

PR #137, opened 2015-09-25
api: don't create a user if you can't send them an email

joeatwork:
Before, when a call was made with a bad URL, the call returned an error but the user was created anyway.


bobbyrullo:
nice! thanks joe. LGTM

PR #138, opened 2015-09-25
server: disable users

joeatwork:
This pull request allows for setting users as "disabled". A disabled user can neither log in, nor can they access dex's administrative functions. This pull request does *not* contain an API for setting or clearing the disabled flag from a user.


bobbyrullo:
looks great, just a couple of things

bobbyrullo:
LGTM

PR #139, opened 2015-09-28
build: vendor quotedprintable to build on Go < 1.5

joeatwork:
quotedprintable is part of the standard library in Go 1.5 but is needed in Godeps in earlier versions.


bobbyrullo:
LGTM

PR #140, opened 2015-09-28
Expose API to enable and disable users

joeatwork:
Diff will clean up once https://github.com/coreos/dex/pull/139 or something like it is merged.


joeatwork:
@bobbyrullo this is ready for another look when you have the time

bobbyrullo:
Almost LGTM - sorry for not mentioning this before, but if you add fields and/or methods to repos you should add a test in the functional/repo package, and run it with and without DEX_TEST_DSN 

https://github.com/coreos/dex/blob/master/functional/repo/user_repo_test.go

joeatwork:
Gotcha - repo tests are added! This pull request is ready for another look!

Issue #116, opened 2015-09-10
Warnings while building


ajanthan:
While executing ./build i am getting following warnings

➜  dex git:(master) ./build
Could not find go-bindata in path, will not generate migrations
\# github.com/coreos/dex/cmd/dex-worker
link: warning: option -X main.version b340660d6d7ec8064d65a67180b1b4f4b346d581 may not work in future releases; use -X main.version=b340660d6d7ec8064d65a67180b1b4f4b346d581
\# github.com/coreos/dex/cmd/dex-overlord
link: warning: option -X main.version b340660d6d7ec8064d65a67180b1b4f4b346d581 may not work in future releases; use -X main.version=b340660d6d7ec8064d65a67180b1b4f4b346d581

My go version 
➜  dex git:(master) go  version
go version go1.5 linux/amd64

bcwaldon:
Fixed via #141

PR #141, opened 2015-09-29
build: add equals sign to avoid link warning when building project

triddell:
With `go version go1.5.1 darwin/amd64`, I was getting the following warning twice when building:

```
link: warning: option -X main.version 6970b481f4d8c6b65d72e6bd9000340fa934187b may not work in future releases; use -X main.version=6970b481f4d8c6b65d72e6bd9000340fa934187b
```

This modification adds the "=" sign (as directed) and prevents the warning.

### DIFF
--- a/build
+++ b/build
@@ -22,7 +22,7 @@ rm -rf $GOPATH/src/github.com/coreos/dex
 mkdir -p $GOPATH/src/github.com/coreos/
 ln -s ${PWD} $GOPATH/src/github.com/coreos/dex
 
-LD_FLAGS="-X main.version $(git rev-parse HEAD)"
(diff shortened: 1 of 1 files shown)


triddell:
This looks like it it addresses issue #116.

The first issue about not having go-bindata can be solved by installing it:

```
go get -u github.com/jteeuwen/go-bindata/...
```

triddell:
I can only assume that the new syntax outlined by the warning message (with the equals sign) more explicitly assigns the main.version variable whereas the previous version (without the equals sign) also worked but is now deprecated.

bobbyrullo:
Thanks @triddell

PR #143, opened 2015-09-30
db: log schema errors, distinguish them from nil results where needed

joeatwork:
gorp schema errors are a big deal, and probably worth some log information. This PRQ also changes the error protocol of some DB functions to distinguish between nil (which seems like a common thing) and type assertion failures (which are super scary)


bobbyrullo:
Thanks @joeatwork

PR #144, opened 2015-09-30
server,cmd: Add flag for disabling registation

bobbyrullo:
For situations where admins add users.

### DIFF
--- a/cmd/dex-worker/main.go
+++ b/cmd/dex-worker/main.go
@@ -44,7 +44,7 @@ func main() {
 	emailFrom := fs.String("email-from", "no-reply@coreos.com", "emails sent from dex will come from this address")
 	emailConfig := fs.String("email-cfg", "./static/fixtures/emailer.json", "configures emailer.")
 
-	enableRegistration := fs.Bool("enable-registration", true, "Allows users to self-register")
+	enableRegistration := fs.Bool("enable-registration", false, "Allows users to self-register")
 
 	noDB := fs.Bool("no-db", false, "manage entities in-process w/o any encryption, used only for single-node testing")
 
--- a/cmd/dex-worker/main.go
+++ b/cmd/dex-worker/main.go
@@ -44,6 +44,8 @@ func main() {
 	emailFrom := fs.String("email-from", "no-reply@coreos.com", "emails sent from dex will come from this address")
 	emailConfig := fs.String("email-cfg", "./static/fixtures/emailer.json", "configures emailer.")
 
+	enableRegistration := fs.Bool("enable-registration", true, "Allows users to self-register")
+
 	noDB := fs.Bool("no-db", false, "manage entities in-process w/o any encryption, used only for single-node testing")
 
 	// UI-related:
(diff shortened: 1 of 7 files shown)


cmd/dex-worker/main.go
```diff
@@ -44,6 +44,8 @@ func main() {
 	emailFrom := fs.String("email-from", "no-reply@coreos.com", "emails sent from dex will come from this address")
 	emailConfig := fs.String("email-cfg", "./static/fixtures/emailer.json", "configures emailer.")
 
+	enableRegistration := fs.Bool("enable-registration", true, "Allows users to self-register")
```
joeatwork: Should this default to true? While we might break existing installs if it defaults to false, it seems safer to have open registration be the explicit choice.

server/config.go
```diff
@@ -183,7 +186,8 @@ func (cfg *MultiServerConfig) Configure(srv *Server) error {
 	return nil
 }
 
-func getTemplates(issuerName, issuerLogoURL string, dir string) (*template.Template, error) {
+func getTemplates(issuerName,
```
joeatwork: Would you consider moving the line break to
```
func getTemplates(issuerName, issuerLogoURL string,
	enableRegister bool....)
```
```

File: server/http_test.go [hunk 3]
```diff
@@ -51,7 +51,7 @@ func (c *fakeConnector) TrustedEmailProvider() bool {
```
joeatwork: It might be worth a test case demonstrating a rejection when an inbound request attempts to submit with `register=1` and handleAuthFunc has been called with `registrationEnabled=false`

server/server.go
```diff
@@ -198,11 +199,15 @@ func (s *Server) HTTPHandler() http.Handler {
...
 	mux := http.NewServeMux()
 	mux.HandleFunc(httpPathDiscovery, handleDiscoveryFunc(s.ProviderConfig()))
-	mux.HandleFunc(httpPathAuth, handleAuthFunc(s, s.Connectors, s.LoginTemplate))
+	mux.HandleFunc(httpPathAuth, handleAuthFunc(s, s.Connectors, s.LoginTemplate, s.EnableRegistration))
 	mux.HandleFunc(httpPathToken, handleTokenFunc(s))
 	mux.HandleFunc(httpPathKeys, handleKeysFunc(s.KeyManager, clock))
 	mux.Handle(httpPathHealth, makeHealthHandler(checks))
-	mux.HandleFunc(httpPathRegister, handleRegisterFunc(s))
+
+	if s.EnableRegistration {
```
joeatwork: SWEET! This is a nice way to ensure that registration can't happen - it isn't even ever wired up!

joeatwork:
Looks great!

cmd/dex-worker/main.go
bobbyrullo: I had the same thought process. yeah, you're probably right.

server/config.go
bobbyrullo: yeah, nicer.

server/http_test.go
bobbyrullo: They don't really get rejected in this case, it just changes the UI. Given that the real protection is below, I don't think it's crucial

bobbyrullo:
Thanks joe, PTAL when you can!

joeatwork:
LGTM

PR #145, opened 2015-10-01
Protect admin api

bobbyrullo:
Requests made to admin API must have an `Authorization` header with 128 bytes encoded base64

### DIFF
--- a/Documentation/getting-started.md
+++ b/Documentation/getting-started.md
@@ -45,17 +45,24 @@ dex needs a 32 byte base64-encoded key which will be used to encrypt the private
 
 The dex overlord and workers allow multiple key secrets (separated by commas) to be passed but only the first will be used to encrypt data; the rest are there for decryption only; this scheme allows for the rotation of keys without downtime (assuming a rolling restart of workers).
 
+# Generate an Admin API Secret
+
+The dex overlord has a an API which is very powerful - you can create Admin users with it, so it needs to be protected somehow. This is accomplished by requiring that a secret is passed via the Authorization header of each request. This secret is 128 bytes base64 encoded, and should be sufficiently random so as to make guessing impractical:
+
+`DEX_OVERLORD_ADMIN_API_SECRET=$(dd if=/dev/random bs=1 count=128 2>/dev/null | base64)`
+
 # Start the overlord
 
 The overlord is responsible for creating and rotating keys and some other adminsitrative tasks. In addition, the overlord is responsible for creating the necessary database tables (and when you update, performing schema migrations), so it must be started before we do anything else. Debug logging is turned on so we can see more of what's going on. Start it up.
 
-`./bin/dex-overlord --db-url=$DEX_DB_URL --key-secrets=$DEX_KEY_SECRET --log-debug=true &`
+`./bin/dex-overlord --admin-api-secret=$DEX_OVERLORD_ADMIN_API_SECRET --db-url=$DEX_DB_URL --key-secrets=$DEX_KEY_SECRET --log-debug=true &`
 
 ## Environment Variables.
 
 Note that parameters can be passed as flags or environment variables to dex components; an equivalent start with environment variables would be:
 
 ```
+export DEX_OVERLORD_ADMIN_API_SECRET=$DEX_OVERLORD_ADMIN_API_SECRET
 export DEX_OVERLORD_DB_URL=$DEX_DB_URL
 export DEX_OVERLORD_KEY_SECRETS=$DEX_KEY_SECRET
 export DEX_OVERLORD_LOG_DEBUG=true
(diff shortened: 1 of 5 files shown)


Documentation/getting-started.md
```diff
@@ -43,19 +43,27 @@ dex needs a 32 byte base64-encoded key which will be used to encrypt the private
 
 `DEX_KEY_SECRET=$(dd if=/dev/random bs=1 count=32 2>/dev/null | base64)`
 
+
```
joeatwork: What does this newline do?

Documentation/getting-started.md
```diff
@@ -43,19 +43,27 @@ dex needs a 32 byte base64-encoded key which will be used to encrypt the private
...
 The dex overlord and workers allow multiple key secrets (separated by commas) to be passed but only the first will be used to encrypt data; the rest are there for decryption only; this scheme allows for the rotation of keys without downtime (assuming a rolling restart of workers).
 
+# Generate a Admin API Secret
+
+The dex overlord has a an API which is very powerful - you can create Admin users with it, so it needs to be protected somehow. This secret is 128 bytes base64 encoded:
```
joeatwork: This could maybe use a connecting sentence that describes how a secret might protect the API, and how it is used. eg "it's information that is very hard to guess that can be shared with approved clients"?

Documentation/getting-started.md
```diff
@@ -43,19 +43,27 @@ dex needs a 32 byte base64-encoded key which will be used to encrypt the private
...
+
+`DEX_OVERLORD_ADMIN_API_SECRET=$(dd if=/dev/random bs=1 count=128 2>/dev/null | base64)`
+
 # Start the overlord
 
 The overlord is responsible for creating and rotating keys and some other adminsitrative tasks. In addition, the overlord is responsible for creating the necessary database tables (and when you update, performing schema migrations), so it must be started before we do anything else. Debug logging is turned on so we can see more of what's going on. Start it up.
 
-`./bin/dex-overlord --db-url=$DEX_DB_URL --key-secrets=$DEX_KEY_SECRET --log-debug=true &`
+`./bin/dex-overlord --admin-api-secret=$DEX_OVERLORD_ADMIN_API_SECRET --db-url=$DEX_DB_URL --key-secrets=$DEX_KEY_SECRET --log-debug=true &`
```
joeatwork: Do the arguments need to be specified if the env variables are named DEX_OVERLORD_*?

Documentation/getting-started.md
```diff
@@ -43,19 +43,27 @@ dex needs a 32 byte base64-encoded key which will be used to encrypt the private
...
 
 ## Environment Variables.
 
 Note that parameters can be passed as flags or environment variables to dex components; an equivalent start with environment variables would be:
 
 ```
+export DEX_OVERLORD_ADMIN_API_SECRET=$DEX_OVERLORD_ADMIN_API_SECRET
```
joeatwork: This line is redundant, it just sets the env variable to it's current value.

contrib/standup-db.sh
```diff
@@ -31,6 +31,7 @@ DEX_KEY_SECRET=$(dd if=/dev/random bs=1 count=32 2>/dev/null | base64)
 export DEX_OVERLORD_DB_URL=$DEX_DB_URL
 export DEX_OVERLORD_KEY_SECRETS=$DEX_KEY_SECRET
 export DEX_OVERLORD_KEY_PERIOD=1h
+export DEX_OVERLORD_ADMIN_API_SECRET=$(dd if=/dev/random bs=1 count=128 2>/dev/null | base64)
```
joeatwork: You might want to set this up by `DEX_KEY_SECRET`, both for consistency and to allow users to specify their own known secrets if they ever want to use the API themselves.

server/admin.go
```diff
@@ -48,7 +51,25 @@ func (s *AdminServer) HTTPHandler() http.Handler {
 	r.GET(AdminGetStateEndpoint, s.getState)
 	r.Handler("GET", httpPathHealth, s.checker)
 	r.HandlerFunc("GET", httpPathDebugVars, health.ExpvarHandler)
-	return r
+
+	return authorizer(r, s.secret, httpPathHealth, httpPathDebugVars)
```
joeatwork: This is a nice "middleware" approach to auth!

joeatwork:
Looks good!

Documentation/getting-started.md
bobbyrullo: removed.

Documentation/getting-started.md
bobbyrullo: Changed. See how you like it.

Documentation/getting-started.md
bobbyrullo: nope, I'm just showing how it works with flags here. I talk about env var equivalency later.

Documentation/getting-started.md
bobbyrullo: yes but it wasn't exported before.

contrib/standup-db.sh
bobbyrullo: There is a method to my madness. I set the `DEX_KEY_SECRET` over by itself because it is shared.

Then for each binary, I set the environment variables close to where the binary is invoked; I think it's more useful there, because it shows in a glance all the things one needs to start up an overlord or worker.

server/admin.go
bobbyrullo: thanks! I originally had a "authorize()" call in each method, then re-did it this way because I knew I was being lazy :)

joeatwork:
LGTM!

bobbyrullo:
Thanks @joeatwork

PR #147, opened 2015-10-06
server: use standard lib http.Request.BasicAuth

gyuho:
Go 1.4+ has https://golang.org/pkg/net/http/#Request.BasicAuth
method for http.Request and it was requested by CoreOS(kelsey) [1]
with the same functionalities. If dex's Go development is being done
in Go 1.4 or later, we should use the standard library.

Thanks!

---
[1] https://codereview.appspot.com/76540043/

### DIFF
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -1,7 +1,6 @@
 package http
 
 import (
-	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


bobbyrullo:
thanks @gyuho  - the best code is less code!

PR #149, opened 2015-10-06
API for granting and revoking admin privileges

joeatwork:
This change exposes a new public API that provides a way for clients to grant or revoke administrative privileges on users.

This change follows the design and implementation of Disable very closely.

### DIFF
--- a/db/user.go
+++ b/db/user.go
@@ -103,13 +103,22 @@ func (r *userRepo) Create(tx repo.Transaction, usr user.User) (err error) {
 }
 
 func (r *userRepo) Disable(tx repo.Transaction, userID string, disable bool) error {
+	return r.setUserField("disabled", tx, userID, disable)
+}
+
+func (r *userRepo) SetAdmin(tx repo.Transaction, userID string, setAdmin bool) error {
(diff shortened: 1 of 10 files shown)


ecnahc515:
Overall it looks good. I'm not 100% sure on the UX of `SetAdmin`. It might be better to have 2 separate API calls for adding/removing of admin permissions, just because default Go values will revoke access if they submit an empty discovery JSON schema. 

Ideally the caller always has to explicitly choose, and two API calls would enforce it. Rather than having 3 options, add admin, remove admin (via boolean value on struct) or 'empty' which will also remove their admin perms.

ecnahc515:
Also, the API is a bit ambiguous (same with disable).
Is SetAdmin(false) mean “set their admin status to false” or “do not set them as an admin” (ie: leave their admin status alone).

joeatwork:
Thanks for your help, @ecnahc515 ! I'd like to table changing our API ergonomics for now (maybe file another issue?) since I don't want to introduce an API that isn't consistent with disable.

ecnahc515:
LGTM

PR #150, opened 2015-10-06
Add auth0 on Similar Software list

woloski:
great work with dex guys! I remember talking to some of you back a year ago for using Auth0 internally for your products but you had some special requirements and it didn't work out. We are using CoreOS extensibly on our product for code sandboxing (https://medium.com/aws-activate-startup-blog/sandboxing-code-in-the-era-of-containers-294edb3a674) and we love it. 

Since you are listing similar software, I thought adding Auth0 would make sense, here. Writing a post on our blog about chaining dex with Auth0 would be nice as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,12 @@ For running the full stack check out the [getting started guide][getting-started
 
 ## Similar Software
 
+### [Auth0](https://auth0.com)
+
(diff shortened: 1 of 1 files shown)


bobbyrullo:
Awesome, can't wait to see the post @woloski

PR #157, opened 2015-10-14
email,api: simplest invitation link that could possibly work

joeatwork:
Generate a really big url that wraps a password change link as the redirect URL inside of an email validation link (with yet another url wrapped inside of the password change link).

### DIFF
--- a/integration/user_api_test.go
+++ b/integration/user_api_test.go
@@ -580,8 +580,8 @@ type testEmailer struct {
 	lastRedirectURL url.URL
 }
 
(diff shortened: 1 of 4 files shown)


joeatwork:
This method generates a 2190-character URL in my test environment, which is asking a lot of browsers, load-balancers, and servers. There are a number of other possible approaches to this, but none I could think of with such a small diff.

joeatwork:
Withdrawing this approach for now in favor of something a bit more deliberate.

PR #158, opened 2015-10-16
server: Share token code

joeatwork:
This change proposes to unify behavior and source code for token generation between email verification and password reset, in order to make it simpler to add other user-facing, token dependent services to dex.


bobbyrullo:
LGTM

PR #160, opened 2015-10-20
build: add install instructions for go-bindata

philips:
Tell the user how to fix ther error! That is the best type of error
message. :)


bcwaldon:
LGTM

PR #179, opened 2015-11-11
New invitation endpoint

joeatwork:
An alternative to https://github.com/coreos/dex/pull/157 - allow newly invited users to simultaneously verify their email addresses and reset their passwords.

### DIFF
--- a/server/password_test.go
+++ b/server/password_test.go
@@ -357,7 +357,7 @@ func TestSendResetPasswordEmailHandler(t *testing.T) {
 
 func TestResetPasswordHandler(t *testing.T) {
 	makeToken := func(userID, password, clientID string, callback url.URL, expires time.Duration, signer jose.Signer) string {
-		pr := user.NewPasswordReset(user.User{ID: "ID-1"},
+		pr := user.NewPasswordReset("ID-1",
 			user.Password(password),
 			testIssuerURL,
 			clientID,
--- a/user/api/api.go
+++ b/user/api/api.go
@@ -88,7 +88,6 @@ type UsersAPI struct {
 }
 
 type Emailer interface {
-	SendResetPasswordEmail(string, url.URL, string) (*url.URL, error)
 	SendInviteEmail(string, url.URL, string) (*url.URL, error)
 }
 
--- a/user/email/email.go
+++ b/user/email/email.go
@@ -23,6 +23,7 @@ type UserEmailer struct {
 
 	passwordResetURL url.URL
 	verifyEmailURL   url.URL
+	invitationURL    url.URL
 }
 
 // NewUserEmailer creates a new UserEmailer.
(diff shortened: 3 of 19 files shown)


joeatwork:
Discussed a simpler approach with @bobbyrullo 

- Invitation endpoint verifies email address, produces a second, pure PasswordChange token, redirects the user.

- Consider adding a method to the InvitationToken that can produce a PasswordChange token, preserving everything it needs

That means you can get rid of handlePOST and a bunch of the new test cases

server/password_test.go
```diff
@@ -355,9 +361,35 @@ func TestSendResetPasswordEmailHandler(t *testing.T) {
 	}
 }
 
+type resetPasswordTestCase struct {
```
bobbyrullo: Why define this outside of testResetPasswordHandler?

server/password_test.go
```diff
@@ -355,9 +361,35 @@ func TestSendResetPasswordEmailHandler(t *testing.T) {
...
+
+	wantFormValues *url.Values
+	wantCode       int
+	wantPassword   string
+}
+
 func TestResetPasswordHandler(t *testing.T) {
-	makeToken := func(userID, password, clientID string, callback url.URL, expires time.Duration, signer jose.Signer) string {
-		pr := user.NewPasswordReset(user.User{ID: "ID-1"},
+	rpwUserID := "ID-1"
```
bobbyrullo: why all these changes? I don't think it's necessary to prefix all this stuff with "rpw" and "pw" - these variables are local to this file anyhow. In addition, it makes this particular commit larger and harder to read - it's difficult to see where the real changes are.

user/email/email.go
```diff
@@ -46,76 +48,103 @@ func NewUserEmailer(ur user.UserRepo,
...
+	}
+
+	resetURL := u.invitationURL
+	q := resetURL.Query()
+	q.Set("token", token)
+	resetURL.RawQuery = q.Encode()
 
 	if u.emailer != nil {
-		err = u.emailer.SendMail(u.fromAddress, subj, tmplName,
+		err = u.emailer.SendMail(u.fromAddress, "Reset Your Password", "password-reset",
```
bobbyrullo: Shouldn't this be "Activate your account" and use the invitation template?

user/api/api.go
```diff
@@ -88,7 +88,6 @@ type UsersAPI struct {
 }
 
 type Emailer interface {
-	SendResetPasswordEmail(string, url.URL, string) (*url.URL, error)
```
bobbyrullo: !!!! why?

server/password_test.go
joeatwork: Made obsolete in a162744

server/password_test.go
joeatwork: Made obsolete in a162744

user/email/email.go
joeatwork: Fixed in 93df33f

PR #187, opened 2015-11-24
*: use example.com instead of coreos.com for defaults

### LINKED ISSUE #86: dex-worker: uses no-reply@coreos.com as default email address
At line https://github.com/coreos/dex/blob/master/cmd/dex-worker/main.go#L37

We probably want to remove this.

philips:
for testing, flag defaults, etc use the RFC recognized example.com
instead of coreos.com

Fixes #86


bcwaldon:
Takin over at https://github.com/coreos/dex/pull/205

Issue #163, opened 2015-10-22
make sure email addresses are trimmed 


bobbyrullo:

lcacciagioni:
And then used as user id?

PR #188, opened 2015-12-01
registration: trim spaces and sanity check user email from form

ericchiang:
Trim spaces when registering emails to address #163.

Also added a basic sanity check to catch invalid emails. Before you could enter an email like "asfddsaf asdf asdf".

### DIFF
--- a/server/register.go
+++ b/server/register.go
@@ -104,7 +104,7 @@ func handleRegisterFunc(s *Server) http.HandlerFunc {
 		trustedEmail := ses.Identity.Email != "" && idpc.TrustedEmailProvider()
 		validate := r.Form.Get("validate") == "1"
 		formErrors := []formError{}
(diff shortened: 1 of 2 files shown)


server/register.go
```diff
@@ -46,6 +47,17 @@ var (
...
 )
 
+// isValidEmail determines if a given email is of form "name@example.com".
+func isValidEmail(email string) bool {
+	addr, err := mail.ParseAddress(email)
+	if err != nil {
+		return false
+	}
+	// mail.ParseAddress will parse the first email if given a list of emails.
+	// Ensure that the given string contains exactly this single email.
```
iyee: why not using `mail.ParseAddressList(list string)` to make sure this func can handle email(s) correctly?

bobbyrullo:
LGTM

PR #190, opened 2015-12-01
user: fix bug in ValidEmail helper

### LINKED ISSUE #189: user: ValidEmail incorrectly marks bad emails valid
Noticed this when submitting #188 

The `user` package's [`ValidEmail`](https://github.com/ericchiang/dex/blob/a9ab63893db219c73a25eed25d1967251b6ce96c/user/user.go#L160-L170) helper sometimes marks bad emails as valid.

```go
package main

import (
	"fmt"
	"net/mail"
	"strconv"
)

func ValidEmail(email string) bool {
	address, err := mail.ParseAddress(email)
	if err != nil {
		return false
	}

	if address.Name != "" || address.Address == "" {
		return false
	}
	return true
}

func main() {
	emails := []string{
		"test@example.com",      // correctly marked valid
		"invalidemail",          // correctly marked invalid
		"test1@example.com \"&", // incorrectly marked valid
	}
	for _, email := range emails {
		fmt.Println(strconv.Quote(email), ValidEmail(email))
	}
}
```

Because `mail.ParseAddress` stops once it identifies a valid email address, a valid email flowed by junk is incorrectly marked valid.

ericchiang:
mail.ParseAddress will stop parsing a string once it finds a valid
email address. This means you could give ValidUser an email
address followed by junk and it would mark it valid.

This commit fixes this behavior and adds some more test cases.

Fixes #189

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -163,10 +163,9 @@ func ValidEmail(email string) bool {
 		return false
 	}
 
-	if address.Name != "" || address.Address == "" {
-		return false
-	}
-	return true
+	// Has mail.ParseAddress parsed the entire string and only found a single
+	// address without a name?
+	return address.Address == email
 }
 
 func ValidPassword(plaintext string) bool {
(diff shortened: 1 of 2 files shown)


bobbyrullo:
LGTM - thanks @ericchiang

PR #191, opened 2015-12-02
connector: add GitHub connector

ericchiang:
DO NOT MERGE

Initial working GitHub connector for review.

Couple of things

1. GitHub's OAuth is _just_ different enough that it wasn't possible to use `github.com/coreos/go-oidc/oauth2` :cry: 
2. There is an example connector in `static/fixtures/connectors.json.sample`. Probably a good time to add documentation for configuring an application through GitHub's website (e.g. what callback URL to use). Where's a good place for that?
3. Is GitHub a trusted email provider? It sometimes doesn't return an email address if your address is hidden.
4. Didn't quite know what to do for `Healthy` and `Sync`.
5. Only tested with `--no-db`. I want to spin up Postgres with a multi-node dex and run this at least once.

screenshot from 2015-12-01 17-12-48


connector/connector_github.go
```diff
@@ -0,0 +1,248 @@
...
+		clientSecret: cfg.ClientSecret,
+		scopes:       strings.Join(cfg.Scopes, ","),
+	}, nil
+}
+
+func (c *GitHubConnector) ID() string {
+	return c.id
+}
+
+func (c *GitHubConnector) Healthy() error {
```
bcwaldon: Is there a readonly API call we can make to verify the client creds configured within dex are valid?

connector/connector_github.go
```diff
@@ -0,0 +1,248 @@
...
+
+func init() {
+	RegisterConnectorConfigType(GitHubConnectorType, func() ConnectorConfig { return &GitHubConnectorConfig{} })
+}
+
+type GitHubConnectorConfig struct {
+	ID           string   `json:"id"`
+	ClientID     string   `json:"clientID"`
+	ClientSecret string   `json:"clientSecret"`
+	Scopes       []string `json:"scopes"`
```
bcwaldon: Why make this configurable?

connector/connector_github.go
```diff
@@ -0,0 +1,248 @@
...
+		return "", fmt.Errorf("no access token in response body: %s", body)
+	}
+
+	if got := q.Get("scope"); got != c.scopes {
+		return "", fmt.Errorf("requested scope %s got %s", c.scopes, got)
+	}
+	return token, nil
+}
+
+type gitHubUser struct {
```
bcwaldon: Could we remove the unused fields below?

connector/connector_github.go
```diff
@@ -0,0 +1,248 @@
...
+func getGitHubUser(token string) (gitHubUser, error) {
+	u := gitHubAPIUserURL + "?" + (url.Values{"access_token": []string{token}}).Encode()
+	resp, err := http.Get(u)
+	if err != nil {
+		return gitHubUser{}, fmt.Errorf("get: %v", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		// TODO (ericchiang): get actual error from response body
```
bcwaldon: Let's go ahead and do this before this PR lands

connector/connector_github.go
```diff
@@ -0,0 +1,248 @@
...
+			q.Set("error", oauth2.ErrorUnsupportedResponseType)
+			q.Set("error_description", "unable to verify auth code with issuer")
+			redirectError(w, errorURL, q)
+			return
+		}
+		user, err := getGitHubUser(token)
+		if err != nil {
+			log.Errorf("Unable to query GitHub API with token: %v", err)
+			q.Set("error", oauth2.ErrorUnsupportedResponseType)
+			// TODO(eric): better error messages
```
bcwaldon: We need to strike a balance of not leaking information, but logging enough that an operator can tell what's going on. I think this TODO isn't necessary, unless you were thinking of something specific.

bcwaldon:
@ericchiang It would be great if some of the code were abstracted out and unit tested. I can see a clear test being written for LoginURL as-is, but if we could break out the response-parsing code, it would be nice to have expected responses documented in the form of tests for future debugging help.

connector/connector_github.go
ericchiang: In case users of dex administrators wanted more than `user:email` in the future. Though admittedly I don't know what that would look like. I'll remove the option.

connector/connector_github.go
ericchiang: [Yep](http://stackoverflow.com/a/20010178). I'll add that.

connector/connector_github.go
bcwaldon: There's nothing that would actually use the information yet, and it would have to be done within dex itself.

connector/connector_github.go
bcwaldon: That seems specific to a given authentication token, which dex doesn't necessarily have.

ericchiang: Yeah I noticed that when attempting to write the function. Seems like a better idea to track rate limiting, since this could actually prevent the connector from doing its job.

bcwaldon: Just making a call that requires client credentials be used for authentication would be good enough here. Ensuring that we log enough information when getting rate-limited would be a good thing to verify.

ericchiang:
@bcwaldon closing in favor of #197

PR #197, opened 2015-12-03
connector: github and bitbucket oauth2 connectors added

ericchiang:
This pull request replaces the github connector in #191 with a more generic oauth2 connector that leverages dex's existing oauth2 library. Site specific logic is now extremely small. As an example two concrete implementations, github and bitbucket, are added.

~~Note: the github connector requires coreos/go-oidc#24 to work.~~

This commit has a huge Godep diff to update github.com/coreos/go-oidc/oauth2  to include coreos/go-oidc#24, which is required for the github connector.

@bcwaldon to address some of the points in #191
* Most of the code is now abstracted out to coreos/go-oidc (which is much better tested).
* Scope is no longer configurable.
* Unused fields in JSON structs are removed.
* Some unit tests are added.

Working on the health check. I think the only way to verify credentials is to generate a login URL and follow it.

TODO:
* Health check (see above)
* Expand unit tests to failure cases
* Documentation for configuring clients
* Styles for front end
* Write better commit message


ericchiang:
@bcwaldon I was using the bitbucket connector for testing since the github one wouldn't work without the patch mentioned above.

Now that that patch is included in this pull request, please let me know if you'd like me to remove the bitbucket connector for a later review.

bcwaldon:
@ericchiang When updating godeps, please separate it into an independent commit for reviewability (example at https://github.com/coreos/dex/pull/129/commits).

bcwaldon:
@ericchiang I would like to see one architectural change to this. Given that there's no generic "oauth2" connector, I believe we should have a user create an impl-specific connector config. This means that in the connector config JSON blob should look something like this:

```
[
  {
    "type": "github",
    ...
  },
  {
    "type": "bitbucket",
    ...
  },
  {
    "type": "oidc",
    ...
  }
]
```

bcwaldon:
...and it's fine to support both github and bitbucket in here - separating them into two commits would be nice, though.

ericchiang:
@bcwaldon

Refactored the connector configurations to remove the generic "oauth2" type. Types are now "github" and "bitbucket".

Split up the commit to (hopefully) be more readable.

For health, github and bitbucket give very good error messages when you use the wrong client secret. For example bitbucket returns the following (github does something very similar):

```
{"error_description": "Invalid OAuth client credentials", "error": "unauthorized_client"}
```

However, github.com/coreos/go-oidc/oauth2 doesn't report these very well. See coreos/go-oidc#25. If we fix that issue, dex will be able to log much better error messages when things are misconfigured.

I couldn't find a way to verify clientID/clientSecret pairs though the github or bitbucket's API without trying to login a user.

ericchiang:
TODO: because github and bitbucket aren't actual email providers, mark them as untrusted. For example if you lose access to your email address you can still login to your github.

bobbyrullo:
in general out commit style is  "$DIR: $WHAT \n $WHY"

if len($DIR) > 1, then comma separate them. If it affects almost all dirs, then use *. For Godep the format is "Godeps: add $PKG"

examples:

one pkg:

```
foo: added IsBaz() method
Clients need to know if foos can baz for backwards compatibility
```

multiple packages: 

```
foo, bar: remove dependency on baz package.
Avoids tightly coupling to baz by using an interface
```

all packages:

```
*: rename authd to dex
because there are too many things called authd, and bcwaldon likes 90s cartoons.
```

ericchiang:
Security concern: GitHub's oauth implementation requires a state token during the user redirect and token callback. See notes [here](https://developer.github.com/v3/oauth/#web-application-flow):

For the redirect you must provide:

> state	string	An unguessable random string. It is used to protect against cross-site request forgery attacks.

When the callback is called, GitHub provides that same state, which must be verified to check against tampering.

> If the user accepts your request, GitHub redirects back to your site with a temporary code in a code parameter as well as the state you provided in the previous step in a state parameter. If the states don't match, the request has been created by a third party and the process should be aborted.

Don't know the exact attack this is preventing, but it's worth verifying.

To address this:
* Create a table in the database to store random UUIDs for state tokens.
* Once a state token is verified remove it from the database.
* Use the `Sync` method on the `Connector` interface to clean up old, unclaimed state tokens.
* Check if bitbucket has a similar concept.

bcwaldon:
We already sent a `state` field that contains the session key. I believe this is a non-issue.

ericchiang:
@bcwaldon ah you're right. Concern retracted.

ericchiang:
@bobbyrullo Cleaned up the commit messages, restructured the tests, and made github and bitbucket untrusted email providers. Users must now verify their email through dex.

@bcwaldon updating Godeps to include coreos/go-oidc#26 made API errors extremely explicit. I accidentally used the wrong client secret and dex logged this:

```
ERROR: Unable to verify auth code with issuer: incorrect_client_credentials: The client_id and/or client_secret passed are incorrect.
```

Bitbucket provides a similar message.

bobbyrullo:
LGTM

PR #202, opened 2015-12-07
connector: filter OIDC identities w/ configured domain

bcwaldon:
Fixes #66

PR #211, opened 2015-12-10
v0.2.0

ericchiang:
Would like to get #202 and #210 merged before releasing. Also probably need to change some of the build scripts to use a version rather than a git SHA.

Is there anything else we want to wait for?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,9 @@
+## v0.2.0
+
+- added github and bitbucket connectors (#190)
+- added invite emails (#168)
+- stronger encryption of token-signing keys (#167)
+- disable user from API (#140)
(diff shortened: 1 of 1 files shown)


bcwaldon:
@ericchiang thanks for writing these notes up, but we use GitHub releases and git tags to manage our release process. When we are ready to cut the release, we will cut a signed tag and use your changelog as the start of the release notes.

ericchiang:
@bcwaldon thanks. Realize now this PR should probably just be "added changelog".

Was planning on doing a proper github release once this got in :smile:

bcwaldon:
@ericchiang We use the GitHub release history as the de facto changelog, so I don't think this is necessary.

ericchiang:
@bcwaldon ah, okay.

PR #216, opened 2015-12-16
static: use --issuer-name as title for public facing pages

### LINKED ISSUE #142: Titles of Public facing dex pages are "Dex"
When logging in, the title of the page the user sees is "dex". It would be cool if this was (maybe just slightly) configurable, like --issuer-logo-url, so we could see "Bananacorp Login"  or just "Bananacorp Authentication" or something in the browser's title bar and in bookmarks.

ericchiang:
Closes #142


bcwaldon:
LGTM

Issue #224, opened 2015-12-17
Update UUID repository reference


Brahmasmi:
The currently referenced uuid repository,
```
code.google.com/p/go-uuid/uuid
```
(for example in https://github.com/coreos/dex/blob/master/user/user.go) is pointing to an old location and may be missing updates. The repo has now moved to https://github.com/pborman/uuid, as mentioned on https://code.google.com/p/go-uuid/.
```
github.com/pborman/uuid
```

We may want to update our references to the uuid repository.

PR #226, opened 2015-12-17
Don't import code.google.com/p/go-uuid/uuid

ericchiang:
As pointed out in #224 code.google.com is dead.

Since we only use it once in the entire code base, this is a good opportunity to get rid of a dependency.


bobbyrullo:
let's just move to where the new thing is.

PR #240, opened 2015-12-22
cmd: reduce backoff max when worker is waiting for connectors

### LINKED ISSUE #177: connectors: higher cap on connectors
Going through the getting started guide it can be rather annoying to have the exp. backoff go off to 1 minute. Is there a reason for the exp backoff? Can we cap it at 1 second or something?

https://github.com/coreos/dex/blob/9172f54fc2e35667f6ac5bb4fd303f5893e2d13b/cmd/dex-worker/main.go#L169

ericchiang:
Reduces the max amount of time between retries when waiting for the database to populate connectors from 1 minute to 8 seconds. 

fixes #177

### DIFF
--- a/cmd/dex-worker/main.go
+++ b/cmd/dex-worker/main.go
@@ -169,7 +169,7 @@ func main() {
 		if len(cfgs) > 0 && err == nil {
 			break
 		}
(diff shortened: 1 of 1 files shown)


bobbyrullo:
LGTM

PR #242, opened 2015-12-22
server: fix flow when user logs in through wrong connector

ericchiang:
This cleans up the code that deals with a user attempting to login through a different connector than they registered with. The only functional change is that `newLoginURLFromSession` is now called with register = false when a user has an existing account.

It fixes weird behavior seen in [this comment](https://github.com/coreos/dex/issues/180#issuecomment-166473092).


bobbyrullo:
nicely done, thanks

LGTM

Issue #166, opened 2015-10-23
feature request: automatically login oidc accounts from registration page


sym3tri:
Currently if a user registers with a Google account, then goes back to the site and tries to _register_ again with the same account they get a vague error message. Instead we should just log them in b/c they almost certainly meant to click "login" instead of "register".

joeatwork:
+1 - let's drop the whole magic register/login thing and just let the user explicitly choose?

bobbyrullo:
+1 to the fix, -1 the dropping the magic. +1 to adding *more* mag :)

What I mean is: It's still useful to have a link to a registration page. But it can be smarter about what happens when someone comes in with an existing account.

FWIW we already have "just let the user explicitly choose" - there's a reg link on login page and vice versa; this bug is asking for more than that.

ericchiang:
![screenshot from 2015-12-18 15-47-31](https://cloud.githubusercontent.com/assets/2342749/11909651/ac1d4688-a59e-11e5-95ab-16f43d34e9f6.png)

PR #246, opened 2015-12-23
server: better UX when remote ID already exists

bobbyrullo:
Instead of cryptic message with nowhere to, give them the choice to
login with that account or register.

### DIFF
--- a/server/register.go
+++ b/server/register.go
@@ -20,14 +20,21 @@ type formError struct {
 	Error string
 }
 
+type remoteExistsData struct {
+	Login string
+
+	Register string
+}
+
 type registerTemplateData struct {
-	Error      bool
-	FormErrors []formError
-	Message    string
-	Email      string
-	Code       string
-	Password   string
-	Local      bool
+	Error        bool
+	FormErrors   []formError
+	Message      string
+	Email        string
+	Code         string
+	Password     string
+	Local        bool
+	RemoteExists *remoteExistsData
 }
 
 var (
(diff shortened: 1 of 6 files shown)


bobbyrullo:
fixes #166 ?

bobbyrullo:
Ok, now you get taken to a page where you can login directly. Not sure how I feel about it though - a new session has to be created, and you get "logged in" in dex's mind even if you don't click the link. Take a look and tell me what you think.

server/register.go
```diff
@@ -92,6 +98,42 @@ func handleRegisterFunc(s *Server) http.HandlerFunc {
...
+				internalError(w, err)
+				return
+			}
+			// make sure to clean up the old session
+			if err = s.KillSession(code); err != nil {
+				internalError(w, err)
+			}
+
+			// finally, we can create a valid redirect URL for them.
+			redirURL, err := s.Login(ses.Identity, newSessionKey)
```
sym3tri: forget to handle this error?

server/register.go
```diff
@@ -304,3 +351,20 @@ func newLoginURLFromSession(issuer url.URL, ses *session.Session, register bool,
 	loginURL.RawQuery = v.Encode()
 	return &loginURL
 }
+
+func remoteIdentityExists(ur user.UserRepo, connectorID, id string) (error, bool) {
```
sym3tri: Wouldn't it be more idiomatic to have the return values be `(bool, error)`?

sym3tri:
I dig it.

server/register.go
bobbyrullo: hah yes. Clearly haven't been writing enough code lately

PR #250, opened 2015-12-29
db, functional: make value column in key not pkey

### LINKED ISSUE #247: postgres can't store RSA keys of length 2048 as primary keys
coreos/go-oidc#31 introduced a change that increased the default key length from 1024 to 2048 bytes. This causes a bug because the key is now too large to use as a primary key.

```
$ DEX_TEST_DSN='postgres://postgres@localhost/postgres?sslmode=disable' ./test-functional 
--- FAIL: TestDBPrivateKeySetRepoSetGet (1.07s)
	db_test.go:169: case 0: Unexpected error: pq: index row size 4008 exceeds maximum 2712 for index "key_pkey"
ERROR: error Decoding client creds 
ERROR: error Decoding client creds 
FAIL
FAIL	github.com/coreos/dex/functional	5.288s
```

Also, we're not running some functional tests through travis ci

bobbyrullo:
go-oidc increased the size of the generated keys, which were too large
to be used as primary keys in postgres.

fixes #247


ericchiang:
LGTM (assuming the functional tests pass)

PR #253, opened 2015-12-30
Add multiple storage backend support to dex

sercand:
Currently dex only supports postgres and in-memory as storage. This PR adds support for implementing additional backend storages.
Some Notes
- Postgres DB implementation is moved to "db/postgresql" and also "db/memory" is created for in-memory storage.
- "--no-db" flag is removed from from dex-worker and "--db" flag is added to dex-worker,dex-overlord and dexctl. 
- Default value of "--db" is "postgresql". 
- For using in-memory storage, user must run dex-worker with "--db memory" flag.
- server/config.go has changed to support multiple storages.
- Necessary changes are made for functional tests.
- Running functional tests for a storage can be done with DEX_TEST_STORAGE env variable.
- Garbage collecting at dex-overlord now runs only if storage backend needs it. Because some DB's support TTL items, I thought garbage collecting should be optional.
- Example work in progress storage backends are 'https://github.com/sercand/dex/tree/mongodb_storage' and 'https://github.com/sercand/dex/tree/etcd_storage'. Since mongodb does not support transactions, adding mongodb support is broken with current system.


ericchiang:
This is really awesome. However, we strongly discourage changes this large without opening proposals beforehand.

We have a lot of hesitations and specific concerns in regard to adding pluggable backends to dex. But again, that conversation is much better suited to a proposal issue than a PR.

I have some notes regarding the actual code, but would like to have a more general design discussion before we dive into the implementation. Can you open a proposal in the issues?

bobbyrullo:
Agree with what @ericchiang said - we are *so* happy to have more contributors, but big changes like this need to be discussed first.

sercand:
Ok, I'm working on it.

ericchiang:
As noted in #254, we'd love for you to put another PR together. Closing this for now.

PR #280, opened 2016-01-19
*: move user API auth to middleware and fix return status

### LINKED ISSUE #152: dex API should return 403, not 401, if users are authenticated but not authorized
For example, attempts to access `/users` by non-admin users should return 403, not 401

ericchiang:
Move client authentication into its own middleware and provide
differentiation between HTTP requests that do not provide
credentials (401) and requests that authenticate as a non-admin
user (403).

Closes #152


bobbyrullo:
LGTM

PR #283, opened 2016-01-20
travis: quote cwd build-docker-push script

dghubble:


ericchiang:
That seems to have fixed it. Though for some reason it's still complaining about ruby: https://travis-ci.org/coreos/dex/jobs/103652830

PR #293, opened 2016-01-27
Fixed db-url flag in standup-db.sh

set321go:
This change is required because otherwise i get ` unknown shorthand flag: 'd'`


ericchiang:
LGTM. Thanks!

Issue #295, opened 2016-01-27
Example app cannot register.


set321go:
I followed the steps in https://github.com/coreos/dex/blob/master/Documentation/getting-started.md and I read the other pages in the documentation but could not get the example app to work. 

I have been using the `contrib/standup-db.sh`. 

Which bits run on the three ports 5555, 5556, 5557?

When I visit localhost:5555 I get the home page with the register and login link.When I click the register link I am presented with a login page (google + local). This is the message I see in my log
```
INFO: URL: http://127.0.0.1:5556/auth?client_id=myclientid&redirect_uri=http%3A%2F%2F127.0.0.1%3A5555%2Fcallback&register=1&response_type=code&scope=openid+email+profile&state=
```

What should the `register=1` parameter do?
Is there any documentation about what parameters dex understands & what they do?

I also opened a pr for the script issue i ran across #293

ericchiang:
The login page and registration page are largely identical. Does the page your directed to say "Create Your dex Account"?

I think `register=1` is the only URL parameter that's dex specific. All the others are defined by OpenID Connect/OAuth2.0 which are documented by their relative specifications.

Looks like we don't currently note that. I'll open an issue to add some documentation around the callbacks.

set321go:
Both pages are the same, they both say "Log in to Dex" . I will have a fiddle but i'm pretty green when it comes to go and i've never tried debugging a go app.

Good to hear things are mostly just spec headers makes it easier to learn.

set321go:
Looks like in the `handleauthfunc` in `http.go` there is also a flag that is checked as well

    register := q.Get("register") == "1" && registrationEnabled

This looks like it comes in from the `Server` type should i have set this flag somewhere?

ericchiang:
Yep the command the getting started guide has you run is.

./bin/dex-worker --db-url=$DEX_DB_URL --key-secrets=$DEX_KEY_SECRET --email-cfg=static/fixtures/emailer.json.sample __--enable-registration=true__ --log-debug=true

Can you confirm if you have this or not?

set321go:
That's what I was missing, thanks. Ive submitted another PR for the script.

PR #296, opened 2016-01-27
worker start command missing registration and emailer config.

set321go:
As discussed in #295 registration was not working because the flag was not set on the worker startup. I also added the emailer config because I had to change the name of this file manually to get it picked up so it would be nice if it just started.

### DIFF
--- a/contrib/standup-db.sh
+++ b/contrib/standup-db.sh
@@ -69,6 +69,8 @@ EOF
 export DEX_WORKER_DB_URL=$DEX_DB_URL
 export DEX_WORKER_KEY_SECRETS=$DEX_KEY_SECRET
 export DEX_WORKER_LOG_DEBUG=1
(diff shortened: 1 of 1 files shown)


contrib/standup-db.sh
```diff
@@ -69,7 +69,7 @@ EOF
 export DEX_WORKER_DB_URL=$DEX_DB_URL
 export DEX_WORKER_KEY_SECRETS=$DEX_KEY_SECRET
 export DEX_WORKER_LOG_DEBUG=1
-./bin/dex-worker &
+./bin/dex-worker --email-cfg=static/fixtures/emailer.json.sample --enable-registration=true &
```
ericchiang: can you make these environment variables like the rest of the script?

```
export DEX_WORKER_EMAIL_CFG=static/fixtures/emailer.json.sample
export DEX_WORKER_ENABLE_REGISTRATION=true
```

ericchiang:
LGTM

PR #304, opened 2016-02-08
move to sqlite3 for --no-db mode and tests

ericchiang:
Currently, dex's state is represented by "repo" interfaces. For testing (and `--no-db` mode) we implement an in-memory version of all these interfaces so we don't have to rely on postgres for testing See the in memory [user repo](https://github.com/coreos/dex/blob/907f536e74edacf3457daddc7b369e5ef86e0bfb/user/user.go#L175-L183) for an example.

These are becoming a pain for a couple reasons.

1. They're a large portion of the code base.
2. Maps and slices don't map well to the SQL data structures, so adding additional columns to the database can result in having to completely restructure the in memory version. 
3. In memory repos can't reference data types from other repos. JOIN statements have no 
analog.
4. We don't test our SQL code when we use them in tests.

Goals of this PR are:

- Move away from implementing our own in-memory implementations
- Begin to experiment with supporting other SQL flavors.
- Have `--no-db` mode use SQLite.

NOTE:

__This PR DOES NOT add support for SQLite3 to dex.__ Only tests and `--no-db` mode.

TODO:

- [x] Add sqlite3 implementation of the database backend.
- [x] Move in memory representations to an in memory DB, including `--no-db` mode.
- [x] Disable sqlite3 support unless dex specifies `--no-db` mode.


ericchiang:
@bobbyrullo This is ready for a review. Large changes are

* Deleted all the "in memory" implementations of repos.
* dex now uses sqlite for tests and --no-db mode (functional tests still use postgres).
* Added a `db/translate` package to translate from Postgres queries to sqlite so we only have to write one kind of queries.
* Moved `SessionManager` to `session/manager` package (so it can import the db package for tests).
* A ton of small fixes to database queries and tests.
* No user facing changes.

On thing to note is that the SQL implementation of `ClientIdentityRepo` expects client secrets to be base64 encoded. Because of this, client secrets explicitly passed to dex during --no-db mode must be base64 encoded. See example [here](https://github.com/ericchiang/dex/blob/3942fdc62acdb7ddcac6a6bc2a1d32cc82b98a19/static/fixtures/clients.json)

ericchiang:
@bobbyrullo this has been updated (see the last two commits).

bobbyrullo:
Cool I'll take a look later on.

bobbyrullo:
LGTM - nice work!

PR #308, opened 2016-02-12
add support for resend an invite email

### LINKED ISSUE #184: api: can't resend an invite email
For administrators and user management tooling - it would be useful for administration clients to be able to stimulate invitation, email verification, and password reset messages, for example when the original messages are spam filtered or dropped.

rsoletob:
This change solves the User's API problem when you want to create an user that its email hasn't been verified yet but it exists.
At now, you can resend invitation email, but the user isn't modified if it was created before.

Fixes #184

### DIFF
--- a/user/api/api.go
+++ b/user/api/api.go
@@ -144,11 +144,6 @@ func (u *UsersAPI) CreateUser(creds Creds, usr schema.User, redirURL url.URL) (s
 		return schema.UserCreateResponse{}, ErrorUnauthorized
 	}
 
-	hash, err := generateTempHash()
-	if err != nil {
-		return schema.UserCreateResponse{}, mapError(err)
-	}
-
 	metadata, err := u.clientIdentityRepo.Metadata(creds.ClientID)
 	if err != nil {
 		return schema.UserCreateResponse{}, mapError(err)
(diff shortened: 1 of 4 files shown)


user/api/api.go
```diff
@@ -159,13 +154,34 @@ func (u *UsersAPI) CreateUser(creds Creds, usr schema.User, redirURL url.URL) (s
...
-	}
-
-	userUser, err := u.manager.Get(id)
-	if err != nil {
+	// Retrieve user to check if it's already created
+	userUser, err := u.manager.GetByEmail(usr.Email)
+
+	if err == nil {
+		// Check if email is verified
+		if userUser.EmailVerified {
```
ericchiang: If the Email is not verified, won't this just return an error?

bobbyrullo: No, why woud it?

bobbyrullo:
Hey there @rsoletob , I see what you're doing here, but why not just have a ResendInvitation api endpoint, and method? 

Otherwise the semantics of CreateUser are weird - "CreateUser creates a user, unless one already exists, in which case it sends an invitation email"

rsoletob:
> Hey there @rsoletob , I see what you're doing here, but why not just have a ResendInvitation api endpoint, and method?

It's sounds reasonable.

> Otherwise the semantics of CreateUser are weird - "CreateUser creates a user, unless one already exists, in which case it sends an invitation email"

I agree with you, I'll do a better PR with your advise.

bobbyrullo:
Awesome, I'll just close this one out then.

PR #316, opened 2016-02-15
connector_ldap: Implement connection pooling for LDAP connections

### LINKED ISSUE #309: LDAP connector: implement connection pooling
The LDAP connector implemented in #178 creates a new LDAP session for every request, closing and discarding the connection after login.

Add connection reuse to the existing logic. After attempting to login, connections should be:
* "reset" by binding anonymously
* kept alive by occasionally pinging the LDAP server (or discarded if this fails)
* reused on future login attempts by other dex end-users

fnordahl:
The initial implementation of the LDAP connector initiates and tears down a connection for each session. Although the creation and tear-down of connections are light weight operations for most LDAP server implementations, it is not hard to imagine that this at some point will become a bottle neck and scaling issue.

This PR addresses this issue by implementing connection pooling for the LDAP connections. The pooled connections are checked for liveliness and kept alive. The number of connections to keep as well as the interval for liveliness checks are configurable.

Fixes #309

### DIFF
--- a/Documentation/connectors-configuration.md
+++ b/Documentation/connectors-configuration.md
@@ -167,6 +167,8 @@ In addition to `id` and `type`, the `ldap` connector takes the following additio
 
 * skipCertVerification: a `boolean`. Skip server certificate chain verification.
 
+* maxIdleConn: a `integer`. Maximum number of idle LDAP Connections to keep in connection pool. Default: `5`
+
 * baseDN: a `string`. Base DN from which Bind DN is built and searches are based.
 
 * nameAttribute: a `string`. Attribute to map to Name. Default: `cn`
--- a/connector/connector_ldap.go
+++ b/connector/connector_ldap.go
@@ -12,6 +12,7 @@ import (
 	"net/url"
 	"path"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/coreos/dex/pkg/log"
@@ -40,6 +41,7 @@ type LDAPConnectorConfig struct {
 	KeyFile              string        `json:"keyFile"`
 	CaFile               string        `json:"caFile"`
 	SkipCertVerification bool          `json:"skipCertVerification"`
+	MaxIdleConn          int           `json:"maxIdleConn"`
 	BaseDN               string        `json:"baseDN"`
 	NameAttribute        string        `json:"nameAttribute"`
 	EmailAttribute       string        `json:"emailAttribute"`
@@ -81,6 +83,8 @@ func (cfg *LDAPConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, tpls *t
 	const defaultEmailAttribute = "mail"
 	const defaultBindTemplate = "uid=%u,%b"
 	const defaultSearchScope = ldap.ScopeWholeSubtree
+	const defaultMaxIdleConns = 5
+	const defaultPoolCheckTimer = 7200 * time.Second
 
 	if cfg.UseTLS && cfg.UseSSL {
 		return nil, fmt.Errorf("Invalid configuration. useTLS and useSSL are mutual exclusive.")
@@ -154,11 +158,22 @@ func (cfg *LDAPConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, tpls *t
 		tlsConfig.Certificates = []tls.Certificate{cert}
 	}
 
+	maxIdleConn := defaultMaxIdleConns
+	if cfg.MaxIdleConn > 0 {
+		maxIdleConn = cfg.MaxIdleConn
+	}
+
+	ldapPool := &LDAPPool{
+		MaxIdleConn:    maxIdleConn,
+		PoolCheckTimer: defaultPoolCheckTimer,
+		ServerHost:     cfg.ServerHost,
+		ServerPort:     cfg.ServerPort,
+		UseTLS:         cfg.UseTLS,
+		UseSSL:         cfg.UseSSL,
+		TLSConfig:      tlsConfig,
+	}
+
 	idp := &LDAPIdentityProvider{
-		serverHost:       cfg.ServerHost,
-		serverPort:       cfg.ServerPort,
-		useTLS:           cfg.UseTLS,
-		useSSL:           cfg.UseSSL,
 		baseDN:           cfg.BaseDN,
 		nameAttribute:    nameAttribute,
 		emailAttribute:   emailAttribute,
@@ -168,7 +183,7 @@ func (cfg *LDAPConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, tpls *t
 		searchBindDN:     cfg.SearchBindDN,
 		searchBindPw:     cfg.SearchBindPw,
 		bindTemplate:     bindTemplate,
-		tlsConfig:        tlsConfig,
+		ldapPool:         ldapPool,
 	}
 
 	idpc := &LDAPConnector{
@@ -188,9 +203,9 @@ func (c *LDAPConnector) ID() string {
 }
 
 func (c *LDAPConnector) Healthy() error {
-	ldapConn, err := c.idp.LDAPConnect()
+	ldapConn, err := c.idp.ldapPool.Acquire()
 	if err == nil {
-		ldapConn.Close()
+		c.idp.ldapPool.Put(ldapConn)
 	}
 	return err
 }
@@ -210,18 +225,145 @@ func (c *LDAPConnector) Register(mux *http.ServeMux, errorURL url.URL) {
 }
 
 func (c *LDAPConnector) Sync() chan struct{} {
-	return make(chan struct{})
+	stop := make(chan struct{})
+
+	go func() {
+		for {
+			select {
+			case <-time.After(c.idp.ldapPool.PoolCheckTimer):
+				alive, killed := c.idp.ldapPool.CheckConnections()
+				if alive > 0 {
+					log.Infof("Connector ID=%v idle_conns=%v", c.id, alive)
+				}
+				if killed > 0 {
+					log.Warningf("Connector ID=%v closed %v dead connections.", c.id, killed)
+				}
+			case <-stop:
+				return
+			}
+		}
+	}()
+	return stop
 }
 
 func (c *LDAPConnector) TrustedEmailProvider() bool {
 	return c.trustedEmailProvider
 }
 
+// A LDAPPool is a Connection Pool for LDAP connections
+// Initialize exported fields and use Acquire() to get a connection.
+// Use Put() to put it back into the pool.
+type LDAPPool struct {
+	m              sync.Mutex
+	conns          map[*ldap.Conn]struct{}
+	MaxIdleConn    int
+	PoolCheckTimer time.Duration
+	ServerHost     string
+	ServerPort     uint16
+	UseTLS         bool
+	UseSSL         bool
+	TLSConfig      *tls.Config
+}
+
+// Acquire removes and returns a random connection from the pool. A new connection is returned
+// if there are no connections available in the pool.
+func (p *LDAPPool) Acquire() (*ldap.Conn, error) {
+	conn := p.removeRandomConn()
+	if conn != nil {
+		return conn, nil
+	}
+	return p.ldapConnect()
+}
+
+// Put makes a connection ready for re-use and puts it back into the pool. If the connection
+// cannot be reused it is discarded. If there already are MaxIdleConn connections in the pool
+// the connection is discarded.
+func (p *LDAPPool) Put(c *ldap.Conn) {
+	p.m.Lock()
+	if p.conns == nil {
+		// First call to Put, initialize map
+		p.conns = make(map[*ldap.Conn]struct{})
+	}
+	if len(p.conns)+1 > p.MaxIdleConn {
+		p.m.Unlock()
+		c.Close()
+		return
+	}
+	p.m.Unlock()
+	// drop to anonymous bind
+	err := c.Bind("", "")
+	if err != nil {
+		// unsupported or disallowed, throw away connection
+		log.Warningf("Unable to re-use LDAP Connection after failure to bind anonymously: %v", err)
+		c.Close()
+		return
+	}
+	p.m.Lock()
+	p.conns[c] = struct{}{}
+	p.m.Unlock()
+}
+
+// removeConn attempts to remove the provided connection from the pool. If removeConn returns false
+// another routine is using the connection and the caller should discard the pointer.
+func (p *LDAPPool) removeConn(conn *ldap.Conn) bool {
+	p.m.Lock()
+	_, ok := p.conns[conn]
+	delete(p.conns, conn)
+	p.m.Unlock()
+	return ok
+}
+
+// removeRandomConn attempts to remove a random connection from the pool. If removeRandomConn
+// returns nil the pool is empty.
+func (p *LDAPPool) removeRandomConn() *ldap.Conn {
+	p.m.Lock()
+	defer p.m.Unlock()
+	for conn := range p.conns {
+		delete(p.conns, conn)
+		return conn
+	}
+	return nil
+}
+
+// CheckConnections attempts to iterate over all the connections in the pool and check wheter
+// they are alive or not. Live connections are put back into the pool, dead ones are discarded.
+func (p *LDAPPool) CheckConnections() (int, int) {
+	var conns []*ldap.Conn
+	var alive, killed int
+
+	// Get snapshot of connection-map while holding Lock
+	p.m.Lock()
+	for conn := range p.conns {
+		conns = append(conns, conn)
+	}
+	p.m.Unlock()
+
+	// Iterate over snapshot, Get and ping connections.
+	// Put live connections back into pool, Close dead ones.
+	for _, conn := range conns {
+		ok := p.removeConn(conn)
+		if ok {
+			err := ldapPing(conn)
+			if err == nil {
+				p.Put(conn)
+				alive++
+			} else {
+				conn.Close()
+				killed++
+			}
+		}
+	}
+	return alive, killed
+}
+
+func ldapPing(conn *ldap.Conn) error {
+	// Query root DSE
+	s := ldap.NewSearchRequest("", ldap.ScopeBaseObject, ldap.NeverDerefAliases, 0, 0, false, "(objectClass=*)", []string{}, nil)
+	_, err := conn.Search(s)
+	return err
+}
+
 type LDAPIdentityProvider struct {
-	serverHost       string
-	serverPort       uint16
-	useTLS           bool
-	useSSL           bool
 	baseDN           string
 	nameAttribute    string
 	emailAttribute   string
@@ -231,26 +373,26 @@ type LDAPIdentityProvider struct {
 	searchBindDN     string
 	searchBindPw     string
 	bindTemplate     string
-	tlsConfig        *tls.Config
+	ldapPool         *LDAPPool
 }
 
-func (m *LDAPIdentityProvider) LDAPConnect() (*ldap.Conn, error) {
+func (p *LDAPPool) ldapConnect() (*ldap.Conn, error) {
 	var err error
 	var ldapConn *ldap.Conn
 
 	log.Debugf("LDAPConnect()")
-	if m.useSSL {
-		ldapConn, err = ldap.DialTLS("tcp", fmt.Sprintf("%s:%d", m.serverHost, m.serverPort), m.tlsConfig)
+	if p.UseSSL {
+		ldapConn, err = ldap.DialTLS("tcp", fmt.Sprintf("%s:%d", p.ServerHost, p.ServerPort), p.TLSConfig)
 		if err != nil {
 			return nil, err
 		}
 	} else {
-		ldapConn, err = ldap.Dial("tcp", fmt.Sprintf("%s:%d", m.serverHost, m.serverPort))
+		ldapConn, err = ldap.Dial("tcp", fmt.Sprintf("%s:%d", p.ServerHost, p.ServerPort))
 		if err != nil {
 			return nil, err
 		}
-		if m.useTLS {
-			err = ldapConn.StartTLS(m.tlsConfig)
+		if p.UseTLS {
+			err = ldapConn.StartTLS(p.TLSConfig)
 			if err != nil {
 				return nil, err
 			}
@@ -273,11 +415,11 @@ func (m *LDAPIdentityProvider) Identity(username, password string) (*oidc.Identi
 	var bindDN, ldapUid, ldapName, ldapEmail string
 	var ldapConn *ldap.Conn
 
-	ldapConn, err = m.LDAPConnect()
+	ldapConn, err = m.ldapPool.Acquire()
 	if err != nil {
 		return nil, err
 	}
-	defer ldapConn.Close()
+	defer m.ldapPool.Put(ldapConn)
 
 	if m.searchBeforeAuth {
 		err = ldapConn.Bind(m.searchBindDN, m.searchBindPw)
@@ -307,16 +449,11 @@ func (m *LDAPIdentityProvider) Identity(username, password string) (*oidc.Identi
 		ldapName = sr.Entries[0].GetAttributeValue(m.nameAttribute)
 		ldapEmail = sr.Entries[0].GetAttributeValue(m.emailAttribute)
 
-		// drop to anonymous bind, prepare for bind as user
-		err = ldapConn.Bind("", "")
+		// prepare LDAP connection for bind as user
+		m.ldapPool.Put(ldapConn)
+		ldapConn, err = m.ldapPool.Acquire()
 		if err != nil {
-			// unsupported or disallowed, reconnect
-			log.Warningf("Re-connecting to LDAP Server after failure to bind anonymously: %v", err)
-			ldapConn.Close()
-			ldapConn, err = m.LDAPConnect()
-			if err != nil {
-				return nil, err
-			}
+			return nil, err
 		}
 	} else {
 		bindDN = m.ParseString(m.bindTemplate, username)
--- a/functional/ldap_test.go
+++ b/functional/ldap_test.go
@@ -7,9 +7,12 @@ import (
 	"net/url"
 	"os"
 	"strconv"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/coreos/dex/connector"
+	"golang.org/x/net/context"
 	"gopkg.in/ldap.v2"
 )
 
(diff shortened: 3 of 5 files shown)


connector/connector_ldap.go
```diff
@@ -154,11 +157,29 @@ func (cfg *LDAPConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, tpls *t
 		tlsConfig.Certificates = []tls.Certificate{cert}
 	}
 
+	var initConn, maxConn int
+	if cfg.InitConn != 0 {
```
iyee: I guess checking whether the config param `> 0` is better.

Documentation/connectors-configuration.md
```diff
@@ -167,6 +167,10 @@ In addition to `id` and `type`, the `ldap` connector takes the following additio
 
 * skipCertVerification: a `boolean`. Skip server certificate chain verification.
 
+* initConn: a `integer`. Number of LDAP Connections to establish on startup. Default: `5`
```
ericchiang: I don't think this is necessary. Let's just let the pool fill as various connections are made.

connector/connector_ldap.go
```diff
@@ -217,11 +238,74 @@ func (c *LDAPConnector) TrustedEmailProvider() bool {
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	for conn := range p.conns {
+		delete(p.conns, conn)
+		return conn, nil
+	}
+	// the pool is empty
+	conn, err := p.ldapConnect()
```
ericchiang: Because you're holding the mutex here you won't be able to connect more than one connection at a time. Need to unlock before this point.

connector/connector_ldap.go
```diff
@@ -217,11 +238,74 @@ func (c *LDAPConnector) TrustedEmailProvider() bool {
...
+
+func (p *LDAPPool) Put(c *ldap.Conn) {
+	p.m.Lock()
+	defer p.m.Unlock()
+
+	if len(p.conns)+1 > p.maxConn {
+		c.Close()
+	} else {
+		// drop to anonymous bind
+		err := c.Bind("", "")
```
ericchiang: Like the comment above, this bind should be performed outside of the mutex lock. The mutex should be used to guard the map and the counter, don't make remote calls while holding it.

connector/connector_ldap.go
```diff
@@ -217,11 +238,74 @@ func (c *LDAPConnector) TrustedEmailProvider() bool {
 	return c.trustedEmailProvider
 }
 
+type LDAPPool struct {
+	m          sync.Mutex
+	conns      map[*ldap.Conn]*ldap.Conn
```
ericchiang: Does this work instead?

```
conns map[*ldap.Conn]struct{} // A bag of connections
```

connector/connector_ldap.go
```diff
@@ -154,11 +157,29 @@ func (cfg *LDAPConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, tpls *t
 		tlsConfig.Certificates = []tls.Certificate{cert}
 	}
 
+	var initConn, maxConn int
+	if cfg.InitConn != 0 {
+		initConn = cfg.InitConn
+	} else {
+		initConn = 5
+	}
+	if cfg.MaxConn != 0 {
```
ericchiang: Can you clean up these initializations to: 

```
const defaultMaxConns = 30

...

if cfg.MaxConn <=0 {
    cfg.MaxConn = defaultMaxConn
}

...

ldapPool := &LDAPPool{
    ...
    maxConn: cfg.MaxConn,
    ...
}
```

fnordahl: I have deliberately created a temporary variable here and other places in the constructor to avoid modifying cfg.

My reasoning behind this is:

1. Do not to modify data structures created by parts of the code that I do not control directly.
2. If we at one point want to implement online reconfiguration of dex and all its connectors we probably do need to know the previous state of cfg to have something to compare with values read from disk to be able to determine whether a update is necessary for each value.

If you still do not agree with it I will change it, but at least you know the thought behind it.

connector/connector_ldap.go
fnordahl: I did consider to change this given that I only use the key of the map throughout the code. The value stored could be nil, so using a pointer type that uses the least amount of memory is probably the thing to look for.

Is your proposal the norm in Go for such usage of maps?

connector/connector_ldap.go
fnordahl: Yes, thx!

Documentation/connectors-configuration.md
fnordahl: It probably will be good enough, yes. LDAP servers are generally really good at connection initialization and teardown.

connector/connector_ldap.go
ericchiang: Sure sounds reasonable. I would still like to see constants for default values.

connector/connector_ldap.go
fnordahl: Yes of course, thanks. One path could be to allocate and store a new ldap.Conn in the map while holding the mutex, release the mutex and then initialize/dial the connection. If that is not allowed by the library I will have to reverse it. Create+Dial the connection and then lock+store+unlock and return. Will fix it.

connector/connector_ldap.go
fnordahl: Duly noted. Will fix it.

Can we go back to process based multi-processing, please? Forks are almost as cheap as threads these days! ;-) (Do not take this comment too seriously, although forks are cheap and in some circumstances adds to security.)

connector/connector_ldap.go
ericchiang: The empty struct takes [zero memory](https://play.golang.org/p/BR2urbScRB). But I wouldn't worry about memory constraints since you'll have at most a 100 connections or so.

I suggested this because `map[*ldap.Conn]*ldap.Conn` looks odd. This just seems more explicit. Wouldn't say it's the norm but I have seen maps combined with the empty struct for things like this.

connector/connector_ldap.go
ericchiang: Maybe this?

```go
func (p *LDAPPool) connFromPool() *ldap.Conn {
    p.mu.Lock()
    defer p.mu.Unlock()

    for conn := range p.conns { // you can range over a nil map
        delete(p.conns, conn)
        return conn
    }
    return nil
}

func (p *LDAPPool) Get() (*ldap.Conn, error) {
    if conn := p.connFromPool(); conn != nil {
        return conn, nil
    }
    return p.ldapConnect()
}
```

fnordahl: Nice. Elegant solution!

ericchiang: Go's net/http.Transport has some pretty neat tricks for [connection caching](https://github.com/golang/go/blob/7c22af830a72adf30b18ee3d1744aab8e3c009a8/src/net/http/transport.go#L668) if you're looking for inspiration. Though this definitely doesn't need to be a sophisticated as that implementation.

connector/connector_ldap.go
fnordahl: From small streams, mighty rivers can flow.

Your suggestion, `conns map[*ldap.Conn]struct{}`, looks neat and you can't do much better than zero. I'll use it.

connector/connector_ldap.go
fnordahl: Yes, there will be constants. I'll post a follow-up PR to turn the other defaults into constants as well.

fnordahl:
This probably should have some sort of keep-alive, reaper after N secs or periodic liveliness-check of the connections in the pool.

Maybe a check before return from Get() if the connection comes from the pool would do?

It would be bad to let random end users suffer from dead-after-idle-too-long connections left over from a previous surge in traffic.

ericchiang:
Is there a ping equivalent for LDAP connections?

fnordahl:
Yes, that + querying the root DSE is considered a "ping". I had it in my original Healthy function. Have it stashed somewhere.

Do we have an existing framework for firing functions periodically? When I think about it there probably should be a maxIdleConn parameter too. No need to keep 30 connections open when there is no pressure.

The connection is already anonymously bound when put in the pool, so I could implement a ping function that queries the root DSE and use it before returning the connection to the caller in Get.

fnordahl:
A follow-up question to this would be, do you think we need this for other stuff as well? There is ample opportunity to write a generic engine that can pool and keep-alive/ping anything, given that there already is or probably will be need for it.

ericchiang:
I think [`Sync()`](https://github.com/coreos/dex/blob/aff47d2cb2d43b8459fae8f4459bef9c85b99db9/connector/interface.go#L31-L34) is what you're looking for.

fnordahl:
Updated version:
- [x] Changed configuration statement from maxConn to maxIdleConn to better reflect what the pool actually does. (There is no cap on how many connections can be made, only how many idle to keep).
- [x] Implemented functions for pinging LDAP connections.
- [x] Implemented periodic liveness check of LDAP connections with configurable timer.

connector/connector_ldap.go
```diff
@@ -217,11 +250,108 @@ func (c *LDAPConnector) TrustedEmailProvider() bool {
...
+	MaxIdleConn    int
+	PoolCheckTimer time.Duration
+	ServerHost     string
+	ServerPort     uint16
+	UseTLS         bool
+	UseSSL         bool
+	TlsConfig      *tls.Config
+}
+
+func (p *LDAPPool) Get(params ...*ldap.Conn) (*ldap.Conn, error) {
```
ericchiang: Can you separate getting a random connection (Get()) and getting a specific connection (Get(params)) into different methods? They have very different uses and it's odd to overload this method.

connector/connector_ldap.go
```diff
@@ -217,11 +250,108 @@ func (c *LDAPConnector) TrustedEmailProvider() bool {
...
+	var conn *ldap.Conn
+
+	p.m.Lock()
```
ericchiang: What's being guarded by the lock and what's intentionally outside the lock isn't obvious from this flow, a lot of conditions, and manual unlocks. Splitting out the functionality of Get() and Get(conn) should help.

connector/connector_ldap.go
```diff
@@ -217,11 +250,108 @@ func (c *LDAPConnector) TrustedEmailProvider() bool {
...
+	for conn := range p.conns {
+		conns = append(conns, conn)
+	}
+	p.m.Unlock()
+
+	// Iterate over snapshot, Get and ping connections.
+	// Put live connections back into pool, Close dead ones.
+	for _, el := range conns {
+		conn, err := p.Get(el)
```
ericchiang: err is being ignored here (I think intentionally). Yeah, this is a bad use for Get.

Maybe just use

```go
// getConn attempts to remove the provided connection from the pool. If getConn returns false
// another routine is using the connection and the caller should discard the pointer.
func (p *LDAPPool) getConn(conn *ldap.Conn) bool {
    p.mu.Lock()
    defer p.mu.Lock()

    _, ok := p.conns[conn]
    delete(p.conns, conn)
    return ok
}
```

fnordahl: Nice suggestion. My original thought was that having two functions that needed to do the locking right would lead to double the opportunity to at some point do it wrong. But I guess that the complexity of the dual use in the one function does just this by itself. I'll split them up and post a new patch set.

Thanks!

connector/connector_ldap.go
```diff
@@ -210,18 +231,131 @@ func (c *LDAPConnector) Register(mux *http.ServeMux, errorURL url.URL) {
...
+			if alive > 0 {
+				log.Infof("Connector ID=%v idle_conns=%v", c.id, alive)
+			}
+			if killed > 0 {
+				log.Warningf("Connector ID=%v closed %v dead connections.", c.id, killed)
+			}
+			time.Sleep(c.idp.ldapPool.PoolCheckTimer)
+		}
+	}()
 	return make(chan struct{})
```
bobbyrullo: The return channel should be a stop chan, so that when you poke it, your process stops. (see connector_oidc.go)

connector/connector_ldap.go
```diff
@@ -154,11 +159,27 @@ func (cfg *LDAPConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, tpls *t
...
+	}
+
+	ldapPool := &LDAPPool{
+		MaxIdleConn:    maxIdleConn,
+		PoolCheckTimer: poolCheckTimer,
+		ServerHost:     cfg.ServerHost,
+		ServerPort:     cfg.ServerPort,
+		UseTLS:         cfg.UseTLS,
+		UseSSL:         cfg.UseSSL,
+		TlsConfig:      tlsConfig,
```
bobbyrullo: nit: TLSConfig

connector/connector_ldap.go
```diff
@@ -210,18 +231,131 @@ func (c *LDAPConnector) Register(mux *http.ServeMux, errorURL url.URL) {
...
+	defer p.m.Unlock()
+
+	_, ok := p.conns[conn]
+	delete(p.conns, conn)
+	return ok
+}
+
+// Ping attempts to iterate over all the connections in the pool and check wheter they are
+// alive or not. Live connections are put back into the pool, dead ones are discarded.
+func (p *LDAPPool) Ping() (int, int) {
```
bobbyrullo: With a name like `Ping`, I don't expect to see mutating behavior. Could you rename to something like "PruneConnections" or "CheckConnections"

connector/connector_ldap.go
```diff
@@ -210,18 +231,131 @@ func (c *LDAPConnector) Register(mux *http.ServeMux, errorURL url.URL) {
...
+	MaxIdleConn    int
+	PoolCheckTimer time.Duration
+	ServerHost     string
+	ServerPort     uint16
+	UseTLS         bool
+	UseSSL         bool
+	TlsConfig      *tls.Config
+}
+
+// Get returns a random connection from the pool. A new connection is returned if there are
```
bobbyrullo: It's not exactly random is it, just picks the first one off the top right?

connector/connector_ldap.go
```diff
@@ -210,18 +231,131 @@ func (c *LDAPConnector) Register(mux *http.ServeMux, errorURL url.URL) {
...
+		delete(p.conns, conn)
+		p.m.Unlock()
+		return conn, nil
+	}
+	p.m.Unlock()
+	return p.ldapConnect()
+}
+
+// Put makes a connection ready for re-use and puts it back into the pool. If the connection
+// cannot be reused it is discarded. If there allready are MaxIdleConn connections in the pool
```
bobbyrullo: spelling: already

connector/connector_ldap.go
```diff
@@ -210,18 +231,131 @@ func (c *LDAPConnector) Register(mux *http.ServeMux, errorURL url.URL) {
...
+		return
+	}
+	p.m.Lock()
+	p.conns[c] = struct{}{}
+	p.m.Unlock()
+}
+
+// getConn attempts to remove the provided connection from the pool. If getConn returns false
+// another routine is using the connection and the caller should discard the pointer.
+func (p *LDAPPool) getConn(conn *ldap.Conn) bool {
```
bobbyrullo: similar issue here: you don't expect side-effects with `getXXX`; try "removeConn"

bobbyrullo:
hey @fnordahl sorry we've neglected this PR so long (if that happens feel free to ping one of us). 

Anyhow, this is mostly good with a few nits. I would like to see some tests though, exercising potential race conditions - for example, what happens when ping checks the liveness of a connection that is no longer in the connection map (since you hold the lock for the map only briefly and then check)

connector/connector_ldap.go
fnordahl: Thx, I'll have a look at connector_oidc and make a new attempt at making it work.

connector/connector_ldap.go
fnordahl: Ack

connector/connector_ldap.go
fnordahl: I see your point. I will fix it.

connector/connector_ldap.go
fnordahl: Ack

connector/connector_ldap.go
fnordahl: I do see your point, but It feels weird to use "remove" to get a connection to do some work with. But I'll change it, I may try to come up with a alternative name that encompasses both concerns.

fnordahl:
Thank you for your review/comments @bobbyrullo . No worries on time passing, I have had a lot of stuff going on elsewhere latelly so there has unfortunatelly not been much time for me to focus on this. I'm back though and will se this through.

The rationale behind the ping function is that it at first should get a snapshot of the map quickly and not hold the lock for too long. After it has a snapshot of the map it starts to iterate over it using the getConn function. The getConn function does hold a lock, checks the existence of the specific connection requested, removes it from the map if it exists and releases the lock. If it does not exist that fact is returned to the caller. Thus there really should not be any room for races here.

Some scale tests and/or tests of behaviour with actual dead/timed out connections is in place. I'll have a look at how this can be implemented.

fnordahl:
Updated version:
- [x] nit: TLSConfig
- [x] *LDAPPool.Ping() -> *LDAPPool.CheckConnections()
- [x] Spelling: allready -> already
- [x] Get() -> Acquire(), getConn(...) -> removeConn(...)

Todo:
- [ ] Make Sync use stop chan

I currently have issues testing and I suspect that something has broken the LDAP connector. I will investigate more tomorrow.

fnordahl:
@ericchiang @bobbyrullo I have looked over the Sync stuff in go-oidc, and for the simple tasks the LDAP connector has to perform, it seems like a lot of code to accomplish little. Could I jump straight to attempt at an implementation outlined in #363 ?

Or are there a lot of infrastructure currently missing in Dex for this to be feasible / usable at this time?

connector/connector_ldap.go
```diff
@@ -210,18 +231,131 @@ func (c *LDAPConnector) Register(mux *http.ServeMux, errorURL url.URL) {
...
+	TLSConfig      *tls.Config
+}
+
+// Acquire removes and returns a random connection from the pool. A new connection is returned
+// if there are no connections available in the pool.
+func (p *LDAPPool) Acquire() (*ldap.Conn, error) {
+	p.m.Lock()
+	for conn := range p.conns {
+		delete(p.conns, conn)
+		p.m.Unlock()
```
philips: please use defer for unlocks unless there is a strong reason to avoid it

connector/connector_ldap.go
```diff
@@ -210,18 +231,131 @@ func (c *LDAPConnector) Register(mux *http.ServeMux, errorURL url.URL) {
...
+	p.m.Lock()
+	p.conns[c] = struct{}{}
+	p.m.Unlock()
+}
+
+// removeConn attempts to remove the provided connection from the pool. If removeConn returns false
+// another routine is using the connection and the caller should discard the pointer.
+func (p *LDAPPool) removeConn(conn *ldap.Conn) bool {
+	p.m.Lock()
+	defer p.m.Unlock()
```
philips: in this case not having the defer feels cleaner

connector/connector_ldap.go
fnordahl: The reasoning behind manual Lock/Unlock in this function is that the call to p.ldapConnect() could make us block while holding the lock. An alternative could be to return nil and let it be up to the user of the function to get a connection himself and later on put it into the pool. At the time of writing it somehow felt better to have all connection related things happen through the pool interface, but there is nothing wrong with doing this separation if you think that fits better.

connector/connector_ldap.go
fnordahl: I'm ok with that. Simple and straight forward function where it is easy to foresee the entire call path with the glance of an eye.

connector/connector_ldap.go
fnordahl: Could of course also do the actual lifting in a private function with Lock() + defer Lock in place, and do a call to that + return p.ldapConnect() in the public function.

fnordahl:
> @ericchiang @bobbyrullo I have looked over the Sync stuff in go-oidc, and for the simple tasks the LDAP connector has to perform, it seems like a lot of code to accomplish little. Could I jump straight to attempt at an implementation outlined in #363 ?
> 
> Or are there a lot of infrastructure currently missing in Dex for this to be feasible / usable at this time?

I believe I found a simpler way of solving it while still following the current scheme. New version up. How does this look?

I still have the issues with something breaking the connector, but I'll continue investigate that and get back with a fix. (I probably have no one else than myself not writing good enough tests to blame, I will make amends.)

Would love some feedback on the intermediate work though.

fnordahl:
After doing a binary search I have identified that this commit breaks the LDAP connector: 821b242c83bd1770af0886899be163bdc77749cd

I will investigate further and of course use the wonderful experience of having my code being broken by other commits to improve my tests! :-)

fnordahl:
Some more detail: It is the "register-maybe" process that breaks. Manually registering before first login attempt still works.

fnordahl:
@philips I have addressed the usage/code-style for locks per your comments.

Should I refactor throughout, moving the parts touching global structures to seperate functions so that the call paths either are obvious at the glimpse of an eye or I can make use of defer Unlock without blocking?

If so, I think @ericchiang have suggested this for me before, so sorry for not listening the first time... ;)

fnordahl:
@ericchiang what's your take on this? Are the rest of the functions with locks OK as is, or do you suggest I refactor them? Refactoring would be move parts touching global structures to separate functions that can either be simple enough so you can spot the entire call path at the blink of an eye, or that they contain no blocking calls so Lock+defer Unlock could be used.

ericchiang:
@fnordahl sorry to take so long following up.

Regarding the actual code, I'm okay with it not being the most "elegant" implementation if it works. However we're going to need a plan to test this logic (particularly with concurrent requests). Would it be okay to move on to that part?

fnordahl:
Thank you for the feedback @ericchiang . Sure, I'll brew up some tests.

fnordahl:
And there we have functional stress tests for the LDAP Connection Pool. What do you think of it @ericchiang

functional/ldap_test.go
```diff
@@ -123,3 +125,52 @@ func TestConnectorLDAPHealthy(t *testing.T) {
...
+		UseSSL:      false,
+	}
+
+	// Excercise pool operations with MaxIdleConn + 10 concurrent goroutines.
+	// The 10 surplus connections should be discarded after use.
+	// Remaining active connections will be verified at the end of the test.
+	numRoutines := ldapPool.MaxIdleConn + 10
+	var wg sync.WaitGroup
+	for i := 0; i < numRoutines; i++ {
+		wg.Add(1)
```
ericchiang: This is kinda a race. Please just do a `wg.Add(numRountines)`

functional/ldap_test.go
```diff
@@ -123,3 +125,52 @@ func TestConnectorLDAPHealthy(t *testing.T) {
...
+					ldapConn, err := ldapPool.Acquire()
+					if err != nil {
+						t.Errorf("Unable to acquire LDAP Connection: %v", err)
+					}
+					s := ldap.NewSearchRequest("", ldap.ScopeBaseObject, ldap.NeverDerefAliases, 0, 0, false, "(objectClass=*)", []string{}, nil)
+					_, err = ldapConn.Search(s)
+					if err != nil {
+						t.Errorf("Search request failed. Dead/invalid LDAP connection from pool?: %v", err)
+					}
+					ldapPool.Put(ldapConn)
```
ericchiang: probably don't want to Put if there was an error

functional/ldap_test.go
```diff
@@ -123,3 +125,52 @@ func TestConnectorLDAPHealthy(t *testing.T) {
...
+					_, _ = ldapPool.CheckConnections()
+				}
+			}
+		}()
+	}
+
+	// Wait for all operations to complete and check status.
+	wg.Wait()
+	alive, killed := ldapPool.CheckConnections()
```
ericchiang: Why do we care how many are alive?

functional/ldap_test.go
```diff
@@ -123,3 +125,52 @@ func TestConnectorLDAPHealthy(t *testing.T) {
 		}
 	}
 }
+
+func TestLDAPPool(t *testing.T) {
```
ericchiang: Might pick a more descriptive name. Like "TestMultipleLDAPConns". You aren't testing the pool, you're testing concurrent access :)

functional/ldap_test.go
```diff
@@ -123,3 +125,52 @@ func TestConnectorLDAPHealthy(t *testing.T) {
...
+
+	// Excercise pool operations with MaxIdleConn + 10 concurrent goroutines.
+	// The 10 surplus connections should be discarded after use.
+	// Remaining active connections will be verified at the end of the test.
+	numRoutines := ldapPool.MaxIdleConn + 10
+	var wg sync.WaitGroup
+	for i := 0; i < numRoutines; i++ {
+		wg.Add(1)
+		go func() {
+			timeout := time.After(5 * time.Second)
```
ericchiang: This can probably be done with a golang.org/x/net/context object.

functional/ldap_test.go
```diff
@@ -123,3 +125,52 @@ func TestConnectorLDAPHealthy(t *testing.T) {
 		}
 	}
 }
+
+func TestLDAPPool(t *testing.T) {
+	server := ldapServer(t)
+	ldapPool := &connector.LDAPPool{
+		MaxIdleConn: 30,
```
ericchiang: Would be cool if this was a benchmark, see how tuning this number effects the throughput. (For a follow up PR)

functional/ldap_test.go
fnordahl: The calls are executed serially by the for loop, so no room for races here. But sure, I agree that calling the function 40 times is less efficient than once :-)

functional/ldap_test.go
fnordahl: The Put function should check the eligibility of the connections for re-use before adding them back to the pool. So it would be a good test. But let's do a explicit test of that instead of randomly putting back known broken connections. Thanks for pointing it out.

functional/ldap_test.go
fnordahl: Checking that there are MaxIdleConn connections active after the test gives us information about two things:
1. Did we indeed have concurrent operations during the test.
2. The test runs more than MaxIdleConn connections concurrently, and checking that the pool stays at MaxIdleConn after the run confirms that the code handling the high watermark of the pool works.

functional/ldap_test.go
fnordahl: We do both. In my head, the reason for the test is to verify that we do not have lock contention while exercising the various pool functions concurrently. But sure, I'll find a more descriptive name.

functional/ldap_test.go
fnordahl: Thx! I'll check it out.

functional/ldap_test.go
fnordahl: Indeed, that is a good idea. Didn't think of that. Thx!

functional/ldap_test.go
ericchiang: It's racy because theoretically the `wg.Done` could race with the `wg.Add` (since the latter's in a goroutine). It's more correct this way.

functional/ldap_test.go
ericchiang: That sound good. But shouldn't a connection be able to error out? Isn't that one of the points of a connection pool (to be able to tolerate bad connections)?

functional/ldap_test.go
fnordahl: Ack

functional/ldap_test.go
fnordahl: Yes indeed. The connections in this test most likely will never error out. I will create a separate test for the handling of a harsh environment with connections going bad.

Documentation/connectors-configuration.md
```diff
@@ -167,6 +167,10 @@ In addition to `id` and `type`, the `ldap` connector takes the following additio
 
 * skipCertVerification: a `boolean`. Skip server certificate chain verification.
 
+* maxIdleConn: a `integer`. Maximum number of idle LDAP Connections to keep in connection pool. Default: `5`
+
+* poolCheckTimer: a `integer`. Time in Seconds for periodic liveness-check of connections in pool. Default: `7200` (Every two hours)
```
ericchiang: This doesn't need to be configurable. Let's drop it for brevity.

ericchiang:
one last comment, but other than that lgtm

Documentation/connectors-configuration.md
fnordahl: Done.

ericchiang:
:+1: as always great work

PR #323, opened 2016-02-18
functional: don't fail if postgres or ldap isn't available

### LINKED ISSUE #318: Allow running a subset of functional tests
With the addition of the ldap connector it's now hard to run functional tests outside of travis. Allow running functional tests against various targets.

```
./test-functional postgres
./test-functional ldap
```

ericchiang:
It's currently very hard to run functional tests because you need
to set both the Postgres and LDAP environment variables or the
init() functions for the various test suites will kill all the
functional tests before they can start.

Use t.Fatal instead of os.Exit in the init().

Closes #318

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,9 @@ go:
 
 env:
   - DEX_TEST_DSN="postgres://postgres@127.0.0.1:15432/postgres?sslmode=disable" ISOLATED=true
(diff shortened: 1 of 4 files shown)


ericchiang:
@fnordahl please take a look at this one. I reworked the LDAP functional tests a bit and want to get your okay.

Issue #261, opened 2016-01-09
feature: revoking refresh tokens


ericchiang:
Notes:

OAuth2.0 spec provides an example with basic auth.

> The client constructs the request by including the following parameters using the "application/x-www-form-urlencoded" format in the HTTP request entity-body:

```
POST /revoke HTTP/1.1
Host: server.example.com
Content-Type: application/x-www-form-urlencoded
Authorization: Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW

token=45ghiukldjahdnhzdauz&token_type_hint=refresh_token
```

Google provides [similar examples](https://developers.google.com/identity/protocols/OAuth2WebServer?hl=en#tokenrevoke):

>To programmatically revoke a token, your application makes a request to https://accounts.google.com/o/oauth2/revoke and includes the token as a parameter:

> curl https://accounts.google.com/o/oauth2/revoke?token={token}

> The token can be an access token or a refresh token. If the token is an access token and it has a corresponding refresh token, the refresh token will also be revoked.

> If the revocation is successfully processed, then the status code of the response is 200. For error conditions, a status code 400 is returned along with an error code.

Salesforce has a similar [endpoint](https://help.salesforce.com/apex/HTViewHelpDoc?id=remoteaccess_revoke_token.htm&language=en).

The API driven revocations I've been able to find are for the OAuth client. End users generally revoke access through web forms like this:

![oauth2revoke](https://cloud.githubusercontent.com/assets/2342749/12213198/3de6c2e8-b629-11e5-89c3-44d199d92e6a.png)

Seems reasonable to add a revoke endpoint for clients like the one specified in the [OAuth2.0 token revocation spec](https://tools.ietf.org/html/rfc7009). Maybe we should consider different endpoints to cover other kinds of situations?

ericchiang:
cc @bobbyrullo, @sym3tri 

# Proposal: revoking refresh tokens

## Motivation:

1. Need a way for users to revoke client's access to their identity.
2. Users never see the refresh token issued to clients. Can't just say "revoke this token."
3. Need a way for as user to "login to dex" and revoke a client.
4. Want to force as little UI as possible onto projects using dex (#255). Things should be API driven.

## Proposal:

* Add new API endpoints that allow any user to see what clients they're logged into and revoke access.
* Have this API require logging in through an "admin" client.
* Generate a client id and secret for dex to use to "login to itself" so it can access this API.

## Details:

### Overview of APIs

[...]

PR #329, opened 2016-02-22
add ability to revoke refresh tokens in user API

ericchiang:
work for #261

### DIFF
--- a/schema/workerschema/README.md
+++ b/schema/workerschema/README.md
@@ -34,20 +34,6 @@ __Version:__ v1
 }
 ```
 
-### ClientWithMetadata
-
-A client with associated metadata. The value of the metadata field is the OpenID Connect dynamic registration Client Metadata object. The definition of that object been omitted here for brevity.
-
-```
-{
-    Credentials: {
-        ID: string
-    },
-    Metadata: {
-    }
-}
-```
-
 ### ClientWithSecret
 
 
@@ -70,6 +56,19 @@ A client with associated metadata. The value of the metadata field is the OpenID
 {
     error: string,
     error_description: string
+}
+```
+
+### RefreshClient
+
+A client with associated public metadata.
+
+```
+{
+    clientID: string,
+    clientName: string,
+    clientURI: string,
+    logoURI: string
 }
 ```
 
@@ -184,7 +183,7 @@ A client with associated metadata. The value of the metadata field is the OpenID
 
 > __Summary__
 
-> List Client
+> List RefreshClient
 
 > __Description__
 
@@ -203,7 +202,7 @@ A client with associated metadata. The value of the metadata field is the OpenID
 
 > __Summary__
 
-> Revoket Client
+> Revoke RefreshClient
 
 > __Description__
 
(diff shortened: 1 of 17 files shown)


ericchiang:
Update: I've implemented the API endpoints. Need to add documentation and test cases.

ericchiang:
@bobbyrullo pushed a commit that cleans up the naming and addresses most of your concerns.

As for clients having multiple refresh tokens for a given user, I commented inline. See [here](https://github.com/coreos/dex/pull/329#commitcomment-16333949)

ericchiang:
Naming updated

ericchiang:
@bobbyrullo ready for review

ericchiang:
updated (see last 4 commits). Will add some tests and squash later.

user/api/api.go
```diff
@@ -99,10 +106,23 @@ type Creds struct {
...
+	return NewUsersAPIWithClock(dbMap, emailer, localConnectorID, clockwork.NewRealClock())
+}
+
+func NewUsersAPIWithClock(dbMap *gorp.DbMap, emailer Emailer, localConnectorID string, clock clockwork.Clock) *UsersAPI {
+
+	userRepo := db.NewUserRepo(dbMap)
+	passwordInfoRepo := db.NewPasswordInfoRepo(dbMap)
+	connectorConfigRepo := db.NewConnectorConfigRepo(dbMap)
+	txFactory := db.TransactionFactory(dbMap)
+	opts := manager.ManagerOptions{}
```
bobbyrullo: please pass in the manager here.

schema/workerschema/README.md
```diff
@@ -34,6 +34,20 @@ __Version:__ v1
 }
 ```
 
+### ClientWithMetadata
+
```
bobbyrullo: please don't mix autogenerated code with handwritten code.

schema/workerschema/v1.json
```diff
@@ -365,23 +378,10 @@
           "response": {
```
bobbyrullo: maybe this should be /account/${user_id}/refresh/clients

schema/workerschema/v1.json
```diff
@@ -254,6 +254,19 @@
...
             "$ref": "ClientWithSecret"
           }
+        },
+        "Revoke": {
+          "id": "dex.Client.Revoke",
+          "description": "Revoke all refresh tokens issues to the client for the authenticated user.",
+          "httpMethod": "DELETE",
+          "path": "client/refresh/{clientid}/revoke",
```
bobbyrullo: why do you need the revoke at the end? Also, this seems to more rightly belong under account no?

ericchiang: Revoke just because we might add other actions later. Happy to take it out.

I believe you asked it to be moved from '/account' to the '/client' prefix (though github seems to have lost that comment).

bobbyrullo: please do remove it - you can still use the path for other things, just not DELETE.

Also, I was asking for the opposite, and some consistency. This refresh token is something that belongs to the user really, not the client. 

so something like

DELETE /account/${user_id}/refresh/${client_id}   for revoking

and 

GET /account/${user_id}/refresh/${client_id}  for listing.

ericchiang:
Opening #386 rather than force pushing here to avoid losing all the inline comments.

Issue #334, opened 2016-02-23
Proposal: store password hashes in remote identities


ericchiang:
Came up when investigating #332.

Storing passwords as part of the remote identities would allow us to:
1. Remove a table and associated repo.
2. Eliminate a configuration file for --no-db mode.
3. Allow multiple local connectors.

```
// RemoteIdentity represents a User's identity at an IDP.
type RemoteIdentity struct {
	// IDPCID is the identifier of the IDP which hosts this identity.
	ConnectorID string

	// ID is the identifier of this User at the IDP.
	ID string

	// Password hash for remote identities which require 
	PasswordHash []byte
}
```

The current config looks like this

```
$ cat static/fixtures/users.json.sample 
[
    {
        "user":{
            "id": "elroy-id",
            "email": "elroy77@example.com",
            "displayName": "Elroy Jonez"
        },
        "remoteIdentities": [
            {
                "connectorId": "local",
                "id": "elroy-id"
            }
        ]
    }
]
$ cat static/fixtures/connectors.json.sample 
[
	{
		"type": "local",
		"id": "local",
		"passwordInfos": [
			{
				"userId":"elroy-id",
				"passwordPlaintext": "bones"
			},
			{
				"userId":"penny",
				"passwordPlaintext": "kibble"
			}
		]
	},
	{
		"type": "oidc",
		"id": "google",
		"issuerURL": "https://accounts.google.com",
		"clientID": "${CLIENT_ID}",
		"clientSecret": "${CLIENT_SECRET}"
	},
	{
		"type": "github",
		"id": "github",
		"clientID": "${CLIENT_ID}",
		"clientSecret": "${CLIENT_SECRET}"
	},
	{
		"type": "bitbucket",
		"id": "bitbucket",
		"clientID": "${CLIENT_ID}",
		"clientSecret": "${CLIENT_SECRET}"
	}
]
```

I'd like to move the passwords to the remote identity which would change users.json.sample to

```
[
    {
        "user":{
            "id": "elroy-id",
            "email": "elroy77@example.com",
            "displayName": "Elroy Jonez"
        },
        "remoteIdentities": [
            {
                "connectorId": "local",
                "id": "elroy-id",
                "passwordPlaintext": "bones"
            }
        ]
    }
]
```

This would also fix #290

PR #335, opened 2016-02-23
user: fix password info JSON encoding to survive round trips

### LINKED ISSUE #332: regression: Not able to log in with local connector when running dex-worker with --no-db
PR #304 introduces a regression.

With dex-worker running with --no-db it is no longer possible to log in using the local connector with credentials configured in static/fixtures/connectors.json and static/fixtures/users.json.

Adding some Printlns to password.go UnmarshalJSON function shows that the users are created twice, the first time with valid password info, the second time with blank password.
<pre>diff --git a/user/password.go b/user/password.go
index 4b93f34..8fb67cf 100644
--- a/user/password.go
+++ b/user/password.go
@@ -153,6 +153,7 @@ func (u *PasswordInfo) UnmarshalJSON(data []byte) error {
        }
 
        u.UserID = dec.UserID
+       fmt.Println("%v", u.UserID)
 
        u.PasswordExpires = dec.PasswordExpires
 
@@ -169,6 +170,7 @@ func (u *PasswordInfo) UnmarshalJSON(data []byte) error {
                        return err
                }
        }
+       fmt.Println("%v", u.Password)
        return nil
 }
 
</pre>

When running a version of dex prior to the commit of #304 it works as expected
``git checkout 9ca27a20352edfdd335b571bc7a6f61258294cdc``

Unfortunately I am prohibited from chasing this to the root cause at this time. Maybe you know what's going on from this description, @ericchiang ?

ericchiang:
PasswordInfos are marshaled when storing them in the database as
part of the local connector. However, the custom unmarsheler
defined could not unmarshal the standard marshling of this struct.

Add a struct tag to the Password field to correct this.

I'd like to see #334 be the long term fix to this. However for now, this works.

Closes #332 

cc @fnordahl

### DIFF
--- a/connector/config_repo_test.go
+++ b/connector/config_repo_test.go
@@ -61,7 +61,7 @@ func TestNewConnectorConfigFromMap(t *testing.T) {
 				"type": "local",
 				"id":   "foo",
 				"passwordInfos": []map[string]string{
-					{"userId": "abc", "passwordHash": "PING"},
+					{"userId": "abc", "passwordHash": "UElORw=="}, // []byte is base64 encoded when using json.Marshasl
 					{"userId": "271", "passwordPlaintext": "pong"},
 				},
 			},
--- a/user/password.go
+++ b/user/password.go
@@ -53,9 +53,9 @@ func NewPasswordFromPlaintext(plaintext string) (Password, error) {
 type PasswordInfo struct {
 	UserID string
 
-	Password Password
+	Password Password `json:"passwordHash"`
 
-	PasswordExpires time.Time
+	PasswordExpires time.Time `json:"passwordExpires"`
 }
 
 func (p PasswordInfo) Authenticate(plaintext string) (*oidc.Identity, error) {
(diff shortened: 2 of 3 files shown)


fnordahl:
I tested the patch and it resolves the situation and I can log in again. :+1: 

I have a question about another change with #304: The string passed to the --client-secret parameter of example-app now also has to be base64-encoded.

Is this intentional?

ericchiang:
#304 moved the --no-db mode to use the postgres logic. The postgres code unfortunately assumes that [client secrets are base64 encoded](https://github.com/coreos/dex/blob/f51125f55538c8fc5a3ad1e780296fd505fb948d/db/client.go#L105).

Should be easy enough to change it back.

fnordahl:
Either that or have the server base64 encode the password submitted from the client before further internal handling. We can't assume or expect that clients in the outside world would base64-encode their passwords.

I have opened PR #345 that addresses this issue.

fnordahl:
Never mind my last comment and PR. I was convinced this behaviour was also present when running normally with a real DB which is not the case. Sorry about that.

PR #339, opened 2016-02-24
db: email comparison should be case insensitive

### LINKED ISSUE #338: Emails in dex are case sensitive
I wasn't 100% this is a bug, but after reading around and asking @brianredbeard , it seems emails should probably be treated as case insensitive. Currently you can register with the same email multiple times if the case differs.

This actually is  causing bugs when using the email in another system I'm working on where we expected the email casing to not matter.

ericchiang:
Going to add a few tests and some additional documentation, but this provides a fix for #338 that does not require a database migration (e.g. lower casing all of the emails currently in the DB).

fixes #338


bobbyrullo:
Why aren't we instead preventing mixed case and doing the DB migration?

ericchiang:
Fine doing it both ways. Thought it might be rough to have a migration with a rolling upgrade. You'd be able to create a user with an invalid email on an old work, for instance.

bobbyrullo:
I think this needs to be two stages:

1) stop allowing mixed case

after people have installed that release they can do a fix.

2) new release with migration which fails if people haven't done the fix
(we don't want this automated I think)

ericchiang:
ready for review @bobbyrullo

bobbyrullo:
Minor issues, otherwise LGTM!

Issue #326, opened 2016-02-20
Add dynamic client registration to bootstrapping (admin) API


ericchiang:
This allows us to provide authentication in front of the dynamic client registration and gets us closer to feature parity through the APIs for #298

bobbyrullo:
Please explain this issue - how does it get us authentication? Also, moving dynamic client registration to the overlord means having to expose the overlord outside of your firewall.

ericchiang:
Sorry. I'd like to add a link to the admin API to allow registration. Not remove it from dex itself.

Currently `dexctl new-client` uses a direct connection to the DB to create that client. We need an alternative API to allow this. I'd like to add an endpoint to the overlord API (protected by the secret token).

ericchiang:
closed in https://github.com/coreos/dex/pull/392

PR #343, opened 2016-02-26
add client registration to the admin API

ericchiang:
work for #326

Need to add documentation and add generated schemas

### DIFF
--- a/admin/api.go
+++ b/admin/api.go
@@ -4,6 +4,11 @@ package admin
 import (
 	"net/http"
 
+	"github.com/coreos/go-oidc/oidc"
+	"github.com/go-gorp/gorp"
+
+	"github.com/coreos/dex/client"
+	"github.com/coreos/dex/db"
 	"github.com/coreos/dex/schema/adminschema"
 	"github.com/coreos/dex/user"
 	"github.com/coreos/dex/user/manager"
@@ -11,22 +16,32 @@ import (
 
 // AdminAPI provides the logic necessary to implement the Admin API.
 type AdminAPI struct {
-	userManager      *manager.UserManager
-	userRepo         user.UserRepo
-	passwordInfoRepo user.PasswordInfoRepo
-	localConnectorID string
+	userManager        *manager.UserManager
+	userRepo           user.UserRepo
+	passwordInfoRepo   user.PasswordInfoRepo
+	clientIdentityRepo client.ClientIdentityRepo
+	localConnectorID   string
 }
 
-func NewAdminAPI(userManager *manager.UserManager, userRepo user.UserRepo, pwiRepo user.PasswordInfoRepo, localConnectorID string) *AdminAPI {
+// TODO(ericchiang): Swap the DbMap for a storage interface. See #278
+
+func NewAdminAPI(dbMap *gorp.DbMap, localConnectorID string) *AdminAPI {
 	if localConnectorID == "" {
 		panic("must specify non-blank localConnectorID")
 	}
 
+	userRepo := db.NewUserRepo(dbMap)
+	passwordInfoRepo := db.NewPasswordInfoRepo(dbMap)
+	connectorConfigRepo := db.NewConnectorConfigRepo(dbMap)
+	txFactory := db.TransactionFactory(dbMap)
+	opts := manager.ManagerOptions{}
+	userManager := manager.NewUserManager(userRepo, passwordInfoRepo, connectorConfigRepo, txFactory, opts)
 	return &AdminAPI{
-		userManager:      userManager,
-		userRepo:         userRepo,
-		passwordInfoRepo: pwiRepo,
-		localConnectorID: localConnectorID,
+		userManager:        userManager,
+		userRepo:           userRepo,
+		passwordInfoRepo:   passwordInfoRepo,
+		clientIdentityRepo: db.NewClientIdentityRepo(dbMap),
+		localConnectorID:   localConnectorID,
 	}
 }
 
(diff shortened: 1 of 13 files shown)


ericchiang:
@bobbyrullo ready for review

admin/api.go
```diff
@@ -11,22 +16,32 @@ import (
...
+	userManager        *manager.UserManager
+	userRepo           user.UserRepo
+	passwordInfoRepo   user.PasswordInfoRepo
+	clientIdentityRepo client.ClientIdentityRepo
+	localConnectorID   string
 }
 
-func NewAdminAPI(userManager *manager.UserManager, userRepo user.UserRepo, pwiRepo user.PasswordInfoRepo, localConnectorID string) *AdminAPI {
+// TODO(ericchiang): Swap the DbMap for a storage interface. See #278
+
```
bobbyrullo: I really don't like the DbMap leaking out of where it's supposed to be, so I am going to raise the priority of #385

admin/api.go
```diff
@@ -11,22 +16,32 @@ import (
...
 	if localConnectorID == "" {
 		panic("must specify non-blank localConnectorID")
 	}
 
+	userRepo := db.NewUserRepo(dbMap)
+	passwordInfoRepo := db.NewPasswordInfoRepo(dbMap)
+	connectorConfigRepo := db.NewConnectorConfigRepo(dbMap)
+	txFactory := db.TransactionFactory(dbMap)
+	opts := manager.ManagerOptions{}
+	userManager := manager.NewUserManager(userRepo, passwordInfoRepo, connectorConfigRepo, txFactory, opts)
```
bobbyrullo: why create the manager here? that seems like a step in the wrong direction; ideally, APIs would talk only to managers not repos.

admin/api.go
```diff
@@ -108,6 +123,33 @@ func (a *AdminAPI) GetState() (adminschema.State, error) {
...
+	if err != nil {
+		return oidc.ClientRegistrationResponse{}, mapError(err)
+	}
+	c, err := a.clientIdentityRepo.New(id, req.Client)
+	if err != nil {
+		return oidc.ClientRegistrationResponse{}, mapError(err)
+	}
+	if req.IsAdmin {
+		// TODO(ericchiang): Put this in a transaction
+		if err := a.clientIdentityRepo.SetDexAdmin(c.ID, req.IsAdmin); err != nil {
```
bobbyrullo: please do that in this PR

admin/api.go
ericchiang: Way fewer arguments to "NewAdminAPI". Makes it way easier to create the AdminAPI.

pkg/gendoc/markdown.go
```diff
@@ -60,7 +60,7 @@ __Version:__ {{ .Version }}
```
bobbyrullo: can you change the comment to accurately reflect the contents of the commit? "updated generated files: documentation, API etc."

admin/api.go
ericchiang: Will make `clientIdentityRepo.New` take a `is_admin` argument and do that there.

admin/api.go
bobbyrullo: ideally, APIs would just get managers and not talk to repos at all....but that's not where we are at yet. Still, no other API does this and I feel like it's a step in the wrong direction. Also, this makes it so you can't set the managers clock which is useful in tests.

integration/admin_api_test.go
```diff
@@ -252,6 +259,73 @@ func TestCreateAdmin(t *testing.T) {
...
+			req := struct {
+				Client  oidc.ClientMetadata `json:"client"`
+				IsAdmin bool                `json:"isAdmin"`
+			}{tt.client, true}
+
+			data, err := json.Marshal(&req)
+			if err != nil {
+				return err
+			}
+			resp, err := f.hc.Post(s.URL+"/api/v1/client", "application/json", bytes.NewReader(data))
```
bobbyrullo: use the admin client to test - that way you know you've implemented the API correctly. See TestGetState for example.

pkg/gendoc/markdown.go
bobbyrullo: also please dont mix generated and handwritten code - the v1.json is in this commit and it should not be.

schema/adminschema/v1.json
```diff
@@ -44,7 +44,26 @@
                   "type": "boolean"
               }
           }
-      }
+      },
+      "ClientCreateRequest": {
```
bobbyrullo: ClientCreate? Or ClientRegister - choose one! :)

schema/adminschema/v1.json
```diff
@@ -44,7 +44,26 @@
...
+          "id": "ClientCreateRequest",
+          "type": "object",
+	      "description": "'client' field is a client registration request as defined by the OpenID Connect dynamic registration spec, and holds fields such as redirect URLs, prefered algorithms, etc. For brevity field names and types of that object have been omitted.",
+          "properties": {
```
bobbyrullo: the client object is necessary or else the generated client doesn't work right?

ericchiang: Seems to work okay.

https://github.com/ericchiang/dex/blob/admin_client_registration/schema/adminschema/README.md#clientcreaterequest

bobbyrullo: I don't mean the generated code. I mean the generated client library.

ericchiang: It creates an empty struct. Any idea how to make it an `interface{}` or something similar?

admin/api.go
bobbyrullo: or just have SetDexAdmin take an tx

ericchiang: Would have to update `New`, and `SetDexAdmin` to take one as well and add a transaction factory to AdminAPI. But either way.

schema/adminschema/v1.json
bobbyrullo: Nope, and that wouldn't be a great API either, taking an `interface{}` when always expects a certain struct.

 Maybe this API doesn't have to have the full OIDC dynamic reg stuff. Or maybe we reimplement it all in that JSON. 

The main thing is that the generated client code should work.

ericchiang: Okay. We need to document that these exist btw; even I didn't know this generated a client. Would be a pity if we did all this work and no one used them.

ericchiang:
@bobbyrullo I've got an update but it requires force pushing (had to undo the `NewAdminAPI` stuff). Want me to open an new PR or force push here?

ericchiang:
Work is here https://github.com/ericchiang/dex/tree/admin_client_registration_2

ericchiang:
Gentle bump

ericchiang:
closing in favor of #392

Issue #337, opened 2016-02-24
client secrets must be base64 encoded in --no-db mode


ericchiang:
Using the SQL code for --no-db mode exposed the fact that the SQL code assumes that client secrets are base64 decodable: https://github.com/coreos/dex/blob/f51125f55538c8fc5a3ad1e780296fd505fb948d/db/client.go#L206

Is there anyway we can get around this?

ericchiang:
Client secrets no longer have encoding issuers

PR #345, opened 2016-02-26
encode pw from client before further processing

fnordahl:
The changes done in #304 makes dex expect that clients send base64
encoded passwords. I believe that this is an assumption we cannot
make.

Changed the server so it base64 encodes the password from the client
before further processing internally in Dex.

Fixed the tests so they create a mockServer with a base64-encoded
password repo and a mockClient that transmits a plain password.

Partially fixes #335

### DIFF
--- a/integration/client_api_test.go
+++ b/integration/client_api_test.go
@@ -12,7 +12,7 @@ import (
 )
 
 func TestClientCreate(t *testing.T) {
-	ci := oidc.ClientIdentity{
+	server_ci := oidc.ClientIdentity{
 		Credentials: oidc.ClientCredentials{
 			ID:     "72de74a9",
 			Secret: base64.URLEncoding.EncodeToString([]byte("XXX")),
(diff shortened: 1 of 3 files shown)


fnordahl:
I guess this is relevant for #337 

I don't know which of the sollutions is right, but this is one possible sollution. :-)

integration/client_api_test.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 func TestClientCreate(t *testing.T) {
-	ci := oidc.ClientIdentity{
+	server_ci := oidc.ClientIdentity{
```
ericchiang: No snake case, all camel :)

server/http.go
```diff
@@ -434,7 +435,7 @@ func handleTokenFunc(srv OIDCServer) http.HandlerFunc {
 			return
 		}
 
-		creds := oidc.ClientCredentials{ID: user, Secret: password}
+		creds := oidc.ClientCredentials{ID: user, Secret: base64.URLEncoding.EncodeToString([]byte(password))}
```
ericchiang: I'm a little confused what the intention of this change is. The issues with the --no-db mode secrets required to be base64 encoded are in the db package.

fnordahl:
D'oh. I was convinced this behaviour was also present when running normally with a real db. Sorry about that. Nothing to see here, move along :-)

cc @ericchiang

Issue #336, opened 2016-02-24
token endpoint basic auth uses wrong decoding strategy


gerson24:
We're working on the integration of our Play/Scala apps with Dex. I'm having problems when I call `/token`,
because when I build the header "Authorization: Basic Y2xpZW50aWQ6c2xka2ZqYXNsa2Rmago=", the library I'm using it first urlencodes the ClientID, and the Secret, separated, and then a base64 of everything.
Example:

`"Authorization: Basic clientID:Secret" --> This will become "Authorization: Basic " + base64encode(urlencode(clientID) + ':' + urlencode(Secret))`

Dex is decoding the base64, but not the urlencode, because it just doesn't expect it.
After doing some research (https://bitbucket.org/connect2id/oauth-2.0-sdk-with-openid-connect-extensions/issues/129/clientsecretbasic-incorrectly-url-encodes),
we think this Java library we're using is encoding it that way, because is specified on the Oauth2 Spec (http://tools.ietf.org/html/rfc6749#section-2.3.1).

This is where Dex should decode the credentials
```go		
		state := r.PostForm.Get("state")

		user, password, ok := r.BasicAuth()
		if !ok {
			log.Errorf("error parsing basic auth")
			writeTokenError(w, oauth2.NewError(oauth2.ErrorInvalidClient), state)
			return
		}

		creds := oidc.ClientCredentials{ID: user, Secret: password}
```
https://github.com/coreos/dex/blob/v0.2.3/server/http.go#L427-L437

If you agree, we're happy to make a PR.

ericchiang:
Yeah, thanks. It does indeed look like it's wrong. It looks like there's the same bug in our client too: https://github.com/coreos/go-oidc/blob/ec2746d2ccb220e81c41b0b0cb2d4a1cc23f7950/oauth2/oauth2.go#L189

I'd like to evaluate what's using our client code first before pushing a patch. This would be a breaking change (all be it a correction), so I just want to be a little cautious.

gerson24:
great, thanks! Let me know when you want me to make the PR

ericchiang:
@gerson24 we'd like to see a fix that accommodates both the correct encoding and the existing encoding for a little bit.

Also our client needs to be updated to do the same. See coreos/go-oidc#54. You're welcome to take this one too if you'd like.

gerson24:
ok, perfect! I'm working on it.

PR #346, opened 2016-02-29
add basic auth url decode

gerson24:
Now it follows oauth2 spec(http://tools.ietf.org/html/rfc6749#section-2.3.1).

Fixes: #336

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -410,7 +410,7 @@ func handleAuthFunc(srv OIDCServer, idpcs []connector.Connector, tpl *template.T
 	}
 }
 
-func handleTokenFunc(srv OIDCServer) http.HandlerFunc {
+func handleTokenFunc(srv OIDCServer, clientIdentityRepo client.ClientIdentityRepo) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


ericchiang:
As discussed in #336 we'd like to have this try both decodings for a little bit (at least for one release) for backward compatibility issues. Can you update your patch to try the correct encoding, and if that doesn't work, try the old one?

gerson24:
Check it now, if you want I can to move this: https://github.com/coreos/dex/blob/master/server/server.go#L413-L422 to http.go and fix the testing in both files

gerson24:
Ok! I'm working on test file

ericchiang:
@gerson24 we're hoping to push a 0.3.0 release today and would very much like to see this in there. Would it be okay to merge as is then add tests in another pr?

adrianlop:
yep, we can do that! thanks @ericchiang

ericchiang:
For last yesterday's release we had a change of heart and pushed the original fix (see #357). Apologies for pushing that ourselves, we were on a tight schedule. Thank you for reporting the issue originally and your work in this PR.

Closing this one since the fix is now in master.

PR #348, opened 2016-02-29
server: add db heatlh checker to server checkers

ericchiang:
Static analysis showed we were never using `db.NewHealthChecker`. Seemed like an oversight.

Add the db health checker to the list of server health checks.


bobbyrullo:
LGTM

PR #353, opened 2016-03-02
vendor: update go-oidc to latest

### LINKED ISSUE #156: don't format NumericDates as floating point in tokens
I got a token with the following claims from dex

`{..."exp":1.444810947e+09, ... "iat":1.444767747e+09 ....}`

These dates happen not to lose any precision, but the scientific notation sure freaked me out for a minute! It might be better to spell these as decimal integers in longform (particularly since you don't save any bytes adding the extra decimal and exponent in this case.)

fnordahl:
coreos/go-oidc#56 addresses floating point NumericDates and has been merged. Update vendored dependency in Dex to make use of it.

Fixes #156


fnordahl:
Some of the tests still fail because of some parts of the code creating integer values for `exp` and `iat` and other parts creating floating point values for `exp` and `iat` and thus the round-trip in the tests does not survive comparison. I'm still looking into the best way to solve it.

fnordahl:
I think I found a way to fix this. (OMG, FINALLY). Please have a look @ericchiang

user/email_verification_test.go
```diff
@@ -127,6 +127,15 @@ func TestEmailVerificationParseAndVerify(t *testing.T) {
 
 	for i, tt := range tests {
 
+		// For compability reasons oidc.NewClaims create Claims with int64 values
+		// jose.ParseJWT will create Claims with float64 values
+		// Convert int64 to float64 so we can compare the Claims
+		tmpexp, _, _ := tt.ev.Claims.Int64Claim("exp")
+		tmpiat, _, _ := tt.ev.Claims.Int64Claim("iat")
+		delete(tt.ev.Claims, "exp")
+		delete(tt.ev.Claims, "iat")
```
ericchiang: I believe these deletes are redundant. `Add` always sets these.

ericchiang:
This is definitely one way of doing this. It leaves a lot of our `map[string]interface{}` usages unresolved, but it does fix a lot of problems by updating.

@bobbyrullo what say you?

user/email_verification_test.go
fnordahl: You're right. New version without the deletes on it's way.

bobbyrullo:
looking!

bobbyrullo:
@fnordahl - thanks for the fixes. The one thing I didn't like was hacking the tests so that the claims end up comparing right - seems like we're not actually testing the real behavior. Take a look at what I did here:

https://github.com/bobbyrullo/dex/commits/bobby_353

Specifically, [this commit](https://github.com/bobbyrullo/dex/commit/a4671b115dcd6470b483d4f34aa35de0b90335a7), which introduces the `timeClaimsToInt` function.

fnordahl:
@bobbyrullo - massaging tests to pass is of course one of the infamous deadly sins. However, in this case we do get (and should expect?) two different claims from different parts of the system. Thank you for the pointer to how you went at it, I'll have a look at it and see if I can make it fit and use it to improve my patches. Thx!

bobbyrullo:
@fnordahl - any progress on this? If you don't have the time to work on it, I can take it over -we're going to need it very soon in dex

fnordahl:
@bobbyrullo thank you for following up. I've been a bit busy, but I would like to take a shot. I will pick it up in a few hours. I'll let you know the outcome.

fnordahl:
I did not notice at first that you actually did fix the tests in-place in the commit you linked to, I thought you were referring to a example from somewhere else. No need for me to redo this. I have just cherry-picked your commit to this PR. I also updated the vendor commit to actually point to the latest version of go-oidc as it has been updated since I last updated this PR.

Thanks @bobbyrullo !

bobbyrullo:
Heh no problem - thanks @fnordahl !

ericchiang:
Terrific work!

PR #369, opened 2016-03-09
move outside dependencies to ./vendor directory

ericchiang:
Move `./Godeps/_workspace/src` to `./vendor` to utilize the Go tool's vendoring mechanism (which defaults to on in 1.6 and will be standard in 1.7)

As a result we now require go1.5+ and our build scripts now error out for earlier versions.


bobbyrullo:
LGTM

PR #375, opened 2016-03-15
Fix response_type missing param

### LINKED ISSUE #370: Dex doesn't return an error when required response_type param is empty or missing
According to OIDC spec in all flows parameter response_type is required ([Authorization code flow authentication request](http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest)). Nowadays dex doesn't return any error when this parameter is missing and works as Authorization code flow.

rsoletob:
This commit fix problem with response_type param, which is required according to OIDC spec, when it is missing.
At now, when connector_id url query param is not set, connector view use response_type that client requested instead of default "code".

Fixes #370

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -255,7 +255,7 @@ func renderLoginPage(w http.ResponseWriter, r *http.Request, srv OIDCServer, idp
 
 		v := r.URL.Query()
 		v.Set("connector_id", idpc.ID())
-		v.Set("response_type", "code")
+		v.Set("response_type", q.Get("response_type"))
 		link.URL = httpPathAuth + "?" + v.Encode()
 		td.Links = append(td.Links, link)
 	}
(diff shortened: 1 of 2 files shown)


ericchiang:
Hmm, looking into why travis CI isn't kicking off a job.

ericchiang:
Quick question, does https://github.com/Tecsisa/dex/blob/671c497513260ed123fe9a39f4224444d1e0abee/server/http.go#L382 not already prevent this?

rsoletob:
Thanks for your comments @ericchiang, I will change it. 

> Quick question, does https://github.com/Tecsisa/dex/blob/671c497513260ed123fe9a39f4224444d1e0abee/server/http.go#L382 not already prevent this?

Yes, it does. I thought it could be better two errors, an error when it's missing and other when it's unsupported, but maybe just one check is the best solution.

> Do we need this struct now?

I think it isn't necessary now, I will change it.

server/http.go
```diff
@@ -273,77 +273,88 @@ func handleAuthFunc(srv OIDCServer, idpcs []connector.Connector, tpl *template.T
...
-				writeAuthError(w, oauth2.NewError(oauth2.ErrorInvalidRequest), acr.State)
+				log.Errorf("Request provided unregistered redirect URL: %s", rURL)
+				writeAuthError(w, oauth2.NewError(oauth2.ErrorInvalidRequest), state)
 				return
 			case (client.ErrorNoValidRedirectURLs):
-				log.Errorf("There are no registered URLs for the requested client: %s", acr.RedirectURL)
-				writeAuthError(w, oauth2.NewError(oauth2.ErrorInvalidRequest), acr.State)
+				log.Errorf("There are no registered URLs for the requested client: %s", rURL)
+				writeAuthError(w, oauth2.NewError(oauth2.ErrorInvalidRequest), state)
 				return
```
ericchiang: It looks like we're missing a default case. can you add one?

rsoletob: Yes of course

ericchiang:
LGTM

PR #399, opened 2016-04-08
replace godep with glide

### LINKED ISSUE #387: "godep save ./..." fails with vendor directory
Now that we use the `vendor` directory the go tool `./...` expression detects all packages in the `vendor` directory. `godep save ./...`, our recommended way of adding new packages, attempts to save all the dependencies of the vendored dependencies. Blah.

Also note that `go list ./... | grep -v vendor` fails so we can't use that ATM to filter out packages.

```
$ go list ./... | grep -v './vendor'
can't load package: /home/eric/src/github.com/coreos/dex/schema/generator_import.go:10:8: import "google.golang.org/api/google-api-go-generator" is a program, not an importable package
```

Either figure out how to get around this with godep and update the dev docs (for example what [rkt does](https://github.com/coreos/rkt/blob/e56895779706097a5fdfc7099003d105cc325638/scripts/godep-save)) or consider being a guinea pig for coreos/docs#775

ericchiang:
Please review the first commit which updates the documentation and adds the glide files.

The second commit is huge as glide does not remove _test.go files and testdata.

See coreos/docs#775 for discussion.

Closes #387


chancez:
@ericchiang We can use `https://github.com/sgotti/glide-vc` to remove some things (we can't use only-go because it will strip files needed by cgo), but it would help with test files, and packages which are not part of the direct import-tree.

ericchiang:
Just to follow up from outside discussions. We like glide-vc but don't like adding an additional dependency in order to work with the `vendor` directory. If this is functionality is important then glide should implement it or we should consider a different tool.

There's some discussion here https://github.com/Masterminds/glide/issues/166

ericchiang:
I'd also like to note that I'm fine with glide not removing unused sub-packages.

bobbyrullo:
how does one do the equivalent of `godep go {build|test}` ?

ericchiang:
Ensure you have either Go 1.6 or Go 1.5 with `GO15VENDOREXPERIMENT=1`. Then just use the go tool. Slice glide puts everything in `./vendor` all dex code will build using that directory.

bobbyrullo:
LGTM

sym3tri:
:tada:

PR #401, opened 2016-04-09
Update k8s example configuration to use 1.2 k8s features

andrewstuart:
- Aggregate yaml files
- Use `kubectl apply` in README
- Update to use Deployments
- Use ConfigMap and Volume for connector json file
- Use Ingress + host name instead of NodePort + hard-coded IP
- Remove use of default namespace in postgres URLs.

Let me know if there are any adjustments you'd like and I'm sure I'd be glad to make them. :-)


contrib/k8s/dex-worker.yaml
```diff
@@ -0,0 +1,113 @@
...
+  template:
+    metadata:
+      labels:
+        app: dex
+        role: worker
+    spec:
+      containers:
+          - image: quay.io/coreos/dex
+            name: dex-worker
+            env:
```
chancez: I think maybe you could replace the `--key-secrets` flag below with an env `DEX_WORKER_KEY_SECRETS`, and then use the downward API to pull the env var from a secret file:

```
- name: DEX_WORKER_KEY_SECRETS
  valueFrom:
    secretKeyRef:
      name: dex
      key: key-secrets
```

andrewstuart: For that, I was actually thinking that it would be best to be read in directly by the program. Otherwise it's a bit less secret, as the `$(cat key)` will show up in `ps aux` from the host or inside the container, and environment can be viewed with `ps auxe`. The only safe way is probably to read and then delete the secret directly from disk.

chancez: I wouldn't really worry about that in this PR. Environment variables while not ideal for these kinds of things, this is no worse than the current `cat` approach. 

There is also an open issue for making dex take a file for all of these configuration needs already and fixing up the example k8s manifest to use that in the future can be done in a separate PR once that lands.

contrib/k8s/README.md
```diff
@@ -27,17 +26,14 @@ dex needs a secret key for encrypting private keys in the database. These can be
 
 [k8s-secrets]: http://kubernetes.io/v1.0/docs/user-guide/secrets.html
 
-```
-kubectl create -f dex-secrets.yaml
-```
+The secret will be created as part of the next command, from `dex-overlord.yaml`
```
ericchiang: Would it be possible to add an example here of creating the secret? For example:

```
dd if=/dev/random bs=1 count=32 2>/dev/null | base64 | tr -d '\n' > key-secrets
kubectl create secret generic dex --from-file=./key-secrets
```

contrib/k8s/dex-overlord.yaml
```diff
@@ -67,10 +81,18 @@ kind: ConfigMap
...
-  local-connector.json: |
+  connector.json: |
     [
       {
-        "type": "local",
-        "id": "local"
+        "id": "local",
+        "type": "local"
       }
+      # {
```
ericchiang: Unfortunately connectors are json not yaml so this comment breaks the deployment.

```
$ kubectl exec $DEX_OVERLORD_POD -- /opt/dex/bin/dexctl --db-url='postgres://postgres@dex-postgres:5432/postgres?sslmode=disable' set-connector-configs '/etc/dex-connectors/connector.json'
Unable to decode connector configs: invalid character '#' after array element
error: error executing remote command: Error executing command in container: Error executing in Docker Container: 1
```

contrib/k8s/README.md
chancez: Be careful using `base64` in examples, it's usage varies based on darwin and linux.

ericchiang: yep. That's why the `tr`'s in there https://github.com/coreos/dex/pull/193

contrib/k8s/dex-overlord.yaml
andrewstuart: Argh. Though that would be ignored by kubectl parsing the yaml -- didn't think about the fact it was a literal block and would make it into the configmap value.

andrewstuart:
Haha, looks like the travis node got shut down while running the tests. :-P

ericchiang:
@andrewstuart rerunning :)

Testing this change now.

Also please not that we normally require that you submit PRs from a branch that's not `master`. See https://github.com/coreos/dex/blob/master/CONTRIBUTING.md#contribution-flow. It makes it easier on your code reviewer because they can checkout your changes without rewriting the branch name. No biggie for this one.

ericchiang:
This is great.

Only one question. Would it be appropriate to include creating the ingress controller node in these instructions? It's really important in the final setup and it's seem hard to start a guide with "goto this other document."

contrib/k8s/dex-overlord.yaml
```diff
@@ -0,0 +1,106 @@
...
+    spec:
+      containers:
+          - image: quay.io/coreos/dex
+            name: dex-overlord
+            env:
+              - name: DEX_OVERLORD_DB_URL
+                value: postgres://postgres@dex-postgres:5432/postgres?sslmode=disable
+              - name: DEX_OVERLORD_ADMIN_LISTEN
+                value: http://0.0.0.0:5557
+              - name: DEX_SECRET
```
chancez: If you rename this to `DEX_OVERLORD_KEY_SECRETS` you can avoid passing the flag below.

contrib/k8s/dex-worker.yaml
```diff
@@ -0,0 +1,104 @@
...
+                value: http://dex.example.com
+                # enable https if you have configured your Ingress with TLS
+                # value: https://dex.example.com
+              - name: DEX_WORKER_DB_URL
+                value: postgres://postgres@dex-postgres:5432/postgres?sslmode=disable
+              - name: DEX_WORKER_EMAIL_CFG
+                value: /opt/dex/email/emailer.json
+              - name: DEX_WORKER_LISTEN
+                value: http://0.0.0.0:5556
+              - name: DEX_SECRET
```
chancez: If you rename this to `DEX_WORKER_KEY_SECRETS` you can avoid passing the flag below.

contrib/k8s/dex-overlord.yaml
andrewstuart: Okay, I had a feeling that was what you meant in an earlier comment, but my cursory `git grep 'DEX_OVERLORD_KEY_SECRETS'` only returned a single usage in the contrib folder, so I wasn't positive that'd be picked up if I just set it.

andrewstuart:
Sorry about the wrong PR target; I often forget to read the contributing.md files, but I've given it a once-over now. :-)

andrewstuart:
I added some quick documentation on setting up an ingress controller directly from the [kubernetes/contrib](/kubernetes/contrib) raw files, which should hopefully mean less upkeep than pulling a yml file into the repo.

contrib/k8s/README.md
```diff
@@ -27,17 +37,23 @@ dex needs a secret key for encrypting private keys in the database. These can be
...
-```
-kubectl create -f dex-secrets.yaml
+The secret will be created as part of the next command, from `dex-overlord.yaml`
+
+If you would like to set up your own secret, you can do the following:
+
+```bash
+# For a real secret (i.e. not checked into this public repo), run the following
+# and comment out the secret in the dex-overlord.yaml file.
+secret=$(dd if=/dev/random bs=1 count=32 2>/dev/null | base64 | tr -d '\n' > key-secrets)
```
ericchiang: The internal command doesn't print anything. You need to remove `> key-secrets` so it prints the secret.

andrewstuart: Yargh, good catch. :-)

ericchiang:
Final notes

* Creating the secret still fails. Unfortunately the secret contains an equals sign so `kubectl` fails.

```
kubectl create secret generic dex --from-literal=key-secrets=$secret
error: invalid literal source key-secrets=BnfYR3tjNyGWppKJmfK2babCqg9UgP5AEMRNVuysgY8=, expected key=value
```

This worked :)

```
echo -n "$secret" | kubectl create secret generic dex --from-file=key-secrets=/dev/stdin
```

But an intermediate file is probably easier.

* We should enable registration for the demo, or the end user won't be able to do much.

```diff
diff --git a/contrib/k8s/dex-worker.yaml b/contrib/k8s/dex-worker.yaml
index 316be43..1b63942 100644
--- a/contrib/k8s/dex-worker.yaml
+++ b/contrib/k8s/dex-worker.yaml
@@ -27,6 +27,8 @@ spec:
                 value: /opt/dex/email/emailer.json
               - name: DEX_WORKER_LISTEN
                 value: http://0.0.0.0:5556
+              - name: DEX_WORKER_ENABLE_REGISTRATION
+                value: "true"
               - name: DEX_WORKER_KEY_SECRETS
                 valueFrom:
                   secretKeyRef:
```

* By default `coreos-kubernetes` has the API server listen on `:443`. This conflicts with the ingress pod, so it's not scheduled until you change the port.

@andrewstuartI I'd like to get this merged so I'm not worried about fixing the ingress port conflict. Can you please update this to fix the secrets and registration and we'll merge it in. Sorry for all the nits!

andrewstuart:
Alright, pushed. :-)

Haha, no worries about the nits. I should have tried running my code (I did this time!) to make sure it ran.

I swapped out the secret creation environment variable for a pipe, with a newline to keep it readable. That feels more secure anyway as there's even less chance for leakage.

Let me know if there's anything else. :-)

ericchiang:
lgtm

ericchiang:
Damn, tried to squash that. One sec.

ericchiang:
Okay, sorry. Quick report of what just happened:

* I accidentally merged all of your commits without squashing them. Attempted to use the new GitHub feature but fat fingered [this button](https://cloud.githubusercontent.com/assets/143418/14119749/7756ff60-f5a4-11e5-8597-12dcbb363746.png).
* I forced pushed to master to revert that merge since it had a dozen or so commits.
* I have no idea how to tell GitHub that this PR is now "unmerged."

Please open a PR with all of these commits squashed into one. With any luck I'll not mess that one up.

andrewstuart:
Sounds good :-)

PR #393, opened 2016-04-05
remove passwordInfos from local connector

ericchiang:
In --no-db mode, load passwords from the users file instead of the
connectors file. This allows us to remove the password infos field
from the local connector and stop loading them during connector
registration, a case that was causing panics when using a real
database (see #286).

Fixes #286
Closes #340

PR #402, opened 2016-04-11
server: add more validation to --no-db static file parsing

ericchiang:
In #393 the format of the static user file in --no-db mode changed.
However, the old format loads without error, which has caused
issues for developers with existing user files.

Add an explicit check to ensure the file is not using the old
format. If they are, print a better error message.

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -184,6 +184,10 @@ func loadUsersFromReader(r io.Reader) (users []user.UserWithRemoteIdentities, pw
 		user.User
 		Password         string                `json:"password"`
 		RemoteIdentities []user.RemoteIdentity `json:"remoteIdentities"`
+
+		// The old format stored all user data under the "user" key.
+		// Attempt to detect that, and print an better error.
+		OldUserFields map[string]string `json:"user"`
 	}
 	if err := json.NewDecoder(r).Decode(&configUsers); err != nil {
 		return nil, nil, err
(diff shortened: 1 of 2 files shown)


server/config.go
```diff
@@ -193,6 +197,10 @@ func loadUsersFromReader(r io.Reader) (users []user.UserWithRemoteIdentities, pw
 	pwis = make([]user.PasswordInfo, len(configUsers))
 
 	for i, u := range configUsers {
+		if u.OldUserFields != nil {
+			return nil, nil, fmt.Errorf("Static user file is using an outdated format. Please refer to example in static/fixtures.")
```
chancez: Perhaps move this inline error to a named variable?

server/config_test.go
```diff
@@ -15,6 +15,8 @@ func TestLoadUsers(t *testing.T) {
 		expUsers []user.UserWithRemoteIdentities
 		// userid -> plaintext password
 		expPasswds map[string]string
+
+		wantErr bool
```
chancez: perhaps change to `expectedErr error` and compare to nil instead, or a specific error?

server/config.go
ericchiang: Why declare a sentinel error for an error that's used once?

server/config_test.go
ericchiang: Because I want this case to cause an error in `loadUsersFromReader` I don't actually care what error it causes.

Without the check above this test doesn't pass, with it it does. Do I need to peg the test to the exact error?

bobbyrullo: I think it's ok as is: if you *were* handling errors differently, than it would be important.

bobbyrullo:
LGTM. Should we have a plan to remove OldUserFields at some point?

ericchiang:
Either way. Does anyone actually look at `// TODO` comments? :)

bobbyrullo:
well no, but we should and we will one day. Doesn't have to be a TODO, could be an issue, or a mark on your calendar or whatever. I would just like to see that not be around forever.

ericchiang:
Once saw a package that turned TODOs into errors if they weren't resolved in a certain amount of time. Maybe we need something like that!

But yeah, I've added a reminder to my calendar to remove this two months from now.

bobbyrullo:
> Once saw a package that turned TODOs into errors if they weren't resolved in a certain amount of time. Maybe we need something like that!
>
That would be sick.
> But yeah, I've added a reminder to my calendar to remove this two months from now.
Brilliant

LGTM

Issue #400, opened 2016-04-08
Error messages aren't very helpful/specific


chancez:
Env:
```
DEX_WORKER_CLIENTS=.dexconfig/clients.json
DEX_WORKER_CONNECTORS=.dexconfig/connectors.json
DEX_WORKER_EMAIL_CFG=.dexconfig/emailer.json
DEX_WORKER_USERS=.dexconfig/users.json
DEX_WORKER_EMAIL_TEMPLATES=templates/dex/emails
DEX_WORKER_HTML_ASSETS=templates/dex/html
DEX_WORKER_NO_DB=true
```

Getting errors like this:

```
FATAL: Unable to build Server: failed to create client identity repo: illegal base64 data at input byte 4
```

This really meant "client secrets must be base64 encoded within the `connectors.json`"

```
FATAL: Unable to build Server: remote identity already in use for another user
```

This wasn't too bad, but it probably could have told me what ID it was that caused this.

```
FATAL: Unable to build Server: invalid ID`
```

This was because the format of the `users.json` changed, and the JSON no longer had a key called `users` within the list of users.

Ideally these get validated before passing them to the server, and provide better context on what was wrong, because it isn't clear what configuration was wrong, or what value specifically it was referring to.

ericchiang:
Going to close this as these specific error messages were cleaned up. @chancez please open a new issue with specific errors if there are any else you want to be addressed.

PR #403, opened 2016-04-11
db: print better error messages for invalid input

ericchiang:
When client secrets are not base64 encoded, print an error message
that's not a generic base64 decode error:

	client secrets must be base64 decodable. See issue #337.
	Please consider replaceing "secret" with "c2VjcmV0"

When a user file is missing a mandatory field print an error message.

	Unable to build Server: user elroy-foo is missing email field

For #400


ericchiang:
cc @chancez how does this and #402 look to you?

chancez:
Looks pretty solid. Haven't tested myself, but it seems like it covers most of those errors I encountered.

PR #405, opened 2016-04-13
connector_oauth2.go: use embedded HTTPClient

### LINKED ISSUE #395: connectors: allow for retrieving OAuth2 client
Given an OAuth2 connector (much like the github connector) to an OAuth2 service _that doesn't use proper TLS certificates_ (self-signed or a CA not trusted by the system cert pool), you might have the following bits in the connector config:

```go
type MyConnectorConfig struct {
	ID           string `json:"id"`
	ClientID     string `json:"clientID"`
	ClientSecret string `json:"clientSecret"`
	URL      string `json:"URL"`
	Cert     string `json:"cert"`
	SkipVerify   bool   `json:"insecureSkipVerify"`
}
```

accompanied by

```go
func newMyConnector(clientID, clientSecret, url, cert string, skipVerify bool, cbURL string) (oauth2Connector, error) {
	config := oauth2.Config{
		Credentials: oauth2.ClientCredentials{ID: clientID, Secret: clientSecret},
		AuthURL:     url + myAuthPath,
		TokenURL:    url + myTokenPath,
		RedirectURL: cbURL,
	}

	var client *http.Client
	if chefCert != "" {
		caPool := x509.NewCertPool()
		caPool.AppendCertsFromPEM([]byte(cert))
		tlsConfig := &tls.Config{
			RootCAs: caPool,
		}

		tr := &http.Transport{TLSClientConfig: tlsConfig}
		client = &http.Client{Transport: tr}
	} else if skipVerify {
		tr := &http.Transport{TLSClientConfig: &tls.Config{
			InsecureSkipVerify: true,
		}}
		client = &http.Client{Transport: tr}
	} else {
		client = http.DefaultClient
	}

	cli, err := oauth2.NewClient(client, config)
	if err != nil {
		return nil, err
	}

	return &myOAuth2Connector{
		clientID:     clientID,
		clientSecret: clientSecret,
		client:       cli,
		url:      url,
	}, nil
}
```

This will work fine _until_  the callback handler uses `http.DefaultClient` again [_here_](https://github.com/coreos/dex/blob/master/connector/connector_oauth2.go#L87):

```go
		ident, err := c.conn.Identity(newAuthenticatedClient(token, http.DefaultClient))
		if err != nil {
			log.Errorf("Unable to retrieve identity: %v", err)
```

...at which point TLS verification fails.

srenatus:
Fixes #395.

(While I tend to agree with https://github.com/coreos/dex/issues/395#issuecomment-209215114, this seems to be the least intrusive way to get to where I want...)

Curious if this works on Travis, because locally I seem to have botched up something around the dependencies...

### DIFF
--- a/connector/connector_oauth2.go
+++ b/connector/connector_oauth2.go
@@ -84,7 +84,7 @@ func (c *OAuth2Connector) handleCallbackFunc(lf oidc.LoginFunc, errorURL url.URL
 			redirectError(w, errorURL, q)
 			return
 		}
-		ident, err := c.conn.Identity(newAuthenticatedClient(token, http.DefaultClient))
+		ident, err := c.conn.Identity(newAuthenticatedClient(token, c.conn.Client().HttpClient()))
 		if err != nil {
 			log.Errorf("Unable to retrieve identity: %v", err)
 			q.Set("error", oauth2.ErrorUnsupportedResponseType)
--- a/glide.lock
+++ b/glide.lock
@@ -1,10 +1,10 @@
-hash: b0a1c1b400f077dac572450cb9602aabba62d9b7c6fc8ef426c2093de7621e8c
-updated: 2016-04-08T11:43:54.191409294-07:00
+hash: 44508a23934668c817df4d2f5d6c3781d25a7896270aae088b9c7efbbebad073
+updated: 2016-04-14T10:21:19.730407106+02:00
 imports:
 - name: github.com/andybalholm/cascadia
   version: 6122e68c2642b7b75c538a63b15168c6c80fb757
 - name: github.com/coreos/go-oidc
-  version: 6039032c0b15517897116d333ead8edf38792437
+  version: dfe8dca68c3c087d3e147aeef30553530e2366bf
   subpackages:
   - http
   - jose
--- a/glide.yaml
+++ b/glide.yaml
@@ -5,7 +5,7 @@ import:
 - package: github.com/andybalholm/cascadia
   version: 6122e68c2642b7b75c538a63b15168c6c80fb757
 - package: github.com/coreos/go-oidc
-  version: 6039032c0b15517897116d333ead8edf38792437
+  version: dfe8dca68c3c087d3e147aeef30553530e2366bf
   subpackages:
   - http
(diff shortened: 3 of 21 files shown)


ericchiang:
Please see our documentation on updating vendored dependencies: https://github.com/coreos/dex/blob/master/Documentation/dev-guide.md#updating-an-existing-package

ericchiang:
Note that we recently switched to https://github.com/Masterminds/glide. I apologize if you've already familiarized yourself with godep, but we've had major issues with it in the past.

srenatus:
Oooh I hadn't noticed that, sorry. I'll update this, of course.

srenatus:
Hmm this would fix #353, too, but without having fixed the tests :confused: 
And https://github.com/coreos/dex/pull/353/commits/dafdfc978681d3e790db205a2f8f02c081af9241 doesn't really solve all problems yet, so... I guess I'll just hold this one until #353 is done with? 

@fnordahl: are you still on it? Feel free to pick c0c46f0 to shortcut the glide-based dependency update, I'll just rework this branch afterwards. :)

ericchiang:
@srenatus we removed a couple fields from the `oidc/http` subpackage to prevent normal packages from importing `net/http/httptest` https://github.com/coreos/go-oidc/pull/59. This will require updating some of dex's code to account for these missing fields.

I'll try to do that update in the near future if you don't get to it first.

Otherwise, please see my comment here https://github.com/coreos/dex/issues/395#issuecomment-209215114. That would be a fix you can do without pulling in your change to coreos/go-oidc, and in my opinion is the preferred fix.

fnordahl:
@srenatus I still have an intention to get #353 in. I have not chased down or solved how ti fix the remaining issues yet though. I will try to re-iterate and see if I can come up with a sollution soon.

srenatus:
@ericchiang ok, I'll just update this in accordance with https://github.com/coreos/dex/issues/395#issuecomment-209215114 then

@fnordahl ok, thanks!

srenatus:
Didn't keep my promise... sorry! It's obsolete now, I suppose, gc'ing.

PR #409, opened 2016-04-14
tests: add HandleTokenFunc test

### LINKED ISSUE #408: HandleTokenFunc needs test

adrianlop:
this PR adds test for HandleTokenFunc

Fixes #408


server/http_test.go
```diff
@@ -332,6 +332,125 @@ func TestHandleAuthFuncResponsesMultipleRedirectURLs(t *testing.T) {
...
+		hdlr := handleTokenFunc(fx.srv)
+		w := httptest.NewRecorder()
+
+		req, err := http.NewRequest("POST", "http://example.com/token", strings.NewReader(tt.query.Encode()))
+		if err != nil {
+			t.Errorf("unable to create HTTP request, error=%v", err)
+		}
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		basicAuthEncode := base64.URLEncoding.EncodeToString([]byte(tt.creds))
+		req.Header.Set("Authorization", "Basic "+basicAuthEncode)
```
ericchiang: Please use [`http.Request.SetBasicAuth`](https://golang.org/pkg/net/http/#Request.SetBasicAuth)

server/http_test.go
```diff
@@ -332,6 +332,125 @@ func TestHandleAuthFuncResponsesMultipleRedirectURLs(t *testing.T) {
...
+
+		sid, err := fx.sessionManager.NewSession("local", "XXX", "", testRedirectURL, "", true, []string{"openid"})
```
ericchiang: Can you add a brief comment here describing what the session manager is doing?

server/http_test.go
```diff
@@ -332,6 +332,125 @@ func TestHandleAuthFuncResponsesMultipleRedirectURLs(t *testing.T) {
...
+	tests := []struct {
+		query    url.Values
+		creds    string
+		wantCode int
+	}{
+		// bad grant_type
+		{
+			query: url.Values{
+				"grant_type": []string{"invalid!"},
+				"state":      []string{""},
```
ericchiang: Is "state" part of the token request? https://tools.ietf.org/html/rfc6749#section-4.1.3

ericchiang:
Some inline comments, but other than that lgtm.

server/http_test.go
```diff
@@ -332,6 +332,125 @@ func TestHandleAuthFuncResponsesMultipleRedirectURLs(t *testing.T) {
 	}
 }
 
+func TestHandleTokenFunc(t *testing.T) {
+
+	fx, err := makeTestFixtures()
+	if err != nil {
+		t.Errorf("could not run test fixtures")
```
ericchiang: Need to `t.Fatal` here since the test can't continue if this fails. Also please print the error:

```
t.Fatalf("could not run test fixtures: %v", err)
```

server/http_test.go
```diff
@@ -332,6 +332,125 @@ func TestHandleAuthFuncResponsesMultipleRedirectURLs(t *testing.T) {
...
+		},
+	}
+
+	for i, tt := range tests {
+		hdlr := handleTokenFunc(fx.srv)
+		w := httptest.NewRecorder()
+
+		req, err := http.NewRequest("POST", "http://example.com/token", strings.NewReader(tt.query.Encode()))
+		if err != nil {
+			t.Errorf("unable to create HTTP request, error=%v", err)
```
ericchiang: Need a `continue` statement here so this skips to the next test case on an error. If not `req` will be nil and this test will produce panics.

server/http_test.go
```diff
@@ -332,6 +332,125 @@ func TestHandleAuthFuncResponsesMultipleRedirectURLs(t *testing.T) {
...
+		}
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		basicAuthEncode := base64.URLEncoding.EncodeToString([]byte(tt.creds))
+		req.Header.Set("Authorization", "Basic "+basicAuthEncode)
+
+		sid, err := fx.sessionManager.NewSession("local", "XXX", "", testRedirectURL, "", true, []string{"openid"})
+		if err != nil {
+			t.Errorf("case %d: cannot create session, error=%v", i, err)
```
ericchiang: Same comment, need a `continue`. You could also have function to do this, so you don't have to do the

```
if err != nil {
    t.Errorf("case %d: foobar: %v", i, err)
    continue
}
```

dance

```
setSession := func() error {
    sid, err := fx.sessionManager.NewSession("local", "XXX", "", testRedirectURL, "", true, []string{"openid"})
    if err != nil {
        return fmt.Errorf("cannot create session: %v", err)
    }
    // etc
}

if err := setSession(); err != nil {
    t.Errorf("csae %d: %v", i, err)
    continue
}
```

adrianlop:
@ericchiang thanks for your comments! I reviewed the PR.

ericchiang:
lgtm. Thanks!

PR #414, opened 2016-04-19
contrib/k8s: update to use 1.2 features

andrewstuart:
- Consolidate files
- Update to Deployments
- Use Ingress, add ingress controller help in README
- Remove hardcoded namespace in postgres URI
- Remove hardcoded IP addresses (replaced with Ingress resources)
- Add readinessProbes

Continuation of #401


ericchiang:
Thanks, this is terrific!

PR #416, opened 2016-04-19
emailVerifyFunc: redirect to redirectURL with valid token

### LINKED ISSUE #229: EmailVerification link should take you to redirectURL with valid access token

### LINKED ISSUE #415: handleEmailVerifyFunc needs test

adrianlop:
tests: add emailVerifyFunc test

Fixes #229
Fixes #415


server/email_verification.go
```diff
@@ -236,6 +328,12 @@ func handleEmailVerifyFunc(verifiedTpl *template.Template, issuer url.URL, keysF
 			}
 			return
 		}
+
+		qr := cbURL.Query()
+		qr.Set("code", code)
+		qr.Set("state", ses.ClientState)
```
ericchiang: Correct me if I'm wrong but I think this state won't be valid for the client being redirected to. Since the clients must generate this state string, they should flat out reject ones they haven't created. So dex can't create a valid state.

server/email_verification.go
```diff
@@ -215,6 +217,96 @@ func handleEmailVerifyFunc(verifiedTpl *template.Template, issuer url.URL, keysF
 			return
 		}
 
+		aud, ok, err := ev.Claims.StringClaim("aud")
+		if !ok {
+			execTemplateWithStatus(w, verifiedTpl, emailVerifiedTemplateData{
+				Error:   "Unable to get 'aud' from token",
```
ericchiang: Since there's a huge amount of duplicated error handling here, you'll definitely want to break this into a separate function to get rid of the

```
execTemplateWithStatus(w, verifiedTpl, emailVerifiedTemplateData{

}, code)
```
Repetition.

ericchiang:
```
server/email_verification_test.go:12: imported and not used: "fmt"
```

You should definitely use https://godoc.org/golang.org/x/tools/cmd/goimports :)

adrianlop:
thanks for suggesting goimports, works like a charm :)

server/email_verification.go
adrianlop: yeah but this state is the one that the client generated, Dex saves this state in the session, using NewSession func. I'm reading that one.
If the user takes too long to verify the email, then maybe the state could expire, but Dex will do his job and verify the email+redirect, but it's on the client's hand to check if the state is the correct one, right?

server/email_verification.go
adrianlop: thanks, revamped!

server/email_verification.go
ericchiang: NewSession is used to create states for federated services, when dex is acting as a client of an upstream service (e.g. when it's federating to accounts.google.com). Notably you're missing the third argument when calling NewSession, the state dex go from it's client.

Those two states, the one dex got from it's client, and the one dex sends to the upstream service, are not interchangeable.

Here dex is attempting to send the user to dex's client by emulate an Authorization Response. That means redirecting back to dex's client with a code and a state. However state is defined as

> state
>         REQUIRED if the "state" parameter was present in the client
>         authorization request.  The exact value received from the
>         client.

https://tools.ietf.org/html/rfc6749#section-4.1.2

However dex never got a state token from the client, it has no way to fill this field correctly. dex's client MUST always reject a state token that it didn't explicitly create.

Does this make sense?

server/email_verification.go
```diff
@@ -190,52 +193,110 @@ type emailVerifiedTemplateData struct {
...
+		if !ok {
+			runtpl("Unable to get 'sub' from token", "That was not a verifiable token.", http.StatusBadRequest)
+			return
+		}
+
+		callback, ok, err := ev.Claims.StringClaim("http://coreos.com/email/verification-callback")
+		if !ok {
+			runtpl("Unable to get callback URL from token", "That was not a verifiable token.", http.StatusBadRequest)
+			return
+		}
```
bobbyrullo: Aren't a lot of these things already checked in ParseAndVerifyEmailVerificationToken ?

adrianlop: indeed, but we need to extract these params here too using `StringClaim`, because `ParseAndVerifyEmailVerificationToken` func does not return them, so I think it's ok if we check errors here again. what do you think? do you prefer returning these params in the `ParseAndVerifyEmailVerificationToken` func?

server/email_verification.go
adrianlop: ok, now I get it, thanks. we have a problem here :P
what would you do in this piece of code then?

ericchiang: The email invitation API call would require code, scope, and nonce parameters. Basically all the parameters the auth endpoint takes on initial request.

adrianlop: ok, i've updated the PR.

adrianlop:
@ericchiang hi there, can you review this PR please? it's been a while, so if it's not going to be merged, you can close it.

ericchiang:
Sorry can you describe how this redirects with a valid token? I would think that the only way to do so would be for the actual API which sent an email to contain a state argument.

server/email_verification.go
```diff
@@ -175,7 +200,7 @@ func handleVerifyEmailResendFunc(
 			}
 		}
 
-		_, err = emailer.SendEmailVerification(usr.ID, clientID, *redirectURL)
+		_, err = emailer.SendEmailVerification(usr.ID, clientID, *redirectURL, state, nonce, scopes)
```
ericchiang: I don't think this is the right place to grab these value from. Sorry am trying to track the email sending code.

server/email_verification.go
```diff
@@ -191,52 +216,128 @@ type emailVerifiedTemplateData struct {
...
 		if err != nil {
-			execTemplateWithStatus(w, verifiedTpl, emailVerifiedTemplateData{
-				Error:   "Bad Email Verification Token",
-				Message: "That was not a verifiable token.",
-			}, http.StatusBadRequest)
+			runtpl("Bad Email Verification Token", "That was not a verifiable token.", http.StatusBadRequest)
+			return
+		}
+
+		aud, ok, err := ev.Claims.StringClaim("aud")
```
ericchiang: Regardless of whatever happens to this PR, we need to figure out a way to reduce the volume of if statement in this block (possibly by defining a helper func).

ericchiang:
Going to have to look into the actual code to catch up on how emails are actually sent in dex. Will follow up later today when I dig through the code a bit more.

server/email_verification.go
adrianlop: using a helperFunc you'll have a problem: `return` statement will exit helper func, not the main func.

alternative: I tried using `goto` but you cannot reference a label outside the helperFunc, so I ran out of alternatives :P

server/email_verification.go
adrianlop: where would you get them?

PR #417, opened 2016-04-19
schema: remove generator_import file used for godep

ericchiang:
generator_import.go existed to trick godep to vendoring a main
package. Since moving to glide, this file is no longer needed.


sym3tri:
LGTM

PR #424, opened 2016-04-21
contrib/k8s: YAML brevity via references, minimum resource requests

andrewstuart:
- Add resource requests for deployments
- Add readiness probe for master
- Use YAML references for easier maintenance/readability


bobbyrullo:
LGTM - I dig the YAML anchors.

andrewstuart:
Sweet! Yeah, I was surprised that they worked right out of the box, and yet I've never really seen any k8s examples/docs using them.

Issue #371, opened 2016-03-10
Allow clients to be issued ID Tokens valid for other clients


ericchiang:
This is what Google calls _["Cross-client Identity"](https://developers.google.com/identity/protocols/CrossClientAuth)_. Allow clients to be issued ID Tokens which are valid for a different client.

Example: "client1" allows "client2" to create ID Tokens on it's behalf. "client2" uses a custom scope to request such a token, and the returned ID Token holds the following claims:

```json
{
    "iss": "{{ dex URL }}",
    "aud": "userID",
    "aud": "client1",
    "azp": "client2"
}
```

Where azp is the authorized party (the party that requested the ID Token) and aud is the client the ID Token is valid for.

bobbyrullo:
Proposal
===

## Data Model

Clients have a new field called "cross_client_authorizers" which is a list of client_ids. If a client_id is in this list it can mint tokens on behalf of the given client. There are no reciprocal minting rights.

## Data Storage

It's probably fine to store as a comma-delimited list of strings; could also be another table, but that seems like overkill.

## CRUD API

REST methods for getting/adding/removing to be added to worker and overlord APIs
* Worker API: full CRUD per client for authorizers - can't add during creation of client (don't want to deviate from dynamic client reg API)
* Admin API: can only add during creation of client (for bootstrapping)

## Authentication
* Worker API: The Client must use their credentials to make changes
* Admin API: The admin key must be used.

## OIDC Interaction

Consider two clients: 
- `Minter`: This is the client that the user is authenticating with
- `Mintee`: This is the client that `Minter` is trying to mint a token for. In `Mintee`'s list of `cross_client_authorizers` is `Minter`'s `client_id`

When the user authenticates with `Minter` minter will add a scope of `audience:server:client_id:$MINTEES_CLIENT_ID`

When the ID token is returned to the user, it will have within it's `aud` the client_id's of the `Minter` and `Mintee`. The `azp` field will have `Minter`'s client_id

## UI 

There will have to be added UI to the auth page letting the user know the extra scope being added

## Refresh Token

Because of the complications w/r/t multiple audiences (eg. refresh tokens are scoped to a single client) I suggest that requests for cross-client minting with offline access will be rejected for now.

bobbyrullo:
cc: @ericchiang @joeatwork @sym3tri

bobbyrullo:
## Alternatives to Not Allowing Refresh Tokens

[...]

PR #426, opened 2016-04-23
Cross client work 

bobbyrullo:
work for #371 

With this commit dex will perform cross-client auth for trusted peers. 

Refresh tokens are not handled yet, nor is updating trusted peers through worker api.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -39,10 +39,6 @@ const (
 	ResetPasswordTemplateName          = "reset-password.html"
 
 	APIVersion = "v1"
-
-	// Scope prefix which indicates initiation of a cross-client authentication flow.
-	// See https://developers.google.com/identity/protocols/CrossClientAuth
-	ScopeGoogleCrossClient = "audience:server:client_id:"
 )
 
 type OIDCServer interface {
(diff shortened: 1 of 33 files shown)


bobbyrullo:
- [x] Add new table, field
- [ ] add API
  - [ ] worker API
  - [ ] admin API
- [ ] OIDC Work
- [ ] Scope UI

server/server.go
```diff
@@ -39,21 +39,37 @@ const (
...
+
 	// RefreshToken takes a previously generated refresh token and returns a new ID token
 	// if the token is valid.
 	RefreshToken(creds oidc.ClientCredentials, token string) (*jose.JWT, error)
+
 	KillSession(string) error
 }
 
+// DexServer is an OIDCServer that also has dex-specific features.
+type DexServer interface {
```
ericchiang: Why a new interface? When would we use OIDCServer in dex over DexServer?

bobbyrullo: Mostly for pedantic reasons - This CrossClient auth stuff is not vanilla OIDC. If we ever pulled out the OIDC server interface, users wouldn't have to implement weird dex features. If we had a test OIDCServer it wouldn't have to implement DexServer stuff for things that don' t need it.

Happy to take it out if you think it's unnecessary..

ericchiang: I think the OIDC server already has very dex specific assumptions. For example it assumes that refresh tokens will be JWT and, in addition, doesn't provide a way to interpret them so it's unclear how we'd use this interface as a programmable entry point in the future.

I don't like introducing more interfaces we have exactly one implementation of.

ericchiang:
Other than our comment thread lgtm.

server/server.go
bobbyrullo: btw - refresh tokens are *not* JWTs - that's the ID token that gets returned when refreshing using a refresh token.

PR #406, opened 2016-04-13
go-docker: update to 1.6 builder

bobbyrullo:

PR #433, opened 2016-05-04
use go 1.6.2 for docker go build

### LINKED ISSUE #404: Update to Go 1.6.1 and 1.5.4
https://groups.google.com/forum/#!topic/golang-announce/9eqIHqaWvck

Dex doesn't use TLS client auth so this doesn't affect the project. However it seems reasonable to use the latest stable release.

EDIT: wording

chris-rock:
I switched to the official golang image. Not sure if there was a reason to use the dex-builder image?

supersedes #406
fixes #404

### DIFF
--- a/Dockerfile-builder
+++ b/Dockerfile-builder
@@ -1,5 +1,0 @@
-FROM golang:1.6
-
-RUN go get golang.org/x/tools/cmd/cover
-RUN go get golang.org/x/tools/cmd/vet
-RUN go get github.com/tools/godep
--- a/go-docker
+++ b/go-docker
@@ -23,4 +23,4 @@ for VAR in ${DOCKER_LINKS//,/ }; do
 done
 
 echo "running with docker, might take a while to pull the image..."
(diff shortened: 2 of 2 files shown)


ericchiang:
Never cared for the official Go image (it's like 750MB for a 80MB Go distro, ships with a python interpreter, etc), but it looks like the dex-builder is just a derivative of it anyway.

```
$ cat Dockerfile-builder
FROM golang:1.6

RUN go get golang.org/x/tools/cmd/cover
RUN go get golang.org/x/tools/cmd/vet
RUN go get github.com/tools/godep
```

Should be fine, we use vendor now, and I believe vet and cover are now built into the tool.

We should also remove Dockerfile-builder if we're going to delete this.

@bobbyrullo any objections?

bobbyrullo:
Seems reasonable to me

chris-rock:
Updated the PR to remove the `Dockerfile-builder` as well

ericchiang:
LGTM. Thanks for the update!

Issue #301, opened 2016-02-03
quay.io public docker images broken? 


preston:
If you `docker pull quay.io/coreos/dex:latest` and then `docker run ...` the image, you'll get a nasty:

> Error response from daemon: No command specified

What gives? It doesn't look like the Dockerfile includes a CMD for the images on quay.io. I'm on Docker v1.9.1.

ericchiang:
For deployments, dex is comprised of several Docker containers working together, each with different CMD fields. `quay.io/coreos/dex` is used as a base for these.

What are you trying to use the image for?

preston:
Just trying to evaluate it at the moment. Is there a different repo with instructions on getting up and running with the correct set of images? Thanks much for your help!

ericchiang:
The [getting started](https://github.com/coreos/dex/blob/master/Documentation/getting-started.md) doc is a good place to look. We currently don't have images available for this, though that might not be a bad idea in the future.

preston:
Yeah I read through that but it doesn't really address Docker. Sounds like I'll need to go the manual way for now. It would indeed be super cool to provide Docker-specific page for folks like me that just want to deploy with a set of provided environment variables.

ericchiang:
Yeah, it'd be pretty easy for us to provide containerized examples in the getting started guide for ease of use.

jmheidly:
Hi guys,

I've been really struggling to dockerize and run dex components along with postgre. Any easy method?!

ericchiang:
This is as far as I got.

```bash
# Run the Postgres container.
POSTGRES_CID=$( sudo docker run -d -P quay.io/coreos/postgres )
# Determine what port Postgres is running on.
POSTGRES_PORT=$( sudo docker inspect $POSTGRES_CID | jq '.[].NetworkSettings.Ports."5432/tcp"' | jq -r .[].HostPort )
# Determine what host to use (boot2docker/dockermachine vs native).
if [ "$DOCKER_HOST" != "" ]; then
    # Look for hostname of docker VM.
    DOCKER_HOSTNAME=$( echo -n ${DOCKER_HOST#tcp://} | cut -f1 -d":" )
else
    # Look for hostname of docker bridge.
    DOCKER_HOSTNAME=$( ip addr show docker0 | grep 'inet ' | awk '{ print $2 }' | cut -f1 -d"/" )
fi

# Create secrets
[...]

PR #436, opened 2016-05-05
add docker-compose example

chris-rock:
This PR address #301 and provides a simple development setup for Docker. It uses docker-compose to start Postgres, dex-overlord, dex-worker and the example app. This should make it easier to understand how components work together.


contrib/compose/docker-compose.yml
```diff
@@ -0,0 +1,60 @@
+version: '2'
+services:
+  dex-overlord:
+    environment:
+    - DEX_OVERLORD_DB_URL=postgres://postgres:postgres@pg:5432/dex?sslmode=disable
+    - DEX_OVERLORD_ADMIN_LISTEN=http://0.0.0.0:5557
+    - DEX_OVERLORD_KEY_SECRETS=888itDzhxF9rw6To9AK77CLg5HoKin9toDned1mZNYc=
```
janekolszak: Add DEX_OVERLORD_ADMIN_API_SECRET

contrib/compose/README.md
```diff
@@ -0,0 +1,41 @@
...
+This demo intends to demonstrate Dex on a development workstation. It is only intended for local test setups and should not be used for production environments. I recommend to use k8s in production.
+
+This setup is composed of 4 containers:
+- dex-worker
+- dex-overlord
+- postgres
+- example app
+
+Notes:
+- I needed to use some hacks to get the timing of all components right (https://github.com/docker/compose/issues/374). You will see `sh -c 'sleep x ;` to enforce the right order. Nevertheless it may happen that the startup fails for some components, since the execution time may depend on the host machine.
```
janekolszak: I think it won't fail, it'll just retry after couple of seconds. At least that's what I see in logs.

contrib/compose/docker-compose.yml
chris-rock: yes. good idea

contrib/compose/README.md
chris-rock: That only works for dex-overlord. dex-worker will not start if the database is not populated. dexctl is not able to create the connectors and the example app will not work, since it requires client credentials

chris-rock:
I've added the overlord admin secret to the compose example. Anything else that you see, we should cover?

janekolszak:
Nope, seems OK.

Are you planning to make a production-like template for docker-compose?

chris-rock:
@janekolszak Not at the moment, since I would go with kubernetes then. But for a local dev/trail setup it's easier

chris-rock:
Anything required to get this merged?

ericchiang:
Sorry a few of us have been pretty busy with CoreOS fest over the last few days. I'll try to review tomorrow.

chris-rock:
@ericchiang oh forgot that, unfortunately I couldn't attend. I hope you guys had fun, read some good things about it in German media

ecordell:
I think this is a good start but there are a couple of things I would like to see change/improve:

 * I think the goal should be a single step `docker-compose up` to get things running. There's no technical reason there needs to be several setup steps to build the binary first and then load it into a container (at least for trying things out locally - there are good reasons to do that otherwise for go programs)
 * We should make an attempt to remove the `sleeps` from the startup. A common solution for this is to have the dependent services retry themselves until they connect. We may also be able to get rid of some of the waiting overhead by loading configuration directly on startup instead of going through dexctl. I do like showing using dex with postgres, but using an in-memory db might speed things up as well (something to think about).

chris-rock:
@ecordell I like your comments
 
1) Currently, we only need to build something because the example app is not part of the dex container (and I like that approach). Did I got you right, that you are proposing to put a container on quay that includes the example app?
2) `sleep` is more a limitation of the docker-compose. We would need to catch a case like: `database is not available`, where you would expect an app to fail immediately during startup, maybe not during runtime. I would love to get rid of `sleep` and I am very happy for proposals. We could customize the containers to achieve this, but I am against bringing a complexity from the runtime to the application.
3) We need to discuss a the purpose, for me this was mainly illustrating all components required to setup a proper production environment. Maybe we should have a `simple` docker-compose and `pg` docker-compose?

contrib/compose/docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+  dex-worker:
+    environment:
+    - DEX_WORKER_ISSUER=http://dex-worker:5556
+    - DEX_WORKER_DB_URL=postgres://postgres:postgres@pg:5432/dex?sslmode=disable
+    - DEX_WORKER_EMAIL_CFG=/opt/dex/static/emailer.json
+    - DEX_WORKER_LISTEN=http://0.0.0.0:5556
+    - DEX_WORKER_ENABLE_REGISTRATION=true
+    - DEX_WORKER_KEY_SECRETS=FzB7DFVUIgByCKrCb7k619Yw8+UVjslEOPBeyL8VF+E=
+    image: quay.io/coreos/dex
+    command: sh -c 'sleep 10 ; /opt/dex/bin/dex-worker'
```
ericchiang: Why sleep here? The dex-worker should just loop until the db comes up

contrib/compose/docker-compose.yml
ericchiang: Really don't like the idea of hard coded secrets :( but I guess it's an example

contrib/compose/docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+    - dex-setup
+    - dex-overlord
+    - pg
+  # once the overlord created the db, we can create a connector
+  dex-setup:
+    image: quay.io/coreos/dex
+    command: sh -c 'sleep 10 ; /opt/dex/bin/dexctl --db-url=postgres://postgres:postgres@pg:5432/dex?sslmode=disable set-connector-configs /opt/dex/static/connectors.json'
+    volumes:
+    - ./config:/opt/dex/static
+    depends_on:
```
ericchiang: Can you elaborate on what the "depends_on" field does here? Why do you need both the sleep 10 and the depends_on?

contrib/compose/README.md
```diff
@@ -0,0 +1,41 @@
...
+cd contrib/compose
+# start the containers
+docker-compose up
+```
+
+To reach the contains from you local machine, you need to update the `/etc/hosts`:
+
+```
+127.0.0.1 front
+127.0.0.1 dex-worker
```
ericchiang: Perhaps this could match the k8s contrib and be dex.example.com. Also, what's doing the hostname mapping?

contrib/compose/README.md
```diff
@@ -0,0 +1,41 @@
+# Running dex in Docker compose
+
+This demo intends to demonstrate Dex on a development workstation. It is only intended for local test setups and should not be used for production environments. I recommend to use k8s in production.
```
ericchiang: Nit: Avoid terms such as "I" and "you"

"I recommend to use k8s in production." vs. "Kubernetes is recommended for production uses."

"You will see `sh -c 'sleepx';` to enforce the right order" vs, "Sleep statements are used to enforce the correct order."

contrib/compose/example-app/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM debian
```
ericchiang: Couple points.

1) We don't need to build from debian. busybox, alpine, or even "SCRATCH" would do fine.
2) We can just build an example app and push it to quay.io under the coreos name. The dockerfile can then live under examples.
3) Why do you need dexctl here?

ericchiang:
I've left some comments inline.

This is a bit hard to judge because none of the dex maintainers use docker-compose so I'm afraid of this example getting outdated.

contrib/compose/docker-compose.yml
chris-rock: @ericchiang depends is there to set the dependencies for docker compose, but that does not mean it waits until the container is up and running.

contrib/compose/docker-compose.yml
chris-rock: as far as I have observed it, it fails at the point when the database not populated with schema

contrib/compose/README.md
chris-rock: Very good feedback

contrib/compose/README.md
chris-rock: +1

contrib/compose/example-app/Dockerfile
chris-rock: 3.) to create a new client on the fly. See `contrib/compose/example-app/run.sh` better would be to have the state-outside, but then I need to pass in the data during startup. At this point, dex is not up and running... Any proposals to improve this?

chris-rock:
@ericchiang please be honest. I understand if you do not want to merge it, because nobody of the core maintainers is using docker-compose. The goal was to setup a demo environment with a one-liner to see  the awesomeness of dex.

ericchiang:
I'm afraid that it'll go the way of "contrib/standup-db.sh" which has gotten out of date a lot as the project develops.

I'm okay with adding this, but I hesitant and just want to make sure you understand our position.

PR #442, opened 2016-05-12
Adds client manager

### LINKED ISSUE #63: make ClientIdentityRepo tx friendly.
...make sure to fix race conditions, eg UserManager.CreateUser

### LINKED ISSUE #419: centralize ClientRepo business logic
We probably want a ClientManager object for that. 

The ClientRepo is weirdly inconsistent compared to the rest of the Repos - most other repos perform just CRUD operations, but ClietRepo is a weird mixture. An example inconsistency is that the Repo creates the secret, but we rely on an ID passed into it. To be like other repos it should not either one.

Also needs txns.

ecordell:
This adds transaction support to the client repo and adds a client manager.

The tests scripts weren't covering the client package, so also added that.

And some refactoring around secrets to isolate the crypto dependencies to one package.

Fixes #419 
Fixes #63


client/manager/manager_test.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	ok, err := f.mgr.Authenticate(*cc)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	} else if !ok {
+		t.Fatalf("Authentication failed for good creds")
+	}
+}
+
+//func TestAuthenticate(t *testing.T) {
```
bobbyrullo: what's all this uncommented code from?

client/manager/manager.go
```diff
@@ -0,0 +1,92 @@
...
+		return false, err
+	}
+
+	return c.Admin, nil
+}
+
+func (m *ClientManager) SetDexAdmin(clientID string, isAdmin bool) error {
+	tx, err := m.begin()
+	if err != nil {
+		return err
```
bobbyrullo: missing a bunch of rollbacks here.

A better pattern we stumbled on in more recent code was to immediately `defer rollback()`, since a rollback will have no effect (just returns ErrTxDone) for a successfully committed tx.

client/manager/manager_test.go
ecordell: Accidentally pushed it up, but fixed now. I was working on pulling the Authenticate method up to the manager and was changing the tests.

client/manager/manager.go
ecordell: Oh nice, didn't know that. I'll update

client/manager/manager.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	ok := secret.CompareHashAndPassword(clientSecret, dec) == nil
+	return ok, nil
+}
+
+func rollback(tx repo.Transaction) {
+	err := tx.Rollback()
+	if err != nil {
+		log.Errorf("unable to rollback: %v", err)
```
ericchiang: Won't this trigger on every successful transaction because you defer the rollback call?

client/manager/manager.go
```diff
@@ -0,0 +1,154 @@
...
+	err = tx.Commit()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (m *ClientManager) Authenticate(creds oidc.ClientCredentials) (bool, error) {
+	clientSecret, err := m.clientRepo.GetSecret(nil, creds.ID)
+	if err != nil || clientSecret == nil {
```
ericchiang: Why are we swallowing this error rather than returning one?

ecordell: I was thinking that we don't want to expose any information about what clients do/don't exist in the db on an auth failure

client/manager/manager.go
ecordell: Yeah that's silly, I'll get rid of it

secret/secret.go
```diff
@@ -0,0 +1,45 @@
+package secret
+
+import (
+	"encoding/base64"
```
bobbyrullo: now that secrets are in one place, do we need this module?

client/client.go
```diff
@@ -27,6 +27,8 @@ type Client struct {
 type ClientRepo interface {
 	Get(tx repo.Transaction, clientID string) (Client, error)
 
+	GetSecret(tx repo.Transaction, clientID string) ([]byte, error)
+
```
bobbyrullo: This returns the hashed secret right? Please document that if that's the case.

db/client.go
```diff
@@ -218,15 +226,9 @@ func (r *clientRepo) SetDexAdmin(clientID string, isAdmin bool) error {
...
-	m, err := r.executor(tx).Get(clientModel{}, creds.ID)
-	if m == nil || err != nil {
-		return false, err
-	}
-
-	cim, ok := m.(*clientModel)
-	if !ok {
-		log.Errorf("expected clientModel but found %v", reflect.TypeOf(m))
-		return false, errors.New("unrecognized model")
+	clientSecret, err := r.GetSecret(nil, creds.ID)
```
bobbyrullo: Can we get rid of this then?

ecordell: It's gone - I think you're looking at an older diff?

ecordell:
@ericchiang @bobbyrullo Most recent update removes the `secret` package, which required moving all client creation into the client manager. Mostly `repo.NewClientRepoWithClients` -> `manager.NewClientManagerWithClients`. This means the repo just deals with storing data, and the manager deals with generating the correct data (generating/hashing secrets, etc) to give to the repo. It also updates the test suite to go through the client id/secret generation rather than mocking everywhere.

I think a decent next cleanup step would be to remove the client and user repos from the server struct and do everything through the managers instead. Up to you guys if you think that should be in this PR or a later one.

bobbyrullo:
@ecordell - awesome - I'll take a look asap. I would do that in a separate PR

ecordell:
It turns out this breaks the example app, which relies on explicitly setting the client id and secret.

It looks like this will break anything relying on loading clients from the config flag like in the examples in the readme. (It will load fine, but then the client id you gave it won't exist in the db)

There are a few options to get around this but I think the best one would be to set the client id directly only when loaded from a config file; otherwise generate the client id by default. Any thoughts?

ecordell:
This should be good to go now

client/manager/manager.go
bobbyrullo: You should return an error here - of course we shouldn't return it to the user, but that's for callers to handle

client/manager/manager.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+	}
+	if err := tx.Commit(); err != nil {
+		return nil, err
+	}
+	return clientManager, nil
+}
+
+func (m *ClientManager) New(meta oidc.ClientMetadata) (*oidc.ClientCredentials, error) {
+	tx, err := m.begin()
```
bobbyrullo: Because you have New() taking a `ClientMetadata` you can't set DexAdmin and other non-metadata properites on it in the same transaction

ecordell:
@bobbyrullo I updated `ClientManager.New()` to take metadata and an admin flag instead of a full client object

bobbyrullo:
@ecordell - why not the full client object, just as `repo.New()` does? We are going to be adding more dex specific metadata to the Client object and we don't want to keep having to change the the `manager.New()` function signature

ecordell:
I thought passing in an actual client object would be an unusual/confusing api, since part of the object is required to exist (`Metadata`) and part of it is required to be empty (`Credentials`) or at least overwritten. Open to suggestions.

`repo.New()` takes the client you give it and stores it in the database, doesn't have this issue.

bobbyrullo:
I agree that it's a bit confusing, but we can document that. Also, this opens us up to later allow for APIs which specify the client_id and secret.

bobbyrullo:
@ecordell see for example https://github.com/coreos/dex/pull/426/commits

client/manager/manager.go
```diff
@@ -94,17 +94,18 @@ func NewClientManagerFromClients(clientRepo client.ClientRepo, txnFactory repo.T
...
-func (m *ClientManager) New(meta oidc.ClientMetadata) (*oidc.ClientCredentials, error) {
+func (m *ClientManager) New(client client.Client) (*oidc.ClientCredentials, error) {
 	tx, err := m.begin()
 	if err != nil {
 		return nil, err
 	}
 	defer tx.Rollback()
 
-	cli, err := m.clientFromMetadata(meta)
+	cli, err := m.clientFromMetadata(client.Metadata)
```
bobbyrullo: Is this still necessary now given that you have a client?

client/manager/manager.go
```diff
@@ -94,17 +94,18 @@ func NewClientManagerFromClients(clientRepo client.ClientRepo, txnFactory repo.T
...
 	if err != nil {
 		return nil, err
 	}
+	cli.Admin = client.Admin
```
bobbyrullo: if you used the client directly you wouldn't need to do this...

client/manager/manager_test.go
```diff
@@ -126,8 +126,12 @@ func TestAuthenticate(t *testing.T) {
 			url.URL{Scheme: "http", Host: "example.com", Path: "/cb"},
 		},
 	}
-
-	cc, err := f.mgr.New(cm)
+	cli := client.Client{
+		Credentials: oidc.ClientCredentials{},
```
bobbyrullo: nit: this line is a noop

client/manager/manager_test.go
```diff
@@ -126,8 +126,12 @@ func TestAuthenticate(t *testing.T) {
...
+		Metadata:    cm,
+		Admin:       false,
```
bobbyrullo: as is this one :)

cmd/dexctl/driver_db.go
```diff
@@ -30,7 +31,12 @@ func (d *dbDriver) NewClient(meta oidc.ClientMetadata) (*oidc.ClientCredentials,
 	if err := meta.Valid(); err != nil {
 		return nil, err
 	}
-	return d.ciManager.New(meta)
+	cli := client.Client{
+		Credentials: oidc.ClientCredentials{},
```
bobbyrullo: noop again....

server/auth_middleware_test.go
```diff
@@ -33,7 +34,12 @@ func TestClientToken(t *testing.T) {
 	dbm := db.NewMemDB()
 	clientRepo := db.NewClientRepo(dbm)
 	clientManager := clientmanager.NewClientManager(clientRepo, db.TransactionFactory(dbm), clientmanager.ManagerOptions{})
-	creds, err := clientManager.New(clientMetadata)
+	cli := client.Client{
+		Credentials: oidc.ClientCredentials{},
```
bobbyrullo: noop

bobbyrullo:
LGTM once tests pass.

PR #446, opened 2016-05-19
examples: add sample for clients

xaka:
Adds consistency to existing samples for users, connectors, etc., as well as protects users from accidental changes to files under Git, which is why there are samples after all.


bobbyrullo:
LGTM - thanks @xaka !

PR #449, opened 2016-05-26
Set and list connectors from admin API

### LINKED ISSUE #360: Add ability to set connectors through bootstrapping API
This is another feature that now goes through the --db-url flag. See #298.

ericchiang:
In addition to adding the necessary endpoints and tests, I've added a small tool to the schema/generate script that formats the JSON schema files. We were mixing tabs and spaces and the diffs were often very ugly.

closes #360


ericchiang:
cc @chancez

ericchiang:
I'll follow up with a dexctl refactoring.

schema/adminschema/v1.json
```diff
@@ -96,77 +96,129 @@
...
-              }
+        }
+      }
+    },
+    "Connectors": {
+      "methods": {
+        "Set": {
+          "id": "dex.admin.Connector.Set",
+          "description": "Set the connectors for the dex system.",
+          "httpMethod": "PUT",
```
sym3tri: shouldn't we have 4 methods to do the full CRUD?

`POST` create a new one
`PUT` update existing
`GET` & `DELETE`

schema/adminschema/v1.json
```diff
@@ -96,77 +96,129 @@
...
+          "path": "connectors",
+          "request": {
```
sym3tri: what is the response?

ericchiang: Nothing. The status code indicates if it was successful.

schema/adminschema/v1.json
ericchiang: You generally always set all the connectors at once. We can split this out into the full CRUD if you'd like and have them deal with individual connectors (maybe based on ID).

ericchiang: On thing that might be confusing is updating vs creating. For instance, what should dexctl do when presented with a list of connectors? Get a list, determine a diff then apply the appropriate create, update, and delete commands?

sym3tri: Sorry, i was assuming creating/updating individual connectors, not the entire set. If we're always dealing with the entire set then having just a "set" command makes sense (no need for "update"). However if we go that route we'll probably need an "add" and "delete" for individual connectors. I'm ok with either approach & will leave it up to you.

schema/adminschema/v1.json
ericchiang: updated the comment

schema/adminschema/v1.json
ericchiang: Going to leave it this way for now and add "add" and "delete" commands later for individual connectors. @bobbyrullo @chancez this work for you?

chancez: Yup

schema/adminschema/v1.json
```diff
@@ -100,6 +100,37 @@
           "$ref": "Client"
         }
       }
+    },
+    "Connector": {
+      "id": "Connector",
+      "type": "any",
+      "description": "A configuration object which describes how dex will federate identity. Because the actual structure differs for each kind of connector, the object here is unspecified."
```
bobbyrullo: you could be a bit more specific here - it's the JSON representation of stuff in connectors/connectors_XXX.go right?

schema/adminschema/v1.json
```diff
@@ -100,6 +100,37 @@
...
+      "properties": {
+        "connectors": {
+          "type": "array",
+          "items": {
+            "$ref": "Connector"
+          }
+        }
+      }
+    },
+    "ConnectorsGetResponse": {
```
bobbyrullo: Maybe it'll be clearer when I see the rest of the commit, but it seems like there should be 1 response: it either set them all or failed to do any.

admin/api_test.go
```diff
@@ -252,3 +253,51 @@ func TestGetState(t *testing.T) {
 		}
 	}
 }
+
+func TestSetConnectors(t *testing.T) {
+	tests := []struct {
+		connectors []connector.ConnectorConfig
+	}{
```
bobbyrullo: please test failures as well

schema/adminschema/v1.json
bobbyrullo: Yeah, I think it's fine this way.

schema/adminschema/v1.json
ericchiang: Can I add a link to https://github.com/coreos/dex/blob/master/Documentation/connectors-configuration.md? This seems like the canonical documentation on connectors.

bobbyrullo: yeah, that's perfect

admin/api_test.go
ericchiang: The actual method being tested is

```go
func (a *AdminAPI) SetConnectors(connectorConfigs []connector.ConnectorConfig) error {
	return a.connectorConfigRepo.Set(connectorConfigs)
}
```

This can only return database or marshaling errors, so it's actually hard to think of a reasonable failure test case.

schema/adminschema/v1.json
ericchiang: updated

admin/api_test.go
bobbyrullo: that's fine - failure is tested in the integration section so I think we have decent coverage.

bobbyrullo:
LGTM

Issue #450, opened 2016-05-27
Add support for TLS in the example application


alexbrand:
Even though it is a sample app, it'd be great if the app could be setup to serve over HTTPS.

ericchiang:
I believe this has been covered in #453. Closing.

PR #451, opened 2016-05-27
Add TLS support to the example application

alexbrand:
Fixes issue #450.

Added two flags for passing the cert and key files. If both flags are set, the server is started using TLS.


ericchiang:
This is going to be tricky since "--listen" and "--callback-url" defaults are "http://".

Can we add some UX to make sure someone specifying these flags also uses the correct scheme for appropriate flags?

alexbrand:
@ericchiang Good call. I'll try to work on this tonight.

ericchiang:
Awesome. A simple solution would be to ensure that the scheme matches the type of listener, and error out if they don't match.

alexbrand:
Looks like I messed up the rebase? :-(

ericchiang:
Yep :(

Don't know how much I can help.

alexbrand:
I'll create a new PR with a proper branch.

PR #453, opened 2016-06-03
Added TLS support to the example application

alexbrand:
Fixes issue #450.

Added two flags for passing the cert and key files. If both flags are set, the server is started using TLS.

Also added UX to ensure that "--listen" and "--callbackURL" are HTTPS when user provides TLS cert and key.

### DIFF
--- a/examples/app/main.go
+++ b/examples/app/main.go
@@ -31,6 +31,9 @@ func main() {
 	clientSecret := fs.String("client-secret", "ZXhhbXBsZS1hcHAtc2VjcmV0", "")
 	caFile := fs.String("trusted-ca-file", "", "the TLS CA file, if empty then the host's root CA will be used")
 
+	certFile := fs.String("tls-cert-file", "", "the TLS cert file. If empty, the app will listen on HTTP")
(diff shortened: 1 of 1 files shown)


examples/app/main.go
```diff
@@ -70,6 +73,16 @@ func main() {
...
 	}
 
+	redirectURLParsed, err := url.Parse(*redirectURL)
+	if err != nil {
+		log.Fatalf("Unable to parse url from --redirect-url flag: %v", err)
+	}
+
+	useTLS := *keyFile != "" && *certFile != ""
+	if useTLS && (redirectURLParsed.Scheme != "https" || l.Scheme != "https") {
+		log.Fatalf("TLS Cert File and Key File were provided. Ensure the listen and redirect URL are using HTTPS.")
```
ericchiang: nit: Maybe

> Ensure listen and redirect URLs are using the "https://" scheme.

ericchiang:
small note, but other than that lgtm

ericchiang:
thanks for your contribution!

PR #455, opened 2016-06-03
*: Update Go versions used for Travis tests and test tip

ericchiang:


ericchiang:
Looks like our env script means we can't build with development versions of Go.

PR #457, opened 2016-06-07
Various client api tweaks

bobbyrullo:
* change generateClientCreds to addClientCreds - more accurate reflection of the API
* add test to make sure that clients in sample file are valid, and that they can be authenticated against (this makes me worry less about screwing up the encoding/hashing of client_secrets)
* in preparation for removing ClientManagerFromClients, fix up a lot of copy-pasta in server tests and use testutil code instead to create fixtures
* finally, remove ClientManagerFromClients, in favor of a similar method in the repo layer

### DIFF
--- a/client/manager/manager.go
+++ b/client/manager/manager.go
@@ -2,7 +2,6 @@ package manager
 
 import (
 	"encoding/base64"
-	"fmt"
 
 	"errors"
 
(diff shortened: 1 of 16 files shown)


client/manager/manager.go
```diff
@@ -64,7 +64,7 @@ func NewClientManager(clientRepo client.ClientRepo, txnFactory repo.TransactionF
 	}
 }
 
-func NewClientManagerFromClients(clientRepo client.ClientRepo, txnFactory repo.TransactionFactory, clients []client.Client, options ManagerOptions) (*ClientManager, error) {
+func NewClientManagerFromClientsX(clientRepo client.ClientRepo, txnFactory repo.TransactionFactory, clients []client.Client, options ManagerOptions) (*ClientManager, error) {
```
ericchiang: ClientsX?

ecordell:
LGTM

client/manager/manager.go
ecordell: I assumed that was because this will be removed?

ericchiang: Hmm, maybe just a comment?

```
// DEPRICATED: Use xyz instead.
func NewClientManagerFromClients(...)
```

bobbyrullo: oops. forgot to delete.

bobbyrullo:
Thanks for the review Evan!

PR #458, opened 2016-06-08
Load trustedPeers in no DB mode, add x-client to example app

bobbyrullo:


db/client.go
```diff
@@ -212,14 +212,23 @@ func (r *clientRepo) All(tx repo.Transaction) ([]client.Client, error) {
...
+		cm, err := newClientModel(c.Client)
 		if err != nil {
 			return nil, err
 		}
 		err = repo.executor(nil).Insert(cm)
+		if err != nil {
+			return nil, err
+		}
+
+		err = repo.SetTrustedPeers(nil, c.Client.Credentials.ID, c.TrustedPeers)
```
ericchiang: Do we not check that the trusted peer exists before creating this relationship? In this case, shouldn't we initialize all the clients then add the trusted peers?

client/client.go
```diff
@@ -94,16 +94,24 @@ func ValidRedirectURL(rURL *url.URL, redirectURLs []url.URL) (url.URL, error) {
 	return url.URL{}, ErrorInvalidRedirectURL
 }
 
-func ClientsFromReader(r io.Reader) ([]Client, error) {
+// LoadableClient contains sufficient information for creating a Client and its related entities.
+type LoadableClient struct {
```
ericchiang: I'm kinda iffy about this as a top level object. It shouldn't really be used outside of ClientsFromReader (which is only used in server/config.go).

db/client.go
bobbyrullo: I took out that check in the next commit because it will make things difficult when bootstrapping.

client/client.go
bobbyrullo: It's also used in db/client.go.

I see what you're saying about the weirdness, but my rationale is this:

1) this is the same model that the user module uses. There's something to be said for consistency.

2) when we add more storage backends (as we will almost certainly do it seems) they will also need something in their implementation which loads such "full" clients.

db/client.go
ericchiang: Okay no worries.

client/client.go
ericchiang: I guess I'd like both to just be private so we aren't encouraged to use this in other packages. But yeah, this seems consistent with a lot of our other code, so sound good.

ericchiang:
Lgtm if you can get the checks green

bobbyrullo:
Thanks eric - was a codegen gofmt issue

PR #459, opened 2016-06-09
Use Github templates for issues/proposals

burdiyan:
Github launched the ability to automatically populate issues and pull requests with custom templates (https://github.com/blog/2111-issue-and-pull-request-templates). 

This eliminates the need to have custom scripts for populating issues with custom templates. It should be easier for contributors to just open a new issue and not worry about copying the template.


ericchiang:
lgtm

ericchiang:
Per conversations with @bobbyrullo these templates were not intended to be for every issue, but only for proposals.

Going to revert this change, though it would be nice to have a generic issue template. Might be worth working with coreos/docs on this.

PR #461, opened 2016-06-09
#400 Add connector id to the registration error message

burdiyan:
Right now it is not clear what connector is failing. It will be easier to debug with more specific error message.

Related to #400


ericchiang:
lgtm, thanks for you contribution!

Issue #460, opened 2016-06-09
Accept client_id and client_secret in request body


teh:
Hi,

Some oauth2 clients (e.g. oauth2_proxy [2]) post the client id and secret in the body, not as a header. But dex only checks the www-authenticate header [1]. I think the fix is simple and I can do it but I wanted to check first whether this is an intentional ommision.

[1]
https://github.com/coreos/dex/blob/a9d854e144b8245d5672d4b92ab497c438283a03/server/http.go#L471

[2] - TEST DATA: no sensitive secrets here.

```
client_id=ULGNrOr4hC9_HakhTBJJzncNUsuef-Y54FjzBj-xPLs%3D%4052.50.74.68&client_secret=EIZRrGT_925P2OlgM571PQoZnhNv59XXRVLXUmU6WtKX29eT8VOntuqzOtOq3jW_eZarmfcthF54ErRH6WNWrUHZRlmQvm1M&code=54P5xRJtfcw%3D&grant_type=authorization_code&redirect_uri=http%3A%2F%2F52.50.74.68%3A4180%2Foauth2%2Fcallback
```

ericchiang:
There was some conversation about this in #320. Going to re-paste this comment from that thread

>    Alternatively, the authorization server MAY support including the
   client credentials in the request-body using the following
   parameters:

>   client_id
         REQUIRED.  The client identifier issued to the client during
         the registration process described by Section 2.2.

>   client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.

>   __Including the client credentials in the request-body using the two
   parameters is NOT RECOMMENDED and SHOULD be limited to clients unable
   to directly utilize the HTTP Basic authentication scheme__ (or other
   password-based HTTP authentication schemes).  The parameters can only
   be transmitted in the request-body and MUST NOT be included in the
   request URI.

https://tools.ietf.org/html/rfc6749#section-2.3 (emphasis my own)

The spec seems pretty clear that clients shouldn't be doing this. You may want to raise a bug with oauth2_proxy, but this gets brought up a lot. There's even a huge list of servers[0] that do this improperly.

It might be easier to be more lax about this.

[0] https://github.com/golang/oauth2/blob/65a8d08c6292395d47053be10b3c5e91960def76/internal/token.go#L92

ericchiang:
@teh can you open an issue upstream with oauth2_proxy? I'd be interested in why they do this in the first place.

[...]

PR #462, opened 2016-06-09
Handle client_id and client_secret in request body.

teh:
I *think* this works but I can't test because I can't figure out how to build this from my github repository. Go seems to use absolute repository paths for imports but I'm not a go programmer so I'm not sure what I'm doing wrong.


ericchiang:
Thank you for your proposal, but I don't think we've agreed to accept this change.

As commented in #460, the spec is very clear that this is "NOT RECOMMENDED" and "SHOULD be limited to clients unable to directly utilize the HTTP Basic authentication scheme."

I'd really rather not add this. Particularly if it's trivial for clients to [use basic auth[(https://golang.org/pkg/net/http/#Request.SetBasicAuth).

teh:
OK no worries. I read

> It might be easier to be more lax about this.

as "let's do this". I'll close

PR #463, opened 2016-06-10
*: add --enable-automatic-registration flag to worker

### LINKED ISSUE #310: Easier registration for first login
This was brought up by @fnordahl in #178. For connectors that are implicitly more restrictive, can we have a better first login experience?

Consider adding additional connector options.

Logging in takes you to a page that asks you if you want to register. Clicking yes takes you back to relying party with an identity.
* Might be a good default for dex instances which enable registration.
* Registration step prevents mis-registration by "low friction" remote connectors (e.g. choosing the wrong google account).

Logging in auto registers you:
* Would be enabled on a per connector basis.
* Good for connectors which require more effort to login (e.g. type in your LDAP username and password).

ericchiang:
When specified the "--register-on-first-login" allows users logging in through remote connectors to skip registration. This make a lot of sense for connectors like LDAP, which have their own registration upstream.

Closes #310

This change also renames "server/testutil.go" to "server/testutil_test.go" to ensure it's not used outside of tests.

cc @sym3tri

### DIFF
--- a/cmd/dex-worker/main.go
+++ b/cmd/dex-worker/main.go
@@ -46,7 +46,9 @@ func main() {
 	emailFrom := fs.String("email-from", "", "emails sent from dex will come from this address")
 	emailConfig := fs.String("email-cfg", "./static/fixtures/emailer.json", "configures emailer.")
 
-	enableRegistration := fs.Bool("enable-registration", false, "Allows users to self-register")
+	enableRegistration := fs.Bool("enable-registration", false, "Allows users to self-register. This flag cannot be used in combination with --enable-automatic-registration.")
+	registerOnFirstLogin := fs.Bool("enable-automatic-registration", false, "When a user logs in through a federated identity service, automatically register them if they don't have an account. This flag cannot be used in combination with --enable-registration.")
+
 	enableClientRegistration := fs.Bool("enable-client-registration", false, "Allow dynamic registration of clients")
 
 	noDB := fs.Bool("no-db", false, "manage entities in-process w/o any encryption, used only for single-node testing")
@@ -88,6 +90,11 @@ func main() {
 	if *printVersion {
 		fmt.Printf("dex version %s\ngo version %s\n", strings.TrimPrefix(version, "v"), strings.TrimPrefix(runtime.Version(), "go"))
 		os.Exit(0)
+	}
+
+	if (*enableRegistration) && (*registerOnFirstLogin) {
+		fmt.Fprintln(os.Stderr, "The flags --enable-registration and --enable-automatic-login cannot both be true.")
+		os.Exit(1)
 	}
 
 	if *logDebug {
@@ -135,6 +142,7 @@ func main() {
 		IssuerLogoURL:            *issuerLogoURL,
 		EnableRegistration:       *enableRegistration,
 		EnableClientRegistration: *enableClientRegistration,
+		RegisterOnFirstLogin:     *registerOnFirstLogin,
 	}
 
 	if *noDB {
--- a/server/server.go
+++ b/server/server.go
@@ -64,28 +64,35 @@ type OIDCServer interface {
 type JWTVerifierFactory func(clientID string) oidc.JWTVerifier
 
 type Server struct {
-	IssuerURL                      url.URL
-	KeyManager                     key.PrivateKeyManager
-	KeySetRepo                     key.PrivateKeySetRepo
-	SessionManager                 *sessionmanager.SessionManager
-	ClientRepo                     client.ClientRepo
-	ConnectorConfigRepo            connector.ConnectorConfigRepo
+	IssuerURL url.URL
+
 	Templates                      *template.Template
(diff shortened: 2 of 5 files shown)


chancez:
I think we probably want to carefully think about how we expose this type of knob. In general, this seems like a per-connector configuration, and maybe it's not even something you can configure per-connector, but each connector should probably say whether or not they require registration.

Thus:
- We should expand the Connector interface to have a `RequiresRegistration() bool` method.
  - LDAP would hard code this to false for now, making this not configurable.
  - Local & Google would hard code this to true right now to keep it inline with the current defaults. This could be something we expose as a configurable value in the connectors config in the future
- the `--enable-registration` flag may conflict with these values. I think we should deprecate that flag (keeping it around until we fix this whole thing up). It may be best to do validation that doesn't allow us to use this flag with connectors that conflict with it, and we simply only support "correct" configurations. This would mean, if you set `--enable-registration=true`, you cannot add a connector which has a `RequiresRegistration()`, so it would only be compatible with Google/Local today. If you have `--enable-registration=false`, then you could only use LDAP (and maybe local with only the admin user).
- Avoid anything which picks something implicitly for a connector, because it may not make sense in the context of future connectors, or even current ones, which means I'm strongly against making a new flag which indiscriminately enables this behavior for all connectors (such as in this PR).

server/server.go
```diff
@@ -325,42 +332,72 @@ func (s *Server) Login(ident oidc.Identity, key string) (string, error) {
...
 	}
 
-	usr, err := s.UserRepo.GetByRemoteIdentity(nil, user.RemoteIdentity{
-		ConnectorID: ses.ConnectorID,
-		ID:          ses.Identity.ID,
-	})
+	remoteIdentity := user.RemoteIdentity{ConnectorID: ses.ConnectorID, ID: ses.Identity.ID}
+
+	// Get the connector used to log the user in.
+	var conn connector.Connector
```
bobbyrullo: Breaking this out into its own method, eg .`(s *Server) GetConnectorByID(id string)` will make this long method a bit easier to read.

cmd/dex-worker/main.go
```diff
@@ -48,6 +48,7 @@ func main() {
 
 	enableRegistration := fs.Bool("enable-registration", false, "Allows users to self-register")
 	enableClientRegistration := fs.Bool("enable-client-registration", false, "Allow dynamic registration of clients")
+	registerOnFirstLogin := fs.Bool("register-on-first-login", false, "When a user logs in through a federated identity service, automatically register them if they don't have an account. This flag is independent of enable-registration.")
```
sym3tri: Independent, true, but this also nullifies the meaning of `enableRegistration` too right? Maybe you should mention that.

ericchiang: No, they're truly independent. enableRegistration only affects if the register screen will be shown, registerOnFirstLogin only affects the behavior of logging in through a upstream connector.

sym3tri: What happens if enableRegistration=true and registerOnFirstLogin=true?

bobbyrullo: > enableRegistration only affects if the register screen will be shown

That's not all it does: much more importantly, it turns off the /registration endpoint. If it was just a matter of the register page being shown or not, that would be a huge security hole.

ericchiang: > What happens if enableRegistration=true and registerOnFirstLogin=true?

You'd be able to use the registration page (and the registration endpoint), as well as be automatically registered via logging in through a remote connector.

bobbyrullo: Should we make these flags mutually exclusive?

ericchiang: That's a lot of what @chancez was getting at. We're trying to express three types of registration with two flags.

* Require registration
* Require local connector and invites
* Require upstream registration

ericchiang: On the other hand, I'm going to advocate that we make these exclusive. There's not a good use cases for having both turned on.

sym3tri:
LGTM after the remaining comments are addressed

ericchiang:
@sym3tri renamed to "enable-automatic-registration" to match "enable-registration" to show similarity and made them exclusive.

sym3tri:
Thanks. LGTM ship it

cmd/dex-worker/main.go
```diff
@@ -90,6 +92,11 @@ func main() {
 		os.Exit(0)
 	}
 
+	if (*enableRegistration) && (*registerOnFirstLogin) {
+		fmt.Fprintln(os.Stderr, "The flags --enable-registration and --register-on-first-login cannot both be true.")
```
sym3tri: Typo here `register-on-first-login` is the old name.

sym3tri:
scratch that LGTM. found a typo

cmd/dex-worker/main.go
ericchiang: fixed

PR #465, opened 2016-06-10
Cross client refresh tokens

bobbyrullo:
Refresh clients can now be used to do cross-client auth.


integration/user_api_test.go
```diff
@@ -147,8 +147,10 @@ func makeUserAPITestFixtures() *userAPITestFixtures {
 	}
 
 	refreshRepo := db.NewRefreshTokenRepo(dbMap)
+	fmt.Println("DEFAULT: ", oidc.DefaultScope)
```
ericchiang: Left over print statement?

refresh/repo.go
```diff
@@ -41,11 +43,11 @@ func DefaultRefreshTokenGenerator() ([]byte, error) {
...
 	// Create generates and returns a new refresh token for the given client-user pair.
 	// On success the token will be return.
-	Create(userID, clientID string) (string, error)
+	Create(userID, clientID string, scope []string) (string, error)
 
-	// Verify verifies that a token belongs to the client, and returns the corresponding user ID.
+	// Verify verifies that a token belongs to the client and is valid for the given scopes. It returns the user ID to which the token belongs.
 	// Note that this assumes the client validation is currently done in the application layer,
-	Verify(clientID, token string) (string, error)
+	Verify(clientID, token string, scopes []string) (string, error)
```
ericchiang: Why does this takes a "scopes" argument? I wouldn't expect those to change between successive refresh requests.

ericchiang: Ah never mind. I didn't realize that scopes could change between refresh requests.

Can we make this comment more explicit about what it's doing? I just want to make sure someone working with this can understand exactly where they are in the refresh token flow. E.g.

```
Verify parses a raw refresh token and returns the user ID for which the token was issued for. It also ensures the requested scopes do not include any scopes not in the original request.
NOTE: Refresh requests require validating the client_id and client_secret pair, which callers MUST do before calling this function.
```

server/http.go
```diff
@@ -518,11 +518,12 @@ func handleTokenFunc(srv OIDCServer) http.HandlerFunc {
 			}
 		case oauth2.GrantTypeRefreshToken:
 			token := r.PostForm.Get("refresh_token")
+			scopes := r.PostForm.Get("scope")
```
ericchiang: What happens when this field is omitted? I'm having trouble following that path in the code.

integration/user_api_test.go
bobbyrullo: removed

refresh/repo.go
bobbyrullo: done.

server/http.go
bobbyrullo: If the "scope" field is omitted then there's no difference from the old behavior; if they add cross-client scopes, then they'll get the cross-client behavior, provided that the refresh token was authorized for that in the first place.

We _could_ be a bit more persnickety here and require that `openid` is specified here, but I'm worried that not all clients do this, and changing behavior from previous versions. We could also check for `email` here before putting email in the claims...but we don't do that elsewhere and that would almost certainly break existing clients.

ericchiang: If the users omit the "scope" field, it's the same as requesting the scopes present in the original requests.[0] How does the code handle this case?

I'm basically looking for some code that looks like:

```
if len(scopes) == 0 {
    scopes = record.Scopes
}
```

Though I can't find it.

I also don't see a test case[1] that tests for that case.

[0] https://tools.ietf.org/html/rfc6749#section-6
[1] https://github.com/bobbyrullo/dex/blob/aa3f0681a9666941b9fa5e9e8ce579715bcb039c/server/server_test.go#L487

bobbyrullo: Good catch - I didn't realize that it was supposed to do that.

ericchiang:
lgtm (pending a squash)

PR #466, opened 2016-06-13
vendor: update go-oidc to add support for Azure AD

tksm:
Update github.com/coreos/go-oidc/ to include coreos/go-oidc#87
which adds support for Azure AD


ericchiang:
lgtm

PR #468, opened 2016-06-15
Remove old client_resource api

### LINKED ISSUE #467: Prep Work for native clients.
## Intro

"Native clients" are soon to be introduced. These clients have essentially public IDs and Secrets, and there are some APIs for which the ID/Secret pair can be used to gain credentials. I wasn't sure what these were, so I made a survey of all the known APIs (except the Admin(bootstrap) APIs and OIDC apis)

## Authn Types

Authn: clientTokenMiddleware
Impl: server/auth_middleware.go
Desc: JWT with client as both sub and aud as bearer token (obtained from token endpoint)
Notes: a bit annoying in that a user has to go to /token using client creds, to obtain these creds, then hit the endpoint they want to hit.

Authn: userJWT
Impl: server/user.go::getCreds()
Desc: JWT obtained from authenticating to a client sub=user, aud=client

## APIS from schema/workerschema

Resource: Clients
Desc: lists, creates OIDC clients
Implementation: server/client\_resource.go
Authn: clientTokenMiddleware
Authz: 
Used: cmd/dexctl/driver_api.go - the dexctl driver NewClient command
Notes: This is used for existing clients to create other clients...not sure if this is really so useful.
Proposal: drop. One day soon create new dextcl driver which uses admin api and admin token

Resource: Users
Desc: User Management API: List, Get, Create, Disable, ResendEmailInvite
Implementation: biz logic: user/api/api.go http: server/user.go
Authn: userJWT 
Authz: user has to be Admin, client has to be admin
Used: console.
Notes: 
Proposal: keep.

Resource: RefreshClient
Desc: List, Revoke refresh tokens for a user.
Implementation: biz logic: user/api/api.go http: server/user.go
Authn: userJWT
Authz: can only see, mutate tokens for self.
Used: nowhere yet - should be in console, maybe dexctl.
Notes: 
Proposal: Keep, but maybe should be in a seperate place from user mgmt server.

## Other APIS, endpoints

API: resendEmail
Desc: resends email
Implementation: server/email_verification
Authn: clientTokenMiddleware + user JWT
Authz: verifies emailResend Token
Used: console, tectonic.com presumably.
Notes: 
Proposal: Keep. Disallow for native clients?

API: sendResetPassowrd
Implementation: 
Authn: session key
Used: from within Dex UI
Proposal: keep

API: resetPassword
Authn: resetPasswordToken 
Used: within dex, part of password reset flow
Proposal: keep

API: invitationAccept
Implementation: 
Authn: invation token
Proposal: keep

API: clientRegistration (OIDC Spec dynamic reg)
Implementation: server/client_registration
Authn: None
Proposal: Keep

## summary

We don't want people native client creds to be able to be able initiate email resends, or to create, list all clients.

In fact, we probably don't want just anyone with a client to be able to list all other clients or create clients. So I propose the following:

1) get rid of driver_api.go - yes, we want to dexctl to use APIs...but not these APIs. As far as I kow no one is using dexctl in this mode

2) get rid of client_resource apis - no one except api\_driver  was using them, and no one is using that

3) when implementing native client, special case the /token endpoint to never give out GrantTypeClientCreds for native clients

4) in the future when we implement methods in the worker api for udpating client metdata by clients use an "registration access token" rather than client_id/secret pairs

bobbyrullo:
Also remove dexctl api_driver
fixes #467


ericchiang:
Looks like a flake. Retesting.

ericchiang:
Never mind. The tests are indeed failing on go 1.5.4.

ericchiang:
lgtm

bobbyrullo:
w00t thanks

ericchiang:
Delete all the code!

PR #471, opened 2016-06-17
Implement Public Clients

bobbyrullo:
fixes https://github.com/coreos/dex/issues/469


client/manager/manager.go
```diff
@@ -192,3 +209,36 @@ func (m *ClientManager) addClientCredentials(cli *client.Client) error {
...
+		}
+	} else {
+		if len(cli.Metadata.RedirectURIs) < 1 {
+			return client.ErrorMissingRedirectURI
+		}
+	}
+
+	err := cli.Metadata.Valid()
+	if err != nil {
+		fmt.Println("manager throwing validation error")
```
ericchiang: left over log statement

db/client.go
```diff
@@ -168,7 +190,7 @@ func isAlreadyExistsErr(err error) bool {
 
 func (r *clientRepo) New(tx repo.Transaction, cli client.Client) (*oidc.ClientCredentials, error) {
 	cim, err := newClientModel(cli)
-
+	fmt.Println("XXX MODEL  new public: ", cli.Public)
```
ericchiang: another one

client/manager/manager.go
```diff
@@ -192,3 +209,36 @@ func (m *ClientManager) addClientCredentials(cli *client.Client) error {
...
+
+func validateClient(cli client.Client) error {
+	if cli.Public {
+		if len(cli.Metadata.RedirectURIs) > 0 {
+			return client.ErrorPublicClientRedirectURIs
+		}
+		if cli.Metadata.ClientName == "" {
+			return client.ErrorPublicClientMissingName
+		}
+		cli.Metadata.RedirectURIs = []url.URL{
```
ericchiang: Is this necessary to make valid pass later?

client/manager/manager.go
bobbyrullo: WIP Means work in progress :)

client/manager/manager.go
bobbyrullo: well, metadata.Valid() does a whole bunch of other checks that we'd like validated right?

ericchiang:
Got a vet error
```
Checking govet...
gopath/src/github.com/coreos/dex/client/manager/manager.go:237: github.com/coreos/dex/client.ValidationError composite literal uses unkeyed fields
```

client/manager/manager.go
```diff
@@ -192,3 +214,35 @@ func (m *ClientManager) addClientCredentials(cli *client.Client) error {
...
+		}
+	} else {
+		if len(cli.Metadata.RedirectURIs) < 1 {
+			return client.ErrorMissingRedirectURI
+		}
+	}
+
+	err := cli.Metadata.Valid()
+	if err != nil {
+		return client.ValidationError{err}
```
ericchiang: This is the vet error btw. Needs to be `client.ValidationError{Err:err}`

bobbyrullo: yeah, I just don't know why mine didn't catch it.

admin/api.go
```diff
@@ -165,6 +163,12 @@ func (a *AdminAPI) GetConnectors() ([]connector.ConnectorConfig, error) {
 }
 
 func mapError(e error) error {
+	switch t := e.(type) {
+	case client.ValidationError:
```
ericchiang: I really dislike type switching for a specific error since it makes the validateClient code really brittle. Would be nice if this was formalized more, such as introducing HTTP specific [error behavior](http://dave.cheney.net/2014/12/24/inspecting-errors). I suppose that's outside of the scope of this PR though.

client/manager/manager.go
```diff
@@ -192,3 +214,35 @@ func (m *ClientManager) addClientCredentials(cli *client.Client) error {
...
+		}
+	} else {
+		if len(cli.Metadata.RedirectURIs) < 1 {
+			return client.ErrorMissingRedirectURI
+		}
+	}
+
+	err := cli.Metadata.Valid()
+	if err != nil {
+		return client.ValidationError{Err: err}
```
ericchiang: Please document that this error it tied to the UI.

db/client.go
```diff
@@ -44,6 +49,16 @@ func newClientModel(cli client.Client) (*clientModel, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if cli.Public {
+		// Metadata.Valid(), and therefore json.Unmarshal(metadata) complains
```
ericchiang: Shouldn't we fix metadata.valid to understand public clients? Maybe add a TODO?

bobbyrullo: I don't think so, because Metadata is in go-oidc.

client/manager/manager.go
bobbyrullo: It doesn't make sense for method to document the users of the method.

db/client.go
bobbyrullo: ...but maybe we could break out metadata.Valid() into different parts, or have it more configurable.

admin/api.go
bobbyrullo: Would it help if I documented manager.New() as to which errors can be returned?

client/manager/manager.go
ericchiang: Users shouldn't depend on it then :)

admin/api.go
ericchiang: Maybe put it on ValidationError? Whichever works best. If it's tested that might be enough.

PR #483, opened 2016-06-21
Clean up LDAP Connector

### LINKED ISSUE #481: LDAP connector doesn't work without searchBeforeAuth
Without search, the LDAP connector never queries for the user's email [0]. Because it returns an identity without one, registration always fails.

[0] https://github.com/coreos/dex/blob/a7b860b9c2dba69b7154feacbd0adbe0c52e2591/connector/connector_ldap.go#L459-L467

ericchiang:
The first commit cleans up the LDAP connector a bit. A follow up one will add the groups implementation.

Cleanup includes:
* Remove some unlikely to be used fields to help configurability.
  * Combined "serverHost" and "serverPort" into "host"
  * Remove "timeout" (just default to 30 seconds).
  * Remove "maxIdleConn" will add it back if users feel the need
    to control the number of cached connections.
  * Remove "trustedEmailProvider" (just always trust).
  * Remove "skipCertVerification" you can't make this connector
    ingore TLS errors.
* Fix configs that don't search before bind (fixes #481).
* Add more examples to Documentation
* Refactor LDAPPool Acquire() and Put() into a Do() function which
  always does the flow correctly.
* Added more comments and renamed some functions.

Edit:

* Moved methods on LDAPIdentityProvider to the LDAPConnector

### DIFF
--- a/Documentation/connectors-configuration.md
+++ b/Documentation/connectors-configuration.md
@@ -15,7 +15,7 @@ The dex connector configuration format is a JSON array of objects, each with an 
         "id": "Google",
         "type": "oidc",
         ...<<more config>>...
-    } 
+    }
 ]
 ```
 
@@ -45,11 +45,8 @@ The configuration for the local connector is always the same; it looks like this
 This connector config lets users authenticate with other OIDC providers. In addition to `id` and `type`, the `oidc` connector takes the following additional fields:
 
 * issuerURL: a `string`. The base URL for the OIDC provider. Should be a URL with an `https` scheme.
-
 * clientID: a `string`. The OIDC client ID.
-
 * clientSecret: a `string`. The OIDC client secret.
-
 * trustedEmailProvider: a `boolean`. If true dex will trust the email address claims from this provider and not require that users verify their emails.
 
 In order to use the `oidc` connector you must register dex as an OIDC client; this mechanism is different from provider to provider. For Google, follow the instructions at their [developer site](https://developers.google.com/identity/protocols/OpenIDConnect?hl=en). Regardless of your provider, registering your client will also provide you with the client ID and secret.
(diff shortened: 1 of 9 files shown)


ericchiang:
Groups implementation added. Need to figure out how to have this work with refresh tokens (hopefully without too much refactoring).

ericchiang:
This PR has been changed to just include the cleanups to the LDAP connector.

New docs here: https://github.com/ericchiang/dex/blob/c629cfaf1d6298fe0f532f80e48a4c852c86cdac/Documentation/connectors-configuration.md#ldap-connector

Ready for review.

ericchiang:
Woops, left some docs unfinished.

ericchiang:
Ready for review

Documentation/connectors-configuration.md
```diff
@@ -136,83 +131,85 @@ Here's an example of a `bitbucket` connector; the clientID and clientSecret shou
...
     }
 ```
 
+"searchFilter" is a format string expanded in a similar manner as "bindTemplate". If the user "janedoe" attempts to authenticate, the connector will run the following query using the service account credentials.
+
+```
+(&(objectClass=person)(uid=janedoe))
+```
+
+If the search finds an entry
```
sym3tri: dangling sentence fragment

connector/connector_ldap.go
```diff
@@ -28,30 +30,62 @@ const (
...
+	ID string `json:"id"`
+
+	// Host and port of ldap service in form "host:port"
+	Host string `json:"host"`
+
+	// UseTLS indicates that the connector should use the TLS port.
+	UseTLS bool `json:"useTLS"`
+	UseSSL bool `json:"useSSL"`
+
+	// CA File indicates a trusted.
```
sym3tri: a trusted what?

connector/connector_ldap.go
```diff
@@ -126,15 +163,18 @@ func (cfg *LDAPConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, tpls *t
 		}
 	}
 
-	if cfg.Timeout != 0 {
-		ldap.DefaultTimeout = cfg.Timeout * time.Millisecond
+	if cfg.Host == "" {
+		// For backward compatibility construct host form old fields.
+		cfg.Host = fmt.Sprintf("%s:%d", cfg.ServerHost, cfg.ServerPort)
```
sym3tri: What if `cfg.ServerHost` is empty. Check for that too?

sym3tri:
found a few minor things, otherwise LGTM

connector/connector_ldap.go
ericchiang: That's checked in the next line implicitly SplitHostPort. There previously wasn't a check if you didn't provide ServerHost, so I omitted one here. I'll add it back in though and be explicit.

ericchiang:
:+1: updated

PR #484, opened 2016-06-24
scope: don't validate empty scopes

bobbyrullo:
If an empty scope is somehow passed along, it shouldn't be validated
when checking refresh token scope.


ericchiang:
Could we add a scopes.Parse method as well? It's probably better to have this centralized than to just do `strings.Split(scopeStr, " ")` everywhere.

bobbyrullo:
agreed but can we do it another PR?

ericchiang:
Yeah totally.

PR #486, opened 2016-06-27
build with go 1.6 and actually push to quay

ericchiang:


.travis.yml
```diff
@@ -39,7 +39,7 @@ deploy:
   skip_cleanup: true
   on:
     branch: master
-    go: '1.5.3'
+    go: '1.6.2'
```
chancez: Why not build with both 1.6.2 and 1.5.3?

PR #487, opened 2016-06-28
*: depricate --email-from flag and move to email config files

### LINKED ISSUE #389: dex-worker: Log warning or fatal exit if -email-from is not set
I just spent a good amount of time figuring out why emails are not firing. I recently migrated from 0.1 or something to 0.3 which was painless. I dont think this `-email-from` was needed in 0.1, don't know when it was introduced. It was then reported to me that the email verification etc. emails were not working.

Anyhow, seemingly no emails are sent if this cmd line flag is not set, even if you specify a valid email config file. I assumed it would be enough as the username is the same as the email address, but I suppose this cant be assumed with all smtp servers (?!).

No warnings during startup. I tried the reset password functionality that is when I saw 

```ERROR: error sending password reset email gomail: could not send email 1: mail: no address:```

Not sure where this is coming from but started looking at the cmd line params after that and once `-email-from` was set it started working.

Either fatal exit the app or print a warning/error to log if a valid (non fake type) emailer config is provided but the `-email-from` is not defined.

The cmd line -h documentation should probably also mention its mandatory if you want emails to work.

### PR DESCRIPTION
**Pull request by ericchiang**
Still need to test with a full installation.

Closes #389

### DIFF
--- a/Documentation/email-configuration.md
+++ b/Documentation/email-configuration.md
@@ -20,6 +20,7 @@ set `auth` to `plain` and specify your username and password.
     "host": "smtp.example.org",
     "port": 587,
     "auth": "plain",
+    "from": "postmaster@example.com",
     "username": "postmaster@example.org",
     "password": "foo"
 }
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

**Comment by ericchiang** [comment]
Seems to work. Ready for review @sym3tri

File: Documentation/email-configuration.md [hunk 1]
```diff
@@ -20,6 +20,7 @@ set `auth` to `plain` and specify your username and password.
     "host": "smtp.example.org",
     "port": 587,
     "auth": "plain",
```
sym3tri: While you're in here can you make this field not required. Currently if it's omitted Dex explodes, but we only support 'plain' and a default. Let's just default to the default if it's not provided.

sym3tri:
LGTM

Documentation/email-configuration.md
ericchiang: Changed this to use auth if "username" and "password" fields are provided.

Also going to let the port be provided in the "host" field.

ericchiang:
Since all changes are backward compatible I'm going to merge unless there are objections.

cc @sym3tri

PR #488, opened 2016-06-28
*: update mailgun-go to remove dependency on deleted simplehttp pkg

ericchiang:
github.com/mbanzon/simplehttp has been deleted from github which mailgun-go depended on. Update mailgun-go to a version which removes this dependency.


ericchiang:
Merging on green

PR #494, opened 2016-06-28
return 409 status code on duplicate email errors

### LINKED ISSUE #492: Return specific error/code from create admin API call
It would be great to get an HTTP 409, and/or a better error when attempting to create a user that already exists. Currently I get the following type back from the generated client:
```
&googleapi.Error{Code:400, Message:"", Body:"{\"error\":\"bad_request\",\"error_description\":\"Email already in use.\"}", Errors:[]googleapi.ErrorItem(nil)}
```

Additionally, having a function I can use like `IsErrorAlreadyExists(err error) bool` would be the preferred way to check, rather than having to guess how to inspect the error type.

ericchiang:
Closes #492

cc @chancez

### DIFF
--- a/admin/api.go
+++ b/admin/api.go
@@ -81,7 +81,7 @@ var (
 		client.ErrorInvalidClientSecret: errorMaker("bad_request", "Secret must be a base64 encoded string", http.StatusBadRequest),
 
 		user.ErrorNotFound:       errorMaker("resource_not_found", "Resource could not be found.", http.StatusNotFound),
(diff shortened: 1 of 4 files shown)


chancez:
Whats the expected way to check for this error then? Should I be type asserting for googleapi.Error and then checking the HTTP code, or?

ericchiang:
I think type asserting for googleapi.Error works or just construct your own request using net/http. I'm not super familiar with the googleapi package.

ericchiang:
Going to merge if this looks good to you.

chancez:
This looks good.

PR #496, opened 2016-06-28
return 409 for duplicate client ids

### LINKED ISSUE #495: Cannot tell if client being created already exists using admin API
When I create a client using the admin API the error I get back if the client already exists is extremely generic:
```
*googleapi.Error verbose: &googleapi.Error{Code:500, Message:"", Body:"{\"error\":\"server_error\"}"
```

This is particularly troublesome as I'm providing my own clientID and clientSecret, so I have no way to check  if it exists already, or to see if it exists when creating (ie: http 409 if it already exists with the given ID).

ericchiang:
Closes #495


ericchiang:
cc @chancez

PR #500, opened 2016-06-30
Added API endpoint for account authentication with login and password

niespodd:
Created JSON API endpoint for user authentication with username and password.

### DIFF
--- a/connector/connector_bitbucket.go
+++ b/connector/connector_bitbucket.go
@@ -39,7 +39,7 @@ func (cfg *BitbucketConnectorConfig) ConnectorType() string {
 	return BitbucketConnectorType
 }
 
-func (cfg *BitbucketConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, tpls *template.Template) (Connector, error) {
+func (cfg *BitbucketConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, _ NewSessionFunc, tpls *template.Template) (Connector, error) {
 	ns.Path = path.Join(ns.Path, httpPathCallback)
 	oauth2Conn, err := newBitbucketConnector(cfg.ClientID, cfg.ClientSecret, ns.String())
 	if err != nil {
--- a/connector/connector_github.go
+++ b/connector/connector_github.go
@@ -39,7 +39,7 @@ func (cfg *GitHubConnectorConfig) ConnectorType() string {
 	return GitHubConnectorType
 }
 
-func (cfg *GitHubConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, tpls *template.Template) (Connector, error) {
+func (cfg *GitHubConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, _ NewSessionFunc, tpls *template.Template) (Connector, error) {
 	ns.Path = path.Join(ns.Path, httpPathCallback)
 	oauth2Conn, err := newGitHubConnector(cfg.ClientID, cfg.ClientSecret, ns.String())
 	if err != nil {
--- a/connector/connector_ldap.go
+++ b/connector/connector_ldap.go
@@ -120,7 +120,7 @@ type LDAPConnector struct {
 
 const defaultPoolCheckTimer = 7200 * time.Second
 
-func (cfg *LDAPConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, tpls *template.Template) (Connector, error) {
+func (cfg *LDAPConnectorConfig) Connector(ns url.URL, lf oidc.LoginFunc, _ NewSessionFunc, tpls *template.Template) (Connector, error) {
 	ns.Path = path.Join(ns.Path, httpPathCallback)
 	tpl := tpls.Lookup(LDAPLoginPageTemplateName)
 	if tpl == nil {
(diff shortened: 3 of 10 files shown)


ericchiang:
Hi @niespodd thanks for your proposal.

Unfortunately, Dex doesn't support this because users can be backed by different identity stores, such as Google or GitHub. For these cases Dex doesn't know your password so it can't perform the logic in your proposed endpoint.

Also as the OAuth2 provider, Dex needs to do basic verification that the end user consents to the permissions a client is requesting, possibly perform two factor authentication, or do session management. These are all impossible if you do this instead.

This has been discussed at length here[0].

At some point we hope to provide backing storage specific APIs. For instance LDAP or GitHub specific calls to hook into organizations. This will be part of larger changes.

Today, the only way we'd entertain this endpoint if it was tied into some one time or [app specific password](https://support.google.com/accounts/answer/185833?hl=en) process. For all of these we recommend a proposal first.

[0] https://groups.google.com/forum/#!topic/dex-dev/KoY0JegNcr4

LEW21:
Hi @ericchiang

I'm @niespodd's coworker. We're using Dex as a internal auth provider in our application, without allowing other parties to register new clients. Our final goal is to never show Dex's HTML pages to our users, and handle everything through our Single-Page App, using Dex as a backend. We're using two connectors: local and Google, and in Google's case with the automatic-registration - it works quite well, users see only the Google's UI.

In local's case, we're forced to redirect them to Dex's login page. We'd prefer to ask for username and password in the SPA, and send them to Dex with AJAX.

BTW, we also want to use the API to automatically log in our app into Dex as an admin user, so it'll be able to access the full worker API.

This patch adds a new endpoint only to the local connector - it's being registered in LocalConnector.register, and the URL is /auth/local/api/login. Other connectors are unaffected.

If you believe that this endpoint could be abused by evil 3rd-party apps - we're OK with limiting it to admin clients.

ericchiang:
Hi @LEW21 I appreciate the use case, but Dex is an OAuth2 provider and as such has requirements about needing to control the user's experience, just like Google. We do provide the ability to customize the actual HTML through our templating system (for example account.tectonic.com), does this work for you?

Though it does seem like you don't actually want OAuth and might be better served by a more customized user store.

LEW21:
That's true, we don't really need OAuth. However, we want to use Dex to get Google, and possibly other connectors in the future; JWT session tokens; key rotation; refresh tokens, and their revocation; and all other cool things that are or will be possible with Dex in the future.

And we're also interested in contributing some new features later on - for example a Facebook connector.

The templating systems works, and we're using it currently, but it's not the optimal user experience we want to achieve. We don't want to redirect our users out of the SPA if that's not absolutely necessary. We want to provide a small area somewhere on the page that lets them input their credentials, and log them in, without reloading the page at all.

ericchiang:
It's true that, and maybe we should work on pulling out the functionality. However, if you're not using Dex as an OAuth provider, you'll have to come to us for any feature you want in your app, which I can't promise we'll be able add (this is a perfect example).

To respond to this PR directly, we don't want to add username/password login for all the reasons that have been stated in my first comment. In addition that endpoint would step around all of the benefits you've listed "JWT session tokens; key rotation; refresh tokens, and their revocation".

LEW21:
It's not like we don't want to use anything from OAuth at all.

This endpoint returns the same code that is being passed as a query parameter to the client's Redirect URL. And we're using it - in the same OAuth way as in the normal flow - to get the JWT token and the refresh token. SPA does not store user's login and password anywhere, so when the refresh token will be revoked - the user will get logged out.

I don't think we're going to need any more strange features than this one and a similar way to register a user with an API call.

And about the reasons from the first post:

> Unfortunately, Dex doesn't support this because users can be backed by different identity stores, such as Google or GitHub. For these cases Dex doesn't know your password so it can't perform the logic in your proposed endpoint.

This affects only the local connector, which supports only passwords.

> Also as the OAuth2 provider, Dex needs to do basic verification that the end user consents to the permissions a client is requesting, 

We're our own client, so we want to grant ourselves all permissions, without anything consent-related. That's why I suggested limiting the API to the admin clients.

> possibly perform two factor authentication

If we'd need it, we would simply use the redirect-based API. REST-based API should simply disallow logging in in such cases.

> or do session management. These are all impossible if you do this instead.

We don't save the password after logging in, so Dex can still manage sessions. (Of course, this also requires limiting the API to the trusted (admin) clients.)

ericchiang:
> This affects only the local connector, which supports only passwords.

Yes, we dislike users writing backend specific code. Passwords are an abstraction that don't work for other backends.

> We're our own client, so we want to grant ourselves all permissions, without anything consent-related. That's why I suggested limiting the API to the admin clients.

We dislike the idea of admin clients and don't want to tie them to more functionality outside of OAuth. In the future we may move away from using OAuth client credentials for accessing the API.

> If we'd need it, we would simply use the redirect-based API. REST-based API should simply disallow logging in in such cases.

You're essentially proposing grant_type=password, which isn't supported with OpenID Connect, isn't implemented by most OAuth2 provider, and still requires confirmation screens.

> We don't save the password after logging in, so Dex can still manage sessions. (Of course, this also requires limiting the API to the trusted (admin) clients.)

Whether or not your app provides best practices with passwords, we don't want to tie Dex's security to outside applications. Dex attempts to provide security guarantees and providing a password endpoint blows a hole in our efforts.

ericchiang:
For the many reasons listed I'm going to close this. We don't want to support password validation from outside applications.

Issue #480, opened 2016-06-21
LDAP connector doesn't escape input


ericchiang:
A hypothetical setup of the LDAP connector is to auth without search against a specific directory.

```
    {
        "type": "ldap",
        "id": "ldap",
        "host": "127.0.0.1:389",
        "useTLS": true,
        "baseDN": "cn=users,cn=accounts,dc=auth,dc=example,dc=com",
        "bindTemplate": "uid=%u,%d"
    }
```

If I enter the username "janedoe" the result bind becomes

```
uid=janedoe,cn=users,cn=accounts,dc=auth,dc=example,dc=com
```

However I can trivially access sub directories by including them in my username. E.g. "janedoe,cn=admins" would become

```
uid=janedoe,cn=admins,cn=users,cn=accounts,dc=auth,dc=example,dc=com
```

Investigate proper escape strategies for LDAP DN entries.

ericchiang:
Further reading: http://www.rlmueller.net/CharactersEscaped.htm

PR #510, opened 2016-07-15
Add groups scope and LDAP implementation

### LINKED ISSUE #432: LDAP: add claims based on groups
It would be nice to put a groups claim into the users identity. The most obvious place to start would be letting someone who is using an LDAP server with an RFC2307(bis) schema to be able to take their posixGroups list and put it into a claim as a JSON list.

cc @fnordahl

ericchiang:
This commit adds the new "groups" scope and an LDAP implementation. Tests are largely updated to accomidate changes to function signatures. Need to add some tests that directly try the groups functionality.

cc @sym3tri @chancez 

Closes #432
Updates #480


session/session.go
```diff
@@ -65,5 +69,9 @@ func (s *Session) Claims(issuerURL string) jose.Claims {
 	if s.Nonce != "" {
 		claims["nonce"] = s.Nonce
 	}
+	log.Println(s.Scope.HasScope(scope.ScopeGroups), s.Groups)
```
ericchiang: woops, need to remove this

ericchiang:
Oh, and cc @johnw188 per our conversation from #sig-auth.

connector/connector_ldap.go
```diff
@@ -447,8 +484,10 @@ func (c *LDAPConnector) Identity(username, password string) (*oidc.Identity, err
 	if c.searchBeforeAuth {
 		err = c.ldapPool.Do(func(conn *ldap.Conn) error {
 			if err := conn.Bind(c.searchBindDN, c.searchBindPw); err != nil {
-				// Don't wrap error as it may be a specific LDAP error.
-				return err
+				if !invalidBindCredentials(err) {
```
chancez: Why remove all these calls to `invalidBindCredentials()`?

server/server.go
```diff
@@ -306,6 +306,15 @@ func (s *Server) NewSession(ipdcID, clientID, clientState string, redirectURL ur
 	return s.SessionManager.NewSessionKey(sessionID)
 }
 
+func (s *Server) connector(id string) (connector.Connector, bool) {
+	for _, c := range s.Connectors {
```
chancez: I know everywhere else uses the slice currently and the number of connectors in generally small, but it might be a good idea to consider adding a map[string]Connectors for an id->connector mapping. Doesn't have to be now though.

connector/connector_ldap.go
ericchiang: Remove? I'm confused, those calls aren't being removed by this PR.

server/server.go
ericchiang: :+1: yeah might be more explicit. And yeah, going to hold off for another PR.

connector/interface.go
```diff
@@ -60,6 +60,11 @@ type ConnectorConfig interface {
 	Connector(ns url.URL, loginFunc oidc.LoginFunc, tpls *template.Template) (Connector, error)
 }
 
+// Groups is an optional interface implemented by a Connector.
+type Groups interface {
```
chancez: Nit: not a fan of the naming of the interface.

connector/interface.go
```diff
@@ -60,6 +60,11 @@ type ConnectorConfig interface {
...
+	Groups(fullUserID string) ([]string, error)
```
chancez: Do we expect this interface to only be implemented by connectors? Perhaps it would be best to embed the connector interface into this one to make that explicit.

chancez: Actually, thinking that through more: it _seems_ better to just add this method to the existing interface, and instead make other connectors which do not support grouping return an error, or something. What do you think?

connector/interface.go
ericchiang: Maybe: "GroupsConnector"? Idk, derivatives of "Directory" or "Membership" seemed a little odd or specific.

server/server.go
```diff
@@ -572,8 +629,11 @@ func (s *Server) RefreshToken(creds oidc.ClientCredentials, scopes scope.Scopes,
 	now := time.Now()
 	expireAt := now.Add(session.DefaultSessionValidityWindow)
 
-	claims := oidc.NewClaims(s.IssuerURL.String(), user.ID, creds.ID, now, expireAt)
-	user.AddToClaims(claims)
+	claims := oidc.NewClaims(s.IssuerURL.String(), usr.ID, creds.ID, now, expireAt)
+	usr.AddToClaims(claims)
+	if groups != nil {
```
chancez: Maybe it would be best to always assign to claims if groups is being asked for, this would mean the key exists, but if there are no groups the value would just be an empty list.

connector/connector_ldap.go
chancez: Oh sorry, still early, eyes are playing tricks on me.

connector/interface.go
ericchiang: Like being able to differentiated between connectors that support this and a call returning an error. Would prefer to either embed the Connector interface or adjust the naming to make it more explicit. 

Would like to use this pattern later for connectors that accept passwords (LDAP and local). So another interface like this

```
type PasswordConnector interface {
    Login(username, password) (identity *oidc.Identity, correctCreds bool, err error)
}
```

Then detect connectors that accept passwords so the connectors themselves stop rendering HTML.

chancez: I see. Yeah, lets keep them separate, and just try to find better naming. I'm not 100% on embedding the other interface either now, because I can imagine not every grouping implementation relies on a particular connector (the group implementation works for _all_ connectors, etc).

ericchiang:
As per in-person conversation with @chancez we might want to have some way for a connector to pass data between logging the user in and getting the groups. An obvious use case is access tokens. E.g. I login to GitHub, and need the access token from the login to get my groups.

Going to defer that point to a different PR, as it involves refactoring the oidc.LoginFunc.

chancez:
👍 This looks fairly solid to me overall.

ericchiang:
Going to squash and merge unless there are objections.

wyattanderson:
Is it possible to restructure this so that group searches can be performed with the user's bind as opposed to requiring a service account? It would simplify deployment of this if we could avoid setting up a service account just for Dex.

(tacking this on to this issue for context; can open a separate issue if desired)

ericchiang:
@wyattanderson not currently. For refresh tokens, we insist on updating the groups information in the returned ID Token, to prevent stale authentication data.

When a client tries to refresh an ID Token, we don't have the users credentials at that point. This is a problem for some possible groups implementations (see #514), but is especially problematic for LDAP since we'd have to hold onto the user's plain text password.

wyattanderson:
Oh yeah, totally forgot about the refresh token situation which makes perfect sense. Thanks!

Also, thanks in general for this project. It's saving me months of work.

ericchiang:
Glad to hear!

PR #511, opened 2016-07-18
*: update Go version used to build Dex

ericchiang:
Due to today's security release of Go update Dex to build with the latest version.[0]

Don't know if this requires dropping Go1.5 and older releases of Go1.6 as I can't say that the vulnerability doesn't impact Dex. Going to defer to @gtank.

[0] https://groups.google.com/forum/#!topic/golang-announce/7jZDOQ8f8tM

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,7 @@ services:
 language: go
 
 go:
-  - 1.5.4
-  - 1.6.2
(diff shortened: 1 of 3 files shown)


ericchiang:
Closing. I don't think this bug impacts us.

PR #517, opened 2016-07-21
cmd: update flag defaults for dev mode

squat:
Updates flags in dex-worker for no-db mode to work with default static
files.


ericchiang:
lgtm

PR #520, opened 2016-07-22
backend: add absolute path to template links

squat:
fixes: #502

WIP

Add a new template function called `absPath` that converts relative
paths to absolute paths based on the issuerURL. The template function
can be used in templates like:
````html
<a href="{{ .MyRelativePath | absPath }}">
````
or simply:
````html
<a href="{{ absPath }}/myrelativepath">
````

PR #521, opened 2016-07-25
Allow dex to work at non base url

### LINKED ISSUE #502: OpenID Provider Configuration Information is stick to root path (`/`)
From [OpenID Connect specs](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig):
OpenID Providers supporting Discovery MUST make a JSON document available at the path formed by concatenating the string /.well-known/openid-configuration to the Issuer.

I just tried to run dex-worker with these parameters:
`/bin/dex-worker --listen=http://0.0.0.0:5556/auth --issuer=http://127.0.0.1:5556/auth ...`

The dex-worker does do as expected and listens & serves the authentication endpoints under `/auth` path. However, apparently, the provider's configuration information (".well-known/openid-configuration" endpoint) is still located under the root path (That is, http://127.0.0.1:5556/.well-known/openid-configuration), and this violates the OpenID Connect clause that is quoted above.

Thanks guys :)

ericchiang:
This change updates #520 by having the handlers register using the issuerURL. It also updates the development environment by having the dex-worker and examle-app default to a non-base URL. This will hopefully catch any future bugs that assume operating from a root path.

Going to have to manually test all the connectors and registration flows :( Until then, do not merge.

Closes #502


ericchiang:
@xaka has noted that most proxies trim the prefix of all paths. For instance the following nginx config trims `/test/` from the path before

```
location /test/ {
  proxy_pass http://localserver.com/;
}
```

Therefore, while URLs rendered in the HTML need to be absolute from the issuer URL, the actual paths that handlers are registered on shouldn't be.

@alon-argus was wondering if you had any comments on this point.

ericchiang:
Have gone through the local and gmail flows with registration and email verification. Think I got all the edge cases.

ericchiang:
Going to do another round of manual testing and add some documentation to the issuer flat.

ericchiang:
@chancez or @xaka want to take a look at this? I've run through GitHub, Google,  LDAP, and local connector and they all work.

xaka:
tested locally with http proxy, etc. so LGTM 👍

PR #522, opened 2016-07-25
connector: fix path that connectors listen on

ericchiang:
When Dex uses a non-root issuer URL, it current assumes that all
path prefixes will be trimmed by an upstream proxy (e.g. nginx).
This means that all paths rendered in HTML will be absolute to the
prefix, but the handlers still listen at the root.

Connectors are currently the only component that registers at a
non-root URL. Make this conform with the rest of Dex by having the
server determine the path the connector listens as rather than the
connector itself.

Updates #502

Note that this is part of a larger conversation about what it means
for Dex to be at a non-root URL path. See #521.

@xaka can you test this to ensure it works with your setup?

### DIFF
--- a/connector/connector_ldap.go
+++ b/connector/connector_ldap.go
@@ -251,9 +251,9 @@ func (c *LDAPConnector) LoginURL(sessionKey, prompt string) (string, error) {
 	return path.Join(c.namespace.Path, "login") + "?" + enc, nil
 }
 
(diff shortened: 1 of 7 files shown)


sym3tri:
LGTM

ericchiang:
One sec. The fix doesn't quite work.

Issue #504, opened 2016-07-07
Getting email from JWT issued by Azure AD


summerwind:
Is there a way to change the claim to get the email in OIDC Connector?

I'd like to use dex with Azure AD. But JWT issued by Azure AD does not contain a "email" claim. Email is in "upn" or "unique_name" claims. So currently dex does not get the email fron JWT issued by Azure AD.

The information of claims in Azure AD is here:
https://azure.microsoft.com/en-us/documentation/articles/guidance-multitenant-identity-claims/

sym3tri:
The root issue here is that Azure has decided to use non-standard claims in their JWTs[1]. They should use `email` b/c that is what's in the spec. Claims named "upn" and "unique_name" are not in the spec. 

This opens up a bigger philosophical discussion over weather or not we support non-standard OIDC providers, and if so how we do overrides. I can think of two possibilities:
  1. A map of upstream provider cliaim ID to Dex claim id
  2. We stuff a map of all upstream claims into a new claim (this may have other implications)

Also this is handled by go-oidc and should be moved here: https://github.com/coreos/go-oidc

[1] http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

summerwind:
Thank you for your comment. As you said, Azure's JWT contains the non-standard claims. This is the root issue.

According to the documentation, there is a way to normalize claims if we use multiple IdP as dex.
https://azure.microsoft.com/en-us/documentation/articles/guidance-multitenant-identity-claims/#claims-transformations

> Claims normalization, or making claims consistent across users. This is particularly relevant if you are getting claims from multiple IDPs, which might use different claim types for similar information. For example, Azure AD sends a "upn" claim that contains the user's email. Other IDPs might send an "email" claim. The following code converts the "upn" claim into an "email" claim:

Adding normalization feature to go-oidc will make more complicated. I thought that  there is a way to adding the settings to specifying the claim name to be normalized to OIDC Connector. My PoC code is below:
https://github.com/summerwind/dex/blob/5547dc874b34d620d83062aba1e45f2ffa9a7015/connector/connector_oidc.go#L155-L172

What are your thoughts on?

ericchiang:
Sorry for the delayed response. I've been out of the office.

I'm actually fine with this for non-required claims.

```
{
    "type": "oidc",
    "id": "azure",
    "issuerURL": "https://accounts.example.com",
    "clientID": "$CLIENT_ID",
[...]

PR #527, opened 2016-08-01
connector: add azure connector

summerwind:
As we discussed in #504, Azure AD contains a non-standard claim. This PR adds the Azure connector to support non-standard claims.


ericchiang:
Thanks for this!

Part of adding a connector is adding additional documentation to `Documentation/connectors-configuration.md`. Can you do that? It also helps me test your change.

summerwind:
Thanks for the comment! I added a document for the `azure` connector. 
To test the azure connector, requires your own Azure Active Directory.
https://azure.microsoft.com/en-us/services/active-directory/

connector/connector_azure.go
```diff
@@ -0,0 +1,145 @@
...
+			log.Errorf("Unable to construct claims: %v", err)
+			q.Set("error", oauth2.ErrorUnsupportedResponseType)
+			q.Set("error_description", "unable to construct claims")
+			redirectError(w, errorURL, q)
+			return
+		}
+
+		// ID Token issued by Azure AD does not contain a email claim.
+		// So we need to copy the value from upn claim to email claim internally.
+		upn, ok, err := claims.StringClaim("upn")
```
ericchiang: Hmm, it seems like this is the only thing unique to this from the OpenID Connect plugin. Could we share code a bit more between those two or even just add a flag to the existing one?

ericchiang:
Great. I'll sign up for an account tomorrow.

Can you take a look at my inline comment? I dislike just copying so much code from the existing OpenID Connect connector. Would like to see that addressed.

ericchiang:
@summerwind yes I would be in favor of a more minimal change. We can still provide azure specific docs like we do with flavors of LDAP[0]. I still have to read a bit into the azure implementation to give good advice on which route to take.

[0] https://github.com/coreos/dex/blob/master/Documentation/connectors-configuration.md#example-searching-a-freeipa-server-with-groups

summerwind:
@ericchiang I agree to take the way to become a minimal change. Please reply if you have any good advice!

summerwind:
Any updates? Shall I modify/improve it somehow?

ericchiang:
Sorry, I'd love to see this connector share code with the existing OpenID Connect implementation. Probably through some sort of configuration flag. Does that sound reasonable?

summerwind:
@ericchiang Sure! I will open another PR when ready.

ericchiang:
Just update this one :)

summerwind:
@ericchiang I added `emailClaim` setting to OIDC Connector as an alternative way. Could you review please?

ericchiang:
The last commit looks really good. Could you clean up your commit history by removing the unused commits? If you have issues with git feel free to open another PR.

summerwind:
@ericchiang Thanks for the review. I have cleaned up my commits and updated PR.
Please tell me if you have another concern ;)

ericchiang:
lgtm

PR #529, opened 2016-08-01
ClientCredentials flow in basic User operations (createUser, getUser, resendInvitationMail)

### LINKED ISSUE #528: Allow admin clients to create users
Hi,

This is the discussion on google group:
https://groups.google.com/forum/#!topic/dex-dev/KWmif8CiCYE

adrianlop:
Fixes #528

### DIFF
--- a/cmd/dex-worker/main.go
+++ b/cmd/dex-worker/main.go
@@ -51,6 +51,9 @@ func main() {
 
 	enableClientRegistration := fs.Bool("enable-client-registration", false, "Allow dynamic registration of clients")
 
+	// Client credentials administration
+	apiUseClientCredentials := fs.Bool("api-use-client-credentials", false, "Forces API to authenticate using client credentials instead of ID token. Clients must be 'admin clients' to use the API.")
+
 	noDB := fs.Bool("no-db", false, "manage entities in-process w/o any encryption, used only for single-node testing")
 
 	// UI-related:
@@ -137,16 +140,17 @@ func main() {
 	}
 
 	scfg := server.ServerConfig{
-		IssuerURL:                *issuer,
-		TemplateDir:              *templates,
-		EmailTemplateDirs:        emailTemplateDirs,
-		EmailFromAddress:         *emailFrom,
-		EmailerConfigFile:        *emailConfig,
-		IssuerName:               *issuerName,
-		IssuerLogoURL:            *issuerLogoURL,
-		EnableRegistration:       *enableRegistration,
-		EnableClientRegistration: *enableClientRegistration,
-		RegisterOnFirstLogin:     *registerOnFirstLogin,
+		IssuerURL:                    *issuer,
+		TemplateDir:                  *templates,
+		EmailTemplateDirs:            emailTemplateDirs,
+		EmailFromAddress:             *emailFrom,
+		EmailerConfigFile:            *emailConfig,
+		IssuerName:                   *issuerName,
+		IssuerLogoURL:                *issuerLogoURL,
+		EnableRegistration:           *enableRegistration,
+		EnableClientRegistration:     *enableClientRegistration,
+		EnableClientCredentialAccess: *apiUseClientCredentials,
+		RegisterOnFirstLogin:         *registerOnFirstLogin,
 	}
 
 	if *noDB {
--- a/integration/user_api_test.go
+++ b/integration/user_api_test.go
@@ -84,6 +84,12 @@ var (
 	userGoodToken = makeUserToken(testIssuerURL,
 		"ID-1", testClientID, time.Hour*1, testPrivKey)
 
+	clientToken = makeClientToken(testIssuerURL,
+		testClientID, time.Hour*1, testPrivKey)
+
+	badClientToken = makeClientToken(testIssuerURL,
+		userBadClientID, time.Hour*1, testPrivKey)
+
 	userBadTokenNotAdmin = makeUserToken(testIssuerURL,
 		"ID-2", testClientID, time.Hour*1, testPrivKey)
 
@@ -97,7 +103,7 @@ var (
 		"ID-4", testClientID, time.Hour*1, testPrivKey)
 )
 
-func makeUserAPITestFixtures() *userAPITestFixtures {
+func makeUserAPITestFixtures(clientCredsFlag bool) *userAPITestFixtures {
 	f := &userAPITestFixtures{}
 
 	dbMap, _, _, um := makeUserObjects(userUsers, userPasswords)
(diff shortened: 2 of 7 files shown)


server/user.go
```diff
@@ -92,7 +94,7 @@ func (s *UserMgmtServer) authAPIHandle(handle authedHandle, requiresAdmin bool)
 			s.writeError(w, err)
 			return
 		}
-		if creds.User.Disabled || (requiresAdmin && !creds.User.Admin) {
+		if creds.User.ID != "" && (creds.User.Disabled || (requiresAdmin && !creds.User.Admin)) {
```
ericchiang: `creds.User.ID != ""` shouldn't be checked here. There should be two paths for this code.

```
if clientcred {

} else {

}
```

server/user.go
```diff
@@ -39,14 +39,16 @@ type UserMgmtServer struct {
 	jwtvFactory JWTVerifierFactory
 	um          *usermanager.UserManager
 	cm          *clientmanager.ClientManager
+	clientcred  bool
 }
 
-func NewUserMgmtServer(userMgmtAPI *api.UsersAPI, jwtvFactory JWTVerifierFactory, um *usermanager.UserManager, cm *clientmanager.ClientManager) *UserMgmtServer {
+func NewUserMgmtServer(userMgmtAPI *api.UsersAPI, jwtvFactory JWTVerifierFactory, um *usermanager.UserManager, cm *clientmanager.ClientManager, clientcred bool) *UserMgmtServer {
```
ericchiang: Can you rename clientcred something more detailed? Like `allowClientCredsAuth`?

cmd/dex-worker/main.go
```diff
@@ -51,6 +51,9 @@ func main() {
 
 	enableClientRegistration := fs.Bool("enable-client-registration", false, "Allow dynamic registration of clients")
 
+	// Client credentials administration
+	enableClientUserCreation := fs.Bool("enable-client-user-creation", false, "allow admin clients to create users using client-credentials flow")
```
ericchiang: I'd prefer this was a hard choice. e.g. Turning this on forced a user to only use client creds (and not an ID Token).

adrianlop:
PR updated. The flag is a hard choice now, it forces you to create users only using client credentials token, if enabled.

adrianlop:
updated PR. now it supports createUser, getUser, resendInvitationMail, not only createUser.

cmd/dex-worker/main.go
```diff
@@ -51,6 +51,9 @@ func main() {
 
 	enableClientRegistration := fs.Bool("enable-client-registration", false, "Allow dynamic registration of clients")
 
+	// Client credentials administration
+	forceClientUserCreation := fs.Bool("force-client-user-creation", false, "force create users using admin client-credentials")
```
ericchiang: This isn't just about client creation, it's about the entire layer of API access.

Let's do `--api-use-client-credentials`

server/server.go
```diff
@@ -95,6 +95,7 @@ type Server struct {
 
 	EnableRegistration       bool
 	EnableClientRegistration bool
+	EnableClientUserCreation bool
```
ericchiang:
```
// If enabled, causes the API to authenticate using client credentials instead of an ID token.
// Clients must be "admin clients" to use the API.
EnableClientCredentialAccess
```

user/api/api.go
```diff
@@ -116,8 +116,7 @@ func NewUsersAPI(userManager *usermanager.UserManager, clientManager *clientmana
 
 func (u *UsersAPI) GetUser(creds Creds, id string) (schema.User, error) {
 	log.Infof("userAPI: GetUser")
-
-	if !u.Authorize(creds) {
+	if creds.User.ID != "" && !u.Authorize(creds) {
```
ericchiang: `creds.User.ID != ""`? `u.Authorize` should just recognize if this flag is enabled and perform the correct authentication.

adrianlop:
thanks, PR updated with your comments.

cmd/dex-worker/main.go
```diff
@@ -51,6 +51,9 @@ func main() {
 
 	enableClientRegistration := fs.Bool("enable-client-registration", false, "Allow dynamic registration of clients")
 
+	// Client credentials administration
+	apiUseClientCredentials := fs.Bool("api-use-client-credentials", false, "Forces API to authenticate using client credentials instead of an ID token.Clients must be 'admin clients' to use the API.")
```
ericchiang: s/token.Clients/token. Clients/g

server/user.go
```diff
@@ -299,36 +318,48 @@ func (s *UserMgmtServer) getCreds(r *http.Request, requiresAdmin bool) (api.Cred
...
-		log.Errorf("userMgmtServer: GetCreds err: %q", err)
-		return api.Creds{}, err
-	}
-
-	i := 0
-	for _, clientID := range clientIDs {
-		// Make sure the client actually exists.
-		isAdmin, err := s.cm.IsDexAdmin(clientID)
+	var usr user.User
+	if s.allowClientCredsAuth && (len(clientIDs) == 1) && (sub == clientIDs[0]) {
```
ericchiang: Wait, why are you extracting this information from a bearer token? I would think that this endpoint would use one of the many client authentication methods provided by OpenID Connect. This just looks like it should be any random ID Token issued to the client.

https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

adrianlop: yep, it's a bearer token obtained from Dex /token endpoint, using grant_type=client_credentials and auth basic (urlencode_base64 client id+secret).
then we extract the clientID from aud and verify it's an admin.

what kind of auth do you want then?

ericchiang: Ah yes, I see.

ericchiang:
this looks good to me, but requires fixing the spelling mistakes in the command line flag and adding some unit/integration tests.

adrianlop:
ok! I'm working on the tests, will update the PR when ready.

adrianlop:
PR updated with tests!

integration/user_api_test.go
```diff
@@ -869,7 +988,7 @@ func (t *testEmailer) SendInviteEmail(email string, redirectURL url.URL, clientI
 	return retURL, nil
 }
 
-func makeUserToken(issuerURL url.URL, userID, clientID string, expires time.Duration, privKey *key.PrivateKey) string {
+func makeToken(issuerURL url.URL, userID, clientID string, expires time.Duration, privKey *key.PrivateKey) string {
```
ericchiang: in the future please don't include changes that aren't relevant to the pr.

ericchiang: ah my bad, I see why you did this. perhaps create a different method that doesn't take a clientID instead?

server/user.go
```diff
@@ -131,6 +133,12 @@ func (s *UserMgmtServer) getUser(w http.ResponseWriter, r *http.Request, ps http
 		return
 	}
 
+	if s.allowClientCredsAuth && creds.User.ID != "" {
```
ericchiang: Why is the block (and the identical other two) needed? This should absolutely be checked already by `getCreds`

server/user.go
```diff
@@ -299,36 +318,48 @@ func (s *UserMgmtServer) getCreds(r *http.Request, requiresAdmin bool) (api.Cred
...
+			log.Errorf("userMgmtServer: GetCreds err: %q", err)
 			return api.Creds{}, err
 		}
-
-		// If the endpoint requires an admin client, filter out clients which are not admins.
 		if requiresAdmin && !isAdmin {
-			continue
+			return api.Creds{}, api.ErrorForbidden
+		}
+	} else {
```
ericchiang: nit: This else statement isn't needed. Just return after your last check in the if statement above an dedent the rest of this code.

adrianlop:
sure no prob, PR updated with your comments.

ericchiang:
lgtm!

PR #531, opened 2016-08-01
user api: accept bearer tokens with multiple audiences

### LINKED ISSUE #530: Dex user API does not like multiple audiences in bearer token
When hitting the `users` API endpoint with a token with multiple values in the audience claim such as 

````json
{
  "aud": [
    "myApp",
    "kubectl"
  ],
}
````
Dex throws the following error
````
userMgmtServer: GetCreds err: "unable to parse claim as string: aud"
````
Dex should allow for an array of strings in the audience claim.

### PR DESCRIPTION
**Pull request by ericchiang**
closes #530

cc @squat @sym3tri @chancez

### DIFF
--- a/server/user.go
+++ b/server/user.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"encoding/json"
-	"errors"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: server/user.go [hunk 1]
```diff
@@ -295,18 +313,30 @@ func (s *UserMgmtServer) getCreds(r *http.Request, requiresAdmin bool) (api.Cred
...
+		}
+		// If the endpoint requires an admin client, filter out clients which are not admins.
+		if isAdmin || !requiresAdmin {
+			clientIDs[i] = clientID
+			i++
+		}
 	}
-	if requiresAdmin && !isAdmin {
+
+	clientIDs = clientIDs[:i]
```
squat: After checking `ok` (line 278), shouldn't we reduce clientIDs to the set of client IDs that verify successfully? Otherwise, I can imagine a case where several clients who claim to be admins but do not verify, and one verified client who is not an admin, would pass the `ok` check; then then the bad admins would get filtered into this new splice even though they do not verify.

ericchiang: The check on line 278 is verifying the token is signed by dex, has the correct issuer, isn't expired, etc. This is a bit different, it's checking if the client itself has "admin" status. I think agree that the logic could be cleaned up a bit.

So maybe, a better comment on line 278, and more obvious logic in this chunk?

squat: Hmm that's what I suspected. Sorry if I misunderstood, I just want to
clarify what I meant:  what if all the clients except for one are
admins and all but one (the non-admin) are expired -> then 278 returns ok.
Afterwards, on this line the clients would be trimmed down to hold only the
admins (which happens to be all expired clients). Then the splice of
clients that is left holds only invalid admins. Would these clients be
given access to endpoints they should not have access to? Let me know
if this is not actually an issue.

ericchiang: The clients themselves can't be expired, only the token which holds a client_id. You can read some details on that token here[0].

Basically, the API gets that token and needs to verify the signature, make sure it hasn't expired, etc. _Normally_ it would also verify that the client_id is for the client we expect, that's why all the current verification methods take a client_id. However, for this this case we don't need to verify a token was issued for a specific client, we just need to check if that client is an admin later on (the line we're having this conversation about), which is done outside this verify call.

Actually the 278 shouldn't be a for loop, probably just a single verification call. The method just needs the client_id being passed is just verifying that the `aud` claim is for a client you expect. 

[0] https://openid.net/specs/openid-connect-core-1_0.html#IDToken

ericchiang:
@squat I've update the code flow. Let me know if that looks better.

squat:
LGTM

Issue #524, opened 2016-07-27
Travis pushes to Quay are broken


ericchiang:
For example https://travis-ci.org/coreos/dex/jobs/147615883#L306

It doesn't even look like an issue with our release scripts, just some issue with Travis.

ericchiang:
builds are green again

PR #534, opened 2016-08-01
*: fix path to email config in dockerfile

ericchiang:
Also, run the build-docker script in the travis file.


ericchiang:
updates #524

PR #535, opened 2016-08-01
*: cleanup containers generated during tests to avoid deploy conflicts

ericchiang:
updates #524

### DIFF
--- a/build-docker
+++ b/build-docker
@@ -35,7 +35,17 @@ push() {
   docker push $repo:latest
 }
 
-build
-if [[ "push" == "$1" ]]; then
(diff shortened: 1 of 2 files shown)


build-docker
```diff
@@ -35,7 +35,17 @@ push() {
   docker push $repo:latest
 }
 
-build
-if [[ "push" == "$1" ]]; then
-  push
+clean() {
+  docker rmi $repo:$version
+  docker rmi $repo:latest
```
sym3tri: may want to add a `-f` to these for good measure

PR #537, opened 2016-08-03
server: when registering a user, set user name if OIDC identity contains one

wyattanderson:
When automatically registering a user from an IP that provides
a `DisplayName`, set it on the created user so that JWT we create
contain a meaningful `name` field.

Added new test and ran existing ones.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -416,6 +416,13 @@ func (s *Server) Login(ident oidc.Identity, key string) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("getting created user: %v", err)
 		}
+
+		if ses.Identity.Name != "" && ses.Identity.Name != usr.DisplayName {
+			err = s.UserManager.SetDisplayName(usr, ses.Identity.Name)
+			if err != nil {
+				return "", fmt.Errorf("couldn't set display name for user: %v", err)
+			}
+		}
 	} else if err != nil {
 		return "", fmt.Errorf("getting user: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


user/manager/manager.go
```diff
@@ -120,6 +120,23 @@ func (m *UserManager) Disable(userID string, disabled bool) error {
 	return nil
 }
 
+func (m *UserManager) SetDisplayName(usr user.User, displayName string) error {
+	tx, err := m.begin()
```
ericchiang: Need to check the error here

user/manager/manager.go
```diff
@@ -120,6 +120,23 @@ func (m *UserManager) Disable(userID string, disabled bool) error {
...
+
+	usr.DisplayName = displayName
+	if err = m.userRepo.Update(tx, usr); err != nil {
+		rollback(tx)
```
ericchiang: Generally, you can just do `defer rollback(tx)` after creating the tx. `rollback` is no-op once `tx.Commit` has been successfully called.

wyattanderson: Gotcha. Was largely copy/pasting from other code here.

server/server.go
```diff
@@ -416,6 +416,13 @@ func (s *Server) Login(ident oidc.Identity, key string) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("getting created user: %v", err)
 		}
+
+		if ses.Identity.Name != "" {
+			err = s.UserManager.SetDisplayName(usr, ses.Identity.Name)
```
ericchiang: One last nit. Isn't this redundant if `usr.DisplayName == ses.Identity.Name`?

ericchiang:
lgtm!

PR #540, opened 2016-08-08
refresh token rotation

### LINKED ISSUE #519: Refresh tokens should only work once and return a new refresh token
Also known as [_"Refresh Token Rotation"_](https://tools.ietf.org/html/rfc6819#section-5.2.2.3)

Currently a refresh token can be redeemed for an access/ID token as many times at the hold of that refresh token likes. E.g.

```json
{
  "access_token":"2YotnFZFEjr1zCsicMWpAA",
  "token_type":"example",
  "expires_in":3600,
  "id_token": "..."
}
```

Instead, Dex should also return a new refresh token every time one is claimed.

```json
{
  "access_token":"2YotnFZFEjr1zCsicMWpAA",
  "token_type":"example",
  "expires_in":3600,
  "refresh_token":"tGzv3JOkF0XG5Qx2TlKWIA",
  "id_token": "..."
}
```

This immediately invalidates the current refresh token, and the client must use the new one. The primary motivation is detecting that a refresh token has be stolen. Summed up well by [rfc 6819](https://tools.ietf.org/html/rfc6819#section-5.2.2.3)

> Refresh token rotation is intended to automatically detect and
> prevent attempts to use the same refresh token in parallel from
> different apps/devices.  This happens if a token gets stolen from the
> client and is subsequently used by both the attacker and the
> legitimate client.

rsoletob:
Added refresh token rotation.

Fixes #519

### DIFF
--- a/db/refresh.go
+++ b/db/refresh.go
@@ -91,101 +91,50 @@ func NewRefreshTokenRepoWithGenerator(dbm *gorp.DbMap, gen refresh.RefreshTokenG
 }
 
 func (r *refreshTokenRepo) Create(userID, clientID, connectorID string, scopes []string) (string, error) {
-	if userID == "" {
-		return "", refresh.ErrorInvalidUserID
-	}
-	if clientID == "" {
-		return "", refresh.ErrorInvalidClientID
-	}
-
-	// TODO(yifan): Check the number of tokens given to the client-user pair.
-	tokenPayload, err := r.tokenGenerator.Generate()
-	if err != nil {
-		return "", err
-	}
-
-	payloadHash, err := bcrypt.GenerateFromPassword(tokenPayload, bcrypt.DefaultCost)
-	if err != nil {
-		return "", err
-	}
-
-	record := &refreshTokenModel{
-		PayloadHash: payloadHash,
-		UserID:      userID,
-		ClientID:    clientID,
-		ConnectorID: connectorID,
-		Scopes:      strings.Join(scopes, " "),
-	}
-
-	if err := r.executor(nil).Insert(record); err != nil {
-		return "", err
-	}
-
-	return buildToken(record.ID, tokenPayload), nil
+	return r.create(nil, userID, clientID, connectorID, scopes)
 }
 
 func (r *refreshTokenRepo) Verify(clientID, token string) (userID, connectorID string, scope scope.Scopes, err error) {
-	tokenID, tokenPayload, err := parseToken(token)
-
-	if err != nil {
-		return
-	}
-
-	record, err := r.get(nil, tokenID)
-	if err != nil {
-		return
-	}
-
-	if record.ClientID != clientID {
-		return "", "", nil, refresh.ErrorInvalidClientID
-	}
-
-	if err = checkTokenPayload(record.PayloadHash, tokenPayload); err != nil {
-		return
-	}
-
-	var scopes []string
-	if len(record.Scopes) > 0 {
-		scopes = strings.Split(record.Scopes, " ")
-	}
-
-	return record.UserID, record.ConnectorID, scopes, nil
+	return r.verify(nil, clientID, token)
 }
 
 func (r *refreshTokenRepo) Revoke(userID, token string) error {
-	tokenID, tokenPayload, err := parseToken(token)
-	if err != nil {
-		return err
-	}
-
 	tx, err := r.begin()
 	if err != nil {
 		return err
 	}
 	defer tx.Rollback()
-	exec := r.executor(tx)
-	record, err := r.get(tx, tokenID)
-	if err != nil {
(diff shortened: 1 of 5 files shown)


server/server_test.go
```diff
@@ -576,23 +576,25 @@ func TestServerRefreshToken(t *testing.T) {
...
-		token         string
-		clientID      string // The client that associates with the token.
-		creds         oidc.ClientCredentials
-		signer        jose.Signer
-		createScopes  []string
-		refreshScopes []string
-		expectedAud   []string
-		err           error
+		token                string
+		expectedRefreshToken string
```
ericchiang: How are you coming up with these values? Can you implement a helper function to do this?

db/refresh.go
```diff
@@ -188,6 +188,66 @@ func (r *refreshTokenRepo) Revoke(userID, token string) error {
...
+		return "", refresh.ErrorInvalidToken
+	}
+
+	// Renew refresh token
+	tokenPayload, err = r.tokenGenerator.Generate()
+	if err != nil {
+		return "", err
+	}
+
+	payloadHash, err := bcrypt.GenerateFromPassword(tokenPayload, bcrypt.DefaultCost)
```
ericchiang: I forget what the point of this payload hash is. Might be beyond the scope of this PR, but we need to add a comment or remove this logic.

db/refresh.go
```diff
@@ -188,6 +188,66 @@ func (r *refreshTokenRepo) Revoke(userID, token string) error {
 	return tx.Commit()
 }
 
+func (r *refreshTokenRepo) RenewRefreshToken(clientID, userID, oldToken string) (newRefreshToken string, err error) {
```
ericchiang: A lot of this logic looks identical to Create and Verify. Can we figure out how to have these share code? E.g. renewing the token should just be "Verify -> Delete -> Create"

server/server.go
```diff
@@ -560,34 +560,34 @@ func (s *Server) CodeToken(creds oidc.ClientCredentials, sessionKey string) (*jo
 	return jwt, refreshToken, nil
 }
 
-func (s *Server) RefreshToken(creds oidc.ClientCredentials, scopes scope.Scopes, token string) (*jose.JWT, error) {
+func (s *Server) RefreshToken(creds oidc.ClientCredentials, scopes scope.Scopes, token string) (*jose.JWT, string, error) {
```
ericchiang: Are the raw Claims ever used or do they immediately call `Encode()`?

ericchiang:
A few comments, but overall looks good. Thanks!

db/refresh.go
rsoletob: I'm going to add a comment. It checks if the hash of payload from received refresh token are equal to the stored in database.

server/server.go
rsoletob: Raw claims are encoded immediately [here](https://github.com/coreos/dex/blob/0e94e7625551a0e2c455de6d2568489e02e00f98/server/http.go#L534). But the interface [OIDCServer](https://github.com/coreos/dex/blob/0e94e7625551a0e2c455de6d2568489e02e00f98/server/server.go#L46) uses jose.JWT in more places. Do you prefer to change it?

rsoletob:
Thanks for your comments. I see an user could have some refresh tokens for the same client and there is a [TODO](https://github.com/coreos/dex/blob/0e94e7625551a0e2c455de6d2568489e02e00f98/db/refresh.go#L101).  Is it correct? I can change it here or in another PR if you think it isn't a valid feature.

ericchiang:
Probably best for a different PR :)

server/server.go
ericchiang: No, this is fine. Again, it's something we can clean up later.

rsoletob:
Ok @ericchiang, revamped

db/refresh.go
```diff
@@ -91,47 +91,108 @@ func NewRefreshTokenRepoWithGenerator(dbm *gorp.DbMap, gen refresh.RefreshTokenG
...
 	}
 
-	// TODO(yifan): Check the number of tokens given to the client-user pair.
-	tokenPayload, err := r.tokenGenerator.Generate()
+	return tx.Commit()
+}
+
+func (r *refreshTokenRepo) RenewRefreshToken(clientID, userID, oldToken string) (newRefreshToken string, err error) {
+	// Verify
+	userID, connectorID, scopes, err := r.verify(nil, clientID, oldToken)
```
ericchiang: Why is verify outside the transaction?

ericchiang:
small nit but other than that tested locally and lgtm

db/refresh.go
rsoletob: Because if dex can't verify the token you save opening a transaction.

ericchiang: okay sure

ericchiang:
lgtm!

PR #544, opened 2016-08-11
*: update examples to hold valid client secrets for no-db mode

ericchiang:
It was noted in #542 that our examples don't hold valid client secrets for no-db mode.

#337 strikes again.


whitlockjc:
There is also inconsistency with how the samples are used.  By default, the `static/fixtures/*.json.sample` files are used so if you follow through with the `cp` commands suggested to create your own versions, you then need to supply the approriate flags to `bin/dex-worker`.  Also, when running `bin/example-app`, you need to use `--discovery=http://127.0.0.1:5556/dex` because you didn't run `bin/dex-worker` with `--issuer=http://127.0.0.1:5556`.

Issue #543, opened 2016-08-10
DEX with multiple oidc connectors is not working 


ghost:
Just started looking into DEX for its ability to federate multiple identity mechanisms. Everything works as documented when I have the google oidc connector configured. As soon as I add another connector for my own OP, nothing works. Here is my connector json, 
```
[
    {
        "type": "local",
        "id": "local"
    },
    {
        "type": "oidc",
        "id": "google",
        "issuerURL": "https://accounts.google.com",
        "clientID": "redacted",
        "clientSecret": "redacted",
        "trustedEmailProvider": true
    },
    {
        "type": "oidc",
        "id": "boxy",
        "issuerURL": "https://auth.qa03.bluetest.com",
        "clientID": "redacted",
        "clientSecret": "redacted",
        "trustedEmailProvider": false
    }
] 
```

With this configuration when I click on the "google" or the "boxy" link in the discovery page, I keep getting redirected to the back to the discovery page in an endless loop I see the following in the dex-worker log

```
ubuntu@allinone2:~/dex$ tail -f dex-worker.log 

INFO: Debug logging enabled.

DEBUG: Debug logging enabled.

2016/08/10 18:36:50 WARN: Running with no limit on: database idle connections

2016/08/10 18:36:50 WARN: Running with no limit on: database open connections

2016/08/10 18:36:50 INFO: Loaded IdP connector: id=local type=local

2016/08/10 18:36:50 INFO: Loaded IdP connector: id=google type=oidc

2016/08/10 18:36:50 INFO: Loaded IdP connector: id=boxpanel type=oidc

2016/08/10 18:36:50 INFO: Binding to 0.0.0.0:5556...

2016/08/10 18:37:30 INFO: 169.44.130.21 - - [10/Aug/2016:18:37:30 +0000] "GET /dex/.well-known/openid-configuration HTTP/1.1" 200 542

[...]

Issue #473, opened 2016-06-20
Improve error logging


ericchiang:
There are a bunch of logging statements that make no sense and need to be removed or clarified. For example:

```
ERROR: IDENTITY: err: <nil>
```

sym3tri:
Another problem that has arisen is not enough log statements when things are not working as expected, making dex hard to troubleshoot.

Logging need to be cleaned up in general. We've been using logrus in other projects and are happy with it so far. https://github.com/Sirupsen/logrus

ericchiang:
Okay. Going to bump the priority and assign to the next milestone.

ericchiang:
This has largely been address. Though we don't use logrus, we do log a lot more meaningful errors.

PR #556, opened 2016-08-16
*: update go-oidc to remove dependency on capnslog

ericchiang:
#543 relieved that we're suppressing the standard logger in dex due to some capnslog side effects. The actual code at fault was not tracked down, but removing the dependency entirely fixes the issue. Since capnslog is largely deprecated, this seems like a reasonable step to take.


ericchiang:
This also updates #473. Apparently we were just completely suppressing outside packages (e.g. go-oidc).

PR #558, opened 2016-08-18
server: fix registration redirect for non-root URLs

### LINKED ISSUE #553: Registration handler doesn't take full issuer URL into account
Here is the problem in handleAuthFunc() function:

```
if register {
	_, ok := idpc.(*connector.LocalConnector)
	if ok {
		q := url.Values{}
		q.Set("code", key)
		ru := httpPathRegister + "?" + q.Encode()
		w.Header().Set("Location", ru)
		w.WriteHeader(http.StatusFound)
		return
	}
}
```

It always redirects to httpPathRegister ("/register") w/o using issuerURL. So it's impossible to register a new user if dex is listening on a non-root path.

ericchiang:
Closes #553

cc @sym3tri @chancez

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"path"
 	"strings"
 	"time"
 
@@ -266,7 +267,7 @@ func renderLoginPage(w http.ResponseWriter, r *http.Request, srv OIDCServer, idp
 	execTemplate(w, tpl, td)
 }
 
-func handleAuthFunc(srv OIDCServer, idpcs []connector.Connector, tpl *template.Template, registrationEnabled bool) http.HandlerFunc {
+func handleAuthFunc(srv OIDCServer, baseURL url.URL, idpcs []connector.Connector, tpl *template.Template, registrationEnabled bool) http.HandlerFunc {
 	idx := makeConnectorMap(idpcs)
 	return func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != "GET" {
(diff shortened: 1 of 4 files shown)


ericchiang:
there's even a test!

server/http.go
```diff
@@ -358,7 +358,7 @@ func handleAuthFunc(srv OIDCServer, idpcs []connector.Connector, tpl *template.T
 			if ok {
 				q := url.Values{}
 				q.Set("code", key)
-				ru := httpPathRegister + "?" + q.Encode()
+				ru := strings.TrimRight(baseURL, "/") + "/" + strings.TrimLeft(httpPathRegister, "/") + "?" + q.Encode()
```
chancez: Could you maybe use `path.Join` or `url.Parse` instead? It should remove some of the need to trim, but is also more likely work correctly with joining paths together.

ericchiang: Sure, I'll switch it to passing a URL and use path.Join

chancez: Actually, it seems weird that you're having to pass the basePath through to so many _different_ places. Is this a thing you only need to do here due to needing to set the `Location` header?

ericchiang: Yes.

ericchiang:
updated

ericchiang:
woops, I have no idea what just happened there. one sec

ericchiang:
cc @chancez ready for another review

server/http_test.go
```diff
@@ -210,6 +226,33 @@ func TestHandleAuthFuncResponsesSingleRedirectURL(t *testing.T) {
...
+			query: url.Values{
+				"response_type": []string{"code"},
+				"redirect_uri":  []string{"http://client.example.com/callback"},
+				"client_id":     []string{"client.example.com"},
+				"connector_id":  []string{"local"},
+				"register":      []string{"1"},
+				"scope":         []string{"openid"},
+			},
+			wantCode:     http.StatusFound,
+			wantLocation: "/register?code=code-2",
```
chancez: perhaps explicitly set baseURL to `/` here?

chancez:
LGTM after nits.

PR #560, opened 2016-08-18
server: fix the path registration for user APIs at non-root URLs

### LINKED ISSUE #557: User API endpoint is not working after non-rootURL changes
Prior to the PR adding support for running dex at non-root URLs, accessing dex at `/api/v1/users` worked, but now it doesn't. Even with the same issuer-url.

I think i see the issue in the code, but cannot figure out what the actual endpoint getting computed is:

on line 290 and 296 here: https://github.com/coreos/dex/blob/b33227bcd4c3b78d3279e2ea22790d57fef5b48c/server/server.go#L290-L296 you'll see you're registering the users handler against `apiBasePath := path.Join(httpPathAPI, APIVersion)` so the users handler is somewhere after `/api/v1`..okay got it.

Let's look at the user hander: https://github.com/coreos/dex/blob/b33227bcd4c3b78d3279e2ea22790d57fef5b48c/server/user.go#L27

Looking at that definition i can see the following:
```
	UsersSubTree                  = "/users"
	UsersListEndpoint             = addBasePath(UsersSubTree)
```

If we look at `addBasePath` we get 
```
func addBasePath(s string) string {
	return path.Join(httpPathAPI, APIVersion, s)
}
```

So our UsersListEndpoint is `/api/v1/users` which is being registered in `server.go` at `/api/v1`. So i expect the endpoint to be roughly: `/api/v1/api/v1/users`, however that isn't working either. But I can see that the endpoints are not what they used to be, in code and out.

ericchiang:
closes #557

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -215,6 +215,8 @@ func (s *Server) HTTPHandler() http.Handler {
 
 	clock := clockwork.NewRealClock()
 	mux := http.NewServeMux()
+
+	// Handler methods which register handlers at prefixed paths.
 	handle := func(urlPath string, h http.Handler) {
 		p := path.Join(s.IssuerURL.Path, urlPath)
 		// path.Join always trims trailing slashes (https://play.golang.org/p/GRr0jDd9P7).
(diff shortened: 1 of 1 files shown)


ericchiang:
cc @sym3tri @chancez

chancez:
Works for me!

PR #562, opened 2016-08-20
dev branch: support the implicit flow

ericchiang:
Friday hack.

cc @sym3tri @chancez @dghubble 

Am going to update this PR with an integration test later.


sym3tri:
Given that implicit flow is a lot less secure I still thing we should have it off by default and only enable it if opted-in by a flag.

ericchiang:
Updated to make optional and integration test added.

PR #567, opened 2016-08-25
dev branch: port templates from master branch

ericchiang:
This change ports the existing templates to the dev branch.

Things this PR does __not__ attempt to accomplish:

* These are not the final templates. This PR only attempts to accomplish enough to get away from the current built in, unstyled templates. Still need to review with the design team before finalizing the look.
* We currently do not have a good strategy for swapping out or styling templates in some custom fashion and this PR does not attempt to improve upon it. Will attempt to make the CSS and structure more modular in follow up efforts.

cc @sym3tri


sym3tri:
templates seem fine for now. we should aim to reduce their complexity over time.

ericchiang:
> templates seem fine for now. we should aim to reduce their complexity over time.

Absolutely. Just need to get this branch to stop using templates defined in inline Go source.

ericchiang:
Also as a note, the data templates are executed with is now defined in a single file[0], which should improve our ability to iterate them without hunting down every `Execute()` call.

[0] https://github.com/ericchiang/dex/blob/2a0c14de8d6a9a3213e0bd8d4bf7c2584008e714/server/templates.go

PR #569, opened 2016-08-25
dev branch: set response types supported in discovery based on server config

ericchiang:


ericchiang:
since this is super small just going to merge

PR #571, opened 2016-09-04
documentation: multiple fixes in examples README

remohammadi:
* In a9dce1c, the defaults are set to the `.example` paths, except for `emailer.json` which is instead committed into the git repository. So there's no need to duplicate those file in order to start dex-worker.
* The default value for discovery is moved from `/` to `/dex` in #521
* Typo in `client-secret` value.


ericchiang:
lgtm. thanks!

Issue #574, opened 2016-09-08
Support of expires_in field in Token endpoint


summerwind:
dex does not return a *expires_in* field from Token endpoint. The spec of OpenID Connect allowed to use *expires_in* fields in a response of Token Endpoint. 
http://openid.net/specs/openid-connect-core-1_0.html#TokenResponse

ReuseTokenSource of `golang.org/x/oauth2` supports *expires_in* field to enable refreshing token automaticaly. 
https://github.com/golang/oauth2/blob/master/internal/token.go#L194

I'd like to add the support of *expires_in* field. Thought?
(I already wrote a patch!)

ericchiang:
Surprised we don't do this already. Yes please, do send in that patch.

summerwind:
@ericchiang Thank you for the quick response!
I sent my patch in #575.

summerwind:
Landed in 9a78dca137c48a1e1fb64de2b59488b476ffdb3d. Thank you for the support!

ericchiang:
Thanks for the work! Think it'll be reasonable to do a release later this week. This seems big enough to warrant one with some other work that's gone in.

summerwind:
Great to hear that!

PR #575, opened 2016-09-08
server: add expires_in field to the response of token endpoint

summerwind:
This PR adds the expires_in field to the response of token endpoint.
Related with #574.

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -491,6 +491,7 @@ func handleTokenFunc(srv OIDCServer) http.HandlerFunc {
 
 		var jwt *jose.JWT
 		var refreshToken string
+		var expiresIn int64
 		grantType := r.PostForm.Get("grant_type")
 
 		switch grantType {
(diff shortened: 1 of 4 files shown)


server/http.go
```diff
@@ -537,6 +538,7 @@ func handleTokenFunc(srv OIDCServer) http.HandlerFunc {
 			IDToken:      jwt.Encode(),
 			TokenType:    "bearer",
 			RefreshToken: refreshToken,
+			ExpiresIn:    int64(session.DefaultSessionValidityWindow.Seconds()),
```
ericchiang: I don't like assuming that we're always using the default validity window, since it ties it to a global variable rather than this server's configuration. Ideally, `CodeToken`, `ClientCredsToken`, and `RefreshToken` would all return the JWT's expiry in it's function signature and the actual configuration of this value lives within those functions.

summerwind: Agree. I'm going to implement an another way.

summerwind:
Fixed the codes. `CodeToken`, `ClientCredsToken`, and `RefreshToken` has changed to return the expiration time. 
@ericchiang Could you please take another look?

server/server.go
```diff
@@ -49,13 +49,13 @@ type OIDCServer interface {
 	Login(oidc.Identity, string) (string, error)
 
 	// CodeToken exchanges a code for an ID token and a refresh token string on success.
-	CodeToken(creds oidc.ClientCredentials, sessionKey string) (*jose.JWT, string, error)
+	CodeToken(creds oidc.ClientCredentials, sessionKey string) (*jose.JWT, string, int64, error)
```
ericchiang: Instead of int64, please return a `time` package type to indicate the exact meaning of the returned value. For instance, you can return a `time.Time` here then convert using `int64(time.Now().Sub(t))` to "expired at".

summerwind: Thank you for the feedback! I'll fix that to use `time.Time`.

summerwind:
Fixed the code to use time.Time. If you have any other concerns, please let me know.

ericchiang:
Thanks for the work. lgtm!

PR #576, opened 2016-09-14
dexctl: Deprecating dexctl's --db-url flag

### LINKED ISSUE #298: Deprecate dexctl's --db-url flag
dexctl shouldn't being connecting directly to the database to perform actions. Deprecate this flag and have all commands go through the APIs.

ajanthan:
dexctl should not connecting directly db instead it should use admin APIs

Fixes #298


ericchiang:
Thanks for the PR!

At first glance this looks like it's removing support rather than deprecating it though. I'd recommend we log a warning for a bit so we don't break compatibility with existing installations.

ajanthan:
Thanks for the feedback.Let me fix it and re-send it

ajanthan:
Tried to update the PR with suggested change but it did not work.Closing this PR and created a new one(https://github.com/coreos/dex/pull/578).

PR #578, opened 2016-09-20
dexctl: Deprecating dexctl's --db-url flag

### LINKED ISSUE #298: Deprecate dexctl's --db-url flag

ajanthan:
dexctl should not connecting directly db instead it should use admin APIs.This is the updated pull request with the feedback received previously .

Fixes #298


cmd/dexctl/command_config.go
```diff
@@ -81,10 +82,36 @@ func runGetConnectorConfigs(cmd *cobra.Command, args []string) int {
 	fmt.Printf("Found %d connector config(s)\n", len(cfgs))
 
 	for _, cfg := range cfgs {
-		fmt.Println()
-		fmt.Printf("ID:   %v\n", cfg.ConnectorID())
-		fmt.Printf("Type: %v\n", cfg.ConnectorType())
+		switch typedCFG := cfg.(type) {
```
ericchiang: I'm a little confused by what this is doing and why it's necessary. Can you expand a little bit?

cmd/dexctl/command_config.go
```diff
@@ -81,10 +82,36 @@ func runGetConnectorConfigs(cmd *cobra.Command, args []string) int {
...
+		case map[string]interface{}:
+
+			id := typedCFG["id"].(string)
```
ericchiang: The following program panics.

```
package main

import "fmt"

func main() {
	m := map[string]interface{}{"id": 4}
	id := m["id"].(string)
	fmt.Println(id)
}
```

Please ensure that `"id"` is actually a string (`id, ok := typedCFG["id"]`)

cmd/dexctl/command_config.go
```diff
@@ -51,7 +52,7 @@ func runSetConnectorConfigs(cmd *cobra.Command, args []string) int {
 		r = f
 	}
 
-	cfgs, err := connector.ReadConfigs(r)
+	cfgs, err := readConfigs(r)
```
ericchiang: Why doesn't connector.ReadConfigs work anymore?

cmd/dexctl/driver_admin_api.go
```diff
@@ -0,0 +1,90 @@
...
+	return nil
+}
+
+type adminAPITransport struct {
+	secret string
+}
+
+func (a adminAPITransport) RoundTrip(r *http.Request) (*http.Response, error) {
+	r.Header.Set("Authorization", a.secret)
+	return http.DefaultTransport.RoundTrip(r)
```
ericchiang: What if dex is running under a custom CA?

cmd/dexctl/driver_admin_api.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	return nil
+}
+
+type adminAPITransport struct {
+	secret string
+}
+
+func (a adminAPITransport) RoundTrip(r *http.Request) (*http.Response, error) {
+	r.Header.Set("Authorization", a.secret)
```
ericchiang: Generally a bad practice since it goes against the docs of http.RoundTripper

```
        // RoundTrip should not modify the request, except for
        // consuming and closing the Request's Body.
```

https://golang.org/pkg/net/http/#RoundTripper

but since everyone does this, it's not a big deal.

cmd/dexctl/driver.go
```diff
@@ -1,13 +1,12 @@
 package main
 
 import (
-	"github.com/coreos/dex/connector"
 	"github.com/coreos/go-oidc/oidc"
 )
 
 type driver interface {
```
ericchiang: Do we still find this interface useful? Could we drop it and just have different paths for different backends?

cmd/dexctl/main.go
```diff
@@ -53,7 +55,9 @@ var (
 func init() {
 	log.EnableTimestamps()
 
-	rootCmd.PersistentFlags().StringVar(&global.dbURL, "db-url", "", "DSN-formatted database connection string")
+	rootCmd.PersistentFlags().StringVar(&global.dbURL, "db-url", "", "DSN-formatted database connection string. --db-url flag is deprecated. Use --base-url and --api-key")
```
ericchiang: Probably need a root CA too.

ericchiang:
Thanks for the update! Sorry for the delayed response, was having laptop issues.

cmd/dexctl/command_config.go
ajanthan: It returns []ConnectorConfig and the admin api requires []interface{} in set connector method.
I wanted to keep clean the new admin api based implementation from haveing code converting ConnectorConfig to interface{}

cmd/dexctl/command_config.go
ajanthan: Existing  direct DB call returns ConnectorConfig and Admin API returns  interface{} .I could not find a way to convert interface{} to ConnectorConfig hence I am doing type switch and printing connector configs different ways based on the type.

cmd/dexctl/command_config.go
ajanthan: yes.You are correct.

cmd/dexctl/driver.go
ajanthan: I agree with you.I didn't find any reasons other than any historical reasons that I am not aware .That will be cleaner.Let me update the code.

cmd/dexctl/driver_admin_api.go
ajanthan: Yes.I haven't thought about it .Let me add tls config as well

cmd/dexctl/main.go
ajanthan: Yes.I agree

ajanthan:
Thanks for the review.Let me update the pull request according to the feedback

ajanthan:
Update.
Added a flag to supply rootCA file and removed driver interface.

ericchiang:
Thanks I'll try to review tomorrow. Please ping me on here if I don't get to it.

ajanthan:
Any feedbacks on latest changes

cmd/dexctl/main.go
```diff
@@ -69,19 +75,41 @@ func wrapRun(run func(cmd *cobra.Command, args []string) int) func(cmd *cobra.Co
 	}
 }
 
-func getDriver() (drv driver) {
+func isDBURLPresent() bool {
```
ericchiang: Since this method is now being used very differently, it shouldn't be exiting on it's own. It should just return an error if not present.

cmd/dexctl/main.go
```diff
@@ -69,19 +75,41 @@ func wrapRun(run func(cmd *cobra.Command, args []string) int) func(cmd *cobra.Co
...
 
-	return
+	return false
+}
+
+func getDBConnector() *dbConnector {
+	dbConnector, err := newDBConnector(global.dbURL)
+	if err != nil {
+		stderr("Unable to configure dexctl databse connector: %v", err)
+		os.Exit(1)
```
ericchiang: Same comment as the DB comment. It's just more idiomatic Go to service these error rather than exiting.

cmd/dexctl/command_config.go
```diff
@@ -71,20 +87,53 @@ func runGetConnectorConfigs(cmd *cobra.Command, args []string) int {
...
 
-	fmt.Printf("Found %d connector config(s)\n", len(cfgs))
+	return 0
+}
 
-	for _, cfg := range cfgs {
-		fmt.Println()
-		fmt.Printf("ID:   %v\n", cfg.ConnectorID())
-		fmt.Printf("Type: %v\n", cfg.ConnectorType())
+func readConfigs(r io.Reader) ([]interface{}, error) {
```
ericchiang: please add a comment about why this is different from connector.ReadConfigs

cmd/dexctl/admin_api_connector.go
```diff
@@ -0,0 +1,116 @@
...
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+
+		tlsConfig := &tls.Config{
+			RootCAs: caCertPool,
+		}
+		tlsConfig.BuildNameToCertificate()
+		transport = &http.Transport{TLSClientConfig: tlsConfig}
+	} else {
+		transport = &http.Transport{}
```
ericchiang: use `http.DefaultTransport` instead. It's got some good default values.

cmd/dexctl/admin_api_connector.go
```diff
@@ -0,0 +1,116 @@
...
+		if err != nil {
+			return adminAPIConnector, err
+		}
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+
+		tlsConfig := &tls.Config{
+			RootCAs: caCertPool,
+		}
+		tlsConfig.BuildNameToCertificate()
```
ericchiang: Not a necessary call. This is done automatically.

cmd/dexctl/admin_api_connector.go
```diff
@@ -0,0 +1,116 @@
...
+func newAdminAPIConnector(adminApiURL string, adminAPIKey string, caFile string) (*AdminAPIConnector, error) {
+	adminAPIConnector := &AdminAPIConnector{baseURL: adminApiURL}
+	var transport *http.Transport
+	if len(caFile) != 0 {
+		caCert, err := ioutil.ReadFile(caFile)
+		if err != nil {
+			return adminAPIConnector, err
+		}
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
```
ericchiang: You'll want to check the return value of this call and error if it doesn't find any certs. (e.g. if someone has mistakenly used a key file)

https://golang.org/pkg/crypto/x509/#CertPool.AppendCertsFromPEM

cmd/dexctl/admin_api_connector.go
```diff
@@ -0,0 +1,116 @@
...
+
+		tlsConfig := &tls.Config{
+			RootCAs: caCertPool,
+		}
+		tlsConfig.BuildNameToCertificate()
+		transport = &http.Transport{TLSClientConfig: tlsConfig}
+	} else {
+		transport = &http.Transport{}
+	}
+	apiClient := &http.Client{
```
ericchiang: nit: you can just assign directly to `adminAPIConnector.client`

cmd/dexctl/admin_api_connector.go
```diff
@@ -0,0 +1,116 @@
...
+
+func newAdminAPITransport(apiSecret string, roundTripper http.RoundTripper) adminAPITransport {
+	adminTransport := adminAPITransport{secret: apiSecret, rt: roundTripper}
+	return adminTransport
+}
+func (a adminAPITransport) RoundTrip(r *http.Request) (*http.Response, error) {
+	if len(r.Header.Get("Authorization")) != 0 {
+		return a.rt.RoundTrip(r)
+	}
+	r.Header.Set("Authorization", a.secret)
```
ericchiang: can you add a comment noting that this is correctly not using a bearer token syntax, just setting the header directly. It looked odd to me and I had to check the server code to confirm. Want to be helpful to whoever has to use this code this next time :)

cmd/dexctl/driver.go
```diff
@@ -1,13 +0,0 @@
-package main
-
-import (
-	"github.com/coreos/dex/connector"
-	"github.com/coreos/go-oidc/oidc"
-)
-
-type driver interface {
```
ericchiang: aww yeah

cmd/dexctl/main.go
```diff
@@ -53,7 +56,10 @@ var (
 func init() {
 	log.EnableTimestamps()
 
-	rootCmd.PersistentFlags().StringVar(&global.dbURL, "db-url", "", "DSN-formatted database connection string")
+	rootCmd.PersistentFlags().StringVar(&global.dbURL, "db-url", "", "DSN-formatted database connection string. --db-url flag is deprecated. Use --base-url and --api-key")
+	rootCmd.PersistentFlags().StringVar(&global.baseURL, "base-url", "", "DSN-formatted dex-overlord base URL")
+	rootCmd.PersistentFlags().StringVar(&global.apiKey, "api-key", "", "API key for Admin API")
+	rootCmd.PersistentFlags().StringVar(&global.rootCA, "root-ca", "", "Location of root CA file.This flag is needed if the server uses custome root CA.")
```
ericchiang: some typos here. Maybe just "File containing a root CA for the Admin API. If unset, the system CAs will be used."

cmd/dexctl/main.go
```diff
@@ -53,7 +56,10 @@ var (
 func init() {
 	log.EnableTimestamps()
 
-	rootCmd.PersistentFlags().StringVar(&global.dbURL, "db-url", "", "DSN-formatted database connection string")
+	rootCmd.PersistentFlags().StringVar(&global.dbURL, "db-url", "", "DSN-formatted database connection string. --db-url flag is deprecated. Use --base-url and --api-key")
+	rootCmd.PersistentFlags().StringVar(&global.baseURL, "base-url", "", "DSN-formatted dex-overlord base URL")
```
ericchiang: probably don't need to specify the "DSN-formatting" maybe just "URL of the dex-overlord." Though we probably want to just add the well-known path of the API to the base url.

ericchiang:
Also please combine these commits and force push over the old one please.

http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

ajanthan:
Incorporated all the changes

PR #579, opened 2016-09-22
static: fix button styles

squat:
This PR fixes some visual issues with the button styles. Here are before and after screenshots for reference:

Before:
screenshot from 2016-09-22 11-11-17

After:
screenshot from 2016-09-22 11-10-40

### DIFF
--- a/static/html/header.html
+++ b/static/html/header.html
@@ -89,14 +89,14 @@
       }
 
       .btn {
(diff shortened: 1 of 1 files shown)


sym3tri:
LGTM

robszumski:
Should this have the same style as our buttons? I know this is open source, but I spotted this in our product, so its making its way all the way through.

squat:
In general, Dex static assets could probably use some TLC, though I think this is a larger topic.

PR #581, opened 2016-09-29
frontend: revert button fix

squat:
This commit reverts the last fix to the button, since it introduced
another issue. Also, it addressed the root cause of the original
problem, which was that one of the templates was missing the correct
class.


ericchiang:
lgtm. merging on green

PR #588, opened 2016-10-04
dev branch: add health check endpoint

ericchiang:
cc @dghubble @squat @kans

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -20,6 +20,36 @@ import (
 	"github.com/coreos/dex/storage"
 )
 
+func (s *Server) handleHealth(w http.ResponseWriter, r *http.Request) {
+	start := s.now()
+	err := func() error {
+		// Instead of trying to introspect health, just try to use the underlying storage.
+		a := storage.AuthRequest{
+			ID:       storage.NewID(),
+			ClientID: storage.NewID(),
+
+			// Set a short expiry so if the delete fails this will be cleaned up quickly by garbage collection.
+			Expiry: s.now().Add(time.Minute),
+		}
+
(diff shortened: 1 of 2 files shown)


server/handlers.go
```diff
@@ -20,6 +20,36 @@ import (
...
+		}
+		if err := s.storage.DeleteAuthRequest(a.ID); err != nil {
+			return fmt.Errorf("delete auth request: %v", err)
+		}
+		return nil
+	}()
+
+	t := s.now().Sub(start)
+	if err != nil {
+		log.Printf("storage health check failed: %v", err)
```
squat: Nit: consistent casing for log statements.

server/handlers.go
```diff
@@ -20,6 +20,36 @@ import (
...
 
+func (s *Server) handleHealth(w http.ResponseWriter, r *http.Request) {
+	start := s.now()
+	err := func() error {
+		// Instead of trying to introspect health, just try to use the underlying storage.
+		a := storage.AuthRequest{
+			ID:       storage.NewID(),
+			ClientID: storage.NewID(),
+
+			// Set a short expiry so if the delete fails this will be cleaned up quickly by garbage collection.
```
dghubble: If I poll this every 30 seconds and the delete fails, what's the expected behavior on the next create? Basically, does the kubernetes manifest health check need to know about this detail?

ericchiang: storage.NewID() always returns a random ID. Each call deals with a unique object so they should never impact one another.

server/handlers.go
ericchiang: thanks, updated

PR #594, opened 2016-10-08
Documentation/proposals: add a proposal for keeping data in-sync during refreshes

ericchiang:
cc @dghubble @sym3tri @squat @kans


ericchiang:
Merging since we'd like to use some of this for the LDAP work

PR #605, opened 2016-10-13
Add bridge development URIs to redirectURIs in config-dev.yaml.

ggreer:
This is just a convenience thing for local development.

### DIFF
--- a/examples/config-dev.yaml
+++ b/examples/config-dev.yaml
@@ -17,6 +17,8 @@ staticClients:
 - id: example-app
   redirectURIs:
   - 'http://127.0.0.1:5555/callback'
+  - 'http://localhost:9000/auth/callback'
+  - 'urn:ietf:wg:oauth:2.0:oob'
   name: 'Example App'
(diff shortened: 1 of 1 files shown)


ericchiang:
@ggreer would this be better in the bridge repo? :)

examples/config-dev.yaml
```diff
@@ -17,6 +17,8 @@ staticClients:
 - id: example-app
   redirectURIs:
   - 'http://127.0.0.1:5555/callback'
+  - 'http://localhost:9000/auth/callback'
+  - 'urn:ietf:wg:oauth:2.0:oob'
```
ericchiang: I don't think this actually has to be registered. It's implied by `public: true`

ggreer: If I don't add that, I get `invalid_request: Unregistered redirect_uri ("urn:ietf:wg:oauth:2.0:oob").` when trying certain features.

But you're right, this shouldn't be in this repo.

PR #618, opened 2016-10-17
contrib/openldap: add an OpenLDAP Docker image for testing

ericchiang:
Still need to add docs on:
- [ ] Which ports to expose.

Also, the entrypoint should support
- [ ] Setting a custom base DN.
- [ ] Using custom TLS assets.
- [ ] Seed DB with additional data.


ericchiang:
Some notes on the approach.

Most variables are seeded by environment variables. For example setting `LDAP_DO MAIN` can be used to customize the base DN of the LDAP directory.

The LDAP directory can be seeded by a simple `slapadd` command from a file mounted into the container. We'll use this to create complex hierarchies beforehand.

All of these behaviors are coordinated by the entrypoint.

Still need to do TLS.

```
    sudo rkt run \
        --volume=examples,kind=host,source=$PWD/examples,readOnly=true \
        --mount=volume=examples,target=/etc/dex/ldap \
        --set-env=LDAP_ADD=/etc/dex/ldap/groups.ldif \
        --set-env=LDAP_DOMAIN=dc=coreos,dc=com \
        --net=default:IP=172.16.28.25 \
        quay.io/coreos/openldap:2.4.44
```

cc @rithujohn191

ericchiang:
Going to stop here and try to get this merged as is. We can add more seed data and TLS stuff later.

rithujohn191:
Ran the OpenLDAP container with rkt and tested out the ldap commands. Everything works fine.

PR #620, opened 2016-10-17
server: fix key rotation polling

### LINKED ISSUE #619: server: key rotation polling wait is wrong
If keys rotate every 24 hours and the server fails to create the initial key it will wait 24 hours to re-poll.[0] This is bad and requires occasionally destroying dex in order to re-sync the keys properly.

Change this to poll at a predefined interval. Say 30 seconds.

cc @dghubble 

[0] https://github.com/coreos/dex/blob/3e94e65b68097c057331381fd0434055a4949c3e/server/rotation.go#L79

ericchiang:
fixes #619

cc @dghubble

### DIFF
--- a/server/rotation.go
+++ b/server/rotation.go
@@ -20,7 +20,7 @@ import (
 // often to rotate them, and how long they can validate signatures after rotation.
 type rotationStrategy struct {
 	// Time between rotations.
-	period time.Duration
+	rotationFrequency time.Duration
 
 	// After being rotated how long can a key validate signatues?
 	verifyFor time.Duration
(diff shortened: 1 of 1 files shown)


dghubble:
What as the `strategy.period` and why not set it?

ericchiang:
How long the key are valid for. Used to set the next expiry[0].

Let me give that a better name. Maybe `keysValidFor`

[0] https://github.com/ericchiang/dex/blob/d8033999d56d9600f431e76bc50831108c920b7a/server/rotation.go#L148

dghubble:
Oh, that is kinda unclear. Maybe `rotatePeriod` or `rotateDuration`? If we're being pedantic, frequency isn't the right name.

ericchiang:
I don't want to be pedantic.

dghubble:
:+1:

PR #623, opened 2016-10-19
*: port oob template

ericchiang:
cc @rithujohn191 @dghubble

### DIFF
--- a/server/templates.go
+++ b/server/templates.go
@@ -15,6 +15,7 @@ const (
 	tmplApproval = "approval.html"
 	tmplLogin    = "login.html"
 	tmplPassword = "password.html"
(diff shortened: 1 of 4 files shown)


server/templates.go
```diff
@@ -175,6 +179,14 @@ func (t *templates) approval(w http.ResponseWriter, state, username, clientName
 	renderTemplate(w, t.approvalTmpl, data)
 }
 
+func (t *templates) oob(w http.ResponseWriter, code string) {
```
dghubble: Is out of band mentioned anywhere?

ericchiang: @dghubble I still need to move the "clients" document over https://github.com/coreos/dex/blob/master/Documentation/clients.md

PR #626, opened 2016-10-22
storage/kubernetes: guess namespace from the service account token

ericchiang:
The in cluster kubernetes client currently requires using the
downward API to determine its namespace. However this value can
be determine by inspecting the service account token mounted into
the pod. As a fallback, use this to guess the current namespace.


rithujohn191:
LGMT

PR #627, opened 2016-10-22
*: expand environment variables in config

ericchiang:
Allow users to define config values which are read form environemnt
variables. Helpful for sensitive variables such as OAuth2 client IDs
or LDAP credentials.


rithujohn191:
LGTM

PR #629, opened 2016-10-23
storage/kubernetes: don't automatically print errors on bad HTTP status codes

ericchiang:
These status codes spam the error logs for events like key rotation
and third party resource creation. In these cases "bad" status codes
are expected and shouldn't be automatically printed.


rithujohn191:
LGTM.

PR #632, opened 2016-10-24
Documentation: add a document on storage options

ericchiang:
fyi @dghubble there's some third party resource docs here.

### DIFF
--- a/Documentation/storage.md
+++ b/Documentation/storage.md
@@ -1,0 +1,173 @@
+# Storage options
+
+Dex requires persisting state to perform various tasks such as track refresh tokens, preventing replays, and rotating keys. This document is a summary of the storage configurations supported by dex.
+
+Storage breaches are serious as they can affect applications that rely on dex. Dex saves sensitive data in its backing storage, including signing keys and bcrypt'd passwords. As such, transport security and database ACLs should both be used, no matter which storage option is chosen.
+
+
+## Caveat: running replicated instances
+
+Tests still need to be written to validate that multiple instances of dex behave correctly when using the same storage. While there aren't any technical limitations, edge cases have been observed and progress on these kind of bugs can be found on the [dex issue tracker][issues-transaction-tests].
+
+The dex team suggests running one dex instance most of the time and two instance during upgrades.
+
+## Kubernetes third party resources
+
+__NOTE:__ Dex requires Kubernetes version 1.4+.
+
+Kubernetes third party resources are a way for applications to create new resources types in the Kubernetes API. This allows dex to run on top of an existing Kubernetes cluster without the need for an external database. While this storage may not be appropriate for a large number of users, it's extremely effective for many Kubernetes use cases.
+
+The `ThirdPartyResource` type acts as a description for the new resource a user wishes to create. The following an example of a resource managed by dex:
+
+```
+kind: ThirdPartyResource
+apiVersion: extensions/v1beta1
+metadata:
+  name: o-auth2-client.oidc.coreos.com
+versions:
+  - v1
+description: "An OAuth2 client."
+```
+
+Once the `ThirdPartyResource` is created, custom resources can be created at a namespace level (though there will be a gap between the `ThirdPartyResource` being created and the API server accepting the custom resource). While most fields are user defined, the API server still respects the common `ObjectMeta` and `TypeMeta` values. For example names are still restricted to a small set of characters, and the `resourceVersion` field can be used for an [atomic compare and swap][k8s-api].
+
+The following is an example of a custom `OAuth2Client` resource:
+
+```
+# Standard Kubernetes resource fields
+kind: OAuth2Client
+apiVersion: oidc.coreos.com/v1
+metadata:
+  namespace: foobar
+  name: client-foo
+
+# Custom fields defined by dex.
+clientID: "aclientid"
+clientSecret: "clientsecret"
+redirectURIs:
+- "https://app.example.com/callback"
+```
+
+The `ThirdPartyResource` type and the custom resources can be queried, deleted, and edited like any other resource using `kubectl`.
+
+```
+kubectl get thirdpartyresources # list third party resources registered on the clusters
+kubectl get --namespace=foobar oauth2clients # list oauth2 clients in a given namespace
+```
+
+To reduce administrative overhead, dex creates and manages its own third party resources and may create new ones during upgrades. While not strictly required we feel this is important for reasonable updates. Though, as a result, dex requires access to the non-namespaced `ThirdPartyResource` type. For example, clusters using RBAC authorization would need to create the following roles and bindings:
+
+```
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+metadata:
+  name: dex
+rules:
+  - apiGroups: ["oidc.coreos.com"] # API group created by dex
+    resources: ["*"]
+    verbs: ["*"]
+    nonResourceURLs: []
+  - apiGroups: ["extensions"]
+    resources: ["thirdpartyresources"]
+    verbs: ["create"] # To manage its own resources identity must be able to create thirdpartyresources.
+    nonResourceURLs: []
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+metadata:
+  name: dex
+subjects:
+  - kind: ServiceAccount
+    name: dex                 # Service account assigned to the dex pod.
+    namespace: demo-namespace # The namespace dex is running in.
+roleRef:
+  kind: ClusterRole
+  name: identity
+  apiVersion: rbac.authorization.k8s.io/v1alpha1
+```
+
+The storage configuration is extremely limited since installations running outside a Kubernetes cluster would likely prefer a different storage option. An example configuration for dex running inside Kubernetes:
+
+```
+storage:
+  type: kubernetes
+  config:
+    inCluster: true
+```
+
+Dex determines the namespace it's running in by parsing the service account token automatically mounted into its pod.
+
+## SQL
+
+Dex supports two flavors of SQL, SQLite3 and Postgres. MySQL and CockroachDB may be added at a later time.
+
+Migrations are performed automatically on the first connection to the SQL server (it does not support rolling back). Because of this dex requires privileges to add and alter the tables for its database.
+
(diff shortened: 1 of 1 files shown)


Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
+# Storage options
+
+Dex requires persisting state to perform various tasks such as track refresh tokens, preventing replays, and rotating keys. This document is a summary of the storage configurations supported by dex.
+
+Regardless of the implementation, dex saves sensitive data in its backing storage: signing keys, bcrypt'd passwords, etc. Storage breaches are serious because attacks may propagate to any application which relies on dex. Whatever storage option used, transport security and database ACLs are both musts.
```
ElijahCaine: I would reword this to be:

**Note:** Storage breaches are serious as they can affect applications that rely on dex. Dex saves a lot of sensitive data in its backing storage, including **signing keys** and **bcrypt'd passwords**. As such, transport security and database ACLs should both be used, no matter which storage option is chosen. (LINK TO SETUP DOCS FOR THIS)

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Caveat: Running replicated instances
+
+Validation is still being done to ensure correctness when multiple instances of dex are using the same backing storage. While everything "should just work", edge cases require time to investigate and write tests for. This work can be followed on the issue tracker [here](https://github.com/coreos/dex/issues/600).
```
ElijahCaine: I would reword this to be:

Dex performs correctness validation when multiple instances use the same backing storage.  This should "Just Work" but when it doesn't you should check out the `[dex issue tracker][dex-issue-tracker-url]` and add or comment on a relevant issue.

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+
+For now, instances should run a single instance of dex and occationally run two during upgrades.
```
ElijahCaine: I would reword this to be (if it's accurate):

The Dex team suggests running **one dex instance most of the time** and **two instance during upgrades**.

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Kubernetes third party resources
+
+NOTE: Dex requires Kubernetes version 1.4+.
+
+Kubernetes third party resources are a way for applications to create new resources types in the Kubernetes API. This allows dex to run on top of an existing Kubernetes cluster without the need for an external database. While this storage may not be appropriate for a large number of users, it's extremely effective for may Kubernetes use cases.
```
ElijahCaine: [...] it's extremely effective for **many** Kubernetes use cases.

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+    inCluster: true
+```
+
+Dex determines the namespace it's running in by parsing the service account token automatically mounted into its pod.
+
+## SQL
+
+Dex supports two flavors of SQL, SQLite3 and Postgres. MySQL and CockroachDB may be added at a later time.
+
+Migrations are performed automatically on the first connection to the SQL server (it does not support rolling back). Because of this, dex requires being able to add and alter the tables for its database.
```
ElijahCaine: I would reword the second sentence to be:

Because of this dex requires privileges to add and alter tables for its database.

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Caveat: Running replicated instances
+
+Validation is still being done to ensure correctness when multiple instances of dex are using the same backing storage. While everything "should just work", edge cases require time to investigate and write tests for. This work can be followed on the issue tracker [here](https://github.com/coreos/dex/issues/600).
+
+For now, instances should run a single instance of dex and occationally run two during upgrades.
+
+## Kubernetes third party resources
+
+NOTE: Dex requires Kubernetes version 1.4+.
```
ElijahCaine: I would make `NOTE` **bold**.

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+
+Dex determines the namespace it's running in by parsing the service account token automatically mounted into its pod.
+
+## SQL
+
+Dex supports two flavors of SQL, SQLite3 and Postgres. MySQL and CockroachDB may be added at a later time.
+
+Migrations are performed automatically on the first connection to the SQL server (it does not support rolling back). Because of this, dex requires being able to add and alter the tables for its database.
+
+NOTE: Previous versions of dex required symmetric keys to encrypt certain values before sending them to the database. This feature has not been ported to V2 yet, though it may be added later (but possibly without a migration path for users of the current implementation).
```
ElijahCaine: I would make `Note` **bold**.

ElijahCaine: I would reword the last second sentence to be

This feature has not yet been ported to dex v2.  If it is added later there may not be a migration path for current v2 users.

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+  type: sqlite3
+  config:
+    file: /var/dex/dex.db
+```
+
+Because SQLite3 uses file locks to prevent race conditions, if the ":memory:" value is provided dex will automatically disable support for concurrent database queries.
+
+### Postgres
+
+Dex authenticates to Postgres with a username and password. As stated above, because dex performs migrations, it requires privileged access to its own database. Database table names are not configurable, and to prevent table name clashes with other applications admins should expect to dedicated a database for dex's sole use.
```
ElijahCaine: I would reword this paragraph to be:

Admins may want to dedicate a database to dex for the following reasons:

1. Dex requires privileged access to its database because it performs migrations.
2. Dex's database table names are not configurable; when shared with other applications there may be table name clashes.

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+
+The SSL "mode" corresponds to the `github.com/lib/pq` package [connection options](https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters). If unspecified, dex defaults to the strictest mode "verify-full".
+
+## Adding a new storage options
+
+Each storage implementation bears a large ongoing maintenance cost and needs to be updated every time a feature requires storing a new type. Bugs often require in depth knowledge of the backing software, and much of this work will be done by developers who are not the original authors. Changes to dex which add new storage implementations are not merged lightly.
+
+Those who still want to construct a proposal for a new storage should review the following packages:
+
+* `github.com/coreos/dex/storage`: Interface definitions which the storage must implement. NOTE: This package is not stable.
```
ElijahCaine: This should be a clickable link.

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+* `github.com/coreos/dex/storage/conformance`: Conformance tests which storage implementations must pass.
```
ElijahCaine: This should be a clickable link.

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+
+Any proposal to add a new implementation must address the following:
+
+* Integration testing setups (Travis and developer workstations).
+* Transactional requirements: atomic deletes, updates, etc.
+* Is there an established, reasonable Go client?
```
ElijahCaine: I would reword this last bullet to be:

* Is there an established and reasonable Go client?

Documentation/storage.md
ericchiang: Thanks! Though this is the document that should have those setups.

Documentation/storage.md
ericchiang: > Dex performs correctness validation when multiple instances use the same backing storage. 

That's not the same thing. This is stating that we still need to write the tests to prove correctness and you probably shouldn't be running these kind of setups in the first place. How does this sound?

> Tests still need to be written to validate that multiple instances of dex behave correctly when using the same storage.
> While there aren't any technical limitations, edge cases have been observed and progress on these kind of bugs can be found on the `[dex issue tracker]`.

Documentation/storage.md
```diff
@@ -0,0 +1,161 @@
...
+      caFile: /etc/dex/postgres.ca
+```
+
+The SSL "mode" corresponds to the `github.com/lib/pq` package [connection options](https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters). If unspecified, dex defaults to the strictest mode "verify-full".
+
+## Adding a new storage options
+
+Each storage implementation bears a large ongoing maintenance cost and needs to be updated every time a feature requires storing a new type. Bugs often require in depth knowledge of the backing software, and much of this work will be done by developers who are not the original authors. Changes to dex which add new storage implementations are not merged lightly.
+
+Those who still want to construct a proposal for a new storage should review the following packages:
```
ericchiang: Wouldn't that imply that it's part of the SQL section (as it's not)?

Documentation/storage.md
ericchiang: We can't link to a godoc.org until we've made this branch the default branch. Part of that getting to that point is writing docs like this :)

ericchiang:
@ElijahCaine thanks. Have a few counter comments, but other than that will rework this PR later today.

Documentation/storage.md
ElijahCaine: Fair enough.  I wanted to click on it and figured readers would too, but if it's technically not an option that's 👍

Documentation/storage.md
ElijahCaine: Ah.  I think I read the thing about signing keys and didn't read anything about how to set that up (because it hasn't been ported yet). That's not required if this is the only doc that covers those topics.

ericchiang:
@ElijahCaine updated with a second commit. Going to leave this here if you want to add any last second notes. Will squash and merge in a bit.

Documentation/storage.md
```diff
@@ -121,7 +140,10 @@ Because SQLite3 uses file locks to prevent race conditions, if the ":memory:" va
 
 ### Postgres
 
-Dex authenticates to Postgres with a username and password. As stated above, because dex performs migrations, it requires privileged access to its own database. Database table names are not configurable, and to prevent table name clashes with other applications admins should expect to dedicated a database for dex's sole use. 
+When using Postgres, admins may want to dedicate a database to dex for the following reasons:
+
+1. Dex requires privileged access ot its database because it performs migrations.
+1. Dex's database table names are not configurable; when shared with other applications there may be table name clashes.
```
ElijahCaine: This should probably be `2.`

ericchiang:
Haha, thanks I'll update again.

Documentation/storage.md
ericchiang: updated, thanks

PR #633, opened 2016-10-25
docs: mention CoreOS docs style guide

ElijahCaine:
That style guide is (or ought to be) enforced in all CoreOS repository documentation (`.md` files).
This change makes that more obvious.


ericchiang:
@ElijahCaine we're actually trying to switch to the "dev" branch eventually, which doesn't have a "contributing" document yet. Will add this to master but we should also think about what we want to put in the next version.

PR #634, opened 2016-10-25
storage/kubernetes: set CurrentContext when the Kubeconfig file contains only one context

rithujohn191:
@ericchiang


ericchiang:
lgtm

PR #649, opened 2016-10-28
api: add gRPC endpoints for creating, updating and deleting passwords

### LINKED ISSUE #639: api: add gRPC endpoints for adding and changing passwords
- The gRPC definitions are in `api/api.proto`. Add new types and calls there.
- Generate `api/api.pb.go` using `make grpc`.
- Update the `server/api.go` definitions.

Questions:
- Should we accept plain text passwords? (ericchiang) Probably not.
- Password changes should not check your current password because this API might be used by someone without a user's password (e.g. an admin).

rithujohn191:
closes #639

cc @ericchiang


server/api.go
```diff
@@ -62,3 +63,89 @@ func (d dexAPI) DeleteClient(ctx context.Context, req *api.DeleteClientReq) (*ap
...
+		return nil, errors.New("no email supplied")
+	}
+	if req.NewHash == nil && req.NewUsername == "" {
+		return nil, errors.New("nothing to update")
+	}
+
+	if req.NewHash != nil {
+		actual, err := bcrypt.Cost(req.NewHash)
+		if err != nil {
+			log.Printf("api: bcrypt.Cost, error: %s", err)
```
ericchiang: This should return an error instead of logging it.

server/api.go
```diff
@@ -62,3 +63,89 @@ func (d dexAPI) DeleteClient(ctx context.Context, req *api.DeleteClientReq) (*ap
...
+		}
+		if actual < bcrypt.MinCost {
+			log.Printf("api: hash does not meet MinCost requirement: %d", bcrypt.MinCost)
```
ericchiang: Same as above. Return an error instead of logging.

Also we don't want `MinCost` we probably want `DefaultCost`. Min is the absolute minimum allowed bcrypt cost, not the minimal secure cost.

server/api.go
```diff
@@ -62,3 +63,89 @@ func (d dexAPI) DeleteClient(ctx context.Context, req *api.DeleteClientReq) (*ap
...
+		}
+	}
+
+	updater := func(old storage.Password) (storage.Password, error) {
+		new := storage.Password{
```
ericchiang: `new` is a keyword. Generally not a good idea to assign to it.

Also don't create a new struct here, just modify `old`. There might be new Password fields added later that you'll zero otherwise. See example in #648.

server/api.go
```diff
@@ -62,3 +63,89 @@ func (d dexAPI) DeleteClient(ctx context.Context, req *api.DeleteClientReq) (*ap
...
+
+		return new, nil
+	}
+
+	if err := d.s.UpdatePassword(req.Email, updater); err != nil {
+		if err == storage.ErrNotFound {
+			return &api.UpdatePasswordResp{NotFound: true}, nil
+		}
+		log.Printf("api: failed to update password: %v", err)
+		return nil, err
```
ericchiang: Probably want to use fmt.Errorf to add context as well. For example

```
return nil, fmt.Errorf("update password: %v", err)
```

server/api.go
```diff
@@ -62,3 +63,89 @@ func (d dexAPI) DeleteClient(ctx context.Context, req *api.DeleteClientReq) (*ap
...
+		return nil, errors.New("no email supplied")
+	}
+
+	err := d.s.DeletePassword(req.Email)
+	if err != nil {
+		if err == storage.ErrNotFound {
+			return &api.DeletePasswordResp{NotFound: true}, nil
+		}
+		log.Printf("api: failed to delete password: %v", err)
+		return nil, err
```
ericchiang: Same comment as above, use `fmt.Errorf`

server/api.go
```diff
@@ -58,7 +60,84 @@ func (d dexAPI) DeleteClient(ctx context.Context, req *api.DeleteClientReq) (*ap
...
 }
+
+func (d dexAPI) CreatePassword(ctx context.Context, req *api.CreatePasswordReq) (*api.CreatePasswordResp, error) {
+	if req.Password == nil {
+		return nil, errors.New("no password supplied")
+	}
+
+	p := storage.Password{
+		Email:    req.Password.Email,
+		Hash:     req.Password.Hash,
```
ericchiang: We should probably check the hash the same way we do in UpdatePassword. Maybe pull out that logic into a helper function and use it here too?

server/api.go
```diff
@@ -58,7 +60,99 @@ func (d dexAPI) DeleteClient(ctx context.Context, req *api.DeleteClientReq) (*ap
...
+		return fmt.Errorf("given hash cost = %d, does not meet minimum cost requirement = %d", actual, bcrypt.DefaultCost)
+	}
+	return nil
+}
+
+func (d dexAPI) CreatePassword(ctx context.Context, req *api.CreatePasswordReq) (*api.CreatePasswordResp, error) {
+	if req.Password == nil {
+		return nil, errors.New("no password supplied")
+	}
+	if req.Password.Hash != nil {
```
ericchiang: They have to provide a password. So if this is nil we should return an error.

rithujohn191: Also the storage package doc recommends a cost value of at least 14. Where as DefaultCost = 10. Should I change the condition to check for a cost value of 14 instead?

rithujohn191:
Changed storage package doc to have min cost value of 10 instead of 14.

rithujohn191:
All the unit tests passed:
[rjohn@e4-b3-18-d3-91-0e server]$ go test -v .
=== RUN   TestCreatePassword
--- PASS: TestCreatePassword (0.08s)
=== RUN   TestUpdatePassword
--- PASS: TestUpdatePassword (0.00s)
=== RUN   TestDeletePassword
--- PASS: TestDeletePassword (0.00s)

server/api_test.go
```diff
@@ -1 +1,60 @@
...
+var serv = NewAPI(s)
+
+func TestCreatePassword(t *testing.T) {
+	ctx := context.Background()
+	p := api.Password{
+		Email:    "test@example.com",
+		Username: "test",
+		UserId:   "test123",
+	}
+	hash, err := bcrypt.GenerateFromPassword([]byte("test1"), 10)
```
ericchiang: Generate this value beforehand instead of requiring each test run to compute a bcrypt.

For example, just use this instead:

```
// bcrypt hash of the value "test1" with cost 10
hash := []byte("$2a$10$XVMN/Fid.Ks4CXgzo8fpR.iU1khOMsP5g9xQeXuBm1wXjRX8pjUtO")
```

server/api_test.go
```diff
@@ -1 +1,60 @@
...
+	"context"
+	"testing"
+
+	"golang.org/x/crypto/bcrypt"
+
+	"github.com/coreos/dex/api"
+	"github.com/coreos/dex/storage/memory"
+)
+
+var s = memory.New()
```
ericchiang: So just have one test that
1. Creates a password.
2. Updates the password.
3. Deletes the password

server/api_test.go
```diff
@@ -1 +1,60 @@
...
+func TestUpdatePassword(t *testing.T) {
+	ctx := context.Background()
+	req := api.UpdatePasswordReq{
+		Email:       "test@example.com",
+		NewUsername: "test1",
+	}
+
+	if _, err := serv.UpdatePassword(ctx, &req); err != nil {
+		t.Errorf("Unable to update password: %v", err)
+	}
```
ericchiang: need to get the user after to ensure the new username was actually set.

server/api_test.go
```diff
@@ -1 +1,60 @@
...
+}
+
+func TestDeletePassword(t *testing.T) {
```
ericchiang: I commented on this above, but two tests shouldn't ever be dependent on one another. If they are, it should probably just be one test :)

rithujohn191: Since both UpdatePassword and DeletePassword rely on CreatePassword to have an existing entry, would you rather have me put them all together?

rithujohn191:
Unit test passed:
$ go test -v .
=== RUN   TestPassword
--- PASS: TestPassword (0.00s)

server/api_test.go
```diff
@@ -1 +1,59 @@
...
+	"testing"
+
+	"github.com/coreos/dex/api"
+	"github.com/coreos/dex/storage/memory"
+)
+
+// Attempts to create, update and delete a test Password
+func TestPassword(t *testing.T) {
+	var s = memory.New()
+	var serv = NewAPI(s)
```
ericchiang: nit: `var serv = ...` is redundant, just use `serv := ...` instead.

ericchiang:
one last nit, other than that lgtm

PR #658, opened 2016-11-01
*: don't error out if a username doesn't exist in the backing connector

### LINKED ISSUE #656: server: don't throw a 500 when a user doesn't exist
Thanks @kans

ericchiang:
Instead of throwing a 500 error if a user enters an invalid name,
display the same text box as if the user had entered the wrong
password.

NOTE: An invalid username now returns much quicker than an invalid
password. Consider adding an arbitrary sleep in the future if we
care about masking which was invalid.

cc @rithujohn191 @kans @dghubble 

closes #656


rithujohn191:
lgmt

PR #665, opened 2016-11-03
cmd/dex: expose IDTokensValidFor and RotateKeysAfter server options in config

### LINKED ISSUE #660: cmd/dex: expose IDTokensValidFor and RotateKeysAfter server options in config
Though those names suck. Perhaps in the config it could look like:

```yaml
expiry:
    idTokens: 48h
    signingKeys: 12h
```

cc @robszumski

rithujohn191:
closes #660.

@ericchiang

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -27,6 +27,7 @@ type Config struct {
 	Web        Web         `json:"web"`
 	OAuth2     OAuth2      `json:"oauth2"`
 	GRPC       GRPC        `json:"grpc"`
+	Expiry     Expiry      `json:"expiry"`
 
 	Templates server.TemplateConfig `json:"templates"`
 
@@ -209,4 +210,13 @@ func (c *Connector) UnmarshalJSON(b []byte) error {
 		Config: connConfig,
 	}
 	return nil
+}
+
+// Expiry holds configuration for the validity period of components.
+type Expiry struct {
+	// SigningKeys defines the duration of time after which the SigningKeys will be rotated.
+	SigningKeys string `json:"signingKeys"`
+
+	// IdTokens defines the duration of time for which the IdTokens will be valid.
+	IDTokens string `json:"idTokens"`
 }
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"time"
 
 	"github.com/ghodss/yaml"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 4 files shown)


rithujohn191:
@ericchiang  I know we discussed that I should ParseDuration before the gRPC code, but since there is a possibility that both the durations might not be provided, I thought  it would be easier to do it after defining serverConfig. Let me know if you want me to change it.

ericchiang:
Travis caught some issues with the struct tags. Please take a look.

cmd/dex/config.go
```diff
@@ -215,3 +216,12 @@ func (c *Connector) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	}
 	return err
 }
+
+// Expiry holds various expiration time durations
```
ericchiang: Maybe "Expiry holds configuration for the validity period of components."

cmd/dex/serve.go
```diff
@@ -154,6 +155,20 @@ func serve(cmd *cobra.Command, args []string) error {
 		TemplateConfig:         c.Templates,
 		EnablePasswordDB:       c.EnablePasswordDB,
 	}
+	if c.Expiry.SigningKeys != "" {
+		signingKeys, err := time.ParseDuration(c.Expiry.SigningKeys)
+		if err != nil {
+			return fmt.Errorf("parsing SigningKeys expiry: %v", err)
```
ericchiang: nit: `signingKeys` (since that's the config field).

ericchiang:
Sorry, I appear to have caused a merge conflict :/

ericchiang:
lgtm

chrisdone:
Has there been any effort to add a webhook for this? The SP doesn't know exactly when Dex refreshes its signing keys, so I was wondering whether there was any interest in making Dex push to its static clients a HTTP request to say its keys have been rotated.

ericchiang:
chrisdone@ key endpoints have expiry data in a Cache-Control HTTP header. You can just use that :) you definitely don't need a webhook for this.

chrisdone:
@ericchiang Thanks. 

Upon reflection, neither the webhook nor the expiry data are sufficient to avoid race conditions:

* Due to time skew where the SP and Dex's times become out of sync, SP can jump the gun. 
* Dex could be restarted or re-provisioned during the expiry interval, meaning that the JWKs would be rotated earlier than expected.

In this situation, the expiry datetime is essentially _advisory_, to reduce cache misses. A webhook would mitigate the time skew and restarts, causing fewer cache misses, but is also _advisory_.

I will have to implement an on-demand system in SP which requests an update from Dex on cache misses (whenever JWKs fail to decode a JWT in an attempted authentication). 

The only missing piece of the puzzle is to avoid DoS attacks causing intentional cache misses. Some kind of rate limit of e.g. 5 minutes will probably have to do. :thinking:

ericchiang:
I don't maintain dex anymore, so you probably want to open up a new issue.

You probably want to figure out if there's a bug before jumping to
solutions (and again a webhook is overkill). For example, dex's key
rotation is based on its database, not an in-memory value, so restarting
won't impact its rotation schedule. You can also look at the go-oidc client
to figure out how it details with caching https://github.com/coreos/go-oidc.
We haven't had any reported issues with that logic so far, so this might be
a bug in your OpenID Connect client that needs fixing.

PR #669, opened 2016-11-04
cmd/dex: only expand from env for storages and connectors

ericchiang:
Bcrypt'd hashes have "$" characters in them. This means that #667
(accepting actually bcrypted values) combined with #627 (expanding
config with environment variables) broke the example config.

For now, allow storages and connectors to expand their configs from
the environment, but don't do this anywhere else.


rithujohn191:
lgtm

PR #672, opened 2016-11-07
api: add version endpoint

### LINKED ISSUE #659: api: add version endpoint
something like this

```proto
message Version {
    // semantic version of the server
    string server
    // numeric version of the API
    int api
}

message VersionReq {}

Version(VersionReq) Version
```

rithujohn191:
Closes #659.
cc @ericchiang


rithujohn191:
Should I also include a definition for GetVerion in server/api.go ?

ericchiang:
@rithujohn191 you can't add new server definitions without their implementation, yes.

PR #683, opened 2016-11-14
api: add gRPC definition for version endpoint.

### LINKED ISSUE #659: api: add version endpoint

rithujohn191:
Closes #659.


api/api.proto
```diff
@@ -80,16 +80,34 @@ message DeletePasswordResp {
...
+
+// VersionResp returns version info.
+message VersionResp {
+  Version version = 1;
+}
+
 // Dex represents the dex gRPC service.
 service Dex {
-  // CreateClient attempts to create the client.
+  // CreateClient creates the client.
```
ericchiang: nit: creates _a_ client.

api/api.proto
```diff
@@ -80,16 +80,34 @@ message DeletePasswordResp {
...
+  string server = 1;
+  // Numeric version of the API.
+  int32 api = 2;
+}
+
+// VersionReq is a request to fetch version info.
+message VersionReq {}
+
+// VersionResp returns version info.
+message VersionResp {
```
ericchiang: This can probably just hold the `Version` fields directly. We don't need the `Version` message.

api/api.proto
```diff
@@ -80,16 +80,34 @@ message DeletePasswordResp {
   bool not_found = 1;
 }
 
+// Version holds the version info of components.
+message Version {
+  // Semantic version of the server.
+  string server = 1;
+  // Numeric version of the API.
```
ericchiang: Please expand this comment to reflect the comment in `server/api.go`. Maybe even just copy it here.

We expect outside users to read this file, not the Go code.

api/api.proto
```diff
@@ -80,16 +80,34 @@ message DeletePasswordResp {
...
-  // CreatePassword attempts to create the password.
+  // CreatePassword creates the password.
   rpc CreatePassword(CreatePasswordReq) returns (CreatePasswordResp) {};
-  // UpdatePassword attempts to modify existing password.
+  // UpdatePassword modifies existing password.
   rpc UpdatePassword(UpdatePasswordReq) returns (UpdatePasswordResp) {};
-  // DeletePassword attempts to delete the password.
+  // DeletePassword deletes the password.
   rpc DeletePassword(DeletePasswordReq) returns (DeletePasswordResp) {};
+  // GetVersion gets the version info.
```
ericchiang: Maybe `GetVersion` returns version information of the server.

ericchiang:
> Small note: .proto requires you to decalre it as int32 or int64, there is no default int type. Chose int32 for now. let me know if you want to change it.

int32 sounds fine. Doubt we'll run out of space.

ericchiang:
lgtm

PR #689, opened 2016-11-15
connector/ldap: Always set tls.Config.ServerName, to support LDAP ser…

cjyar:
…vers with public CA certs.

It seems if you're going to use TLS at all, you always want to set ServerName. This way, you can talk to a remote LDAPS server and use the local host's default CA certificates, with no need to specify the rootCA ldap config.

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -212,7 +212,7 @@ func (c *Config) OpenConnector() (interface {
 		}
 	}
 
-	tlsConfig := new(tls.Config)
+	tlsConfig := &tls.Config{ServerName: host}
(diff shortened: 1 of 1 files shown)


ericchiang:
It's really weird that we have to set this in the first place, this should be inferred as part of the dial call[0].

It looks like the LDAP client actually does its own handshake instead of using tls.Dial[1]. Will merge this and open an issue on the LDAP client.


[0] https://github.com/golang/go/blob/go1.7.3/src/crypto/tls/tls.go#L134-L141
[1] https://github.com/go-ldap/ldap/blob/d0a5ced67b4dc310b9158d63a2c6f9c5ec13f105/conn.go#L124-L138

cjyar:
Thanks!

PR #690, opened 2016-11-15
Documentation: LDAP connector documentation.

rithujohn191:


Documentation/ldap-connector.md
```diff
@@ -0,0 +1,86 @@
+# Overview of the `ldap` Connector
+
+The `ldap` connector allows email/password based authentication hosted by dex, backed by a LDAP directory. All the configuration parameters required by the `ldap` connector are held in [Config struct](../connector/ldap/ldap.go).
```
ericchiang: Don't need `hosted by dex`

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,86 @@
...
+
+The connector executes two primary queries:
+
+1. Finding the user based on the end user's credentials.
+2. Searching for groups using the user entry.
+
+User entries are expected to have an email attribute (configurable through "emailAttr"), and a display name attribute (configurable through "nameAttr").
```
ericchiang: `emaliAttr` and `nameAttr` (use a backquote)

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,86 @@
+# Overview of the `ldap` Connector
```
ericchiang: Title should be something like "Authentication through LDAP"

Then we would have a `## Overview` subsection

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,86 @@
...
+The `ldap` connector allows email/password based authentication hosted by dex, backed by a LDAP directory. All the configuration parameters required by the `ldap` connector are held in [Config struct](../connector/ldap/ldap.go).
+
+The connector executes two primary queries:
+
+1. Finding the user based on the end user's credentials.
+2. Searching for groups using the user entry.
+
+User entries are expected to have an email attribute (configurable through "emailAttr"), and a display name attribute (configurable through "nameAttr").
+
+### Setting the Configuration
```
ericchiang: This can just be combined with the next section. Also we can probably just use "Configuration" as the header. Also two "#", not three.

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,86 @@
...
+
+To set the `ldap` connector configuration in dex, create a config file at `example/config-ldap.yaml`, then use the following command to start an instance of dex:
+
+```
+./bin/dex serve examples/config-dev.yaml
+```
+
+### Example: User authentication
+
+The following is an example config file that can be used by the `ldap` connector to submit a [SearchRequest](https://godoc.org/gopkg.in/ldap.v2#SearchRequest) to authenticate a user.
```
ericchiang: We don't want to refer to the upstream client. It's an implementation detail that we don't expose.

rithujohn191:
@ericchiang.

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,86 @@
...
+  connectors:
+   - type: ldap
+     config:
+       # host and port of the LDAP server in form "host:port".
+       host: ldap.example.com:636
+       # The following field is required if using port 389.
+       # insecureNoSSL: true
+       # Default: use the host's root CA set.
+       rootCA: /etc/dex/ldap.ca
+       # DN to bind as for search operations.
```
ericchiang: Note that these are optional and not required if your LDAP server provides sufficient access for anonymous auth.

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,86 @@
...
+       userSearch:
+         # Would translate to the query "(&(objectClass=person)(uid=<username>))".
+         baseDN: cn=users,dc=example,dc=com
+         filter: "(objectClass=person)"
+         username: uid
+         idAttr: uid
+         # Required. Attribute to map to Email.
+         emailAttr: mail
+         # Maps to display name of users.
+         nameAttr: name
```
ericchiang: Why no group search?

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,86 @@
...
+./bin/dex serve examples/config-dev.yaml
+```
+
+### Example: User authentication
+
+The following is an example config file that can be used by the `ldap` connector to submit a [SearchRequest](https://godoc.org/gopkg.in/ldap.v2#SearchRequest) to authenticate a user.
+
+```yaml
+
+  connectors:
```
ericchiang: Un-indent this so users can just copy paste it into their config.

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,86 @@
...
+The connector executes two primary queries:
+
+1. Finding the user based on the end user's credentials.
+2. Searching for groups using the user entry.
+
+User entries are expected to have an email attribute (configurable through "emailAttr"), and a display name attribute (configurable through "nameAttr").
+
+### Setting the Configuration
+
+To set the `ldap` connector configuration in dex, create a config file at `example/config-ldap.yaml`, then use the following command to start an instance of dex:
```
ericchiang: We don't need to explain this here. This should be something like "in your dex config file" with a link to the getting started guide. https://github.com/coreos/dex/blob/master/Documentation/getting-started.md

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,86 @@
...
+       # The following field is required if using port 389.
+       # insecureNoSSL: true
+       # Default: use the host's root CA set.
+       rootCA: /etc/dex/ldap.ca
+       # DN to bind as for search operations.
+       bindDN: uid=seviceaccount,cn=users,dc=example,dc=com
+       bindPW: password
+       userSearch:
+         # Would translate to the query "(&(objectClass=person)(uid=<username>))".
+         baseDN: cn=users,dc=example,dc=com
```
ericchiang: We need to comment these fields waaaay more.

rithujohn191:
YAML does not allow extra lines between attributes :(

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,99 @@
+# Authentication through LDAP
+
+## Overview
+
+The LDAP connector allows email/password based authentication, backed by a LDAP directory. All the configuration parameters required by the connector are held in [Config struct](../connector/ldap/ldap.go).
```
ericchiang: We don't need to link to source code here. Probably just cut the second sentence.

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,99 @@
...
+
+The connector executes two primary queries:
+
+1. Finding the user based on the end user's credentials.
+2. Searching for groups using the user entry.
+
+## Example: User authentication
```
ericchiang: Just name this section `Configuration`

Documentation/ldap-connector.md
```diff
@@ -0,0 +1,99 @@
...
+
+User entries are expected to have an email attribute (configurable through `emailAttr`), and a display name attribute (configurable through `nameAttr`). The following is an example config file that can be used by the LDAP connector to authenticate a user.
+
+```yaml
```
ericchiang: For inline comments, don't add a newline a the end of every sentence, add it add it at about 80 characters.

For example, don't do this:

```
# The DN and password for an application service account.
# The connector uses these credentials to search for users and groups.
# Optional. Not required if the LDAP server provides access for anonymous auth.
```

Do this:

```
# The DN and password for an application service account. The connector uses
# these credentials to search for users and groups. Not required if the LDAP
# server provides access for anonymous auth.
```

Calpicow:
I've found that the `id` attribute also needs to be set for it to work at all.

ericchiang:
Ah you're totally right, I think we even check that earlier now that https://github.com/coreos/dex/pull/686 is in.

rithujohn191:
Thanks @Calpicow

ericchiang:
lgtm

Issue #679, opened 2016-11-10
api: add call to list passwords


ericchiang:
The storage implementation doesn't currently support a ListPasswords call. I've been holding off on adding this because of https://github.com/coreos/dex/issues/613, however we need this call.

We should:

- [ ] Add a ListPasswords() method to the storage interface (will require updating all implementations).
- [ ] Add a gRPC call for listing passwords.

We can rework the internal bits once we have pagination/search.

PR #695, opened 2016-11-17
api: add call to list passwords

rithujohn191:
1.  Added a ListPasswords() method to the storage interface 
        1.1 updated all implementations of the interface.
2.  Added a gRPC call for listing passwords.


storage/conformance/conformance.go
```diff
@@ -285,13 +285,30 @@ func testPasswordCRUD(t *testing.T, s storage.Storage) {
...
+	var passwordList []storage.Password
+	passwordList = append(passwordList, password)
+
+	listAndCompare := func(want []storage.Password) {
+		passwords, err := s.ListPasswords()
+		if err != nil {
+			t.Errorf("list password: %v", err)
+			return
+		}
+		if diff := pretty.Compare(want, passwords); diff != "" {
```
ericchiang: We probably need to sort these before comparing them, right?

server/api.go
```diff
@@ -171,3 +171,27 @@ func (d dexAPI) GetVersion(ctx context.Context, req *api.VersionReq) (*api.Versi
...
+	if err != nil {
+		log.Printf("api: failed to list passwords: %v", err)
+		return nil, fmt.Errorf("list passwords: %v", err)
+	}
+
+	var passwords []*api.Password
+	for _, password := range passwordList {
+		p := api.Password{
+			Email:    password.Email,
+			Hash:     password.Hash,
```
ericchiang: Can we not return the Hash? I don't think there's a good reason for us to do this.

rithujohn191:
Closes #679.

@ericchiang

ericchiang:
lgtm. Can you run the Kubernetes conformance tests and report back?

rithujohn191:
PASS
ok  	github.com/coreos/dex/storage/kubernetes	1.584s

Kubernetes conformance tests passed

Issue #536, opened 2016-08-03
LDAP groups can only be queried by full DN


wyattanderson:
Our LDAP implementation uses POSIX-style groups (`objectClass=posixGroup`) which store a `memberUid` (e.g. `wyatt`) instead of a `member` (e.g. `uid=wyatt,ou=People,dc=example,dc=com`). Since the `ID` of the remote identity is set to the user's DN and the `searchGroupFilter` parameter only seems to support searching by the `fullUserID`, we can't leverage the groups functionality here because we can't search group membership by DN. Additionally, if I hard-code the `searchGroupFilter` for testing purposes, the group entries that come back are the full DN of the group which is mildly cumbersome for consumers to handle. It might be nice for this to be configurable.

I'm beginning to think that for our purposes it might be better to maintain a fork of Dex with a custom connector that suits our needs (and peculiarities) better, but I thought I'd throw these feature requests out there anyway if the maintainers think it would be a net positive:
- Allow configuration of the stored identity `ID` field to be derived from an attribute (say, `uid`) instead of the DN
- Allow group search by DN or `memberUid` (see [this random JIRA issue from the internet](https://www.jfrog.com/jira/browse/RTFACT-8011) that has an interesting approach)
- Allow groups to be represented by attribute (e.g. `cn`) instead of DN

If these are reasonable requests, I'd be happy to submit a PR. Thanks!

ericchiang:
> Our LDAP implementation uses POSIX-style groups (objectClass=posixGroup) which store a memberUid (e.g. wyatt) instead of a member (e.g. uid=wyatt,ou=People,dc=example,dc=com). Since the ID of the remote identity is set to the user's DN and the searchGroupFilter parameter only seems to support searching by the fullUserID, we can't leverage the groups functionality here because we can't search group membership by DN. 

We could allow the format string to have a "last component" field to accomplish this. E.g. `(memberUid=%u)` would give you the full ID, but `(memberUid=%U)` would give you just the last component of the ID.

> Additionally, if I hard-code the searchGroupFilter for testing purposes, the group entries that come back are the full DN of the group which is mildly cumbersome for consumers to handle. It might be nice for this to be configurable.

We considered doing this in the initial approach. Would be happy to have the groups returned also just use the last component.

> Allow groups to be represented by attribute (e.g. cn) instead of DN

This would also be a reasonable option.

[...]

PR #698, opened 2016-11-18
Allow getAttr to return DN

Calpicow:
Specify "DN" as attribute name to return, but will only work if not present in ldap.Entry.Attributes
Use when full DN is stored in groupSearch's userAttr

Related to #536

### DIFF
--- a/Documentation/ldap-connector.md
+++ b/Documentation/ldap-connector.md
@@ -11,7 +11,9 @@ The connector executes two primary queries:
 
 ## Configuration
 
-User entries are expected to have an email attribute (configurable through `emailAttr`), and a display name attribute (configurable through `nameAttr`). The following is an example config file that can be used by the LDAP connector to authenticate a user.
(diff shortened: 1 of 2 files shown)


ericchiang:
Seems reasonable, but most LDAP implementations we've dealt with actually embed the DN in the entry. What tech are you trying to query that doesn't?

Also I'd be helpful to include a note in the LDAP documentation for this behavior.

Calpicow:
I'm not too familiar with LDAP implementations elsewhere, but the one we have doesn't contain the DN path as any other attribute.

ericchiang:
Okay np. As part of this PR can you add a comment to the following two locations to document this behavior?

* https://github.com/Calpicow/dex/blob/677f8263b3959100cf604c30c6735103ed86e0ae/connector/ldap/ldap.go#L37
* https://github.com/coreos/dex/blob/master/Documentation/ldap-connector.md

ericchiang:
lgtm

PR #701, opened 2016-11-18
connector/ldap: use gopkg.in/ldap.v2's escape filter

ericchiang:
Use the escape filter method provided by the upstream LDAP package
instead of rolling our own.


rithujohn191:
LGTM

PR #708, opened 2016-11-23
Documentation: clarify difference between LDAP ports and security guarentees

ericchiang:
Now that LDAP supports an `insecureSkipVerify` option, clarify that
`insecureNoTLS` is an extremely bad choice and as such we may drop
support for 389 in the future.

However, since we send plain text passwords from our frontend to our
backend, this probably gets us into a bigger conversation about dex's
TLS story. For example when terminiation is approporate. cc'ing
@dghubble for thoughts on how that might apply to our internal uses.

We probably want an overaching security doc at some point, but that
can be another PR.

cc @ElijahCaine since this is a docs review.

### DIFF
--- a/Documentation/ldap-connector.md
+++ b/Documentation/ldap-connector.md
@@ -9,6 +9,12 @@ The connector executes two primary queries:
 1. Finding the user based on the end user's credentials.
 2. Searching for groups using the user entry.
 
+## Security considerations
+
+Dex attempts to bind with the backing LDAP server using the end user's _plain text password_. Though some LDAP implementations allow passing hashed passwords, dex doesn't support hashing and instead _strongly recommends that all administrators just use TLS_. This can often be achieved by using port 636 instead of 389, and administrators that choose 389 are actively leaking passwords.
+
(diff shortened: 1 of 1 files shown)


Documentation/ldap-connector.md
```diff
@@ -9,25 +9,44 @@ The connector executes two primary queries:
 1. Finding the user based on the end user's credentials.
 2. Searching for groups using the user entry.
 
+## Security considerations
+
+Dex attempts to bind with the backing LDAP server using the end user's _plain text password_. Though some LDAP implementations allow passing hashed passwords, dex doesn't support hashing and instead _strongly recommends that all administrators just use TLS_. This can often be achieved by using port 636 instead of 389, and administrators that choose 389 are actively leaking passwords.
```
benley: Is this actually correct?  My understanding is that ldap with STARTTLS is on port 389, and that port 636 is for the older (and somewhat depcrecated?) LDAPS protocol, with a direct SSL transport.

ericchiang: Yes, you're correct. I think we never worked out the STARTTLS vv LDAPS strategy. Biggest goal of this section was to prevent users from just using port 389 without any encryption. Please feel free to send a follow up or open an issue if you're interested.

PR #711, opened 2016-11-30
*: add theme based frontend configuration

ericchiang:
This PR reworks the web layout so static files can be provided and
a "themes" directory to allow a certain degree of control over logos,
styles, etc.

This PR does NOT add general support for frontend customization,
only enough to allow us to start exploring theming internally.
The dex binary also must now be run from the root directory since
templates are no longer "compiled into" the binary.

The docker image has been updated with frontend assets.


ericchiang:
cc @rithujohn191 @amrutac 

Tested the docker image with:

```
make docker-image
sudo docker run -p 5556:5556 -d \
    -v $PWD/examples/config-dev.yaml:/examples/config-dev.yaml |
    quay.io/coreos/dex:$( ./scripts/git-version ) |
    serve /examples/config-dev.yaml
./bin/example-app
```

server/server.go
```diff
@@ -56,7 +56,32 @@ type Config struct {
...
+	// It is expected to contain the following directories:
+	//
+	//   * static - Static static served at "( issuer URL )/static".
+	//   * templates - HTML templates controlled by dex.
+	//   * themes/(theme) - Static static served at "( issuer URL )/theme".
+	//
+	Dir string
+
+	// Defaults to "( issuer URL )/theme/logo.png"
+	LogoURL string
```
amrutac: Just realized that I need a fav icon url as well. Any chance you can add a <code>FavIconURL</code> as well?

amrutac: Here's the default one.
favicon

ericchiang: Follow up PR?

ericchiang: Plus we want to either put the fav icon URL on the connector or derive it from the ID, so it'll probably take a different implementation.

ericchiang:
per irl conversation with @amrutac we're going to punt on the fav icon. @rithujohn191 want to take a quick look?

server/templates.go
```diff
@@ -18,74 +20,117 @@ const (
...
-			return nil, fmt.Errorf("parse files: %v", err)
+		return fmt.Errorf("stat directory %q: %v", dir, err)
+	}
+	if !stat.IsDir() {
+		return fmt.Errorf("path %q is a file not a directory", dir)
+	}
+	return nil
+}
+
+func loadWebConfig(c webConfig) (static, theme http.Handler, templates *templates, err error) {
```
rithujohn191: Maybe a brief comment for this function would be helpful? Just clarifying the default behavior.

ericchiang: thanks, will do

amrutac:
👍

Issue #712, opened 2016-12-01
possible bug: handleConnectorCallback with sql storage fails -> id queried instead of state


evanluc:
I'm not sure if I've misunderstood the purpose of this function but in the case of the **handleConnectorCallback** [the state is checked for in the handler](https://github.com/coreos/dex/blob/9d9ad4a5b30038dc1dd530177361fced06042353/server/handlers.go#L261), [passed to getAuthRequest](https://github.com/coreos/dex/blob/9d9ad4a5b30038dc1dd530177361fced06042353/server/handlers.go#L267) but in the case of the sql storage option [state is never queried for](https://github.com/coreos/dex/blob/9d9ad4a5b30038dc1dd530177361fced06042353/storage/sql/crud.go#L183), just id, so this query always fails to retrieve a row as the state != req id.

If this is indeed the expected behaviour could I get an explanation as to why. It seems counter intuitive.

ericchiang:
We happen to use the ID of the auth request (a random string) as the state when passing clients onto upstream OAuth2 based identity providers. That auth request happens to also have an internal field called "state" that reflects the downstream state dex's client passed. These two aren't the same "state" so the code is correct, even if counter intuitive.

Is there a behavioral bug that you're seeing such as logins with OAuth2 providers not working? Or is this just lack of documentation in the code?

evanluc:
That clears things up. 

My problem was caused by trying to get my token at the route for that handler. It was fixed by redeeming the token at the right route `/token`. I'm using dex paired with python application using [oauthlib](https://oauthlib.readthedocs.io/en/latest/) to communicate between the dex server and my app instead of the golang client. 

Really enjoying this project, my suggestion is to maybe have clearer documentation for the `/auth` and `/token` routes and where in the authentication flow they they fit. It takes some digging in the source code for folks that are not using the golang client.

Eg.
`/auth` : To get the authentication code
`/token`: To redeem our token

I may be up for doing this over the weekend. Does this sound like a good fit for the getting-started doc?

ericchiang:
Ah, thanks for clarifying.

Most OpenID Connect clients get auth and token endpoints from discovery[0], but yeah I can understand OAuth2 needing dex to use that value and stick to it. Beyond documentation, part of this is writing tests or conveying that these shouldn't change too.

We can definitely do that. Thanks for clarifying.

[0] https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

PR #713, opened 2016-12-01
cmd/example-app: use a non-empty state

ericchiang:
Use a non-empty state in the example-app to ensure dex is properly
preserving the state for the code flow.

Updates #712

cc @rithujohn191


rithujohn191:
lgtm

PR #721, opened 2016-12-03
Fallback to RelayState form value if state not in url query

Calpicow:


ericchiang:
Which provider does this? This is a clear violation of the OAuth2 spec: https://tools.ietf.org/html/rfc6749#section-4.1.2

Calpicow:
This is for SAML compatibility. Comments mentioning it are mentioned a little above the code change.

ericchiang:
Ah right. Since we don't support SAML I don't think it's appropriate for us to add this. Plus whenever we do support SAML, we'll want to ensure that "RelayState" is only used with a SAML login and not a OAuth2 based login.

To support SAML, we need to send and be able to parse the correct XML payloads. Unless I'm mistaken on my end.

Calpicow:
Since this handler has no knowledge of which connector is using, what would be the best way to do this?

ericchiang:
We would add a SAML connector implementation: https://github.com/coreos/dex/tree/master/connector

Then we'd have to do some sort of type cast.

Calpicow:
I have a preliminary connector here: https://gist.github.com/Calpicow/e2e71b038c4ee53c5de5a3899e35aecf

Calpicow:
Ok, I see what you're saying now. I'll roll the changes when the connector is finalized.

Thanks.

Calpicow:
There appears to be a circular dependency for this. You need the state parameter to get the connectorID, but you need the connectorID to know how where to read the state. Thoughts?

ericchiang:
hell yeah, thanks!

ericchiang:
1. Read the state from both fields.
2. Get connector from connectorID.
3. Make sure the field matches the connector type.

PR #723, opened 2016-12-06
cmd/dex: add logging config and launch logger for server module.

rithujohn191:
Breaking down this PR into smaller modules so its easier to review.

This commit only includes cmd/dex changes which allow you to add logging config to the ConfigMap. This logger is then passed on to the server module with field "unit" set to "server".

TODO: pass logger on to other components (connectors & storage) with the "unit" field set to component name.

@ericchiang


cmd/dex/config.go
```diff
@@ -29,6 +29,7 @@ type Config struct {
 	OAuth2     OAuth2      `json:"oauth2"`
 	GRPC       GRPC        `json:"grpc"`
 	Expiry     Expiry      `json:"expiry"`
+	Logger     Logger      `json:"logger"`
```
ericchiang: Maybe we could call this section "debug"? That way we can at least add other fields here later. I feel like we're getting a lot of config sub-sections. What do you think?

foo.txt
```diff
@@ -0,0 +1 @@
+bar
```
ericchiang: please remove

cmd/dex/serve.go
```diff
@@ -145,6 +149,8 @@ func serve(cmd *cobra.Command, args []string) error {
 		s = storage.WithStaticPasswords(s, passwords)
 	}
 
+	serverLogger := logger.WithField("unit", "server")
```
ericchiang: don't know if unit is good since systemd already has a "unit". Maybe we just drop this for now?

cmd/dex/serve.go
```diff
@@ -111,6 +113,8 @@ func serve(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	logger := newLogger(c.Logger.Level, c.Logger.Format)
+
 	connectors := make([]server.Connector, len(c.Connectors))
```
ericchiang: We should probably pass these to the connectors too.

foo.txt
rithujohn191: Hmmm..this seems to have sneaked in during the github outage. Sorry

ericchiang:
@rithujohn191 fyi you need to vendor in logrus using glide

```
cmd/dex/serve.go:15:2: cannot find package "github.com/Sirupsen/logrus" in any of:
```

cmd/dex/serve.go
ericchiang: maybe we can annotate connectors for now. E.g. `logger.WithField("connector", "ldap")` since we'll only have one storage I don't think we'll need it

ericchiang: Also these things can change in the future.

rithujohn191:
Yes will submit a cleaner version of this. Just wanted to ensure that you are ok with the general design of it before I went further.

cmd/dex/config.go
rithujohn191: Are there other debug attributes you have in mind that you would like to make part of this struct?

ericchiang:
Thanks for updating the `vendor` stuff. Once you get back to a build able state I'll review. FYI you can run tests with

```
make testall
```

Or even just with `go test -v`

rithujohn191:
Brief summary of a discussion we had about the design of the logging structure in dex:

1. Users can customize the logging level(debug, info, error) & logging format(text or json) via the dex `ConfigMap` .
2. A newly created `logrus.Logger` will be passed on to the "server", "connectors" & "storage" modules. For the "connectors" & "storage" module the logger will be passed in as an argument to the 'Open' function, where as for the "server" module it will be embedded as a `Config` struct field.
3. Add info calls to log "success" state of dex.
4. Update current log calls to use logrus.

NOTE: In the future, we might have to create a new struct type that will hold the "logger" attribute along with other "non-configurable" attributes i.e Eventually split the `Config` struct into 2 parts:
- Configuration data passed on by the user.
- Other attributes required for the "open" function by different modules. 

@ericchiang hope I didn't miss out anything.

ericchiang:
@rithujohn191 notes look good. I'd only expand on

> Add info calls to log "success" state of dex.

We were discussing if we can contain logging to the server package. However, the server only gets good reports from the connectors and storage when there are errors. By definition, requests that went through, even ones that the user doesn't consider a "success" (e.g. a bad password), don't return an error message. Because of this, there's no really way of getting around passing loggers to connectors and storages if we want to provide debug info.

rithujohn191:
@ericchiang made changes to pass the "logger" object to connectors and storage. Updated the tests too. If you are okay with the change, I'll  go ahead and update the error messages to use logrus.

cmd/dex/config.go
ericchiang: Just in the future there might be.

ericchiang:
> If you are okay with the change

looks awesome.

ericchiang:
fyi you need to run `glide-vc` after you revendor. See https://github.com/coreos/dex/blob/master/Documentation/dev-dependencies.md

rithujohn191:
@ericchiang I actually did run glide-vc after I ran glide up -v

ericchiang:
odd, let me look into that.

rithujohn191:
NOTE: I had to modify the signature of func NewAPI() in server/api.go since there is no other way to pass the logger to it.

server/templates.go
```diff
@@ -234,15 +233,14 @@ func (w *writeRecorder) Write(p []byte) (n int, err error) {
 	return w.w.Write(p)
 }
 
-func renderTemplate(w http.ResponseWriter, tmpl *template.Template, data interface{}) {
+func renderTemplate(w http.ResponseWriter, tmpl *template.Template, data interface{}) error {
```
ericchiang: could we just put this on the server?

ericchiang:
looks good.

server/templates.go
ericchiang: yep, you're right. hmm that's a tricky one. let me think about it.

ericchiang: I'm going to recant and say that this is more correct.

ericchiang:
caught the last couple log statements

https://github.com/coreos/dex/blob/63bba20a059da8955603402fee126fdd9232bbdf/server/rotation.go#L157
https://github.com/coreos/dex/blob/63bba20a059da8955603402fee126fdd9232bbdf/server/rotation.go#L97

Maybe just throw a logger on the keyRotator?

Other than that lgtm

rithujohn191:
@ericchiang I figured those two log statements would change to logger.Info messages and be part of the info messages PR we discussed. That's why I didn't update them yet. Would you like me to address them as a part of this PR itself?

ericchiang:
Yeah. Sorry I thought you meant "add more logger.Info" messages as part of that conversation, not converting the existing one over.

rithujohn191:
@ericchiang  oh ok. Made the change.

cmd/dex/serve.go
```diff
@@ -182,27 +181,27 @@ func serve(cmd *cobra.Command, args []string) error {
...
 	if c.Web.HTTPS != "" {
-		log.Printf("listening (https) on %s", c.Web.HTTPS)
+		logger.Errorf("listening (https) on %s", c.Web.HTTPS)
 		go func() {
 			errc <- http.ListenAndServeTLS(c.Web.HTTPS, c.Web.TLSCert, c.Web.TLSKey, serv)
 		}()
 	}
 	if c.GRPC.Addr != "" {
-		log.Printf("listening (grpc) on %s", c.GRPC.Addr)
+		logger.Errorf("listening (grpc) on %s", c.GRPC.Addr)
```
rithujohn191: On it

PR #729, opened 2016-12-07
Add basic SAML support

Calpicow:
`server/handlers.go:302` looks questionable, but not sure what the proper way to do it is.

Please review.


Dockerfile
```diff
@@ -9,7 +9,8 @@ MAINTAINER Rithu John <rithu.john@coreos.com>
 # experience when this doesn't work out of the box.
 #
 # OpenSSL is required so wget can query HTTPS endpoints for health checking.
-RUN apk add --update ca-certificates openssl
+RUN apk add --update ca-certificates openssl xmlsec \
```
ericchiang: xmlsec? Does this use a cgo dependency?

Documentation/saml-connector.md
```diff
@@ -0,0 +1,35 @@
+# Authentication through SAML
+
+## Overview
+
+This connector adds basic support for receiving authentication data through SAML. It uses a fork of `github.com/RobotsAndPencils/go-saml` hosted at `github.com/Calpicow/go-saml`.
```
ericchiang: Can you articulate a little bit of why you need a fork?

server/handlers.go
```diff
@@ -280,12 +290,26 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
...
+
 	callbackConnector, ok := conn.Connector.(connector.CallbackConnector)
 	if !ok {
 		s.notFound(w, r)
 		return
 	}
 
+	// If using SAML connector, oauthState should be empty
+	// Else, relayState should be empty
+	isSaml := reflect.TypeOf(callbackConnector).String() == "*saml.samlConnector"
```
ericchiang: blah. Yeah we should probably add another method to the callback connector interface. For example

```
// State returns the response field with the OAuth2 or SAML state. Allowed values are "state" or "RelayState".
func State() string
```

Then use this here.

Documentation/saml-connector.md
```diff
@@ -0,0 +1,35 @@
...
+  name: saml
+  config:
+    spIssuerUrl: $SP_ISSUER_URL
+    spSignRequest: false
+    # Typically urn:oasis:names:tc:SAML:2.0:nameid-format:transient or below
+    spNameIDPolicyFormat: urn:oasis:names:tc:SAML:2.0:nameid-format:persistent
+    # Set to empty string to omit RequestedAuthnContext block
+    spRequestedAuthnContext: urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport
+    idpSSOURL: $IDP_SSO_URL
+    idpPublicCertPath: /etc/dex/saml.crt
```
ericchiang: "idp" and "sp" are probably redundant here. You can probably just use `issuerURL` and `ca`.

Documentation/saml-connector.md
```diff
@@ -0,0 +1,35 @@
...
+- type: saml
+  id: saml
+  name: saml
+  config:
+    spIssuerUrl: $SP_ISSUER_URL
+    spSignRequest: false
+    # Typically urn:oasis:names:tc:SAML:2.0:nameid-format:transient or below
+    spNameIDPolicyFormat: urn:oasis:names:tc:SAML:2.0:nameid-format:persistent
+    # Set to empty string to omit RequestedAuthnContext block
+    spRequestedAuthnContext: urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport
```
ericchiang: What are the expected values of this field and the one above it? When would one change these values?

Documentation/saml-connector.md
```diff
@@ -0,0 +1,35 @@
...
+    idpSSOURL: $IDP_SSO_URL
+    idpPublicCertPath: /etc/dex/saml.crt
+    idAttr: uid
+    userAttr: cn
+    emailAttr: mail
+    groupsAttr: groups
+    trustedEmailProvider: true
```
ericchiang: Can we drop this for now? The SAML provider should indicate this or dex should assume that all emails are valid.

connector/saml/saml.go
```diff
@@ -0,0 +1,171 @@
...
+	identity = connector.Identity{
+		UserID:        userID,
+		Username:      username,
+		Email:         email,
+		EmailVerified: c.trustedEmailProvider,
+	}
+
+	if s.Groups {
+		groups := response.GetAttribute(c.groupsAttr)
+		identity.Groups = strings.Split(groups, ",")
```
ericchiang: What's an example of a provider that returns groups? Why would we be able to assume that they'd use comma separated values?

Dockerfile
Calpicow: The library uses external calls. See
https://github.com/Calpicow/go-saml/blob/master/xmlsec.go#L63
https://github.com/Calpicow/go-saml/blob/master/xmlsec.go#L113

Documentation/saml-connector.md
Calpicow: I've made a number of improvements to the existing code to support different SAML implementations and to perform sanity checks for security. All changes have a pending pull request. The maintainer is rather slow/reluctant to merge them, and I'd like to see progress now rather than later.

Documentation/saml-connector.md
Calpicow: Service providers could also have their own public/private keys as well. But will remove redundancy of other attributes.

connector/saml/saml.go
Calpicow: I can see this behavior with our SSO implementation (Oracle Access Manager), but can't confirm that this behavior is consistent among others.

Dockerfile
ericchiang: No dependency we have does this and it's generally a red flag when evaluating Go packages. I'll continue looking into go-saml, but generally we require a pure Go implementation.

Since most SAML login flows only use a few XML structs, I'd think we'd prefer just having these structs in dex and doing the XML decoding/signature validation ourselves. dex doesn't need to implement all of SAML

Calpicow:
A little progress on this:

Response verification has been ported to pure Go by using `github.com/ma314smith/signedxml`. Signing of requests is still WIP.

After this I think we could start evaluating whether we want to continue using third-party libraries or rewrite/bring into the dex codebase.

I'm not too sure about testing strategy.

ericchiang:
I'm also looking into this internally. I may have some code for you to look at too.

Calpicow:
Dependencies have been updated to perform both signing and verification in pure Go.

ericchiang:
Let me look through this dependency list. I was working on a version of this connector using https://github.com/russellhaering/goxmldsig since I know the author. I'm always very hesitant to import crypto libraries that aren't well established. Signature validation on the JOSE end has bitten us soooo often.

Also as a general note, if we're forking the SAML library we might just want to internalize it in dex.

Calpicow:
Understandable. We could always swap out the code as necessary.

Calpicow:
I've had a brief look at his SAML library as well, and it looks like I may be able to switch it out easily. It will need a bit of work to add support for our IDP, so let's see if he's open to pull requests.

Issue #745, opened 2016-12-15
storage/sql: postgres doesn't perist timezone


jwmaag:
Using the latest build and fresh postgres database as the storage backend, I was unable to login. 

```
Authorization request period has expired.
```

Digging a bit deeper I noticed that the table was using the __timestamp__ column type (w/o timezone). So it seemed to me that dex was losing some precision between the insert and the select. 

Also telling was the frequent key rotations
```
2016/12/13 14:43:36 keys expired, rotating
2016/12/13 14:43:36 keys rotated, next rotation: 2016-12-13 20:43:36.554887954 -0600 CST
2016/12/13 14:43:36 listening (http) on 0.0.0.0:5556
2016/12/13 14:44:06 keys expired, rotating
2016/12/13 14:44:06 keys rotated, next rotation: 2016-12-13 20:44:06.965425402 -0600 CST
2016/12/13 14:44:36 keys expired, rotating
2016/12/13 14:44:37 keys rotated, next rotation: 2016-12-13 20:44:37.116072981 -0600 CST
2016/12/13 14:45:07 keys expired, rotating
2016/12/13 14:45:07 keys rotated, next rotation: 2016-12-13 20:45:07.321215505 -0600 CST
2016/12/13 14:45:37 keys expired, rotating
2016/12/13 14:45:37 keys rotated, next rotation: 2016-12-13 20:45:37.567846803 -0600 CST
2016/12/13 14:46:07 keys expired, rotating
2016/12/13 14:46:07 keys rotated, next rotation: 2016-12-13 20:46:07.875356915 -0600 CST
2016/12/13 14:46:37 keys expired, rotating
2016/12/13 14:46:38 keys rotated, next rotation: 2016-12-13 20:46:38.063993965 -0600 CST
2016/12/13 14:47:08 keys expired, rotating
2016/12/13 14:47:08 keys rotated, next rotation: 2016-12-13 20:47:08.349067898 -0600 CST
2016/12/13 14:47:38 keys expired, rotating
2016/12/13 14:47:38 keys rotated, next rotation: 2016-12-13 20:47:38.889484596 -0600 CST
2016/12/13 14:48:08 keys expired, rotating
2016/12/13 14:48:09 keys rotated, next rotation: 2016-12-13 20:48:09.681262225 -0600 CST
```

Not sure if i missed something, but my resolution was to wrap the default server.now function with one that returns UTC.

ericchiang:
[...]

PR #746, opened 2016-12-15
server: wrap the default server.now function to always return UTC

jwmaag:
This change was my resolution to the issue described in #745

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -174,7 +174,7 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
 
 	now := c.Now
 	if now == nil {
-		now = time.Now
+		now = func() time.Time { return time.Now().UTC() }
 	}
 
 	s := &Server{


ericchiang:
This is a small fix but isn't the real solution to store the timestamps in postgres?

ericchiang:
Sorry, I mean storing timezones.

jwmaag:
Right, You could migrate the database to timestamptz, but it seemed simpler to just take timezone out of the equation all together.

does the sqlite backend take the timestamp column as is?

ericchiang:
> Right, You could migrate the database to timestamptz, but it seemed simpler to just take timezone out of the equation all together.

It'd be quicker to switch everything over to UTC, but I'd rather the server not be dependent on us never using timezones. E.g. it's weird for logs to spit out timestamps in two different timezones (one for the logging metadata, another for the printed time). I think it should be a quick enough fix.

In terms of migrations I don't even know if this one is safe. If the server's been storing timestamps in one format then suddenly switches a few hours ahead or behind that's going to have some really bad consequences I think the fix for this might have to be breaking.  

> does the sqlite backend take the timestamp column as is?

Yes, we have to investigate if that's an issue.

jwmaag:
Agree that shoehorning something globally to fix a storage backend issue is probably not the correct solution. Closing.

PR #749, opened 2016-12-16
storage: fix postgres timezone handling

### LINKED ISSUE #745: storage/sql: postgres doesn't perist timezone

ericchiang:
Dex's Postgres client currently uses the `timestamp` datatype for
storing times. This lops of timezones with no conversion, causing
times to lose locality information.

We could convert all times to UTC before storing them, but this is
a backward incompatible change for upgrades, since the new version
of dex would still be reading times from the database with no
locality.

Because of this intrinsic issue that current Postgres users don't
save any timezone data, we chose to treat any existing installation
as corrupted and change the datatype used for times to `timestamptz`.
This is a breaking change, but it seems hard to offer an
alternative that's both correct and backward compatible.

Additionally, an internal flag has been added to SQL flavors,
`supportsTimezones`. This allows us to handle SQLite3, which doesn't
support timezones, while still storing timezones in other flavors.
Flavors that don't support timezones are explicitly converted to
UTC.

Closes #745

cc @jwmaag @rithujohn191

### DIFF
--- a/storage/sql/sql.go
+++ b/storage/sql/sql.go
@@ -4,6 +4,7 @@ package sql
 import (
 	"database/sql"
 	"regexp"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/cockroachdb/cockroach-go/crdb"
@@ -28,6 +29,9 @@ type flavor struct {
 	//
 	// See: https://github.com/cockroachdb/docs/blob/63761c2e/_includes/app/txn-sample.go#L41-L44
 	executeTx func(db *sql.DB, fn func(*sql.Tx) error) error
+
+	// Does the flavor support timezones?
+	supportsTimezones bool
 }
 
 // A regexp with a replacement string.
@@ -69,6 +73,8 @@ var (
 			}
 			return tx.Commit()
 		},
+
+		supportsTimezones: true,
 	}
 
 	flavorSQLite3 = flavor{
@@ -80,7 +86,7 @@ var (
 			{matchLiteral("boolean"), "integer"},
 			// Translate other types.
 			{matchLiteral("bytea"), "blob"},
-			// {matchLiteral("timestamp"), "integer"},
+			{matchLiteral("timestamptz"), "timestamp"},
 			// SQLite doesn't have a "now()" method, replace with "date('now')"
 			{regexp.MustCompile(`\bnow\(\)`), "date('now')"},
 		},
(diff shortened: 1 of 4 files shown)


ericchiang:
Still need to test the Kubernetes storage.

ericchiang:
Kubernetes storage passes the new conformance tests.

storage/sql/sql.go
```diff
@@ -107,6 +113,19 @@ func (f flavor) translate(query string) string {
 	return query
 }
 
+func (c *conn) translateArgs(args []interface{}) []interface{} {
```
rithujohn191: Could you please add a comment about what translateArgs does

PR #750, opened 2016-12-16
server: use UTC timestamps in dex.

rithujohn191:
Updated server now() func and logger to use UTC timestamps.

Log messages now use UTC:
time="2016-12-16T22:44:47Z" level=info msg="config issuer: http://127.0.0.1:5556/dex" 
time="2016-12-16T22:44:47Z" level=info msg="config connector: mock" 
time="2016-12-16T22:44:47Z" level=info msg="config connector: local passwords enabled" 
time="2016-12-16T22:44:47Z" level=info msg="config storage: sqlite3" 
time="2016-12-16T22:44:47Z" level=info msg="config static client: example-app" 
time="2016-12-16T22:44:47Z" level=info msg="listening (http) on 0.0.0.0:5556" 

@ericchiang

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -180,6 +180,9 @@ func serve(cmd *cobra.Command, args []string) error {
 		logger.Infof("config skipping approval screen")
 	}
 
+	// explicitly convert to UTC.
+	now := func() time.Time { return time.Now().UTC() }
+
 	serverConfig := server.Config{
 		SupportedResponseTypes: c.OAuth2.ResponseTypes,
 		SkipApprovalScreen:     c.OAuth2.SkipApprovalScreen,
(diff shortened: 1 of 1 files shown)


cmd/dex/serve.go
```diff
@@ -251,6 +255,15 @@ var (
 	logFormats = []string{"json", "text"}
 )
 
+type customFormatter struct {
```
ericchiang: utcFormatter?

server/server_test.go
```diff
@@ -159,7 +159,7 @@ func TestOAuth2CodeFlow(t *testing.T) {
 	clientSecret := "testclientsecret"
 	requestedScopes := []string{oidc.ScopeOpenID, "email", "profile", "groups", "offline_access"}
 
-	t0 := time.Now()
+	t0 := time.Now().UTC()
```
ericchiang: Maybe we should change our comparison rather than our initialization?

rithujohn191: This value gets passed on to the now() func which gets used in newKeyCacher(). So this flow would mimic the actual behavior of the server now() function.

cmd/dex/serve.go
rithujohn191: done

server/server_test.go
rithujohn191: Discussed offline. Removing the UTC to increase test coverage to include cases when time zones are involved.

ericchiang:
lgtm

PR #751, opened 2016-12-20
replace bcrypt hash with a working one

h0me:
replace bcrypt hash with an appropriate value for the string "password"


ericchiang:
lgtm, thanks!

Issue #112, opened 2015-09-04
connectors: add a SAML


philhug:
Being able to proxy requests to a SAML IdP is useful sometimes.
I already have a semi-working implementation and will provide a pull request as soon as it's cleaned up.

bobbyrullo:
that's awesome phil - I can't wait to see it.

philhug:
I'm currently using https://github.com/RobotsAndPencils/go-saml as the SAML libary.
it has some disadvantages as it executes the xmlsec1 binary to verify SAML signatures, but I haven't seen a native xmlsec go library yet.

philhug:
https://github.com/philhug/dex/tree/saml

nphase:
This would be awesome!

philips:
@philhug how is this going?

russellhaering:
@philhug At ScaleFT we recently encountered a need for pure Go SAML implementation that doesn't exec `xmlsec1`.

We ended up writing this: https://github.com/russellhaering/gosaml2

I'd love to get more eyes on it, if there is anything I can do to facilitate its use in dex let me know.

philips:
@philhug What do you think of working on Dex + SAML using the new pure Go libraries that @russellhaering wrote?

andrewstuart:
@philips I've actually been doing a bit of work recently integrating some of my own saml2 work (for encrypted assertions) with @russellhaering's library, since they ended up being pretty orthogonal. One of my short-term goals is to submit a PR for a saml2 dex connector, since it seems like some of the previous work might have dropped off.

ericchiang:
@andrewstuart would love to see this and would be happy to review! Please feel free to open a PR or comment here if you have any further questions.

pgchamberlin:
Hey @andrewstuart, I'm just wondering are you still looking at this? My team would be really interested in a SAML2 connector and may be able to contribute some time and energy. Are you still looking at this?

andrewstuart:
TBH I haven't looked at it in quite a while. I've contributed a few times to the saml2 library and it definitely worked for validating and decrypting identity assertions last I checked, as well as making signed AuthN requests.

Unfortunately, my main drive for working on it was personal and once the prospect of getting to use it at work dried up, I didn't have quite as much motivation to keep working on it, though I'm definitely still interested in improving the SAML2 library, and helping with a dex integration effort.

pgchamberlin:
[...]

PR #752, opened 2016-12-21
Add an experimental SAML connector

ericchiang:
cc @Calpicow please take a look

cc @rithujohn191 

updates #112


ericchiang:
Needs a ton of inline comments and docs. Am testing with Okta.

Some notable points:
* Doesn't support refresh tokens. Because we can't follow up with a provider we can't update the claims.
* Gets ID from `NameID` element of the assertion's subject. This is probably not always correct.
* Okta returns groups by returning multiple attribute values, not through a single string.
* Uses the POST binding rather than the redirect binding.
* Uses github.com/russellhaering/goxmldsig for XML signature validation.
* Does not support XML signed authnrequest.
* Does not support encrypted attributes.
* Probably needs a ton more configuration.
* Has type of `"samlExperimental"` instead of just `"saml"` encase we need to make backward incompatible changes, or if we're verifying signatures wrong.

connector/saml/saml.go
```diff
@@ -0,0 +1,300 @@
...
+		ProtocolBinding: bindingPOST,
+		ID:              "_" + uuidv4(),
+		IssueInstant:    xmlTime(p.now()),
+		Destination:     p.ssoURL,
+		Issuer: &issuer{
+			Issuer: p.issuer,
+		},
+		NameIDPolicy: &nameIDPolicy{
+			AllowCreate: true,
+			Format:      "urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified",
```
Calpicow: I'd like to see NameIDPolicy configurable.

connector/saml/saml.go
```diff
@@ -0,0 +1,300 @@
...
+		return ident, fmt.Errorf("no attribute with name %q", p.emailAttr)
+	}
+	ident.EmailVerified = true
+
+	if ident.Username = attributes.get(p.usernameAttr); ident.Username == "" {
+		return ident, fmt.Errorf("no attribute with name %q", p.usernameAttr)
+	}
+
+	if s.Groups && p.groupsAttr != "" {
+		groups, ok := attributes.all(p.groupsAttr)
```
Calpicow: Maybe add a groupsDelimiter config for IDPs that return groups as comma-separated

connector/saml/saml.go
```diff
@@ -0,0 +1,300 @@
...
+		IssueInstant:    xmlTime(p.now()),
+		Destination:     p.ssoURL,
+		Issuer: &issuer{
+			Issuer: p.issuer,
+		},
+		NameIDPolicy: &nameIDPolicy{
+			AllowCreate: true,
+			Format:      "urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified",
+		},
+		AuthnContextClass: &authnContextClass{
```
Calpicow: My IDP doesn't seem to like it when this element is present. Could we add some kind of mechanism so it could be omitted?

connector/saml/saml.go
ericchiang: Just the format or do you need things like `AllowCreate`?

connector/saml/saml.go
ericchiang: Yeah totally.

connector/saml/saml.go
Calpicow: Just the format

connector/saml/saml.go
ericchiang: Maybe it just want's a different value? I'm actually a little unclear about its exact use. Might be safe to omit by default?

Calpicow: It looks like the correct value from looking at my SAML response (https://gist.github.com/Calpicow/fd1d03ded8dc9da1570155bc14aa689a). I'm fine with omit by default, with a config to set the value.

connector/saml/saml.go
```diff
@@ -0,0 +1,300 @@
...
+
+	// NOTE(ericchiang): If we can't follow up with the identity provider, can we
+	// support refresh tokens?
+	if s.OfflineAccess {
+		return "", "", fmt.Errorf("SAML does not support offline access")
+	}
+
+	r := &authnRequest{
+		ProtocolBinding: bindingPOST,
+		ID:              "_" + uuidv4(),
```
ericchiang: Don't know if we should possibly be matching this on the response.

Calpicow:
There are a few holes with `goxmldsig` when validating SAML responses, but looks good once the issues are resolved.

ericchiang:
ref for goxmldsig issue: https://github.com/russellhaering/goxmldsig/issues/16

connector/saml/saml.go
ericchiang: I think it's an invalid element. e.g. this creates

```xml
<AuthnRequest>
    <AuthnContextClassRef>
    </AuthnContextClassRef>
</AuthnRequest>
```

but the spec wants

```xml
<AuthnRequest>
    <RequestedAuthnContext>
        <AuthnContextClassRef>
        </AuthnContextClassRef>
    </RequestedAuthnContext>
</AuthnRequest>
```

let me fix that.

connector/saml/saml.go
ericchiang: added a groupsDelim parameter.

connector/saml/saml.go
ericchiang: This is now customizable.

ericchiang:
TODOs as I see it:

* Documentation on how to use this (before merge).
* Documentation on work to be done to drop "experimental" tag (before merge).
* General security review of the HTTP POST binding (after merge): https://docs.oasis-open.org/security/saml/v2.0/saml-sec-consider-2.0-os.pdf
* XML signature validation issues (after merge?)

gebi:
If you ever need to make it work with ADFS (microsoft) SAML. We needed that feature a while back, it's not generic and for the "old" go saml library but it shows what's needed to make it run, hope it helps.

https://github.com/mgit-at/go-saml/tree/adfs

PR #754, opened 2016-12-22
build fails

szuecs:
I followed https://github.com/coreos/dex/blob/master/Documentation/getting-started.md
and make fails, because scripts/check-go-version references the wron shell in it's shebang.
sh has no arrays


ericchiang:
lgtm thanks!

For a bit of background, we reference `/bin/sh` so these scripts run in our more minimal containers. Don't know why this every worked on my machine.

PR #756, opened 2016-12-22
*: add 'make revendor' and tests to catch incorrect glide usage

ericchiang:
Introducing glide-vc caused us to unknowingly removed our Go protobuf
compiler (since it's a main). Add flags to glide-vc usage to remedy
this.

Since we now require several glide and glide-vc flags, add a Makfile
target and tests to catch when PRs don't use the correct flags.

cc @rithujohn191


Makefile
```diff
@@ -35,6 +35,11 @@ bin/example-app: check-go-version
 release-binary:
 	@go build -o _output/bin/dex -v -ldflags $(LD_FLAGS) $(REPO_PATH)/cmd/dex
 
+.PHONY: revendor
+revendor:
```
sym3tri: Might be worth checking the version of glide used too, if you're trying to catch mistakes.

ericchiang: This is kinda already addressed in our documentation.

> Dex uses glide and glide-vc to manage its vendor directory. A recent version of these are preferred but dex doesn't require any bleeding edge features. Either install these tools using go get or take an opportunity to update to a more recent version.

If we ever need something that was added in a specific version of glide we can add a script similar to `scripts/check-go-version`, but I don't really want to write a bunch of `sed` commands right now :/

glide_test.go
```diff
@@ -96,13 +97,23 @@ func TestGlideYAMLPinsAllDependencies(t *testing.T) {
...
 	err := filepath.Walk("vendor", func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			t.Fatalf("walk: stat path %s failed: %v", path, err)
 		}
 		if info.IsDir() && filepath.Base(path) == ".git" {
-			t.Fatalf(".git directory detected in vendor: %s. Revendor packages and remove version control data with 'glide update -s -v -u'", path)
+			t.Fatalf(".git directory detected in vendor: %s. Revendor packages using 'make revendor' to use the correct glide and glide-vc flags", path)
+		}
+		if !info.IsDir() && strings.HasSuffix(path, "_test.go") {
+			t.Fatalf("'_test.go' file detected in vendor: %s. Revendor packages using 'make revendor' to use the correct glide and glide-vc flags", path)
```
rithujohn191: Is this so we avoid vendors from introducing their own tests into our code?

ericchiang: Yep. glide-vc has a mode that deletes all test files from the `vendor` directory. We don't run those tests anyway, so there's no good reason to hold onto them.

rithujohn191:
lgtm

PR #757, opened 2016-12-23
*: update refresh tokens instead of deleting and creating another

ericchiang:
The server implements a strategy called "Refresh Token Rotation" to
ensure refresh tokens can only be claimed once.

ref: https://tools.ietf.org/html/rfc6819#section-5.2.2.3

Previously "refresh_token" values in token responses where just the
ID of the internal refresh object. To implement rotation, when a
client redeemed a refresh token, the object would be deleted, a new
one created, and the new ID returned as the new "refresh_token".

However, this means there was no consistent ID for refresh tokens
internally, making things like foreign keys very hard to implement.
This is problematic for revocation features like showing all the
refresh tokens a user or client has out.

This PR updates the "refresh_token" to be an encoded protobuf
message, which holds the internal ID and a nonce. When a refresh
token is used, the nonce is updated to prevent reuse, but the ID
remains the same. Additionally it adds the timestamp of each
token's last use.

cc @rithujohn191


ericchiang:
rebase for good measure.

ericchiang:
@rithujohn191 please take a look

cmd/example-app/main.go
```diff
@@ -273,6 +268,10 @@ func (a *app) handleCallback(w http.ResponseWriter, r *http.Request) {
 	oauth2Config := a.oauth2Config(nil)
 	switch {
 	case code != "":
+		if state := r.FormValue("state"); state != exampleAppState {
```
rithujohn191: Just wondering if there is a reason why we should not perform this check for "state" in case of 'refresh'

ericchiang: "state" is only part of a auth code response[0]. it's not part of a refresh event.

I think it's confusing because we're handling both the auth request redirect response (GET) and a frontend post to re-render the page (POST). Maybe we should switch this so GETs and POSTs are what determine the course of action? e.g.

```
switch r.Method {
case "GET":
    // ... handle state and do code exchange
case "POST":
    // ... try to refresh token
}
```

[0] https://tools.ietf.org/html/rfc6749#section-4.1.2

rithujohn191: Yes that would definitely make the flow more clear

ericchiang:
@rithujohn191 Updated the example-app.

rithujohn191:
Might need a rebase?

PR #758, opened 2016-12-23
connector/ldap: don't query for groups if no base DN is provided

ericchiang:


connector/ldap/ldap.go
```diff
@@ -445,6 +445,11 @@ func (c *ldapConnector) Refresh(ctx context.Context, s connector.Scopes, ident c
 }
 
 func (c *ldapConnector) groups(ctx context.Context, user ldap.Entry) ([]string, error) {
+	if c.GroupSearch.BaseDN == "" {
+		c.logger.Infof("groups requested but no baseDN provided")
```
philips: What would a support person or user do with this information and how they would find information to rectify if this is not their desired configuration? Perhaps:

"No groups returned for ${username} because no groups baseDN has been configured"

Plus, how often will this log line be printed out? On every login?

ericchiang:
You're right, we should probably bump the level to debug and print the
user's email. @rithujohn191 want to open a PR  with that? (Sonce I'm out)

rithujohn191:
On it

PR #762, opened 2016-12-29
Fixed refresh tokens to work correctly with kubernetes storage

moosilauke18:
It was missing connectorData for RefreshToken under Kubernetes storage. Fixed this to allow refresh to work correctly without 500 server error.


rithujohn191:
Thank you for your fix. Could kindly also update the unit test at https://github.com/coreos/dex/blob/master/storage/conformance/conformance.go#L208 and add a connectorData field.

ericchiang:
bump @moosilauke18

PR #766, opened 2017-01-09
server: fixes for the implicit and hybrid flow

### LINKED ISSUE #761: Response type for implicit flow
Hello,
I'm trying to use a js openid connect client (https://github.com/IdentityModel/oidc-client-js) and I'm facing issue with the response types for the implicit flow.
According to the oidc spec (http://openid.net/specs/openid-connect-core-1_0.html#Authentication) the response types that should trigger the implicit flow are:
- id_token
- id_token token

It seems to me that dex only supports the "token" response type in order to trigger this implicit flow.
I may be misreading both oidc specs and dex code, thus please correct me if I'm wrong.
Otherwise I will try to fix this behavior in order to allow the 2 above mentioned response_type values

Thanks,
Simon

ericchiang:
Accept the following response_type for the implicit flow:

    id_token
    token id_token

And the following for hybrid flow

    code id_token
    code token
    code token id_token

This corrects the previous behavior of the implicit flow, which
only accepted "token" (now correctly rejected).

cc @rithujohn191 @xeonx 

@xeonx can you see if this branch works with your client?

closes #761

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -144,12 +144,23 @@ func (s *Server) discoveryHandler() (http.Handler, error) {
 
 // handleAuthorization handles the OAuth2 auth endpoint.
 func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
-	authReq, err := s.parseAuthorizationRequest(s.supportedResponseTypes, r)
+	authReq, err := s.parseAuthorizationRequest(r)
 	if err != nil {
 		s.logger.Errorf("Failed to parse authorization request: %v", err)
-		s.renderError(w, http.StatusInternalServerError, "Failed to connect to the database.")
+		if handler, ok := err.Handle(); ok {
+			// client_id and redirect_uri checked out and we can redirect back to
+			// the client with the error.
+			handler.ServeHTTP(w, r)
+			return
+		}
+
+		// Otherwise render the error to the user.
+		//
+		// TODO(ericchiang): Should we just always render the error?
+		s.renderError(w, err.Status(), err.Error())
 		return
 	}
+
 	authReq.Expiry = s.now().Add(time.Minute * 30)
 	if err := s.storage.CreateAuthRequest(authReq); err != nil {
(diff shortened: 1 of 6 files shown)


xeonx:
Thanks for this PR.
I was able to follow the flow up to token retrieval by client, thus it is a great improvement.
I'm now facing new issues regarding id_token validation on client side:
- I had to allow CORS on /keys endpoint (I will send a PR after cleaning the code, certainly next weekend)
- at_hash is required in id_token by the oidc-client-js client even if optional according to oidc spec

ericchiang:
> I had to allow CORS on /keys endpoint (I will send a PR after cleaning the code, certainly next weekend)

Why don't we make the current CORS option encompass this.

> at_hash is required in id_token by the oidc-client-js client even if optional according to oidc spec

opened https://github.com/coreos/dex/issues/771

ericchiang:
going to add a few inline comments then merge

ericchiang:
does this client use the userinfo_endpoint? we'll have to implement that too (which won't be impossible) https://github.com/coreos/dex/issues/376

xeonx:
Yes userinfo_endpoint is the following call in the flow but it is optional, configurable with a flag (resulting in using claims from the id_token.profile only), thus having it implemented is not mandatory for this client.

server/handlers.go
```diff
@@ -144,12 +144,23 @@ func (s *Server) discoveryHandler() (http.Handler, error) {
...
+			// client_id and redirect_uri checked out and we can redirect back to
+			// the client with the error.
+			handler.ServeHTTP(w, r)
+			return
+		}
+
+		// Otherwise render the error to the user.
+		//
+		// TODO(ericchiang): Should we just always render the error?
+		s.renderError(w, err.Status(), err.Error())
```
ericchiang: Part of the oauth2 spec is returning errors to the client https://tools.ietf.org/html/rfc6749#section-4.1.2.1

PR #772, opened 2017-01-11
server: add at_hash claim support

### LINKED ISSUE #771: server: include at_hash value in ID Token

ericchiang:
The "at_hash" claim, which provides hash verification for the
"access_token," is a required claim for implicit and hybrid flow
requests. Previously we did not include it (against spec). This
PR implements the "at_hash" logic and adds the claim to all
responses.

As a cleanup, it also moves some JOSE signing logic out of the
storage package and into the server package.

For details see:

https://openid.net/specs/openid-connect-core-1_0.html#ImplicitIDToken

closes #771

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -467,6 +467,8 @@ func (s *Server) sendCodeResponse(w http.ResponseWriter, r *http.Request, authRe
 		// Only valid for implicit and hybrid flows.
 		idToken       string
 		idTokenExpiry time.Time
+
+		accessToken = storage.NewID()
 	)
 
 	for _, responseType := range authReq.ResponseTypes {
@@ -502,7 +504,7 @@ func (s *Server) sendCodeResponse(w http.ResponseWriter, r *http.Request, authRe
 		case responseTypeIDToken:
 			implicitOrHybrid = true
 			var err error
-			idToken, idTokenExpiry, err = s.newIDToken(authReq.ClientID, authReq.Claims, authReq.Scopes, authReq.Nonce)
+			idToken, idTokenExpiry, err = s.newIDToken(authReq.ClientID, authReq.Claims, authReq.Scopes, authReq.Nonce, accessToken)
 			if err != nil {
 				s.logger.Errorf("failed to create ID token: %v", err)
 				s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
@@ -513,7 +515,7 @@ func (s *Server) sendCodeResponse(w http.ResponseWriter, r *http.Request, authRe
 
 	if implicitOrHybrid {
 		v := url.Values{}
-		v.Set("access_token", storage.NewID())
+		v.Set("access_token", accessToken)
 		v.Set("token_type", "bearer")
 		v.Set("state", authReq.State)
 		if idToken != "" {
@@ -623,7 +625,8 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 		return
 	}
 
-	idToken, expiry, err := s.newIDToken(client.ID, authCode.Claims, authCode.Scopes, authCode.Nonce)
+	accessToken := storage.NewID()
+	idToken, expiry, err := s.newIDToken(client.ID, authCode.Claims, authCode.Scopes, authCode.Nonce, accessToken)
 	if err != nil {
 		s.logger.Errorf("failed to create ID token: %v", err)
 		s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
@@ -674,7 +677,7 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 			return
 		}
 	}
-	s.writeAccessToken(w, idToken, refreshToken, expiry)
+	s.writeAccessToken(w, idToken, accessToken, refreshToken, expiry)
 }
 
 // handle a refresh token request https://tools.ietf.org/html/rfc6749#section-6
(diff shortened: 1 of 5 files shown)


ericchiang:
@xeonx I didn't add a very thorough test because I don't have a client that implements this logic. Can you test this with your client?

ericchiang:
cc @rithujohn191

server/server_test.go
```diff
@@ -222,6 +222,31 @@ func TestOAuth2CodeFlow(t *testing.T) {
...
+					return fmt.Errorf("failed to verify id token: %v", err)
+				}
+
+				var claims struct {
+					AtHash string `json:"at_hash"`
+				}
+				if err := idToken.Claims(&claims); err != nil {
+					return fmt.Errorf("failed to decode raw claims: %v", err)
+				}
+				if claims.AtHash == "" {
```
rithujohn191: The at_hash is used to verify that the ID Token <-> Access Token pair belongs together right? so will we have to modify the go-oidc Verify() function to check if received at_hash == expected at_hash?

server/oauth2.go
```diff
@@ -125,6 +135,92 @@ func parseScopes(scopes []string) connector.Scopes {
 	return s
 }
 
+func signatureAlgorithm(jwk *jose.JSONWebKey) (alg jose.SignatureAlgorithm, err error) {
```
rithujohn191: Would be helpful to write up a brief comment describing the following two functions.

server/server_test.go
ericchiang: at_hash is optional and there are lots of times you'd verify the id_token without having the access token. Don't know what the best way to introduce this to go-oidc is but I've opened https://github.com/coreos/go-oidc/issues/126 for good measure.

server/oauth2.go
ericchiang: thanks, will do.

xeonx:
Quick test result: at_hash was received but validation failed. I do not have much more info, sorry.

ericchiang:
@xeonx thanks that's useful.

xeonx:
I'm tried to find the issue in the accessTokenHash method using some examples. The following algorithm seems to work (to be adapted with the various possibilities of algorithms):

func accessTokenHash(accessToken string) string {

	hash := sha256.Sum256([]byte(accessToken))
	halfhash := hash[:len(hash)/2]
	return base64.RawURLEncoding.EncodeToString(halfhash)
}

server/oauth2.go
```diff
@@ -125,6 +135,92 @@ func parseScopes(scopes []string) connector.Scopes {
...
+		return "", err
+	}
+	newHash, ok := hashForSigAlg[alg]
+	if !ok {
+		return "", fmt.Errorf("unsupported signature algorithm: %s", alg)
+	}
+
+	hash := newHash().Sum([]byte(accessToken))
+	if len(hash) > 128 {
+		hash = hash[:128]
```
ericchiang: bug: this is bytes, I need bits

ericchiang:
@xeonx updated the at_hash computation (don't have time to rebase right now).

server/handlers.go
```diff
@@ -465,6 +465,8 @@ func (s *Server) sendCodeResponse(w http.ResponseWriter, r *http.Request, authRe
 		// Only valid for implicit and hybrid flows.
 		idToken       string
 		idTokenExpiry time.Time
+
+		accessToken = storage.NewID()
```
xeonx: This accessToken should be reused at line 516:

    v.Set("access_token", accessToken)

instead of

    v.Set("access_token", storage.NewID())

server/oauth2.go
```diff
@@ -125,6 +135,92 @@ func parseScopes(scopes []string) connector.Scopes {
...
+	alg, err := signatureAlgorithm(k.SigningKey)
+	if err != nil {
+		return "", err
+	}
+	newHash, ok := hashForSigAlg[alg]
+	if !ok {
+		return "", fmt.Errorf("unsupported signature algorithm: %s", alg)
+	}
+
+	hash := newHash().Sum([]byte(accessToken))
```
xeonx: The Sum method does not compute the sum of the parameter but append the current internal sum to the parameter. A correct construct will be:

    hasher := sha256.New()
    io.WriteString(hasher, accessToken)
    hash := hasher.Sum(nil)

server/oauth2.go
```diff
@@ -125,6 +135,92 @@ func parseScopes(scopes []string) connector.Scopes {
...
+	if len(hash) > 16 {
```
xeonx: If I correctly understand the spec, we should always take half of the hash, whatever its size
Something like that should work:

    hash = hash[:len(hash)/2]

ericchiang: you're right. let me fix that.

server/oauth2.go
ericchiang: well I've been hashing stuff wrong for white a while. Let me update.

ericchiang:
@xeonx thanks for the comments. pushed another change.

xeonx:
I confirm at_hash validation is ok with the updated pr

server/oauth2_test.go
```diff
@@ -148,3 +150,20 @@ func TestParseAuthorizationRequest(t *testing.T) {
 		}()
 	}
 }
+
+const (
+	// at_hash value and access_token returned by Google.
```
ericchiang: Grabbed some at_hash values from Google. We now test against these.

server/server_test.go
```diff
@@ -222,6 +223,38 @@ func TestOAuth2CodeFlow(t *testing.T) {
...
+					return errors.New("no at_hash value in id_token")
+				}
+				wantAtHash, err := accessTokenHash(jose.RS256, token.AccessToken)
```
ericchiang: Actually have a test that verifies the at hash value.

server/oauth2.go
```diff
@@ -125,6 +136,88 @@ func parseScopes(scopes []string) connector.Scopes {
...
 }
 
+// Determine the signature algorithm for a JWT.
+func signatureAlgorithm(jwk *jose.JSONWebKey) (alg jose.SignatureAlgorithm, err error) {
+	if jwk.Key == nil {
+		return alg, errors.New("no signing key")
+	}
+	switch key := jwk.Key.(type) {
+	case *rsa.PrivateKey:
+		// Because OIDC mandates that we support RS256, we always return that
```
ericchiang: @rithujohn191 comments expanded here.

PR #774, opened 2017-01-12
Support for environment variables support in config files

cdemers:
This small pr enables one to use the ${VAR} notation within the yaml configuration file. As we run dex in containers, being able to use environment variables as configuration is pretty useful for us.  Would this functionality make sense for dex? If so I could make a more complete version of this pr and submit it again. Thanks!

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
@@ -39,6 +40,17 @@ func commandServe() *cobra.Command {
 	}
 }
 
+func substituteEnvVars(text string) string {
+	re := regexp.MustCompile("\\${([a-zA-Z0-9\\-_]+)}")
+	matches := re.FindAllStringSubmatch(text, -1)
+	for _, val := range matches {
+		envVar := os.Getenv(val[1])
+		// fmt.Printf("%q %q %q\n", val[0], val[1], envVar)
+		text = strings.Replace(text, val[0], envVar, -1)
+	}
+	return text
+}
+
 func serve(cmd *cobra.Command, args []string) error {
 	switch len(args) {
 	default:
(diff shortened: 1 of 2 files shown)


ericchiang:
You can already use env variables in a subset of the file. See: https://github.com/coreos/dex/blob/79c21f9b0c884165393e80319e9dbf3117b8342b/examples/config-dev.yaml#L59-L63

cmd/dex/serve.go
```diff
@@ -39,6 +40,17 @@ func commandServe() *cobra.Command {
 	}
 }
 
+func substituteEnvVars(text string) string {
```
ericchiang: I don't think this works https://play.golang.org/p/fDj4Rr5hpU

ericchiang: an nvm. I see that you're requiring them to be surrounded by `${}`

ericchiang:
The current limitation of only letting users substitute out a subset of the config is kind of intentional. I'm a little terrified of people doing things like:

```
export DEX_CONNECTORS="connectors:\n- type: mockCallback\n  id: mock\n  name: Example"
```

cdemers:
Yes, I was being lazy and avoided having to walk the config struct tree values to apply the substitutions, but we could make it so that only key yaml values would be parsed for substitution. Would that be acceptable?

ericchiang:
You can already do this for `connectorConfig` and `storageConfig` values. This is because the most likely thing to hold separate from the actual config are things like database passwords or client secrets. What other values are you looking to substitute? How are you running your containers?

Much of this could be achieved yourself. e.g. right a custom entrypoint that generates a config using the current env. For example we generate dynamic Kubernetes configmaps internally using templates.

I dislike spitting the configuration up more. The old dex combined flags with several config files. I think it's much more sane to endorse a single configuration point, then let people generate them as they wish.

cdemers:
Thanks! I understand. I didn't think about config maps, I'll have a look.

I had multiple small use cases, but my main one was about dealing with deployments in multiple environments (multiple domain names) and the associated TLS certificates that all have to be different based on the DN.
The concerned keys I found are:
- issuer
- web:tls*
- grpc:tls*

(The log level would also be nice)

I was trying to avoid custom entrypoint scripts because I feel that they hide part of the container working mechanic (and they seem to be have bad press in the kubernetes circles).

Maybe I could also just put the whole configuration file in a kubernetes secret, that would be pretty straightforward and it would make kind of sense to an administrator afterward.

Thanks again!

cdemers:
As I understand and agree with you, I will close this PR. Thanks again for having taken the time to comment.

PR #775, opened 2017-01-14
Allow CORS on keys and token endpoints

xeonx:
As discussed in https://github.com/coreos/dex/pull/766 this PR is to allow a js client to access keys and token end points.


server/server.go
```diff
@@ -218,11 +218,11 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
 	if err != nil {
 		return nil, err
 	}
-	handle("/.well-known/openid-configuration", discoveryHandler)
+	handleWithCORS("/.well-known/openid-configuration", discoveryHandler)
 
 	// TODO(ericchiang): rate limit certain paths based on IP.
-	handleFunc("/token", s.handleToken)
-	handleFunc("/keys", s.handlePublicKeys)
+	handleWithCORS("/token", s.handleToken)
```
ericchiang: why the token endpoint?

server/server.go
```diff
@@ -116,7 +116,7 @@ type Server struct {
 
 	supportedResponseTypes map[string]bool
 
-	discoveryAllowedOrigins []string
+	allowedOrigins []string
```
ericchiang: since this is only used when setting up the handlers I don't think we need this on the server.

server/handlers.go
```diff
@@ -130,18 +130,19 @@ func (s *Server) discoveryHandler() (http.Handler, error) {
...
-	})
-	if len(s.discoveryAllowedOrigins) > 0 {
-		corsOption := handlers.AllowedOrigins(s.discoveryAllowedOrigins)
-		discoveryHandler = handlers.CORS(corsOption)(discoveryHandler)
-	}
+	}), nil
+}
 
-	return discoveryHandler, nil
+func (s *Server) setupCORS(handler http.Handler) http.Handler {
```
ericchiang: maybe move this to where the handlers are setup?

server/server.go
xeonx: token endpoint could be used by a js application to get refresh token or to get a token in the hybrid or code flows.
In my use case I don't need it (implicit flow is enough for me), no problem to not handle CORS on it if you prefer.

server/server.go
xeonx: ok, will do

server/handlers.go
xeonx: Do you mean in server/server.go ?

ericchiang: Yes. Maybe even inside of newServer as a closure.

xeonx: ok, will do

server/server.go
ericchiang: Let's skip this for now. At some point I think we'll want to just allow specific flows instead of response_types. E.g.

```
allowedFlows := []string{"hybrid", "code", "implicit"}
```

When we do that, we can enable /token if "hybrid" is specified

rithujohn191:
LGTM

rithujohn191:
@xeonx, it would be great if you could add your implicit flow example as an example-app for dex (refer: https://github.com/coreos/dex/tree/master/cmd/example-app).

PR #785, opened 2017-01-24
Improve SAML Signature and Response Validation

holgerkoser:
In oder to make `dex` work with our IDP I had to change the following things:

1. **Improve Order of Namespace Declarations and Attributes in Canonical XML**<br>
This is related to an issue in goxmldsig for which I created an [pull request](https://github.com/russellhaering/goxmldsig/pull/17). This pull request has been merged.

2. **Do not compress the AuthnRequest if `HTTP-POST` binding is used**<br>
In case of HTTP-POST binding the SAML message should not be compressed. Some IDPs are fault tolerant but some consider this as an invalid AuthnRequest.

3. **SAML Response is valid if the Message and/or the Assertion is signed**<br>
Many IDPs sign only the Assertion element because only the assertion. This should should also be considered as a valid signed Response.

4. **Validate `NotBefore` and `NotOnOrAfter` Condition with allowed ClockDrift**<br>
When xml times are validated it should be possible to allow a small clock drift between the IDP and the assertion consumer.

5. **Add `AssertionConsumerServiceURL` to `AuthnRequest`**<br>
In many cases there are multiple acs URLs configured at the IDP. In this case the IDP uses the default acs URL which is probably the wrong one. Providing the ACS Url in the Request prevents this Situations.

6. **Started to add more Response validations**<br> 
Validate also the response Status, Conditions and SubjectConfirmationData


ericchiang:
Thanks, this is great.

> Improve Order of Namespace Declarations and Attributes in Canonical XML

We use glide to manage dependencies so if you want to update anything in `vendor` you'll need to use that. Please see https://github.com/coreos/dex/blob/master/Documentation/dev-dependencies.md

> Do not compress the AuthnRequest if HTTP-POST binding is used

Missed that one, thanks.

Looking into the other notes. Will comment inline.

connector/saml/saml.go
```diff
@@ -334,19 +353,97 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse string) (ident co
...
+	if err = doc.ReadFromBytes(data); err != nil {
 		return nil, fmt.Errorf("parse document: %v", err)
 	}
-
-	result, err := validator.Validate(doc.Root())
-	if err != nil {
-		return nil, err
+	verified := false
+	response := doc.Root()
+	transformedResponse, err := validator.Validate(response)
```
ericchiang: can you add some comments articulating what this code block is doing?

connector/saml/saml.go
```diff
@@ -334,19 +353,97 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse string) (ident co
...
+func (p *provider) validateStatus(resp *response) error {
+	status := resp.Status
+	if status == nil {
+		return fmt.Errorf("response did not contain a Status")
+	}
+	statusCode := status.StatusCode
+	if statusCode == nil {
+		return fmt.Errorf("response did not contain a StatusCode")
+	}
+	if statusCode.Value != "urn:oasis:names:tc:SAML:2.0:status:Success" {
```
ericchiang: Is this error status important? E.g. should we be returning the status code for the end user?

connector/saml/saml.go
```diff
@@ -334,19 +353,97 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse string) (ident co
...
+		return fmt.Errorf("status code of the response was not Success")
+	}
+	return nil
+}
+
+func (p *provider) validateSubjectConfirmation(subject *subject) error {
```
ericchiang: Can we add some notes that point to where this is discussed in the spec? It seems arbitrary otherwise.

connector/saml/saml_test.go
```diff
@@ -22,21 +28,82 @@ func loadCert(ca string) (*x509.Certificate, error) {
...
+func TestVerifyUnsignedMessageAndUnsignedAssertion(t *testing.T) {
+	runVerify(t, "testdata/idp-cert.pem", "testdata/idp-resp.xml", false)
+}
+
+func TestHandlePOST(t *testing.T) {
+	logger := logrus.New()
+	p := &provider{
+		issuer:       "http://localhost:5556/dex/callback",
+		ssoURL:       "http://idp.org/saml/sso",
+		now:          time.Now,
```
ericchiang: We generally don't use `time.Now` in tests. At most

```
now := time.Now()

timeFunc := func() time.Time { return now }
```

That way we can ensure the test doesn't depend on some arbitrary amount of time having passed (or not passed) between components of the test.

connector/saml/saml.go
```diff
@@ -348,19 +353,97 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse string) (ident co
...
+	return nil
+}
+
+func (p *provider) validateConditions(assertion *assertion) error {
+	conditions := assertion.Conditions
+	if conditions == nil {
+		return fmt.Errorf("response did not contain a Conditions element")
+	}
+	now := p.now()
+	notBefore := time.Time(conditions.NotBefore)
```
ericchiang: `NotBefore` and `NotOnOrAfter` appear to be marked as `[optional]` in the spec, but doesn't seem allowed here.

connector/saml/saml.go
```diff
@@ -106,6 +104,9 @@ type Config struct {
 	//		urn:oasis:names:tc:SAML:2.0:nameid-format:persistent
 	//
 	NameIDPolicyFormat string `json:"nameIDPolicyFormat"`
+
+	// Allow a clock_drift when checking dates
+	AllowedClockDrift int64 `json:"allowedClockDrift"`
```
ericchiang: Hmmm, why would this be configurable? In other packages we just choose a somewhat arbitrary value and role with it.

holgerkoser:
Many thanks for your quick reply,

I am relatively new to go and therefore was not aware of the dependencies management with glide. I will change this in the PR.

One general remark upfront because I have not mentioned yet. As a referenced for Validation and Signing I have used the OneLogin [Online Tools](https://www.samltool.com/online_tools.php). As an   implementation reference I have used [python-saml](https://github.com/onelogin/python-saml) and [ruby-saml](https://github.com/onelogin/ruby-saml).

Today I will add my comments to your inline notes.

connector/saml/saml.go
holgerkoser: You are right. The timestamps in the conditions element are optional and should only be validated if present.
I will change the code accroding to https://github.com/onelogin/python-saml/blob/master/src/onelogin/saml2/response.py#L317-L335

connector/saml/saml_test.go
holgerkoser: You are right. I will change it to a fixed timestamp even if I have set the data to 2116 in the example document.

connector/saml/saml.go
holgerkoser: In the saml-python implementation https://github.com/onelogin/python-saml/blob/master/src/onelogin/saml2/constants.py#L22 it is a constant value of 300 seconds.

In the saml-ruby implementation https://github.com/onelogin/ruby-saml/blob/15fb4789f8cd119fdbd722d58c659129d23a256d/lib/onelogin/ruby-saml/response.rb#L327-L329 it is not part of the settings but can be changed via options on the `Response class` and has a default value of 0 seconds.

Therefore I agree it should not be part of the configuration. I will remove it there but keep it as optional attribute in the provider struct. But which value should be used? 300 seconds seems pretty high to me. Should we go with the python value of 300 seconds?

connector/saml/saml.go
holgerkoser: I think it make sense to return the `StatusMessage` if present to the end user. I will add this element to the status struct. If not present a human reable message containing the statusCode should be returned. This way it is done in the [python library](https://github.com/onelogin/python-saml/blob/master/src/onelogin/saml2/response.py#L317-L335). 

As these elements are optional in the response of course it should not give an error if they are not found. I will change it accordingly.

connector/saml/saml.go
holgerkoser: The relevant part in the spec [Profiles for the OASIS Security
Assertion Markup Language](https://www.oasis-open.org/committees/download.php/35389/sstc-saml-profiles-errata-2.0-wd-06-diff.pdf) is section 3.3 Bearer.

> URI: urn:oasis:names:tc:SAML:2.0:cm:bearer
> The subject of the assertion is [E47]the bearer ofconsidered to be an acceptable attesting entity for the
> assertion by the asserting party, subject to optional constraints on confirmation using the attributes that
> MAY be present in the <SubjectConfirmationData> element, as defined by [SAMLCore].
> If the intended bearer is known by the asserting party to be an entity other than the subject, then the
> asserting party SHOULD identify that entity to the relying party by including a SAML identifier representing
> it in the enclosing <SubjectConfirmation> element.
> If multiple attesting entities are to be permitted to use the assertion based on bearer semantics, then
> multiple <SubjectConfirmation> elements SHOULD be included.
> Example: The bearer of the assertion can confirm itself as the subject, provided the assertion is delivered
> in a message sent to "https://www.serviceprovider.com/saml/consumer" before 1:37 PM GMT on March
> 19th, 2004, in response to a request with ID "_1234567890".
```xml
<SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer">
  <SubjectConfirmationData InResponseTo="_1234567890" Recipient="https://www.serviceprovider.com/saml/consumer" NotOnOrAfter="2004-03-19T13:27:00Z"
  </SubjectConfirmationData>
</SubjectConfirmation>
```

This is inline with the python implementation https://github.com/onelogin/python-saml/blob/master/src/onelogin/saml2/response.py#L234-L263. 

Multiple subject SubjectConfirmation can be in the assertion and at least one SubjectConfirmation must be valid. I will change this accordingly. Currently the request id is not available if I remember correctly.

connector/saml/saml.go
holgerkoser: In the old code only the signature of the document root / response element has been verfied. The `Validate` function of the goxmldsig library only looks for signatures on the root level. Therefore I have added the validation of the assertion element and replace it with the transformed one in case of success. The corresponding code in the python implementation is here https://github.com/onelogin/python-saml/blob/master/src/onelogin/saml2/response.py#L283-L305.

connector/saml/saml.go
holgerkoser: Just looked at the [saml-schema-protocol-2.0.xsd](https://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd) and found that Status and the first level StatusCode are required on the Response.

connector/saml/saml.go
ericchiang: Sorry, I meant a code comment :)

connector/saml/saml.go
ericchiang: Sorry, again a inline code comment.

connector/saml/saml.go
ericchiang: 5 minutes seems like a lot, we do 30 seconds in our Go client[0]. Can we start with 30 second value, ensure we print good errors when the times disagree, and then let users open bugs?

[0] https://github.com/coreos/go-oidc/blob/2cc7913f9f6f26eff95b633c29bfe04895d38121/jwks.go#L17-L25

connector/saml/saml.go
holgerkoser: Understood. I will add it.

connector/saml/saml.go
holgerkoser: Understood. I will add it.

connector/saml/saml.go
```diff
@@ -348,19 +355,151 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse string) (ident co
...
+			if method != "" && method != subjectConfirmationMethodBearer {
+				continue
+			}
+			subjectConfirmationData := subjectConfirmation.SubjectConfirmationData
+			if subjectConfirmationData == nil {
+				continue
+			}
+			inResponseTo := subjectConfirmationData.InResponseTo
+			if inResponseTo != "" {
+				// TODO also validate InResponseTo if present
```
ericchiang: Yep, we've got  a note on the saml interface[0]. Thanks for adding the note here.

[0] https://github.com/coreos/dex/blob/v2.1.0/connector/connector.go#L77

connector/saml/saml.go
```diff
@@ -348,19 +355,151 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse string) (ident co
...
+			}
+			// only validate that subjectConfirmationData is not expired
+			now := p.now()
+			notOnOrAfter := time.Time(subjectConfirmationData.NotOnOrAfter)
+			if !notOnOrAfter.IsZero() {
+				if now.After(notOnOrAfter) {
```
ericchiang: no clock skew allowed here?

connector/saml/saml.go
```diff
@@ -348,19 +355,151 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse string) (ident co
...
+	conditions := assertion.Conditions
+	if conditions == nil {
+		return nil
+	}
+	// Validates Assertion timestamps
+	now := p.now()
+	notBefore := time.Time(conditions.NotBefore)
+	if !notBefore.IsZero() {
+		if now.Add(allowedClockDrift).Before(notBefore) {
+			return fmt.Errorf("timestamp \"NotBefore\" in Conditions is not yet valid. Check system clock")
```
ericchiang: Maybe a better error message would be

```
fmt.Errorf("at %s got response that cannot be processed before %s", now, notBefore)
```

And the same for `NotOnOrAfter`

connector/saml/saml.go
```diff
@@ -348,19 +355,151 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse string) (ident co
...
+		if audiences != nil && len(audiences) > 0 {
+			issuerInAudiences := false
+			for _, audience := range audiences {
+				if audience.Value == p.issuer {
+					issuerInAudiences = true
+					break
+				}
+			}
+			if !issuerInAudiences {
+				return fmt.Errorf("%q is not a valid audience for this Response", p.issuer)
```
ericchiang: Could we ensure that the actual audiences are printed?

```
fmt.Errorf("required audience %s was not in Response audiences %s", p.issuer, audiences)
```

ericchiang:
Also can you please squash?

connector/saml/saml.go
holgerkoser: They don't do it in the [python-saml](https://github.com/onelogin/python-saml/blob/master/src/onelogin/saml2/response.py#L252-L256). I don't now why but wanted to be in line? For the expiration case `NotOnOrAfter` the problem is anyway not so critical as for the not valid yet case which I currently do not validate for subjectConfirmations.

holgerkoser: But of course I can add it if you think it would be more consistant.

connector/saml/saml.go
holgerkoser: I will change this as you suggested.

connector/saml/saml.go
holgerkoser: dito

holgerkoser:
I have changed the error messages and pushed. 

Do I have to squash the commits? I think you should be able to do it when  accepting your PR as described in ["Squash your commits"](https://github.com/blog/2141-squash-your-commits) and its [new documentation](https://help.github.com/articles/about-pull-request-merge-squashing/)

ericchiang:
> I have changed the error messages and pushed.
>
> Do I have to squash the commits? I think you should be able to do it when accepting your PR as described in "Squash your commits" and its new documentation

We prefer to keep the merge commits. It helps us when doing release notes. If we squash I don't think we get a merge commit.

Also we like to keep `vendor/` updates as part of a different commit. Not a huge deal because it's only one line here though.

If you need the git commands, this should work:

```
git reset ( SHA before your commits )
git add .
git commit
git push -f origin ( branch name )
```

holgerkoser:
Should I add all files to one commit or keep the vendor update in a first
```
M	glide.lock
M	glide.yaml
M	vendor/github.com/russellhaering/goxmldsig/canonicalize.go
```
and the saml files in a second
```
M	connector/saml/saml.go
M	connector/saml/saml_test.go
M	connector/saml/types.go
```

ericchiang:
Make sure these:

```
M	glide.yaml
M	connector/saml/saml.go
M	connector/saml/saml_test.go
M	connector/saml/types.go
```

are separate from

```
M	glide.lock
M	vendor/github.com/russellhaering/goxmldsig/canonicalize.go
```

It's so we can differentiate manual changes from generated ones :)

holgerkoser:
Thanks. I just saw that you do it this way in your previous commits. Now we only have two commits.

rithujohn191:
Tested SAML connector with Dex example-app. :+1:

PR #789, opened 2017-01-26
Documentation/proposals: Add a proposal for refresh token revocation.

rithujohn191:
Outline the design for handling refresh token revocation in Dex v2.

### DIFF
--- a/Documentation/proposals/token-revocation.md
+++ b/Documentation/proposals/token-revocation.md
@@ -1,0 +1,82 @@
+# Proposal: design for revoking refresh tokens.
+
+Refresh tokens are issued to the client by the authorization server and are used
+to request a new access token when the current access token becomes invalid or expires.
+It is a common usecase for the end users to revoke client access to their identity.
+This proposal defines the changes needed in Dex v2 to support refresh token revocation.
+
+## Motivation
+
+1. Currently refresh tokens are not associated with the user. Need a new "session object" for this.
+2. Need an API to list refresh tokens based on the UserID.
+3. We need a way for users to login to dex and revoke a client.
+4. Limit the number refresh tokens for each user-client pair to 1.
+
+## Details
+
+Currently in Dex when an end user successfully logs in via a connector and has the OfflineAccess
+scope set to true, a refresh token is created and stored in the backing datastore. There is no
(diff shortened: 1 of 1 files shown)


Documentation/proposals/token-revocation.md
```diff
@@ -0,0 +1,72 @@
...
+//
+// Will have to handle garbage collection i.e. if no refresh token exists for a user,
+// this object must be cleaned up.
+type OfflineSession struct {
+        // UserID of an end user who has logged in to the server.
+        UserID        string
+        // The ID of the connector used to login the user.
+        ConnID     string
+        // List of pointers to RefreshTokens issued for SessionID
+        Refresh         []*RefreshToken
```
ericchiang: "RefreshToken" should only contain metadata about the refresh token. It's not the actual refresh token type, right?

rithujohn191: should I clarify this to be internal.RefreshToken?

ericchiang: We should probably just create a `RefresTokenhRef` object.

Documentation/proposals/token-revocation.md
```diff
@@ -0,0 +1,72 @@
...
+		* Create an OfflineSession for the user and add the new `RefreshToken` pointer to
+		  the `Refresh` list.
+                
+2. Refresh token rotation:
+   There will be no change to this codeflow. When the client refreshes a refresh token, the `TokenID`
+   still remains intact and only the `RefreshToken` obj gets updated with a new nonce. We do not need
+   any additional checks in the OfflineSession objects as the `RefreshToken` pointers still remain intact.
+
+3. User revokes a refresh token (New functionality):
+   A user that has been authenticated via a connector will have the ability to revoke their refresh tokens.
```
sym3tri: How is authentication implemented? A valid JWT for this dex instance (iss) with a matching user ID?

ericchiang: Oh, we should remove this. Our API doesn't do authentication or authorization. We'd want to leave it to another app.

rithujohn191: Oh I see how that can be misleading. I will clarify that an external app will have to handle the authentication.

PR #790, opened 2017-01-26
Fixes #706

### LINKED ISSUE #706: connector/github: API calls do not account for pagination.

givia:


connector/github/github.go
```diff
@@ -221,40 +221,52 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
+			if team.Org.Login == org {
+				groups = append(groups, team.Name)
+			}
+		}
+
+		if len(teams) < 100 {
+			break
+		}
+
+		page = page + 1
```
ericchiang: Why not use `Link: rel="next"`? https://developer.github.com/v3/orgs/teams/#response-13

givia: I think `rel="next"` is a part of response and increasing the page remains unchanged. But we can change the break condition to `rel="last"`.

givia: I changed the code to use link header as you suggested.

rithujohn191:
lgtm. tested out the change.

PR #793, opened 2017-02-01
storage: Add OfflineSession object to backend storage.

rithujohn191:
This change adds session objects for users with refresh tokens known as `OfflineSession` objects to the backend storage.

Having `OfflineSession` objects will allow us to access refresh tokens on a per User basis. This will come in handy when we implement refresh token revocation.

In order to uniquely identify these objects we need both the UserID of the end user and the ConnectionID of the connector the user is on.

Added conformance tests to ensure the functionality works.


storage/kubernetes/client.go
```diff
@@ -58,11 +58,24 @@ func (c *client) idToName(s string) string {
...
+func (c *client) idToNameComposite(s1 string, s2 string) string {
+	return idToNameComposite(s1, s2, c.hash)
+}
+
 // Kubernetes names must match the regexp '[a-z0-9]([-a-z0-9]*[a-z0-9])?'.
 var encoding = base32.NewEncoding("abcdefghijklmnopqrstuvwxyz234567")
 
 func idToName(s string, h func() hash.Hash) string {
-	return strings.TrimRight(encoding.EncodeToString(h().Sum([]byte(s))), "=")
+	h().Write([]byte(s))
```
ericchiang: This is wrong but I don't think we can change this.

storage/sql/migrate.go
```diff
@@ -153,6 +153,13 @@ var migrations = []migration{
 				signing_key_pub bytea not null,   -- JSON object
 				next_rotation timestamptz not null
 			);
+
```
ericchiang: probably needs to be part of a new migration

storage/kubernetes/client.go
```diff
@@ -58,11 +58,24 @@ func (c *client) idToName(s string) string {
 	return idToName(s, c.hash)
 }
 
+// idToNameComposite maps two arbitrary IDs, to a single Kubernetes object name.
+// This is used when more than one field is used to uniquely identify the object.
+func (c *client) idToNameComposite(s1 string, s2 string) string {
```
ericchiang: Maybe `offlineTokenName(userID, connID string) string`?

rithujohn191: Made it generic incase other objects wanted to use it later on. But will make it userID and connID for now since only OfflineSession makes use of it.

storage/sql/migrate.go
rithujohn191: Done

storage/kubernetes/client.go
rithujohn191: Changing it back to maintain backward compatibility.

storage/kubernetes/client.go
ericchiang: It's not generic until you're using it twice :)

rithujohn191: Yep you're right. Changed this to what you suggested.

rithujohn191:
Added changes to the dex code flow when a Client requests a refresh token. This involves creating and maintaining the OfflineSession object per user.

Change Details:
1) When a refresh token is requested by a client, try to retrieve the `OfflineSession` object for the User with the given `UserID + ConnID`.
2) If the object exists update it with new refresh token reference object and delete the old refresh token in the storage.
3) If the object does not exists create a new `OfflineSession` object for the user.

Note: I have also changed the `Refresh` attribute in `OfflineSession` struct to be a single `RefreshTokenRef` value instead of an array since we plan to only allow a single refresh token per client-user pair.

storage/storage.go
```diff
@@ -241,6 +245,26 @@ type RefreshToken struct {
...
+type OfflineSession struct {
+	// UserID of an end user who has logged in to the server.
+	UserID string
+
+	// The ID of the connector used to login the user.
+	ConnID string
+
+	// RefreshTokenRef object which is a reference object
+	// for refresh tokens issued to a user with ID = UserID.
+	Refresh RefreshTokenRef
```
ericchiang: Sorry I missed this, but a user can have multiple RefreshTokens out for different clients.

server/handlers.go
```diff
@@ -668,6 +668,44 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
 			return
 		}
+
+		tokenRef := storage.RefreshTokenRef{
+			ID:        refresh.ID,
+			CreatedAt: refresh.CreatedAt,
+			LastUsed:  refresh.LastUsed,
+		}
+
+		if session, err := s.storage.GetOfflineSession(refresh.Claims.UserID, refresh.ConnectorID); err != nil {
```
ericchiang: This should be isolated to only when `err == storage.ErrNotFound`. Other errors need to be reported as errors.

server/handlers.go
```diff
@@ -668,6 +668,44 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
+				ConnID:  refresh.ConnectorID,
+				Refresh: tokenRef,
+			}
+			if err := s.storage.CreateOfflineSession(offlineSession); err != nil {
+				s.logger.Errorf("failed to create offline session: %v", err)
+				s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
+				return
+			}
+		} else {
+			oldTokenRef := session.Refresh
```
ericchiang: Per comment below, while a client can only have one refresh token per user, a user might be logged into multiple clients.

ericchiang:
Also needs to be some logic when a client attempts to refresh a token.

storage/storage.go
```diff
@@ -241,6 +245,26 @@ type RefreshToken struct {
 	Nonce string
 }
 
+// RefreshTokenRef is a reference object that contains metadata about refresh tokens.
+type RefreshTokenRef struct {
+	ID        string
+	CreatedAt time.Time
+	LastUsed  time.Time
```
ericchiang: We might be able to get this info from the underlying refresh token. I guess we just have to figure out if we want to save database "gets".

rithujohn191: I put this as a part of this struct in case we want to list the refresh tokens per user and know the time-stamps related to each. In that case it will save us several database retrievals (per user refresh token).

ericchiang: Yep, definitely. But be aware we'll need to update `LastUsed` on every refresh.

We might also put things that are guaranteed never to change like the client ID here.

rithujohn191: hmmm... if the client ID is put into this struct, then we would only have 1 refresh token per user-client pair right? that would mean that if we need to list the refresh token per user we would have to query several `OfflineSession` objects. Or do you mean we should have a list of Client IDs per object?

ericchiang: One `OfflineSession` per user (maybe even rename it `OfflineSessions`?). But the `RefreshTokenRef` needs to contain a client ID. The steps for creating a new refresh token should be:

1. Get an existing OfflineSisson(s) object.
2. Is there a refresh token for the client currently requesting the new refresh token?
3. If so remove that refresh token from the list, add a new one.

rithujohn191: Changed the `Refresh` attribute to be of type map[string]*RefreshTokenRef indexed by the ClientID of the refresh token. 

Pros: 
* efficient access of object using ClientID of refresh token since it is a hash table indexed by ClientID
* easy update of attributes (like LastUsed) since it a pointer to `RefreshTokenRef`

Also renamed OfflineSession to OfflineSessions as you recommended

server/handlers.go
rithujohn191: Made this change.

server/handlers.go
```diff
@@ -668,6 +668,54 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
+
+		tokenRef := storage.RefreshTokenRef{
+			ID:        refresh.ID,
+			ClientID:  refresh.ClientID,
+			CreatedAt: refresh.CreatedAt,
+			LastUsed:  refresh.LastUsed,
+		}
+
+		if session, err := s.storage.GetOfflineSessions(refresh.Claims.UserID, refresh.ConnectorID); err != nil {
+			if err == storage.ErrNotFound {
```
ericchiang:
```
if err != storage.ErrNotFound {
    // handle error
    return
}
// create new token
```

server/handlers.go
```diff
@@ -668,6 +668,54 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
 			return
 		}
+
```
ericchiang: If anything fails here, you'll need to clean up that refresh token you just created.

Maybe a `defer`?

```
defer func() {
    if !createdOfflineSession {
        // delete refresh token.
    }
}()
```

server/handlers.go
```diff
@@ -668,6 +668,54 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
+					s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
+					return
+				}
+			} else {
+				s.logger.Errorf("failed to get offline session: %v", err)
+				s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
+				return
+			}
+		} else {
+			if oldTokenRef, ok := session.Refresh[tokenRef.ClientID]; ok {
```
ericchiang: Do we know `session.Refresh` isn't going to be `nil` here?

server/handlers.go
```diff
@@ -814,14 +863,25 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
...
 	if err := s.storage.UpdateRefreshToken(refresh.ID, updater); err != nil {
 		s.logger.Errorf("failed to update refresh token: %v", err)
 		s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
 		return
 	}
+
+	// Update LastUsed time stamp in refresh token reference object
+	// in offline session for the user.
+	if err := s.storage.UpdateOfflineSessions(refresh.Claims.UserID, refresh.ConnectorID, func(old storage.OfflineSessions) (storage.OfflineSessions, error) {
+		old.Refresh[refresh.ClientID].LastUsed = lastUsed
```
ericchiang: Need to check if this has the same refresh token ID here. What between the call to get the refresh token and this call the refresh token was deleted?

server/handlers.go
rithujohn191: session.Refresh gets initialized upon creation of the OfflineSession object. And we only hit this else case if the session object already exists.

ericchiang: Sure, but what about when this map is empty? Does that get properly translated for storages? Maybe we should have a conformance test to ensure an empty map doesn't get unmarshaled as nil?

ericchiang:
It would be really good if we could figure out a testing strategy for this.

server/handlers.go
rithujohn191: In the current code base since we don't ever delete the refresh token, this case would not arise. I was planning on addressing these corner cases once I add the ability to delete refresh tokens. I could get to them now if you'd like.

server/handlers.go
rithujohn191: okay will add that test

server/handlers.go
ericchiang: We do delete refresh tokens when clients request a second one.

server/handlers.go
rithujohn191: Tested to see that this works fine. It gets interpreted as an empty map after its unmarshaled. Also, once we add token revocation, if there is ever a case when the `OfflineSession` object has an empty `Refresh` map, we would clean up the entire `OfflineSession` object.

ericchiang: Thanks. But yeah we need to codify that assumption, not just use a map that could potentially be nil.

rithujohn191:
Had an offline discussion with @ericchiang and we decided to rename the `OfflineSession` object to `User`. The motivation behind this being that this object will eventually serve as the sole user object which will have more functionality than just being a link between the users and their refresh tokens.

sym3tri:
Discussed with @rithujohn191 IRL. My concern with using a User object is:
1. We don't have a use-case we're solving with it in this PR. I'd liked to avoid speculation.
2. Once we introduce a User object, it's very important and cannot be deleted. Session objects are more transient and matter less.

rithujohn191:
Ed, also brought up the point that when Dex is reconfigured to use a different connector, we will have to clean up the stale `OfflineSession` objects that correspond to the old connector. Need a new garbage collection story.

ericchiang:
> Ed, also brought up the point that when Dex is reconfigured to use a different connector, we will have to clean up the stale OfflineSession objects that correspond to the old connector. Need a new garbage collection story.

Why? What's the concern here?

ericchiang:
> We don't have a use-case we're solving with it in this PR. I'd liked to avoid speculation.

The motivation was that we need to figure out how to map the "sub" we're returning in the ID token to these offline sessions. At that point, it seems like a user object. By keeping the name "offline session" We're kinda speculating that this object won't evolve into that.

> Once we introduce a User object, it's very important and cannot be deleted. Session objects are more transient and matter less.

The object being added here tracks all offline sessions by a particular "user" that logged in through a connector. It will be kept around even when those individual session get renewed or revoked.

I think there's a confusion around the term "User" which is horribly overloaded. I don't get the "it's very important" remark. "User" in the case that I'm proposing just means "user ID returned by a given connector."

rithujohn191:
> It will be kept around even when those individual session get renewed or revoked.

Once a refresh token is revoked, if it is the last refresh token in the `Refresh` list the `OfflineSession` will be cleaned up. As of now all the `OfflineSession` does is provide a link between users and their refresh tokens.  Wouldn't be very valuable to keep around a session object with no refresh token associated with it

ericchiang:
> Once a refresh token is revoked, if it is the last refresh token in the Refresh list the OfflineSession will be cleaned up. As of now all the OfflineSession does is provide a link between users and their refresh tokens. Wouldn't be very valuable to keep around a session object with no refresh token associated with it

Sure. But there'd be no functional change if we did keep it around. It's just a semantic point about what we want this object do (or not do) in the future.

Anyway, I'm fine with calling it "OfflineSessions" for now if "User" is contentions.

rithujohn191:
Maybe we can stick to calling it `OfflineSession` for now to avoid the confusion and transition it to `User` at a later stage when we have a clear cut use-case for it

ericchiang:
Sure, but the backing storage types will still be called "OfflineSession", which isn't the end of the world.

server/handlers.go
rithujohn191: Added this

server/handlers.go
rithujohn191: Added a check to verify refresh token ID is the same before updating the LastUsed timestamp.

server/handlers.go
```diff
@@ -814,14 +884,33 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
...
+		old.LastUsed = lastUsed
 		return old, nil
 	}
+
+	// Update LastUsed time stamp in refresh token reference object
+	// in offline session for the user.
+	if err := s.storage.UpdateOfflineSessions(refresh.Claims.UserID, refresh.ConnectorID, func(old storage.OfflineSessions) (storage.OfflineSessions, error) {
+		if old.Refresh[refresh.ClientID].ID != refresh.ID {
+			s.logger.Errorf("refresh token with id %s is invalid", refresh.ID)
+			s.tokenErrHelper(w, errInvalidRequest, "Refresh token is invalid.", http.StatusBadRequest)
```
ericchiang: When you're in the middle of an update function, think of it as holding a transaction. You shouldn't wait to print or respond to web requests until you're out of the update.

ericchiang:
looks good. going to test it manually on monday then we should be good to merge.

server/handlers.go
rithujohn191: Oh ok. Removed the error logging and the wed request response.

rithujohn191:
Tested this change with example-app.

rithujohn191:
Added a unit test to server/server_test.go to test out the refresh token code flow.

server/server_test.go
```diff
@@ -971,3 +971,123 @@ func TestKeyCacher(t *testing.T) {
...
+		RedirectURL:  redirectURL,
+	}
+
+	if resp, err = http.Get(oauth2Client.server.URL + "/login"); err != nil {
+		t.Fatalf("get failed: %v", err)
+	}
+
+	q := resp.Request.URL.Query()
+	// Grab code, exchange for token.
+	if code := q.Get("code"); code != "" {
```
ericchiang: This needs to be done in the server handler (the client). I realize that this technically works, but it's doing a weird introspection into the request the end user sent, when the client is the thing that's supposed to be doing the code exchange.

rithujohn191: Moved the "code exchange" into the server handler of the client.

PR #802, opened 2017-02-14
api: adding a gRPC call for revoking refresh tokens.

rithujohn191:
Added a gRPC call that will enable users to revoke client specific refresh tokens.

Also added a test in api_test.go. Since we do not have APIs in place for creating refresh tokens and offline session objects, this has been done using the storage methods. If this test case seems like an overkill, I could get rid of it.


server/api.go
```diff
@@ -228,3 +228,42 @@ func (d dexAPI) ListRefresh(ctx context.Context, req *api.ListRefreshReq) (*api.
...
 }
+
+func (d dexAPI) RevokeRefresh(ctx context.Context, req *api.RevokeRefreshReq) (*api.RevokeRefreshResp, error) {
+	id := new(internal.IDTokenSubject)
+	if err := internal.Unmarshal(req.UserId, id); err != nil {
+		d.logger.Errorf("api: failed to unmarshal ID Token subject: %v", err)
+		return nil, err
+	}
+
+	offlineSessions, err := d.s.GetOfflineSessions(id.UserId, id.ConnId)
```
ericchiang: Why isn't this done in an `Update`?

server/api.go
```diff
@@ -228,3 +228,42 @@ func (d dexAPI) ListRefresh(ctx context.Context, req *api.ListRefreshReq) (*api.
...
+		d.logger.Errorf("api: user does not have a refresh token for the client = %s", req.ClientId)
+		return nil, fmt.Errorf("user does not have a refresh token for the client = %s", req.ClientId)
+	}
+
+	// Remove entry from Refresh list of the OfflineSession object.
+	delete(offlineSessions.Refresh, req.ClientId)
+
+	// If the user has no more refresh tokens, the OfflineSession object should be cleaned up.
+	if len(offlineSessions.Refresh) == 0 {
+		if err := d.s.DeleteOfflineSessions(id.UserId, id.ConnId); err != nil {
```
ericchiang: Can we just keep this object around? Why delete it?

ericchiang:
thanks for the test. couple nits

server/api.go
rithujohn191: Discussed offline and decided to keep the Offline Session object around even if the user does not have active refresh tokens. Might be helpful later on for auditing purposes.

server/api.go
rithujohn191: Changed this to an update

api/api.proto
```diff
@@ -120,6 +120,19 @@ message ListRefreshResp {
...
+message RevokeRefreshReq {
+  // The "sub" claim returned in the ID Token.
+  string user_id = 1;
+  string client_id = 2;
+}
+
+// RevokeRefreshResp determines if the refresh token is revoked successfully. 
+message RevokeRefreshResp {
+  // Set to true is refresh token was not found and token could not be revoked.
+  bool not_found = 1;
```
ericchiang: this is never set?

rithujohn191: Set this to true when UpdateOfflineSession returns err == storage.ErrNotFound.

PR #809, opened 2017-02-21
storage: Surface "already exists" errors.

### LINKED ISSUE #808: Creating a password that already exist doesn't fill the response struct with already_exist flag
the err variable consist of UNIQUE CONSTRAINT. It should probably fill the already_exist flag in response.

rithujohn191:
This fixes #808.

If a database `Exec` command fails due to an "already exists" error, it is translated to a storage.ErrAlreadyExists error and propagated forward.

Added additional conformance tests to check for this scenario.


ericchiang:
Awesome!

Thanks for this, especially for adding a test.

Does Kubernetes storage already pass?

rithujohn191:
Yep it does:
PASS
ok  	github.com/coreos/dex/storage/kubernetes	1.034s

Issue #811, opened 2017-02-22
dex.spec for building a rpm?


SEJeff:
Figured it was worth asking about a spec file for building dex into a rpm. I'm happy to do this if it would be accepted.

I've got a policy of only allowing docker containers (on premise) that come from a RHEL based internal image, and generally still package things as rpm, even dex. Figured it made sense to contribute it back if that seems sensible?

Ditto for a systemd service file if someone might want to run dex standalone on a server.

ericchiang:
> I'm happy to do this if it would be accepted.

Yeah absolutely. It's been a while since I built an RPM, but would be happy to review.

We haven't been doing active development on standalone deployments mostly because we don't use them internally, not because we don't want them.

SEJeff:
Great, once I've got it working, I'll send a PR. I figured you'd be ok with a systemd service file once it is done since I know CoreOS :heart: systemd.

nabokihms:
I will close this one since providing package builds for different Linux distributions has yet to be planned. Feel free to open a new issue or PR if you still need one.

PR #813, opened 2017-02-23
[Makefile] Allow specifying VERSION as an env var

SEJeff:
This makes specifying the VERSION when building native operating system packages require less hacks.

Refs: #811

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ ORG_PATH=github.com/coreos
 REPO_PATH=$(ORG_PATH)/$(PROJ)
 export PATH := $(PWD)/bin:$(PATH)
 
-VERSION=$(shell ./scripts/git-version)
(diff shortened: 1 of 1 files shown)


SEJeff:
```bash
jschroeder@omniscience:~/git/dex (master *)$ make VERSION=2.1.0-ohyeah
github.com/coreos/dex/cmd/dex
github.com/coreos/dex/cmd/example-app
jschroeder@omniscience:~/git/dex (master *)$ bin/dex version
dex Version: 2.1.0-ohyeah
Go Version: go1.8
Go OS/ARCH: linux amd64
```

SEJeff:
My first thought was patching it. My second thought was running `go install ... cmd/dex cmd/example-app` in the spec file. Then I remembered you can do this in Makefiles. Thanks!

PR #819, opened 2017-02-24
[storage.md] Fix the ThirdPartyResource syntax

### LINKED ISSUE #818: kubernetes 1.5.2 doesn't like the TRP syntax in the storage docs
```bash
jschroeder@omniscience:~/projects/dex$ cat tpr.yaml
kind: ThirdPartyResource
apiVersion: extensions/v1beta1
metadata:
  name: o-auth2-client.oidc.coreos.com
versions:
  - v1
description: "An OAuth2 client."
jschroeder@omniscience:~/projects/dex$ kubectl create -f tpr.yaml
error: error validating "tpr.yaml": error validating data: field versions[0]: expected object of type map[string]interface{}, but the actual type is string; if you choose to ignore these errors, turn validation off with --validate=false
```

SEJeff:
This makes manually creating the `o-auth2-client.oidc.coreos.com` actually work.

Fixes #818

### DIFF
--- a/Documentation/storage.md
+++ b/Documentation/storage.md
@@ -20,7 +20,7 @@ apiVersion: extensions/v1beta1
 metadata:
   name: o-auth2-client.oidc.coreos.com
 versions:
-  - v1
+  - name: v1
 description: "An OAuth2 client."
 ```
 


rithujohn191:
Thanks for the fix.

PR #815, opened 2017-02-24
storage/kubernetes: fix kubernetes storage conformance test failures

ericchiang:
fixes #814

cc @rithujohn191

PR #823, opened 2017-02-27
storage/kubernetes: fix conflict error detection in TRP creation

### LINKED ISSUE #822: error resources: ID already exists.
Im running dex v2.2.2 on kubernetes, after a pod restarts, i see multiple lines of failed creating third party resources: ID already exists.

should this be an error? should it log it once and maybe as info? or is it standard that I go in after each restart and update the client-id?

ericchiang:
PR #815 fixed the Kubernetes storage implementation by correctly
returning storage.ErrAlreadyExists on POST conflicts. This caused a
regression in TPR creation (#822) when some, but not all, of the
resources already existed. E.g. for users upgrading from old
versions of dex.

Fixes #822

### DIFF
--- a/storage/kubernetes/storage.go
+++ b/storage/kubernetes/storage.go
@@ -3,7 +3,6 @@ package kubernetes
 import (
 	"errors"
 	"fmt"
-	"net/http"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


storage/kubernetes/storage.go
```diff
@@ -108,27 +113,29 @@ func (c *Config) open(logger logrus.FieldLogger) (*client, error) {
...
+	ok = true
 	for _, r := range thirdPartyResources {
 		err := cli.postResource("extensions/v1beta1", "", "thirdpartyresources", r)
 		if err != nil {
-			if e, ok := err.(httpError); ok {
-				if e.StatusCode() == http.StatusConflict {
-					cli.logger.Errorf("third party resource already created %q", r.ObjectMeta.Name)
-					continue
-				}
+			if err == storage.ErrAlreadyExists {
```
rithujohn191: Do we want to have a similar check for storage.ErrNotFound and have a separate message for that too?

ericchiang: `storage.ErrNotFound` would mean that TPRs weren't enabled on the server. hmmm, maybe? Let me go add that.

ericchiang:
@rithujohn191 updated the logging message.

Issue #828, opened 2017-03-01
kubectl refresh issues


zihaoyu:
We use `dex` as OIDC provider for `kube-apiserver`, and configure `kubeconfig` to use [OIDC authenticator](https://kubernetes.io/docs/admin/authentication/#option-1---oidc-authenticator) in user config. Our expiration settings look like the following:
```
expiry:
  signingKeys: 6h
  idTokens: 24h
```
However we noticed that when we come into office the next day, `kubectl` can't authenticate us some times:
```
error: You must be logged in to the server (the server has asked for the client to provide credentials)
```

So we need to get new tokens and put them in `kubeconfig` file.

Is it because `signingKeys` were rotated before `idTokens` were re-issued, so `dex` didn't know how to decrypt the tokens?

ericchiang:
@zihaoyu we've been seeing this issue crop up recently.

Dex's logic will keep old signing keys around as long as there are id tokens that it could have signed. There will no longer be used to sign, but will be kept around to verify. I don't think this is the source of the issue, but yes, we've seen this response from kubectl a bit lately.

What version of dex are you using? What logins (LDAP, GitHub, etc.)?

Anecdotally we've seen these bugs reported less with our recent release v2.2.3. Could you try that?

zihaoyu:
@ericchiang We use GitHub connector, and `quay.io/coreos/dex:v2.0.0-beta.1` image. Will try v2.2.3. Thanks.

joshrosso:
Experiencing an identical behavior with:

Dex: 2.1.0
Connector: LDAP

ericchiang:
I think I can see an issue https://github.com/coreos/dex/blob/3797a71ec9b5cbd254908f9d509a9dd21d0cfb96/server/rotation.go#L135

ericchiang:
Yep, the expiry bug is causing verification keys to arbitrarily be deleted. I think this is causing the issue. Will send a fix ASAP.

ericchiang:
Okay I was finally able to reproduce this

```
$ kubectl get pods
I0301 13:16:35.917132    1893 logs.go:41] RoundTripper returned a response & error; ignoring response
Unable to connect to the server: No valid id-token, and cannot refresh without refresh-token
```

joshrosso:
Thanks @ericchiang.

So #829 fixes the above message?

ericchiang:
[...]

PR #829, opened 2017-03-01
server: fix expiry detection for verification keys

ericchiang:
updates #828

### DIFF
--- a/server/rotation.go
+++ b/server/rotation.go
@@ -22,8 +22,9 @@ type rotationStrategy struct {
 	// Time between rotations.
 	rotationFrequency time.Duration
 
-	// After being rotated how long can a key validate signatues?
-	verifyFor time.Duration
+	// After being rotated how long should the key be kept around for validating
+	// signatues?
+	idTokenValidFor time.Duration
 
 	// Keys are always RSA keys. Though cryptopasta recommends ECDSA keys, not every
 	// client may support these (e.g. github.com/coreos/go-oidc/oidc).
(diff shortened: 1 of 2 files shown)


server/rotation.go
```diff
@@ -132,7 +132,7 @@ func (k keyRotater) rotate() error {
 		i := 0
 
 		for _, key := range keys.VerificationKeys {
-			if !key.Expiry.After(tNow) {
```
dghubble: // key is still valid, include it in the VerificationKeys

server/rotation.go
```diff
@@ -132,7 +132,7 @@ func (k keyRotater) rotate() error {
...
+			if key.Expiry.After(tNow) {
```
dghubble: // key is still valid, include it in the VerificationKeys

ericchiang:
@rithujohn191 can you take a look at the PR again? I've added a few more comments to the rotation code.

server/rotation.go
```diff
@@ -128,22 +129,30 @@ func (k keyRotater) rotate() error {
...
-		i := 0
+		expired := func(key storage.VerificationKey) bool {
+			return tNow.After(key.Expiry)
+		}
 
+		// Remove any verification keys that have expired.
+		i := 0
 		for _, key := range keys.VerificationKeys {
-			if !key.Expiry.After(tNow) {
+			if !expired(key) {
```
ericchiang: Only functional change. Everything else is comments and renaming.

server/rotation.go
```diff
@@ -128,22 +129,30 @@ func (k keyRotater) rotate() error {
...
-			// Move current signing key to a verification only key.
+			// Move current signing key to a verification only key, throwing
+			// away the private part.
 			verificationKey := storage.VerificationKey{
 				PublicKey: keys.SigningKeyPub,
-				Expiry:    tNow.Add(k.strategy.verifyFor),
+				// After demoting the signing key, keep the token around for at least
+				// the amount of time an ID Token is valid for. This ensures the
+				// verification key won't expire until all ID Tokens is could have
+				// signed have expired as well.
```
rithujohn191: "This ensures the verification key won't expire until all ID Tokens it has signed have expired as well"?

ericchiang: thanks, updated.

PR #842, opened 2017-03-09
*: only use docker when releasing, update to Go 1.8, remove aci scripts

ericchiang:
This change modifies our release process to only require Docker
when building a release and updates our released binary to use Go
1.8. It also removes our .aci scripts, which we've not been
regularly building.

A nice consequence is that OSX users can now build a release image.

cc @rithujohn191


Documentation/dev-releases.md
```diff
@@ -86,18 +81,3 @@ sudo setenforce Permissive
 make docker-image
```
rithujohn191: We can get rid of the "rkt doesn't play nice with SELinux" comment under "Building the Docker image"

ericchiang: thanks for catching that! removed the reference.

PR #844, opened 2017-03-10
update kubernetes example-app explanation

dmmcquay:
Clarify some potentially confusing issues with how to run and build the
example-app binary.

fixes https://github.com/coreos/dex/issues/843

### DIFF
--- a/Documentation/kubernetes.md
+++ b/Documentation/kubernetes.md
@@ -93,11 +93,19 @@ __Caveats:__ No health checking is configured because dex does its own TLS termi
 
 ## Logging into the cluster
 
-The example app can be used to log into the cluster. Choose the GitHub option and grant access to dex to view your profile.
+The `example-app` can be used to log into the cluster and get an ID Token.
+
+To build the app, you can run `make` and it will build the `example-app` binary in the root of the repo in the `bin` directory. To build the `example-app` requires at least a 1.7 version of Go. 
 
 ```
 $ ./bin/example-app --issuer https://dex.example.com:32000 --issuer-root-ca examples/k8s/ssl/ca.pem
 ```
+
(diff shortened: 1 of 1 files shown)


Documentation/kubernetes.md
```diff
@@ -93,12 +93,20 @@ __Caveats:__ No health checking is configured because dex does its own TLS termi
...
 
 ```
 $ ./bin/example-app --issuer https://dex.example.com:32000 --issuer-root-ca examples/k8s/ssl/ca.pem
 ```
 
+Please note that the `example-app` will listen at http://127.0.0.1:5555 and can be changed with the `--listen` flag.
+
+Once the example app is running, choose the GitHub option and grant access to dex to view your profile.
+
+The default redirect uri is http://127.0.0.1:5555/callback and can be changed with the `--redirect-uri` flag and should correspond with your GitHub OAuth2 client.
```
ericchiang: The example-app's redirect-uri isn't impacted by the the GitHub OAuth2 setup. The example app is dex's client, so this flag needs to map to the value in the configmap.

Documentation/kubernetes.md
```diff
@@ -93,12 +93,20 @@ __Caveats:__ No health checking is configured because dex does its own TLS termi
 
 ## Logging into the cluster
 
-The example app can be used to log into the cluster. Choose the GitHub option and grant access to dex to view your profile.
+The `example-app` can be used to log into the cluster and get an ID Token.
+
+To build the app, you can run `make` and it will build the `example-app` binary in the root of the repo in the `bin` directory. To build the `example-app` requires at least a 1.7 version of Go.
```
ericchiang: Probably doesn't need to be a new paragraph.

Also we need to specify that `make` is also run at the root of the directory.

PR #851, opened 2017-03-15
*: make cloning to GOPATH optional

ericchiang:
This PR causes dex to setup its own GOPATH internally using symlinks.
As a result, users should be able to clone dex to any directory and
just built or test it. E.g. this should work:

    git clone https://github.com/coreos/dex
    cd dex
    make

The on host dependencies for building dex are now just a recent Go
installation. Testing still requires golint.

This change can be tested by cloning this PR and seeing if it builds.

    git clone https://github.com/ericchiang/dex
    cd dex
    make testall

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 bin
 dist
 _output
+_gopath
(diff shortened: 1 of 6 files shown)


rithujohn191:
Are you missing something in travis script thats causing the build to fail?

ericchiang:
Having trouble with the makefile. Let's hold off for now. Will revisit when I get some time.

PR #855, opened 2017-03-16
storage: make static storages query real storages for some actions

### LINKED ISSUE #735: storage: static storages should also query real storages
For example, if I specify `staticClients` in my config, I should also be able to add clients through the API and union the results.

ericchiang:
If dex is configured with static passwords or clients, let the API
still add or modify objects in the backing storage, so long as
their IDs don't conflict with the static ones. List options now
aggregate resources from the static list and backing storage.

Closes #735

Still needs unit tests.


storage/static.go
```diff
@@ -36,25 +32,50 @@ func (s staticClientsStorage) GetClient(id string) (Client, error) {
...
+			n++
+		}
+	}
+	return append(clients[:n], s.clients...), nil
 }
 
 func (s staticClientsStorage) CreateClient(c Client) error {
-	return errors.New("static clients: read-only cannot create client")
+	if s.isStatic(c.ID) {
+		return errors.New("static clients: read-only cannot create client")
```
rithujohn191: Should we show a different error message that implies that the client already exists as a static client? Similar comment for CreatePassword

storage/memory/static_clients_test.go
```diff
@@ -15,41 +15,83 @@ func TestStaticClients(t *testing.T) {
...
-	s := New(logger)
+	backing := New(logger)
 
 	c1 := storage.Client{ID: "foo", Secret: "foo_secret"}
 	c2 := storage.Client{ID: "bar", Secret: "bar_secret"}
-	s.CreateClient(c1)
-	s2 := storage.WithStaticClients(s, []storage.Client{c2})
+	c3 := storage.Client{ID: "spam", Secret: "spam_secret"}
+
+	backing.CreateClient(c1)
```
rithujohn191: Do we want to add a similar test for Passwords?

storage/static.go
ericchiang: I don't like adding yet another sentinel error :( Maybe we could introduce a `storage.Error` interface instead? Something like

```go
type ErrKind int

const (
    ErrAlreadyExists ErrKind = iota
    ErrNotFound
    ErrNotModifiable
    ErrUnknown
)

type Error interface {
    Kind() ErrKind
}
```

Then use that to do all of this custom detection?

ericchiang: Nice article on these kind of errors here: https://dave.cheney.net/2014/12/24/inspecting-errors

rithujohn191: Yep that would be nice.

ericchiang: Something we want to address in this PR or later? I'm leaning towards later.

rithujohn191: ok

PR #860, opened 2017-03-17
connector/oidc: expose oauth2.RegisterBrokenAuthHeaderProvider

### LINKED ISSUE #859: odic connector: Expose x/oauth2#RegisterBrokenAuthHeaderProvider to clients
Using the OIDC connector with [Okta](https://okta.com) produces the following error 

```console
level=error msg="Failed to authenticate: oidc: failed to get token: oauth2: cannot fetch token: 401 Unauthorized\nResponse: {\"error\":\"invalid_request\",\"error_description\":\"Cannot supply client credentials in both the Authentication header and the request-body.\"}"
```

It looks like dex needs a way to call `oauth2.RegisterBrokenAuthHeaderProvider` before 
[this line](https://github.com/coreos/dex/blob/master/connector/oidc/oidc.go#L105) is invoked. 

#### Addational context 
See https://github.com/golang/oauth2/issues/224

ericchiang:
closes #859

@curtisallen if you can confirm the fix I'll add docs and such.

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -6,6 +6,9 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
+	"strings"
+	"sync"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/coreos/go-oidc"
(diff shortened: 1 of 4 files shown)


ericchiang:
@curtisallen do you have any suggestions for setting up okta as an OpenID Connect provider? I've tried with their free tier, but only see options for SAML and their SOO solution. We'd like to test this internally.

ericchiang:
Basically, I've tried following this doc but it looks outdated or not available on the free tier https://support.okta.com/help/Documentation/Knowledge_Article/Using-OpenID-Connect

curtisallen:
@ericchiang yeah absolutely setting up an  OIDC app is hidden this guide is really useful https://help.okta.com/en/prev/Content/Topics/Apps/Apps_App_Integration_Wizard.htm#OIDCWizard

curtisallen:
You'll basically end up with a app like this 
screen shot 2017-03-17 at 2 00 04 pm

ericchiang:
Yeah that's really weird, I don't see those options on my dev account. Blah.

curtisallen:
Oh strange maybe you have a make a new account or something (perhaps it's out of beta) ¯\_(ツ)_/¯

ericchiang:
@curtisallen @rithujohn191 ptal

rithujohn191:
Do you wanna add an example of its usage here: https://github.com/coreos/dex/blob/master/examples/config-dev.yaml#L58 ?

connector/oidc/oidc.go
```diff
@@ -21,9 +23,34 @@ type Config struct {
...
+	BasicAuthUnsupported *bool `json:"basicAuthUnsupported"`
+
 	Scopes []string `json:"scopes"` // defaults to "profile" and "email"
 }
 
+// Detect auth header provider issues for known providers. This lets users
+// avoid having to explicitly set "basicAuthUnsupported" in their config.
+//
+// Setting the config field always overrides values returned by this function.
+func knownBrokenAuthHeaderProvider(issuerURL string) bool {
```
curtisallen: IMO implementing a known list of broken providers isn't a good idea. This list will grow and grow over time just like it has in [oauth2](https://github.com/golang/oauth2/blob/master/internal/token.go#L92) violating the open close principle. 

I think it's better to rely on clear documentation in this case

curtisallen: conversely, not all IDPs have nice error messages like Okta so things could fail and operators will have no idea why... So I can see both sides of it

ericchiang: If we can detect cases where someone will _always_ have to set the `basicAuthUnsupported` flag then I think we should just avoid the overhead and do it automatically. These providers being weird shouldn't cost our users mental overhead, and maintaining a list isn't that bad.

ericchiang:
@rithujohn191 let me add a note here https://github.com/coreos/dex/blob/master/Documentation/openid-connect.md

ericchiang:
@rithujohn191 docs added.

connector/oidc/oidc.go
```diff
@@ -21,9 +23,34 @@ type Config struct {
...
+	if u, err := url.Parse(issuerURL); err == nil {
+		// Detect Okta.
+		//
+		// https://github.com/coreos/dex/issues/859
+		for _, host := range []string{"okta.com", "oktapreview.com"} {
```
curtisallen: Perhaps this list should be defined separately above?

PR #862, opened 2017-03-17
api: Update timestamp type for RefreshTokenRef to int64.

### LINKED ISSUE #861: api: times should be represented as unix times or proto times, not strings

rithujohn191:
Fixes #861


ericchiang:
@rithujohn191 can we add a test and bump the api version?

rithujohn191:
Added a unit test and incremented the api version to 2

PR #869, opened 2017-03-21
*: validate InResponseTo SAML response field and make issuer optional

ericchiang:


rithujohn191:
Tested this change with Okta :+1. The instructions are great! thanks.

connector/saml/saml_test.go
```diff
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	defaultIssuer      = "http://localhost:5556/dex/callback"
+	defaultIssuer      = "http://www.okta.com/exk91cb99lKkKSYoy0h7"
 	defaultRedirectURI = "http://localhost:5556/dex/callback"
+	responseID         = "_fd1b3ef9-ec09-44a7-a66b-0d39c250f6a0"
```
rithujohn191: Maybe add a note where this value came from?

PR #872, opened 2017-03-23
connector: Connectors without a RefreshConnector should not error out

### LINKED ISSUE #871: Connectors that don't support refresh tokens shouldn't return one instead of erroring
If a connector doesn't implement `connector.RefreshConnector` it should silently not return a refresh token when `offline_access` is requested instead of erroring out, like our current SAML connector.

For backwards compatibility we'll have to make the OIDC connector implement this method, so it continues to support refresh tokens, despite not being able to refresh the users identity (https://github.com/coreos/dex/issues/863).

```
// Refresh is implemented for backwards compatibility, even though it's a no-op.
func (o *oidcConnector) Refresh(ctx context.Context, s connector.Scopes, identity connector.Identity) (connector.Identity, error) {
    return identity, nil
}
```

cc @rithujohn191

rithujohn191:
fixes #871.

Tested this change by running the example-app with okta. The "Creating SAML data: SAML does not support offline access" error is no longer displayed. The example app does not display a refresh token on the final out put page.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -646,6 +646,20 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 	}
 
 	reqRefresh := func() bool {
+		// Ensure the connector supports refresh tokens.
+		//
+		// Connectors like `samlExperimental` do not implement RefreshConnector.
+		conn, ok := s.connectors[authCode.ConnectorID]
+		if !ok {
+			s.logger.Errorf("connector ID not found: %q", authCode.ConnectorID)
+			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
+			return false
+		}
+		_, ok = conn.Connector.(connector.RefreshConnector)
+		if !ok {
+			return false
+		}
+
 		for _, scope := range authCode.Scopes {
 			if scope == scopeOfflineAccess {
 				return true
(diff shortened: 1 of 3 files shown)


server/handlers.go
```diff
@@ -647,7 +647,15 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 
 	reqRefresh := func() bool {
 		for _, scope := range authCode.Scopes {
-			if scope == scopeOfflineAccess {
+			conn, ok := s.connectors[authCode.ConnectorID]
```
ericchiang: Can you add a comment here?

ericchiang: Also this check needs to happen outside the for loop :)

rithujohn191: Done

server/handlers.go
```diff
@@ -646,8 +646,19 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
+		// Ensure the connector supports refresh tokens.
+		//
+		// Connectors like `samlExperimental` do not implement RefreshConnector.
+		conn, ok := s.connectors[authCode.ConnectorID]
+		if !ok {
+			s.logger.Errorf("connector ID not found: %q", authCode.ConnectorID)
+			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
+			return false
+		}
+		_, ok = conn.Connector.(connector.RefreshConnector)
```
ericchiang: You can short circuit this.

```
if !ok {
    return false
}
```

This avoids the loop in some cases.

rithujohn191: Done

PR #877, opened 2017-03-27
re-enable deep linking of connectors

ibloat:
It was possible to directly link to connectors in v1, continuing authentication directly with the connector indicated by the connector_id parameter.
Unless there was a reason to remove that feature it would be nice to have it back.

Since the other two redirects had the same comment I c/p it over.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -177,6 +177,13 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 	}
 
 	connectorInfos := make([]connectorInfo, len(s.connectors))
+	if conn, ok := s.connectors[r.Form.Get("connector_id")]; ok {
+		// TODO(ericchiang): Make this pass on r.URL.RawQuery and let something latter
+		// on create the auth request.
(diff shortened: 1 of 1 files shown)


ericchiang:
@ibloat out of curiosity why do you need this feature? I don't want to add features just because they were in v1.

ericchiang:
Basically this seems like a really hacky solution. Why not associate this information with the client itself, or just run multiple dex instances?

ibloat:
The goal is to bypass the connector selection page (let the user select the connector within the app, outside dex)
Currently dex already has two behaviours, showing the selection page if there is more than one connector, otherwise directly redirecting to the sole connector.
Adding the possibility to explicitly specify the provider does not seem to introduce any new flow from the end-user perspective but requires knowledge of connector IDs on part of the client.
Running multiple dex instances each with a single provider will work fine for our use case as we plan on having only three connectors enabled but maybe a bit cumbersome if one enabled 20 connectors.
Is enabling certain connectors for certain clients a desired/planned feature or how would you go about associating this information with the client?

We will go with multiple single-connector dex instances as it seems the most straight-forward, thank you for that suggestion.
Bypassing connector selection might be too niche to justify adding that 'feature' back in especially if it wasn't really intended the first time around, closing this PR.

PR #879, opened 2017-03-28
api: update message definition numbered tag value

rithujohn191:
protocol buffers expect the tag numbers for attributes to be continuous.


rithujohn191:
This is not an issue!

PR #882, opened 2017-03-28
api: add gRPC calls to create, list and delete connectors.

rithujohn191:
This change defines 4 new API calls: 
* CreateConnector
* UpdateConnector
* ListConnectors
* DeleteConnector

Also added unit tests for each.


rithujohn191:
Added the following changes:
* Added static storage support. The idea is that the connectors that are in the Dex ConfigMap will be treated as static resources which cannot be modified in anyway.
* The server list of connectors must have both static connectors and regular storage connectors (which can be manipulated via API calls)
* Unit tests to test static connector behavior.

I have not changes out the server struct to have connectors of type Storage.Connector yet. Let me know if you would still like me to make that switch.

ericchiang:
FYI this still needs to have the server dynamically query the storage when listing connectors (e.g. for the login page) or when a user hits a specific connector.

I think that's a prerequisite for adding these API calls, because without that, the API calls don't do anything until you restart the server.

rithujohn191:
Yes. Working on those changes now.

rithujohn191:
#913 is a pre-req for this PR

rithujohn191:
Breaking this up into smaller PRs

PR #887, opened 2017-03-30
server/api.go: avoid invalid memory address error when checking API response.

rithujohn191:
If the user explicitly checks for the `AlreadyExists` or `NotFound` error, it errors out with "runtime error: invalid memory address or nil pointer dereference" since we do not set these values to false if no error occurred and instead return an empty struct.

This change sets these attributes to false if no error was encountered.

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -64,7 +64,8 @@ func (d dexAPI) CreateClient(ctx context.Context, req *api.CreateClientReq) (*ap
 	}
 
 	return &api.CreateClientResp{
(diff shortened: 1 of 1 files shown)


ericchiang:
@rithujohn191 do have a test to demonstrate this? Those values default to false, setting them explicitly shouldn't be necessary.

rithujohn191:
Sorry wrong diagnoses of a separate issue. Closing this out.

PR #888, opened 2017-04-03
Add userinfo_endpoint provider metadata to discovery endpoint 

wiliamsouza:
This adds userinfo_endpoint provider metadata to discovery endpoint .well-known/openid-configuration.
 and  make it editable in config file.

This is used for some smart OpenID connect authentication to provide user data profile or infomations.

https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -47,7 +47,8 @@ type Config struct {
 	// StaticPasswords cause the server use this list of passwords rather than
 	// querying the storage. Cannot be specified without enabling a passwords
 	// database.
-	StaticPasswords []password `json:"staticPasswords"`
+	StaticPasswords  []password `json:"staticPasswords"`
+	UserInfoEndpoint string     `json:"userInfoEndpoint"`
 }
 
 type password storage.Password
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -91,16 +91,17 @@ func (s *Server) handlePublicKeys(w http.ResponseWriter, r *http.Request) {
 }
 
 type discovery struct {
-	Issuer        string   `json:"issuer"`
-	Auth          string   `json:"authorization_endpoint"`
-	Token         string   `json:"token_endpoint"`
-	Keys          string   `json:"jwks_uri"`
-	ResponseTypes []string `json:"response_types_supported"`
-	Subjects      []string `json:"subject_types_supported"`
-	IDTokenAlgs   []string `json:"id_token_signing_alg_values_supported"`
-	Scopes        []string `json:"scopes_supported"`
-	AuthMethods   []string `json:"token_endpoint_auth_methods_supported"`
-	Claims        []string `json:"claims_supported"`
+	Issuer           string   `json:"issuer"`
+	Auth             string   `json:"authorization_endpoint"`
+	Token            string   `json:"token_endpoint"`
+	Keys             string   `json:"jwks_uri"`
+	ResponseTypes    []string `json:"response_types_supported"`
+	Subjects         []string `json:"subject_types_supported"`
+	IDTokenAlgs      []string `json:"id_token_signing_alg_values_supported"`
+	Scopes           []string `json:"scopes_supported"`
+	AuthMethods      []string `json:"token_endpoint_auth_methods_supported"`
+	Claims           []string `json:"claims_supported"`
+	UserInfoEndpoint string   `json:"userinfo_endpoint"`
 }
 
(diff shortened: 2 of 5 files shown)


server/handlers.go
```diff
@@ -91,16 +91,17 @@ func (s *Server) handlePublicKeys(w http.ResponseWriter, r *http.Request) {
...
+	Auth             string   `json:"authorization_endpoint"`
+	Token            string   `json:"token_endpoint"`
+	Keys             string   `json:"jwks_uri"`
+	ResponseTypes    []string `json:"response_types_supported"`
+	Subjects         []string `json:"subject_types_supported"`
+	IDTokenAlgs      []string `json:"id_token_signing_alg_values_supported"`
+	Scopes           []string `json:"scopes_supported"`
+	AuthMethods      []string `json:"token_endpoint_auth_methods_supported"`
+	Claims           []string `json:"claims_supported"`
+	UserInfoEndpoint string   `json:"userinfo_endpoint"`
```
wiliamsouza: Maybe add `omitempty` here?

cmd/dex/config.go
```diff
@@ -47,7 +47,8 @@ type Config struct {
 	// StaticPasswords cause the server use this list of passwords rather than
 	// querying the storage. Cannot be specified without enabling a passwords
 	// database.
-	StaticPasswords []password `json:"staticPasswords"`
+	StaticPasswords  []password `json:"staticPasswords"`
+	UserInfoEndpoint string     `json:"userInfoEndpoint"`
```
rithujohn191: Plz add a comment to describe the UserInfoEndpoint. A link to https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata would also be helpful.

server/handlers.go
rithujohn191: Not necessary since it is only a "recommended"  attribute.

ericchiang:
@rithujohn191 we don't implement this feature https://github.com/coreos/dex/issues/376 Why would we return it in the metadata?

rithujohn191:
I was under the impression that the claims in ID Token could serve the info needed. Sorry about that. That would make https://github.com/coreos/dex/issues/376 a pre-requisite for this PR.

wiliamsouza:
This metadata is useful when you already have this endpoint implemented as other service. For now I'm solving this in other place https://github.com/ByteInternet/drf-oidc-auth/pull/20.
To make it clear the use case for userinfo_endpoint is.

```
                                                  +------------------------------------------+
                                                  |      populate OAuth clients              |
                                                  |      users login and passwords           |
                                                  |                                   +------+------+
                                                  |                                   |             |
                                                  |                                   | sellers api <----------+
                                 2       +--------v-------+                           |             |          |
                          request token  |                |                           +-------------+          |
                        +----------------> openid connect |                                                    |
                        |                |                <------------------------------------------------+   |
                        |                +-----+---^------+              7 validate id_token               |   |
                        |                      |   |                       with oidc provider              |   |
                        |                      |   |                                                       |   |         8
              1         |                      |   |                                                       |   | retrieve userinfo
            authN       |             3        |   |                                                       |   | from users rest api
+--------+  authZ  +----+----+      receive    |   |    5                             +--------------+     |   |
|        |         |         |   token and jwt |   | validate                    jwt  |              +----->   |
| seller +--------->  webapp <-----------------+   | token                     +------> products api |     |   |
|        |         |         |                     |                           |      |              +--------->
+--------+         +----+----+                     |                           |      +--------------+     |   |
                        |                          |                           |                           |   |
                        |                          |                           |                           |   |
                        |                          |                           |                           |   |
                        |                          |                           |      +------------+       |   |
                        |                          |                           | jwt  |            +-------+   |
                        |                          |                           +------> orders api |           |
                        |                          |                           |      |            +-----------+
                        |                          |                           |      +------------+
                        |        4      +----------+--+                        |
                        |       use     |             |           6            |
                        +---------------> api gateway +------------------------>
                          token and jwt |             |
                                        +-------------+
                                    Do access_token offload
                                    bypass id_token only

```
I'm in doubt what to do with this PR. feel free to close it.

rithujohn191:
We have been getting a lot of requests to support the user info endpoint but the general idea seems to be that users want this feature so that they can avoid the use of tokens. Which might not be the best practice in terms of security.

I will go ahead and close out this PR for now until we make a final decision regarding the user info endpoint. Thanks again for sending this in.

wiliamsouza:
In our case we use the same JWT token to get userinfo from endpoint. When you make a decision let me know I would love to help.

PR #898, opened 2017-04-07
connector/saml: clean up SAML verification logic and comments

ericchiang:
Intention is to make this code a little more understandable. Clean up some loops, add comments, etc.

Will add a bunch of tests tomorrow.

cc @rithujohn191 @squat

### DIFF
--- a/connector/saml/saml.go
+++ b/connector/saml/saml.go
@@ -2,10 +2,8 @@
 package saml
 
 import (
-	"crypto/rand"
 	"crypto/x509"
 	"encoding/base64"
-	"encoding/hex"
 	"encoding/pem"
 	"encoding/xml"
 	"errors"
@@ -270,12 +268,22 @@ func (p *provider) POSTData(s connector.Scopes, id string) (action, value string
 	return p.ssoURL, base64.StdEncoding.EncodeToString(data), nil
 }
 
+// HandlePOST interprets a request from a SAML provider attempting to verify a
+// user's identity.
+//
+// The steps taken are:
+//
+// * Verify signature on XML document (or verify sig on assertion elements).
+// * Verify various parts of the Assertion element. Conditions, audience, etc.
+// * Map the Assertion's attribute elements to user info.
+//
 func (p *provider) HandlePOST(s connector.Scopes, samlResponse, inResponseTo string) (ident connector.Identity, err error) {
 	rawResp, err := base64.StdEncoding.DecodeString(samlResponse)
 	if err != nil {
 		return ident, fmt.Errorf("decode response: %v", err)
 	}
 
+	// Root element is allowed to not be signed if the Assertion element is.
 	rootElementSigned := true
 	if p.validator != nil {
 		rawResp, rootElementSigned, err = verifyResponseSig(p.validator, rawResp)
@@ -289,6 +297,8 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse, inResponseTo str
 		return ident, fmt.Errorf("unmarshal response: %v", err)
 	}
 
+	// If the root element isn't signed, there's no reason to inspect these
+	// elements. They're not verified.
 	if rootElementSigned {
 		if p.ssoIssuer != "" && resp.Issuer != nil && resp.Issuer.Issuer != p.ssoIssuer {
 			return ident, fmt.Errorf("expected Issuer value %s, got %s", p.ssoIssuer, resp.Issuer.Issuer)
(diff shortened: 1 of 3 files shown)


connector/saml/saml.go
```diff
@@ -270,12 +268,22 @@ func (p *provider) POSTData(s connector.Scopes, id string) (action, value string
...
 }
 
+// HandlePOST interprets a request from a SAML provider attempting to verify a
+// user's identity.
+//
+// The steps taken are:
+//
+// * Verify signature on XML document (or verify sig on assertion elements).
+// * Verify various parts of the Assertion element. Conditions, audience, etc.
+// * Map the Assertion's attribute elements to
```
rithujohn191: Incomplete sentence.

connector/saml/saml.go
```diff
@@ -315,16 +329,25 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse, inResponseTo str
 	if assertion == nil {
 		return ident, fmt.Errorf("response did not contain an assertion")
 	}
+
+	// Even though subject is optional, without it we can pull the user ID,
```
rithujohn191: "Even though subject is optional, without it we can --> cannot pull the user ID,"

connector/saml/saml.go
```diff
@@ -336,53 +359,57 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse, inResponseTo str
 		return ident, fmt.Errorf("subject does not contain an NameID element")
 	}
 
+	// After verifying the assertion, map data in the attribute statements to
+	// varous user info.
```
rithujohn191: varous --> various?

connector/saml/saml.go
```diff
@@ -396,96 +423,105 @@ func (p *provider) validateStatus(resp *response) error {
 	return nil
 }
 
-// Multiple subject SubjectConfirmation can be in the assertion
-// and at least one SubjectConfirmation must be valid.
+// validateSubject ensure the response is to the request we expect.
```
rithujohn191: ensure --> ensures

connector/saml/saml.go
```diff
@@ -396,96 +423,105 @@ func (p *provider) validateStatus(resp *response) error {
...
+			return nil
+		}()
+		if err == nil {
+			return nil
 		}
+		errs = append(errs, err)
 	}
-	if !validSubjectConfirmation {
-		return fmt.Errorf("no valid SubjectConfirmation was found on this Response")
+	if len(errs) == 1 {
```
rithujohn191: Can this value never be greater than 1?

connector/saml/saml_test.go
```diff
@@ -355,152 +354,3 @@ func TestVerifySignedMessageAndSignedAssertion(t *testing.T) {
 func TestVerifyUnsignedMessageAndUnsignedAssertion(t *testing.T) {
 	runVerify(t, "testdata/idp-cert.pem", "testdata/idp-resp.xml", false)
 }
```
rithujohn191: Are we planning to add other tests in its place?

connector/saml/types.go
```diff
@@ -115,7 +116,7 @@ type conditions struct {
 	NotBefore    xmlTime `xml:"NotBefore,attr,omitempty"`
 	NotOnOrAfter xmlTime `xml:"NotOnOrAfter,attr,omitempty"`
 
-	AudienceRestriction *audienceRestriction `xml:"AudienceRestriction,omitempty"`
+	AudienceRestriction []*audienceRestriction `xml:"AudienceRestriction,omitempty"`
```
rithujohn191: Is this one not meant to be an "attr"?

connector/saml/saml.go
ericchiang: it's either going to be 1 or greater than 1.

connector/saml/saml_test.go
ericchiang: yes

connector/saml/types.go
ericchiang: It's an element, not an attribute. So the XML will look like

```xml
<SubjectConfirmationData InResponseTo="foo">
  <AudienceRestriction>
  </AudienceRestriction>
  <AudienceRestriction>
  </AudienceRestriction>
</SubjectConfirmationData>
```

(with a lot more namespace annotations)

connector/saml/saml.go
rithujohn191: Should we change the condition to if len(errs) >= 1

connector/saml/types.go
rithujohn191: ah ok got it

connector/saml/saml.go
ericchiang: that's handled by the next return. This is just saying "if there's only one element in the slice, return that element, rather than the entire slice."

rithujohn191: :+1:

ericchiang:
@rithujohn191 tests added.

PR #901, opened 2017-04-10
connector/github: add support for github enterprise.

rithujohn191:
Fixes https://github.com/coreos/dex/issues/622.

This change enables Dex to support login via GitHub enterprise. Users have to include the hostname of the GitHub enterprise account in the Dex ConfigMap.

This change has been tested with GitHub Enterprise edition.

### DIFF
--- a/Documentation/github-connector.md
+++ b/Documentation/github-connector.md
@@ -31,4 +31,37 @@ connectors:
     org: my-oranization
 ```
 
+## GitHub Enterprise
+
+Users can use their GitHub Enterprise account to login to dex. The following configuration can be used to enable a GitHub Enterprise connector on dex:
+
+```yaml
+connectors:
+- type: github
+  # Required field for connector id.
+  id: github
+  # Required field for connector name.
+  name: GitHub
+  config:
+    # Required fields. Dex must be pre-registered with GitHub Enterprise
+    # to get the following values.
+    # Credentials can be string literals or pulled from the environment.
+    clientID: $GITHUB_CLIENT_ID
+    clientSecret: $GITHUB_CLIENT_SECRET
+    redirectURI: http://127.0.0.1:5556/dex/callback
+    # Optional organization to pull teams from, communicate through the
+    # "groups" scope.
+    #
+    # NOTE: This is an EXPERIMENTAL config option and will likely change.
+    org: my-oranization
+
+    # Required ONLY for GitHub Enterprise.
+    # This is the Hostname of the GitHub Enterprise account listed on the
+    # management console. Ensure this domain is routable on your network.
+    hostName: git.example.com
+    # ONLY for GitHub Enterprise. Optional field.
+    # Used to support self-signed or untrusted CA root certificates.
+    rootCA: /etc/dex/ca.crt
+```
+
 [github-oauth2]: https://github.com/settings/applications/new
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -3,13 +3,18 @@ package github
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"regexp"
 	"strconv"
+	"strings"
+	"time"
 
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/github"
@@ -19,7 +24,7 @@ import (
 )
 
 const (
-	baseURL    = "https://api.github.com"
+	apiURL     = "https://api.github.com"
 	scopeEmail = "user:email"
 	scopeOrgs  = "read:org"
 )
@@ -30,17 +35,45 @@ type Config struct {
 	ClientSecret string `json:"clientSecret"`
 	RedirectURI  string `json:"redirectURI"`
 	Org          string `json:"org"`
+	HostName     string `json:"hostName"`
+	RootCA       string `json:"rootCA"`
 }
 
 // Open returns a strategy for logging in through GitHub.
 func (c *Config) Open(logger logrus.FieldLogger) (connector.Connector, error) {
-	return &githubConnector{
+	g := githubConnector{
 		redirectURI:  c.RedirectURI,
 		org:          c.Org,
 		clientID:     c.ClientID,
 		clientSecret: c.ClientSecret,
+		apiURL:       apiURL,
 		logger:       logger,
-	}, nil
+	}
+
+	if c.HostName != "" {
+		// ensure this is a hostname and not a URL or path.
+		if strings.Contains(c.HostName, "/") {
+			return nil, errors.New("invalid hostname: hostname cannot contain `/`")
+		}
+
+		g.hostName = c.HostName
+		g.apiURL = "https://" + c.HostName + "/api/v3"
+	}
+
+	if c.RootCA != "" {
+		if c.HostName == "" {
+			return nil, errors.New("invalid connector config: Host name field required for a root certificate file")
+		}
+		g.rootCA = c.RootCA
+
+		var err error
+		if g.httpClient, err = newHTTPClient(g.rootCA); err != nil {
+			return nil, fmt.Errorf("failed to create HTTP client: %v", err)
+		}
+
+	}
+
+	return &g, nil
 }
 
 type connectorData struct {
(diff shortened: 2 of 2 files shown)


connector/github/github.go
```diff
@@ -95,15 +134,52 @@ func (e *oauth2Error) Error() string {
...
+		Transport: &http.Transport{
+			TLSClientConfig: &tlsConfig,
+			Proxy:           http.ProxyFromEnvironment,
+			Dial: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}).Dial,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+		},
```
rithujohn191: Are the above default timeouts ok?

ericchiang: I generally use whatever the default transport supplies https://golang.org/pkg/net/http/#RoundTripper

Documentation/github-connector.md
```diff
@@ -31,4 +31,36 @@ connectors:
...
+    # Required fields. Dex must be pre-registered with GitHub Enterprise
+    # to get the following values.
+    # Credentials can be string literals or pulled from the environment.
+    clientID: $GITHUB_CLIENT_ID
+    clientSecret: $GITHUB_CLIENT_SECRET
+    redirectURI: http://127.0.0.1:5556/dex/callback
+    # Required for GitHub Enterprise.
+    # This is the Hostname of the GitHub Enterprise account listed on the
+    # management console. Ensure this domain is routable on your network.
+    hostName: git.example.com
```
ericchiang: Can you separate these from the other fields to stress that it's optional and only needs to be used for enterprise? E.g. put this after "org" and with an empty newline in between.

connector/github/github.go
```diff
@@ -30,17 +35,34 @@ type Config struct {
...
 		org:          c.Org,
 		clientID:     c.ClientID,
 		clientSecret: c.ClientSecret,
+		apiURL:       apiURL,
 		logger:       logger,
-	}, nil
+	}
+
+	if c.HostName != "" {
+		g.hostName = c.HostName
```
ericchiang: Can we enforce this is a hostname and not a URL?

connector/github/github.go
```diff
@@ -95,15 +134,52 @@ func (e *oauth2Error) Error() string {
...
 		return identity, &oauth2Error{errType, q.Get("error_description")}
 	}
 
 	oauth2Config := c.oauth2Config(s)
-	ctx := r.Context()
 
+	var ctx context.Context
+	// GitHub Enterprise account
+	if c.hostName != "" && c.rootCA != "" {
+		httpClient, err := newHTTPClient(c.rootCA)
```
ericchiang: Create this client at init, not on each request.

connector/github/github.go
```diff
@@ -95,15 +134,52 @@ func (e *oauth2Error) Error() string {
...
+		if err != nil {
+			return identity, fmt.Errorf("failed to create HTTP client: %v", err)
+		}
+		ctx = context.WithValue(r.Context(), oauth2.HTTPClient, httpClient)
+	} else {
+		ctx = r.Context()
+	}
+
+	log.Printf("Response: q = %v", q)
```
ericchiang: nit: need to remove this.

connector/github/github.go
rithujohn191: as in add a comment to highlight it?

ericchiang: Like, ensure it doesn't have a schema or a path.

connector/github/github.go
rithujohn191: Done

connector/github/github.go
rithujohn191: Done

Documentation/github-connector.md
rithujohn191: Done

connector/github/github.go
rithujohn191: Done. If url.ParseRequestURI executes without an error report a problem.

connector/github/github.go
```diff
@@ -30,17 +33,46 @@ type Config struct {
...
+		apiURL:       apiURL,
 		logger:       logger,
-	}, nil
+	}
+
+	if c.HostName != "" {
+		// make sure this is a hostname and not a URL.
+		// Ensure it doesn't have a schema or a path.
+		if _, err := url.ParseRequestURI(c.HostName); err == nil {
+			return &g, errors.New("invalid hostname: URL provided instead of hostname")
```
ericchiang: nit: always return nil for error cases.

connector/github/github.go
```diff
@@ -95,14 +146,39 @@ func (e *oauth2Error) Error() string {
...
+	rootCABytes, err := ioutil.ReadFile(rootCA)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read root-ca: %v", err)
+	}
+	if !tlsConfig.RootCAs.AppendCertsFromPEM(rootCABytes) {
+		return nil, fmt.Errorf("no certs found in root CA file %q", rootCA)
+	}
+
+	return &http.Client{
+		Transport: &http.Transport{
```
ericchiang: This actually doesn't use the default values of the http.Transport. You probably need to copy the fields directly.

connector/github/github.go
```diff
@@ -95,14 +146,39 @@ func (e *oauth2Error) Error() string {
...
 func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (identity connector.Identity, err error) {
 	q := r.URL.Query()
 	if errType := q.Get("error"); errType != "" {
 		return identity, &oauth2Error{errType, q.Get("error_description")}
 	}
 
 	oauth2Config := c.oauth2Config(s)
-	ctx := r.Context()
+
+	var ctx context.Context
```
ericchiang: nit: you can avoid the "else" condition by initializing this with the "else" value.

```
ctx := r.Context()
if c.httpClient != nil {
    ctx = context.WithValue(ctx, oauth2.HTTPClient, c.httpClient)
}
```

connector/github/github.go
```diff
@@ -68,10 +108,20 @@ func (c *githubConnector) oauth2Config(scopes connector.Scopes) *oauth2.Config {
 	} else {
 		githubScopes = []string{scopeEmail}
 	}
+
+	var endpoint oauth2.Endpoint
```
ericchiang: nit: you can avoid the "else" condition by initializing this with the "else" value.

Documentation/github-connector.md
```diff
@@ -31,4 +31,37 @@ connectors:
...
+    clientID: $GITHUB_CLIENT_ID
+    clientSecret: $GITHUB_CLIENT_SECRET
+    redirectURI: http://127.0.0.1:5556/dex/callback
+    # Optional organization to pull teams from, communicate through the
+    # "groups" scope.
+    #
+    # NOTE: This is an EXPERIMENTAL config option and will likely change.
+    org: my-oranization
+
+    # Required ONLY for GitHub Enterprise.
```
ericchiang: :+1:

rithujohn191:
Addressed all the feedback comments.

connector/github/github.go
```diff
@@ -30,17 +35,46 @@ type Config struct {
...
 		clientSecret: c.ClientSecret,
+		apiURL:       apiURL,
 		logger:       logger,
-	}, nil
+	}
+
+	if c.HostName != "" {
+		// make sure this is a hostname and not a URL.
+		// Ensure it doesn't have a schema or a path.
+		if _, err := url.ParseRequestURI(c.HostName); err == nil {
```
ericchiang: Why this over url.Parse?

ericchiang: Also in 1.7 I think this has bad error handling. E.g. it'll incorrectly report "google.com" as invalid.

ericchiang: Maybe just

```
// Maybe a URL and not a hostname?
if strings.Contains(c.Hostname, "/") {
    // not a hostname
}
```

rithujohn191: for ParseRequestURI the string rawurl is assumed not to have a #fragment suffix. Should we account for that as well?

ericchiang: I think `url.Parse` and `url.ParseRequestURI` are the wrong tools here since they can both produce false positives (e.g. say something is a URL when it isnt').

Ideally we'd have a `IsHostname` function. But lacking that, we should be conservative about the validation.

rithujohn191: Ok will go ahead and switch this out for strings.Contains(c.Hostname, "/")

PR #902, opened 2017-04-11
*: promote SAML to stable

ericchiang:
This means we no longer refer to it as "experimental" and wont make
breaking changes.


ericchiang:
@rithujohn191 thanks for catching that. updated.

PR #903, opened 2017-04-11
connector/ldap: fix case where groups are listed on the user entity

ericchiang:
Support schemas that determine membership by having fields on the
user entity, instead of listing users on a groups entity. E.g. the
following schema is now supported when it wasn't previously:

    cn=eric,cn=user,dn=example,dn=com
    objectClass=myPerson
    cn: eric
    uid: eric
    email: eric@example.com
    memberOf: foo
    memberOf: bar

    cn=foo,cn=group,dn=example,dn=com
    objectClass=myGroup
    cn: foo

    cn=bar,cn=group,dn=example,dn=com
    objectClass=myGroup
    cn: bar

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -256,18 +256,22 @@ func (c *ldapConnector) do(ctx context.Context, f func(c *ldap.Conn) error) erro
 	return f(conn)
 }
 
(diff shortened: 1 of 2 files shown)


ericchiang:
> Were you able to test this with OneLogin?

No. No account.

PR #910, opened 2017-04-13
*: update grpc and correct protobuf generation

ericchiang:
Turns out that manually setting PATH in the Makefile doesn't work
so we've been using the protobuf plugins installed on the host. Fix
this by specifying plugins by path.


Makefile
```diff
@@ -16,10 +16,6 @@ user=$(shell id -u -n)
 group=$(shell id -g -n)
 
 export GOBIN=$(PWD)/bin
-# Prefer ./bin instead of system packages for things like protoc, where we want
-# to use the version dex uses, not whatever a developer has installed.
-export PATH=$(GOBIN):$(shell printenv PATH)
-export GO15VENDOREXPERIMENT=1
```
squat: about time :stuck_out_tongue:

Makefile
```diff
@@ -74,10 +70,10 @@ docker-image: clean-release _output/bin/dex
 proto: api/api.pb.go server/internal/types.pb.go
 
 api/api.pb.go: api/api.proto bin/protoc bin/protoc-gen-go
-	@protoc --go_out=plugins=grpc:. api/*.proto
+	@./bin/protoc --go_out=plugins=grpc:. --plugin=protoc-gen-go=./bin/protoc-gen-go api/*.proto
```
squat: it would make sense to replace occurrences of `./bin` with `$(GOBIN)` instead of hardcoding this path in more places

Makefile
```diff
@@ -74,10 +70,10 @@ docker-image: clean-release _output/bin/dex
...
 
 server/internal/types.pb.go: server/internal/types.proto bin/protoc bin/protoc-gen-go
-	@protoc --go_out=. server/internal/*.proto
+	@./bin/protoc --go_out=. --plugin=protoc-gen-go=./bin/protoc-gen-go server/internal/*.proto
```
squat: here too

Makefile
ericchiang: If we did do that, I think we'd prefer to have a variable like

```
PROTOC=$(GOBIN)/protoc
```

not super opinionated it now. I think it's fine this way if that's okay.

squat: I think it's OK either way. $(GOBIN) would be sufficient IMO since we'd be installing protoc relative to GOBIN either way so $(GOBIN)/protoc should always be safe. but, LGTM either way

PR #913, opened 2017-04-17
server: account for dynamically changing connector object in storage.

rithujohn191:
This change enables connector objects to be dynamically modifiable. The connectors defined by the user in the Dex ConfigMap will be deemed as static connectors and will be 'read-only'. 

Dynamic connectors can be created, updated or deleted via gRPC API calls from a dex client (api calls yet to be implemented). This essentially means that every time the dex server refers to a connector object it needs to ensure that it has retrieved the latest version of this connector object from the storage.

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -9,13 +9,6 @@ import (
 	"github.com/Sirupsen/logrus"
 	"golang.org/x/crypto/bcrypt"
 
-	"github.com/coreos/dex/connector"
-	"github.com/coreos/dex/connector/github"
-	"github.com/coreos/dex/connector/gitlab"
-	"github.com/coreos/dex/connector/ldap"
-	"github.com/coreos/dex/connector/mock"
-	"github.com/coreos/dex/connector/oidc"
-	"github.com/coreos/dex/connector/saml"
 	"github.com/coreos/dex/server"
 	"github.com/coreos/dex/storage"
 	"github.com/coreos/dex/storage/kubernetes"
@@ -25,16 +18,19 @@ import (
 
 // Config is the config format for the main application.
 type Config struct {
-	Issuer     string      `json:"issuer"`
-	Storage    Storage     `json:"storage"`
-	Connectors []Connector `json:"connectors"`
-	Web        Web         `json:"web"`
-	OAuth2     OAuth2      `json:"oauth2"`
-	GRPC       GRPC        `json:"grpc"`
-	Expiry     Expiry      `json:"expiry"`
-	Logger     Logger      `json:"logger"`
+	Issuer  string  `json:"issuer"`
+	Storage Storage `json:"storage"`
+	Web     Web     `json:"web"`
+	OAuth2  OAuth2  `json:"oauth2"`
+	GRPC    GRPC    `json:"grpc"`
+	Expiry  Expiry  `json:"expiry"`
+	Logger  Logger  `json:"logger"`
 
 	Frontend server.WebConfig `json:"frontend"`
+
+	// StaticConnectors are user defined connectors specified in the ConfigMap
+	// Write operations, like updating a connector, will fail.
+	StaticConnectors []Connector `json:"connectors"`
 
 	// StaticClients cause the server to use this list of clients rather than
 	// querying the storage. Write operations, like creating a client, will fail.
@@ -170,24 +166,7 @@ type Connector struct {
 	Name string `json:"name"`
 	ID   string `json:"id"`
 
-	Config ConnectorConfig `json:"config"`
-}
-
-// ConnectorConfig is a configuration that can open a connector.
-type ConnectorConfig interface {
-	Open(logrus.FieldLogger) (connector.Connector, error)
-}
-
-var connectors = map[string]func() ConnectorConfig{
-	"mockCallback": func() ConnectorConfig { return new(mock.CallbackConfig) },
-	"mockPassword": func() ConnectorConfig { return new(mock.PasswordConfig) },
-	"ldap":         func() ConnectorConfig { return new(ldap.Config) },
-	"github":       func() ConnectorConfig { return new(github.Config) },
-	"gitlab":       func() ConnectorConfig { return new(gitlab.Config) },
-	"oidc":         func() ConnectorConfig { return new(oidc.Config) },
-	"saml":         func() ConnectorConfig { return new(saml.Config) },
-	// Keep around for backwards compatibility.
-	"samlExperimental": func() ConnectorConfig { return new(saml.Config) },
+	Config server.ConnectorConfig `json:"config"`
 }
 
 // UnmarshalJSON allows Connector to implement the unmarshaler interface to
@@ -203,7 +182,7 @@ func (c *Connector) UnmarshalJSON(b []byte) error {
 	if err := json.Unmarshal(b, &conn); err != nil {
 		return fmt.Errorf("parse connector: %v", err)
 	}
-	f, ok := connectors[conn.Type]
+	f, ok := server.ConnectorsConfig[conn.Type]
 	if !ok {
 		return fmt.Errorf("unknown connector type %q", conn.Type)
 	}
@@ -222,6 +201,21 @@ func (c *Connector) UnmarshalJSON(b []byte) error {
 		Config: connConfig,
 	}
 	return nil
+}
+
+// ToStorageConnector converts an object to storage connector type.
+func ToStorageConnector(c Connector) (storage.Connector, error) {
+	data, err := json.Marshal(c.Config)
+	if err != nil {
+		return storage.Connector{}, fmt.Errorf("failed to marshal connector config: %v", err)
+	}
+
+	return storage.Connector{
+		ID:     c.ID,
+		Type:   c.Type,
+		Name:   c.Name,
+		Config: data,
+	}, nil
 }
 
 // Expiry holds configuration for the validity period of components.
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -74,7 +74,6 @@ func serve(cmd *cobra.Command, args []string) error {
 		errMsg string
 	}{
 		{c.Issuer == "", "no issuer specified in config file"},
-		{len(c.Connectors) == 0 && !c.EnablePasswordDB, "no connectors supplied in config file"},
 		{!c.EnablePasswordDB && len(c.StaticPasswords) != 0, "cannot specify static passwords without enabling password db"},
 		{c.Storage.Config == nil, "no storage suppied in config file"},
 		{c.Web.HTTP == "" && c.Web.HTTPS == "", "must supply a HTTP/HTTPS  address to listen on"},
@@ -128,34 +127,6 @@ func serve(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	connectors := make([]server.Connector, len(c.Connectors))
-	for i, conn := range c.Connectors {
-		if conn.ID == "" {
-			return fmt.Errorf("invalid config: no ID field for connector %d", i)
-		}
-		if conn.Config == nil {
-			return fmt.Errorf("invalid config: no config field for connector %q", conn.ID)
-		}
-		if conn.Name == "" {
-			return fmt.Errorf("invalid config: no Name field for connector %q", conn.ID)
-		}
-		logger.Infof("config connector: %s", conn.ID)
-
-		connectorLogger := logger.WithField("connector", conn.Name)
-		c, err := conn.Config.Open(connectorLogger)
-		if err != nil {
-			return fmt.Errorf("failed to create connector %s: %v", conn.ID, err)
-		}
-		connectors[i] = server.Connector{
-			ID:          conn.ID,
-			DisplayName: conn.Name,
-			Connector:   c,
-		}
-	}
-	if c.EnablePasswordDB {
-		logger.Infof("config connector: local passwords enabled")
-	}
-
 	s, err := c.Storage.Config.Open(logger)
 	if err != nil {
 		return fmt.Errorf("failed to initialize storage: %v", err)
@@ -176,6 +147,35 @@ func serve(cmd *cobra.Command, args []string) error {
 		s = storage.WithStaticPasswords(s, passwords)
 	}
 
+	if c.EnablePasswordDB {
+		c.StaticConnectors = append(c.StaticConnectors, Connector{
+			ID:   server.LocalConnector,
+			Name: "Email",
+			Type: server.LocalConnector,
+		})
+		logger.Infof("config connector: local passwords enabled")
+	}
+
+	storageConnectors := make([]storage.Connector, len(c.StaticConnectors))
+	for i, c := range c.StaticConnectors {
+		if c.ID == "" || c.Name == "" || c.Type == "" {
+			return fmt.Errorf("invalid config: ID, Type and Name fields are required for a connector")
+		}
+		if c.Config == nil {
+			return fmt.Errorf("invalid config: no config field for connector %q", c.ID)
+		}
+		logger.Infof("config connector: %s", c.ID)
+
+		// convert to a storage connector object
+		conn, err := ToStorageConnector(c)
+		if err != nil {
+			return fmt.Errorf("failed to initialize storage connectors: %v", err)
+		}
+		storageConnectors[i] = conn
+
+	}
+	s = storage.WithStaticConnectors(s, storageConnectors)
+
 	if len(c.OAuth2.ResponseTypes) > 0 {
 		logger.Infof("config response types accepted: %s", c.OAuth2.ResponseTypes)
 	}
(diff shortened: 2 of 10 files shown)


storage/storage.go
```diff
@@ -333,3 +345,45 @@ type Keys struct {
...
+	"github":       func() ConnectorConfig { return new(github.Config) },
+	"gitlab":       func() ConnectorConfig { return new(gitlab.Config) },
+	"oidc":         func() ConnectorConfig { return new(oidcConnector.Config) },
+	"saml":         func() ConnectorConfig { return new(saml.Config) },
+	// Keep around for backwards compatibility.
+	"samlExperimental": func() ConnectorConfig { return new(saml.Config) },
+}
+
+// OpenConnector will parse the connector config and open the connector.
+func OpenConnector(logger logrus.FieldLogger, conn Connector) (connector.Connector, error) {
```
rithujohn191: Note sure if this is the right place for this method

ericchiang: Probably in server, right?

server/server.go
```diff
@@ -362,3 +364,51 @@ func (s *Server) startGarbageCollection(ctx context.Context, frequency time.Dura
...
+	storageConnectors, err := s.storage.ListConnectors()
+	if err != nil {
+		return fmt.Errorf("failed to list connector objects from storage: %v", err)
+	}
+
+	for _, conn := range storageConnectors {
+		var err error
+		if _, ok := s.connectors[conn.ID]; ok {
+			// Update existing connector
+			if conn.ResourceVersion > s.connectors[conn.ID].ResourceVersion {
```
ericchiang: probably just `conn.ResourceVersion != s.connectors[conn.ID].ResourceVersion`

server/server.go
```diff
@@ -362,3 +364,51 @@ func (s *Server) startGarbageCollection(ctx context.Context, frequency time.Dura
...
+		}
+		if err != nil {
+			return fmt.Errorf("failed to update connector list in the server: %v", err)
+		}
+	}
+	return nil
+}
+
+func toServerConnector(conn storage.Connector) (Connector, error) {
+	logger := &logrus.Logger{
```
ericchiang: Need to reuse the server's logger.

server/server.go
```diff
@@ -103,7 +105,7 @@ func value(val, defaultValue time.Duration) time.Duration {
 type Server struct {
 	issuerURL url.URL
 
-	// Read-only map of connector IDs to connectors.
+	// Map of connector IDs to connectors.
 	connectors map[string]Connector
```
ericchiang: Since you write to this map now, we need a mutex or something.

server/server.go
```diff
@@ -362,3 +364,51 @@ func (s *Server) startGarbageCollection(ctx context.Context, frequency time.Dura
 	}()
 	return
 }
+
+// getUpdatedConnectorList retrieves a current list of connector objects available in
+// the storage. It gets the latest configuration of connectors that are dynamically
+// updated via the API.
+func (s *Server) getUpdatedConnectorList() error {
```
ericchiang: This should probably just return the list of connectors. It feels weird to have to call this before accessing the map.

cmd/dex/config.go
```diff
@@ -178,7 +181,7 @@ type ConnectorConfig interface {
 	Open(logrus.FieldLogger) (connector.Connector, error)
 }
 
-var connectors = map[string]func() ConnectorConfig{
+var connectorsConfig = map[string]func() ConnectorConfig{
```
ericchiang: Maybe we can move this logic in the server too?

cmd/dex/serve.go
```diff
@@ -175,6 +147,50 @@ func serve(cmd *cobra.Command, args []string) error {
...
+				return fmt.Errorf("failed to initialize connectors: %v", err)
+			}
+			connectors[i] = conn
+		}
+		s = storage.WithStaticConnectors(s, connectors)
+	}
+
+	// Retrieves connector objects in backend storage. This list includes the static connectors
+	// defined in the ConfigMap and dynamic connectors retrieved from the storage.
+	storageConnectors, err := s.ListConnectors()
```
ericchiang: This seems like this should be something done by the server. cmd is just for validating inputs, not checking existing storage values.

storage/storage.go
rithujohn191: Ok will move it there. This method is also used by cmd/dex/serve.go and requires the connectorsConfig variable defined. It does seem redundant to define the connectorsConfig variable in both cmd/dex/config and here, but not really sure how to avoid that since we can't import the main pkg.

ericchiang: Actually just commented on that :) Maybe we just centralize this logic in the server?

server/server.go
rithujohn191: So are you suggesting getting rid of the connectors map in the server struct all together and just making use of this function whenever we need the list of connectors?
Maybe returning a map here will make it easier to access a particular connection object.

server/server.go
rithujohn191: done

server/server.go
rithujohn191: done

server/server.go
rithujohn191: done

cmd/dex/config.go
rithujohn191: done

cmd/dex/serve.go
rithujohn191: moved this to server/server.go

cmd/dex/serve.go
```diff
@@ -176,6 +148,34 @@ func serve(cmd *cobra.Command, args []string) error {
...
+		if c.ID == "" || c.Name == "" || c.Type == "" {
+			return fmt.Errorf("invalid config: ID, Type and Name fields are required for a connector")
+		}
+		if c.Config == nil {
+			return fmt.Errorf("invalid config: no config field for connector %q", c.ID)
+		}
+		logger.Infof("config connector: %s", c.ID)
+
+		// convert to a storage connector object
+		conn, err := ToStorageConnector(c, "1")
```
ericchiang: Maybe just leave the resourceVersion blank? I don't see any reason to always set it to "1"

server/server.go
```diff
@@ -17,14 +20,19 @@ import (
...
 	"github.com/coreos/dex/storage"
 )
 
-// Connector is a connector with metadata.
+// Connector is a connector with resource version metadata.
 type Connector struct {
-	ID          string
-	DisplayName string
-	Connector   connector.Connector
+	ResourceVersion string
```
ericchiang: Why is this field here? Does anything else use it?

server/server.go
```diff
@@ -362,3 +382,84 @@ func (s *Server) startGarbageCollection(ctx context.Context, frequency time.Dura
...
+	conn := s.connectors[id]
+	s.mu.Unlock()
+
+	if storageConnector.ResourceVersion != conn.ResourceVersion {
+		// Connector object has been updated in the storage. Need to get latest.
+		if err := s.OpenConnector(storageConnector); err != nil {
+			return Connector{}, fmt.Errorf("failed to open connector: %v", err)
+		}
+	}
+	return conn, nil
```
ericchiang: This is going to be the connector from line 455, not the new one.

Maybe don't split `OpenConnector` into its own method?

server/server.go
```diff
@@ -122,6 +127,9 @@ type Server struct {
 	idTokensValidFor time.Duration
 
 	logger logrus.FieldLogger
+
+	// mutex for the connectors map.
+	mu sync.Mutex
```
ericchiang: nit: It's generally good style to put the mutex above what it's protecting. E.g.

```
type foo struct {
   a int

    mu sync.Mutext // protects b
    b int

    c int
}
```

server/server.go
```diff
@@ -195,8 +193,30 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
...
-	for _, conn := range c.Connectors {
-		s.connectors[conn.ID] = conn
+	// Retrieves connector objects in backend storage. This list includes the static connectors
+	// defined in the ConfigMap and dynamic connectors retrieved from the storage.
+	storageConnectors, err := c.Storage.ListConnectors()
+	if err != nil {
+		return nil, fmt.Errorf("server: failed to list connector objects from storage: %v", err)
+	}
+
+	if len(storageConnectors) == 0 {
```
ericchiang: This won't count the passwordDB (which is specified below).

server/server.go
```diff
@@ -195,8 +193,30 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
...
+		return nil, errors.New("server: no connectors specified")
+	}
+
+	for _, conn := range storageConnectors {
+		if err := s.OpenConnector(conn); err != nil {
+			return nil, errors.New("server: Failed to open connector")
```
ericchiang: nit: return the actual error and maybe print the id of the connector you were trying to open.

cmd/dex/serve.go
```diff
@@ -74,7 +74,7 @@ func serve(cmd *cobra.Command, args []string) error {
 		errMsg string
 	}{
 		{c.Issuer == "", "no issuer specified in config file"},
-		{len(c.Connectors) == 0 && !c.EnablePasswordDB, "no connectors supplied in config file"},
+		{len(c.StaticConnectors) == 0 && !c.EnablePasswordDB, "no connectors supplied in config file"},
```
ericchiang: Probably should remove this, right? The connectors can be specified in the storage.

server/server.go
ericchiang: Also this check has the un-fun result of you being able to mess up dex by deleting the last in-storage connector. Any thoughts?

cmd/dex/serve.go
rithujohn191: Done

server/server.go
rithujohn191: Used by func (s *Server) getConnector()

server/server.go
rithujohn191: We wont let the user delete the in-storage connector for the static connectors right? So we technically will never have the problem of deleting the last connector since we cant delete static connectors.

server/server.go
ericchiang: :+1:

server/server.go
ericchiang: I mean dex would have different behavior depending on what's in it's backing storage. No biggie, please just add the passwordDB to this count.

server/server.go
ericchiang: Or have OpenConnector return a connector.

server/server.go
rithujohn191: I will print the actual error 'err' which already prints the id of the connector that failed to open.

server/server.go
rithujohn191: The reason I split OpenConnector into its own method was to avoid redundant code. Will make it return a connector.

cmd/dex/serve.go
rithujohn191: yep will get rid of this. The check in server.go handles the case when 0 connectors are available.

server/server.go
```diff
@@ -362,3 +381,89 @@ func (s *Server) startGarbageCollection(ctx context.Context, frequency time.Dura
...
+// getConnector retrieves the connector object with the given id from the storage
+// and updates the connector list for server if necessary.
+func (s *Server) getConnector(id string) (Connector, error) {
+	storageConnector, err := s.storage.GetConnector(id)
+	if err != nil {
+		return Connector{}, fmt.Errorf("failed to get connector object from storage: %v", err)
+	}
+
+	s.mu.Lock()
+	conn := s.connectors[id]
```
ericchiang: this ID could also not map to anything. Need to check for that case.

```
s.mu.Lock()
conn, ok := s.connectors[id]
s.mu.Unlock()
if !ok {
    // take action
}
```

server/server.go
```diff
@@ -195,8 +192,30 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
...
+	// defined in the ConfigMap and dynamic connectors retrieved from the storage.
+	storageConnectors, err := c.Storage.ListConnectors()
+	if err != nil {
+		return nil, fmt.Errorf("server: failed to list connector objects from storage: %v", err)
+	}
+
+	if c.EnablePasswordDB {
+		// The passwordDB connector is an internal connector maintained by the server.
+		s.mu.Lock()
+		s.connectors["local"] = Connector{
```
ericchiang: This needs to be part of the static connectors, or you won't include it when you list all connectors.

rithujohn191: Moved this to cmd/dex/serve.go

server/server.go
rithujohn191: Combined the if condition to the following:
" if !ok || storageConnector.ResourceVersion != conn.ResourceVersion " since the action we take for both cases are the same.

Had to separately declare the variable conn to avoid nil pointer dereferencing

cmd/dex/serve.go
```diff
@@ -176,6 +147,39 @@ func serve(cmd *cobra.Command, args []string) error {
...
+		conn, err := ToStorageConnector(c)
+		if err != nil {
+			return fmt.Errorf("failed to initialize storage connectors: %v", err)
+		}
+		storageConnectors[i] = conn
+
+	}
+	s = storage.WithStaticConnectors(s, storageConnectors)
+
+	if err != nil {
```
ericchiang: Are you missing a call that creates this error?

server/server.go
```diff
@@ -362,3 +370,100 @@ func (s *Server) startGarbageCollection(ctx context.Context, frequency time.Dura
...
+
+	return c, nil
+}
+
+// OpenConnector updates server connector map with specified connector object.
+func (s *Server) OpenConnector(conn storage.Connector) (Connector, error) {
+	var c connector.Connector
+
+	// The local passwordDB connector is an internal connector maintained by the server.
+	if conn.Type == "local" {
```
ericchiang: nit: If you're going to reference this string in multiple places, add a constant.

server/handlers.go
```diff
@@ -167,24 +167,30 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if len(s.connectors) == 1 {
-		for id := range s.connectors {
+	connectors, e := s.storage.ListConnectors()
+	if e != nil {
+		s.logger.Errorf("Failed to get list of connectors: %v", err)
+		s.renderError(w, http.StatusInternalServerError, "Failed to retrieve connector list.")
```
ericchiang: need a return after this.

server/handlers.go
```diff
@@ -339,9 +345,9 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	conn, ok := s.connectors[authReq.ConnectorID]
-	if !ok {
-		s.renderError(w, http.StatusInternalServerError, "Requested resource does not exist.")
+	conn, err := s.getConnector(authReq.ConnectorID)
+	if err != nil {
+		s.renderError(w, http.StatusInternalServerError, "Requested resource does not existr")
```
ericchiang: spelling "existr"

server/handlers.go
```diff
@@ -649,13 +655,14 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 		// Ensure the connector supports refresh tokens.
 		//
 		// Connectors like `saml` do not implement RefreshConnector.
-		conn, ok := s.connectors[authCode.ConnectorID]
-		if !ok {
+		conn, err := s.getConnector(authCode.ConnectorID)
+		if err != nil {
```
ericchiang: nit: need to include the error in the log message.

server/handlers.go
```diff
@@ -841,8 +848,8 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
 		scopes = requestedScopes
 	}
 
-	conn, ok := s.connectors[refresh.ConnectorID]
-	if !ok {
+	conn, err := s.getConnector(refresh.ConnectorID)
+	if err != nil {
```
ericchiang: nit: need to include the error in the log message.

server/handlers.go
ericchiang: Also you're going to want to log an error.

server/server.go
```diff
@@ -362,3 +370,100 @@ func (s *Server) startGarbageCollection(ctx context.Context, frequency time.Dura
...
+	var c connector.Connector
+
+	f, ok := ConnectorsConfig[conn.Type]
+	if !ok {
+		return c, fmt.Errorf("unknown connector type %q", conn.Type)
+	}
+
+	connConfig := f()
+	if len(conn.Config) != 0 {
+		data := []byte(os.ExpandEnv(string(conn.Config)))
```
ericchiang: Maybe leave thes expandenv in the cmd package? If you set a connector through the API it probably shouldn't interact with the environment.

server/server.go
rithujohn191: In the server package it is only used this one time. It is used once more in the cmd package but thats it. Would you like me to add a constant for these 2 occurrences?

rithujohn191:
Addressed all the feedback

server/server.go
ericchiang: your call, not a blocker.

rithujohn191: made this change

PR #916, opened 2017-04-19
README: add section about reporting security vulnerabilities

ericchiang:
It was pointed out by @davidillsley that we don't point this out
anywhere in our docs.


ericchiang:
cc @radhikapc for docs.

PR #917, opened 2017-04-20
Documentation: add a doc describing how to use dex

### LINKED ISSUE #755: Documentation: explain how to write an app that uses dex
Brought up by @set321go in #753

This can include:
* App that wants to talk to the Kubernetes API on behalf of a user.
* App that uses OpenID Connect to authenticate a user (it does its own authz).
* App that uses a proxy headers to authenticate a user going through an OpenID Connect aware proxy.
* Putting an app that doesn't do authentication or authorization behind an OpenID Connect aware proxy.

p1 because this kind of documentation is a blocker for users who want to use dex.

ericchiang:
cc @ElijahCaine @radhikapc @rithujohn191 @squat 

Closes #755


Documentation/using-dex.md
```diff
@@ -0,0 +1,186 @@
+# Writing apps that use dex
+
+Once you have dex up and running, the next step is to write applications that use dex to drive authentication. Apps that interact with dex generally fall into one of two categories:
+
+1. Apps that request OpenID Connect ID tokens to authenticate users.
+  * Used for authenticating an end user.
```
ericchiang: blah, this didn't format very nicely.

ElijahCaine: You might need:
```
1. Thing
    * four spaces in
    * something else
2. Even more
    * etc, etc.
```
Formats to:

1. Thing
    * four spaces in and one newline break above/below
    * something else
2. Even more
    * etc, etc.

At least that renders correctly in the PR comments 😐

Documentation/using-dex.md
```diff
@@ -0,0 +1,186 @@
...
+The dex repo contains a small [example app][example-app] as a working, self contained app that performs this flow.
+
+The rest of this section explores the code sections which to help explain how to implementing this logic in your own app.
+
+### Configuring your app
+
+The example app uses the following Go packages to perform the code flow:
+
+* [github.com/coreos/go-oidc](https://godoc.org/github.com/coreos/go-oidc)
+* [golang.org/x/oauth2](https://godoc.org/golang.org/x/oauth2)
```
ElijahCaine: These should also be reference-based links, not embedded links. `</pedantic>`

Documentation/using-dex.md
```diff
@@ -0,0 +1,188 @@
...
+        // handle error
+    }
+}
+```
+
+### State tokens
+
+The state parameter is an arbitrary string that dex will always return with the callback. It plays a security role, preventing certain kinds of OAuth2 attacks. Specifically it can be used by clients to ensure:
+
+* The user who started the flow is the one who finished it by cooking the user with some data that can be associated with the state.
```
radhikapc: The statement is unclear to me. Do you mean:
"The user who started the flow is the one who finish it by associating the user data with the state"

ericchiang:
@ElijahCaine thanks, updated.

Documentation/using-dex.md
ericchiang: > The use who started the flow is the one who finished it. This can be accomplished by cooking the user with some data that can be associated with the state (possibly the entire state token itself).

That work?

radhikapc: "cooking the user with some data" is precisely where my confusion is :-)

ericchiang: @radhikapc as in a HTTP cookie. Maybe:

> The use who started the flow is the one who finished it, by linking the user's session with the state token. For example, by setting the state as an HTTP cookie, then comparing it when the user returns to the app.

radhikapc: I think you have a typo there - The user - 'r' is missing

ericchiang: Thanks, updated.

PR #924, opened 2017-04-23
MySQL storage backend

pborzenkov:
The pull request adds MySQL storage backend.

### DIFF
--- a/storage/sql/config.go
+++ b/storage/sql/config.go
@@ -5,9 +5,11 @@ import (
 	"fmt"
 	"net/url"
 	"strconv"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/coreos/dex/storage"
+	"github.com/go-sql-driver/mysql"
 	"github.com/lib/pq"
 	sqlite3 "github.com/mattn/go-sqlite3"
 )
@@ -15,6 +17,12 @@ import (
 const (
 	// postgres error codes
 	pgErrUniqueViolation = "23505" // unique_violation
+)
+
+const (
+	// MySQL error codes
+	mysqlErrDupEntry            = 1062
+	mysqlErrDupEntryWithKeyName = 1586
 )
 
 // SQLite3 options for creating an SQL db.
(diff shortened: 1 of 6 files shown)


pborzenkov:
Hmm, looks like Tarvis is using MySQL 5.5 which doesn't support fractional DATETIME values. 5.6 is available in 'trusty' CI environment. I'll remove MySQL backend from CI until the switch is made.

storage/sql/config.go
```diff
@@ -59,14 +70,33 @@ func (s *SQLite3) open(logger logrus.FieldLogger) (*conn, error) {
...
+const (
+	// MySQL SSL modes
+	mysqlSSLTrue       = "true"
+	mysqlSSLFalse      = "false"
+	mysqlSSLSkipVerify = "skip-verify"
+	mysqlSSLCustom     = "custom"
+)
+
+// NetworkDB contains options common to SQL databases accessed over network.
+type NetworkDB struct {
```
rithujohn191: Please add individual comments describing each attribute.

storage/sql/config.go
```diff
@@ -145,3 +170,105 @@ func (p *Postgres) open(logger logrus.FieldLogger) (*conn, error) {
...
+func (s *MySQL) open(logger logrus.FieldLogger) (*conn, error) {
+	cfg := mysql.Config{
+		User:   s.User,
+		Passwd: s.Password,
+		DBName: s.Database,
+
+		Timeout: time.Second * time.Duration(s.ConnectionTimeout),
+
+		ParseTime: true,
+		Params: map[string]string{
```
rithujohn191: Please add a comment to explain this

pborzenkov:
@rithujohn191 The problem is that conformance tests require the storage to provide millisecond precision right now. I see several ways to fix this:

- Keep the precision requirement and require MySQL 5.6. This way either CI has to be switched to 'trusty' or MySQL >= 5.6 has to be started in a docker container;
- Relax the precision requirement in the tests and agree to full second precision;
- Extend the Storage interface with the 'TimestampPrecision() time.Duration' method which will be used to report the maximum timestamp precision supported by the storage. And use the method throughout the test code.

Personally, I think the third one is the way to go and it does not require a lot of coding. What do you think?

michelaquino:
Hi guys, what the PR status?

pborzenkov:
Unfortunately, I didn't have time to finish it. I'll try to complete it in the coming days. I'll go with the item 2 from my proposal (relaxing timestamp precision requirements to full second).

ericchiang:
I don't think the dex team has the bandwidth to maintain another storage implementation. If someone wants to step up as a maintainer, we would consider merging, But without someone dedicated to this storage, I don't think this is going to happen today.

sagikazarmark:
Dex supports MySQL now.

PR #925, opened 2017-04-25
api: add gRPC calls to create, list and delete connectors

rithujohn191:
This change defines 4 new API calls:
* CreateConnector
* UpdateConnector
* ListConnectors
* DeleteConnector

Also added unit tests for each.

Logic used for resource versioning:
When a connector is created via an API call its resource version default to "1". Every time the connector is updated via an API call, its ResourceVersion is incremented by 1. For static connectors the resource version is left blank as it cannot be updated


server/api.go
```diff
@@ -272,3 +272,111 @@ func (d dexAPI) RevokeRefresh(ctx context.Context, req *api.RevokeRefreshReq) (*
...
+	}
+
+	c := storage.Connector{
+		ID:              req.Connector.Id,
+		Type:            req.Connector.Type,
+		Name:            req.Connector.Name,
+		ResourceVersion: "1",
+		Config:          req.Connector.Config,
+	}
+	if err := d.s.CreateConnector(c); err != nil {
```
ericchiang: Probably want to make sure the config actually parses too, right? E.g. so people can't provide invalid arguments or JSON.

server/api.go
```diff
@@ -272,3 +272,111 @@ func (d dexAPI) RevokeRefresh(ctx context.Context, req *api.RevokeRefreshReq) (*
...
+		currentVersion, err := strconv.Atoi(old.ResourceVersion)
+		if err != nil {
+			return storage.Connector{}, errors.New("failed to covert ResourceVersion string to int")
+		}
+		old.ResourceVersion = strconv.Itoa(currentVersion + 1)
+
+		return old, nil
+	}
+
+	if err := d.s.UpdateConnector(req.Id, updater); err != nil {
```
ericchiang: Same comment as "create", probably want to make sure this actually parses.

server/api.go
```diff
@@ -272,3 +272,111 @@ func (d dexAPI) RevokeRefresh(ctx context.Context, req *api.RevokeRefreshReq) (*
...
+
+		if req.Name != "" {
+			old.Name = req.Name
+		}
+
+		if len(req.Config) > 0 {
+			old.Config = req.Config
+		}
+
+		currentVersion, err := strconv.Atoi(old.ResourceVersion)
```
ericchiang: Maybe just use a random string? I don't see a reason for these to actually have to be numbers.

rithujohn191: I thought it would make debugging easier in case something goes wrong and would be easier to spot that an older version of the resource is being used

ericchiang: Hmm, we probably wanted to make this an int then. Your call.

rithujohn191: which would mean we need to change it in the backend storage too and alter the db tables. I looked at grpc code and they seem to do it the same way. I could switch it all out to int if it seems like a design issue.

ericchiang: I don't think it's worth the effort to change now. Incrementing it like this seems fine.

server/api.go
rithujohn191: added a method parseConnConfig() that does this

server/api.go
rithujohn191: Done

server/api.go
```diff
@@ -272,3 +273,150 @@ func (d dexAPI) RevokeRefresh(ctx context.Context, req *api.RevokeRefreshReq) (*
...
+// parseConnConfig will parse the provided connector config to ensure it is correct.
+func parseConnConfig(connType string, config []byte) error {
+	f, ok := ConnectorsConfig[connType]
+	if !ok {
+		return fmt.Errorf("unknown connector type %q", connType)
+	}
+
+	connConfig := f()
+	if len(config) != 0 {
+		if err := json.Unmarshal(config, connConfig); err != nil {
```
ericchiang: probably want to open it too?

server/api.go
```diff
@@ -272,3 +273,150 @@ func (d dexAPI) RevokeRefresh(ctx context.Context, req *api.RevokeRefreshReq) (*
...
+}
+
+func (d dexAPI) UpdateConnector(ctx context.Context, req *api.UpdateConnectorReq) (*api.UpdateConnectorResp, error) {
+	if req.Id == "" {
+		return nil, errors.New("no connector ID supplied")
+	}
+	if req.Type == "" && req.Name == "" && len(req.Config) == 0 {
+		return nil, errors.New("nothing to update")
+	}
+	if len(req.Config) != 0 {
```
ericchiang: this check is redundant. you've already made it a couple lines above.

server/api.go
rithujohn191: OpenConnector() is already called in the getConnector() (https://github.com/coreos/dex/blob/master/server/server.go#L448) which is called in ther server/handlers.go.  Wouldn't it be redundant to call it again here?

server/api.go
rithujohn191: A user can update any one of these values req.Type, req.Name, req.Config or all of them. In the case that they just want to update the config we need to make the check again. It could be the case that they just want to update the type or the name.

server/api.go
ericchiang: I'd imagine you want to reject bad configs at the API call if you can, instead of by the time server is reading it.

server/api.go
ericchiang: You probably want to make that an or statement then and require a Type, Name and Config.

rithujohn191: The connector is looked up via the id attribute we don't to mandate that the user provide the Type and Name

ericchiang: It needs a type to parse the config and a name to display to the user.

rithujohn191:
Looking into a cleaner way of doing this!  Possibly by using component config.

fishioon:
These api will merge again？
In my case，we need these api. 
Make a different with your code. Change resource version from add int to timestamp.

PR #928, opened 2017-04-28
cmd/dex/serve: allow connector to have no config attribute

rithujohn191:
Connectors like "local" connector will not have a config


rithujohn191:
Going to close this out since most connectors require config to be non-empty and it seems like a reasonable check for all connectors except "local"

PR #943, opened 2017-05-11
Expand env vars in staticClients

jungziege:
This is a PR to address https://github.com/coreos/dex/issues/889.

I ran the tests both as currently written, and manually put an env var into the config test. If you have suggestions for how to add the env var explansion testing into that file permanently, I'm happy to add them (not sure how to inject env vars into the test env using Travis).

Also, my original version of this PR accidentally called `WithStaticClients` with an empty client list, but all the tests still passed. I've since fixed that line, but just wanted to call out that functionality doesn't seem to be tested.

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -34,7 +34,7 @@ type Config struct {
 
 	// StaticClients cause the server to use this list of clients rather than
 	// querying the storage. Write operations, like creating a client, will fail.
-	StaticClients []storage.Client `json:"staticClients"`
+	StaticClients []client `json:"staticClients"`
 
 	// If enabled, the server will maintain a list of passwords which can be used
 	// to identify a user.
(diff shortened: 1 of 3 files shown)


jungziege:
@rithujohn191 Wanted to ping you about this PR, since I assume you're the best person to review this. Is this the right direction? We'd like to use this to supply secrets via kubernetes env vars, but if this isn't something you're interested in accepting we'd like to know sooner so we can start work on creating clients with gRPC.

yawboateng:
any update on this?

ericchiang:
Sorry for taking so long but it's because we really don't want to keep doing this "arbitrary expand env var" pattern. Here's my counter proposal https://github.com/coreos/dex/issues/1099

jungziege:
Closing in favor of proposed alternative.

PR #951, opened 2017-05-22
cmd/dex: add a pprof endpoint for profiling CPU and memory

ericchiang:
cc @rithujohn191

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/http/pprof"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


cmd/dex/serve.go
```diff
@@ -252,6 +253,20 @@ func serve(cmd *cobra.Command, args []string) error {
 			}()
 		}()
 	}
+	if c.PprofAddr != "" {
+		logger.Infof("listening (grpc) on %s", c.PprofAddr)
```
ericchiang: grpc?

lhotrifork:
@ericchiang @rithujohn191 what would it take to get this PR merged? Any reason why it stranded? :)

sagikazarmark:
This PR is probably outdated and needs some rework.

The pprof handlers could be registered onto the telemetry endpoint.

PR #960, opened 2017-06-02
Email verified for local connector

DusanKasan:
This PR implements 'email_verified' for the local connector (password DB) and resolves https://github.com/coreos/dex/issues/957 

As it stands right now, Dex will always return 'email_verified' as true for any locally stored accounts. Presumably this is so that registration and eventual email verification is done outside and is out of scope of Dex. While this is understandable, this causes a problem where after user registration we want to show a specific "waiting for email verification" error on login (before the verification is done).

As this flow is pretty much the standard, I propose supporting it by means of a config boolean switch. 

### What is done:
- email_verification added to the local password DB
- gRPC API for passwords updated to reflect the field addition
- when logging in via local account (email) return the stored value of `email_verification` field
  - **should this return true when "requireEmailVerification" config value is false/ommited?**
- requireEmailVerification config value introduced. False by default, it will only allow verified accounts to be logged in.
  - also a special error on login message added for this purpose

This feature is not covered by tests yet.


ericchiang:
I don't think we should do this. I'd rather have external systems do this kind of user management.

PR #961, opened 2017-06-03
Added dockerfile for example app

stijndehaes:


Makefile
```diff
@@ -34,6 +36,10 @@ bin/grpc-client: check-go-version
 release-binary:
 	@go build -o /go/bin/dex -v -ldflags $(LD_FLAGS) $(REPO_PATH)/cmd/dex
```
derekjwilliams: Does it make sense to reference $GOPATH here instead of hardcoding '/go/bin/dex' ?

nabokihms:
The linked issue is marked as resolved, so I'm going to close this PR. If there are problems left, please open a separate issue.

Issue #892, opened 2017-04-04
Include connector UserID in ID Token


Calpicow:
The JWT sub is being strangely encoded. I expect it to be equal to the UserID attribute set by the connector.

Looks like the result of https://github.com/coreos/dex/blob/master/server/oauth2.go#L269-L278

ericchiang:
Because different backends can clash, we combined the UserID returned by the connector and combine it with the connector ID. This also gives us a consistent ID to track things like refresh tokens.

Why do you need the UserID? Could we possibly map it to a different scope?

Calpicow:
For example, if I wanted to map permissions on the user account in Kubernetes via RBAC, it would be pretty tedious and difficult to track if `sub` is a base64 encoded string.

If any, I think we should encode it into a human-readable format or yes, create another claim with the concatenated information (or both).

ericchiang:
So I think we should add a scope which adds this to claims. E.g. if you provide the `uid` scope, you get a claim like `dex.coreos.com/uid` returned. How does that sound?

cc @rithujohn191 @squat

Calpicow:
I think that would suffice for the Kubernetes use-case. I could imagine some OIDC consumers would have `sub` hardcoded into its logic and/or expect the value to be backed by LDAP or otherwise. As a consumer, I'd also expect the `uid` to be present in the JWT payload without having to explicitly request it.

Perhaps we could automatically include a `dex.coreos.com/connector_id` scope and make the behavior of the `sub` scope configurable.

blakebarnett:
What would the output in the API server logs be?  We'd just like to be able to search logs for usernames, preferably by `username` in the ldap userSearch.

chancez:
For what it's worth, what `auth0` does is they just concat the `connectorID` + `|` + `userID`.

ericchiang:
I would strongly encourage we add an additional scope (or set of scopes) and surface this as a claim. Please don't rely on the sub field's encoding.

chancez:
And also, the regular userID for the given connection is in a non-standard `identities` claim like so:

Ex:
```
  "sub": "google-oauth2|110272483197731336751",
  "user_id": "google-oauth2|110272483197731336751",
  "identities": [
    {
      "provider": "google-oauth2",
      "user_id": "110272483197731336751",
      "connection": "google-oauth2",
      "isSocial": true
    },
    {
      "profileData": {
        "email": "chance.zibolski@coreos.com",
        "email_verified": true
      },
[...]

Issue #831, opened 2017-03-01
Add a claim indicating which connector the user logged in through


ericchiang:
It's extremely useful to know which connector a user logged in through. Something like

```
{
  "iss": "https://eric-dex-testing.staging.core-os.net/identity",
  "sub": "ChZRUUtDenltbzdWSUhhNFp4U2pONHFnEgVsb2NhbA",
  "aud": "tectonic-kubectl",
  "exp": 1488403357,
  "iat": 1488403057,
  "at_hash": "yT3YPoXRBDywYpHlYm9xhA",
  "email": "eric.chiang@coreos.com",
  "email_verified": true,
  "login": "ldap",
  "name": "admin"
}
```

nabokihms:
There is a way to know which connector is used by requesting the `federated:id` scope.
See https://dexidp.io/docs/custom-scopes-claims-clients/#custom-claims for more details.

PR #970, opened 2017-06-15
Include uid and connector in token

Calpicow:
Ref #892 #831


srenatus:
@Calpicow @ericchiang What's the status of this? (Can I contribute anything to making this happen?)

I'm not sure about the `uid` part (since it seems to be a convention at best -- should this be configurable?); but the connector ID would be very useful.

PR #971, opened 2017-06-16
connector/ldap/ldap.go: Resolve gid, when LDAP groupsearch is incomplete

marcosterland:
Some LDAP servers exclude the main group from the groupsearch. To deliver
a complete list of groups, the numeric group id (gid) of a user can be
resolved to the group name and be appended.


sprohaska:
I think the second search should be optional, so that it can be disabled if it is not needed.

marcosterland:
@sprohaska 
The second search is returns an empty result, if the parameters aren't set.
I'll add a check, just to be safe.

ericchiang:
Why isn't this possible with the existing group search parameters? Can you provide a LDAP schema that isn't possible to query today but is with this change?

ericchiang:
Closing for now since we haven't heard back. Happy to re-open if someone wants to post some example schemas.

PR #972, opened 2017-06-19
Documentation/github-connector: warn user that GitHub email id should be public.

rithujohn191:
Looks like a lot of users are trying the GitHub connector with private email ids.

### DIFF
--- a/Documentation/github-connector.md
+++ b/Documentation/github-connector.md
@@ -6,6 +6,11 @@ One of the login options for dex uses the GitHub OAuth2 flow to identify the end
 
 When a client redeems a refresh token through dex, dex will re-query GitHub to update user information in the ID Token. To do this, __dex stores a readonly GitHub access token in its backing datastore.__ Users that reject dex's access through GitHub will also revoke all dex clients which authenticated them through GitHub.
 
+## Caveats
+
+* Please note that in order for a user to be authenticated via GitHub, the user needs to mark their email id as public on GitHub. This will enable the API to return the user's email to Dex.
+* Currently, authentication via GitHub allows users outside of the `Org` specified in the connector to login. This is being tracked by [issue #920][issue-920].
+
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


Documentation/github-connector.md
```diff
@@ -8,7 +8,7 @@ When a client redeems a refresh token through dex, dex will re-query GitHub to u
 
 ## Configuration
 
-Register a new application with [GitHub][github-oauth2] ensuring the callback URL is `(dex issuer)/callback`. For example if dex is listening at the non-root path `https://auth.example.com/dex` the callback would be `https://auth.example.com/dex/callback`.
+Register a new application with [GitHub][github-oauth2] ensuring the callback URL is `(dex issuer)/callback`. For example if dex is listening at the non-root path `https://auth.example.com/dex` the callback would be `https://auth.example.com/dex/callback`. Please note that a user needs to mark their email id as public in GitHub, so that the ID token that gets returned for this user contains an email id field in the ID token claims.
```
ericchiang: It's kinda sounds like you're saying that ID tokens are returned by GitHub.

Also might be worth marking this as a known caveat. e.g. https://github.com/coreos/dex/blob/master/Documentation/oidc-connector.md#caveats or https://github.com/coreos/dex/blob/master/Documentation/saml-connector.md#caveats

rithujohn191: Should I just leave at "users should have their email ids as public"?

rithujohn191: Yep will create a new caveat section for it. Maybe also add our GitHub orgs problem there.

ericchiang: I think detail the fact that GitHub's API won't return an email if the end user's email isn't public. However you want to word that, but the current sentence isn't super clear.

rithujohn191: :+1:

rithujohn191:
Made changes according to the feedback.

simonbyrne:
Note that you should be able to get non-public emails with user:email scope, but you need to explicitly query [`api.github.com/user/emails`](https://developer.github.com/v3/users/emails/#list-email-addresses-for-a-user)

e.g. see https://github.com/thephpleague/oauth2-client/issues/9#issuecomment-20653139

Issue #959, opened 2017-06-01
Google open id HD param


stijndehaes:
The google open id specification specifies the follow hd-param: https://developers.google.com/identity/protocols/OpenIDConnect#hd-param

This parameters allow you to specify for which domain users can log in. I can not find how to configure this in dex. I think however that this param is google specific so maybe that's why it is not used? I would love to use it to make sure that only user from our company can use this login.

Thanks for looking at my issue.
Stijn

rithujohn191:
Hi @Steniaz ,

As of now we do not support the optional hd-param in our [OIDC connector](https://github.com/coreos/dex/blob/master/Documentation/oidc-connector.md) and it is not on our road map. However we would be more than happy to review a PR for this change if you can submit one.

thanks

Steniaz:
@rithujohn191 Ok thanks for the help :) I'll see if I can do something in the coming weeks.

rithujohn191:
Great thank you :)

roguePanda:
I saw that it had been a while and needed this for my own purposes, so I put together a quick PR (#974). Let me know if it works for your use case @Steniaz!

Steniaz:
@roguePanda Had a short look at your PR, this is exactly what I need thank you! Hopefully this can be merged pretty fast :)

ericchiang:
I'd like this to be in a google specific connector and not in the OpenID Connect connector https://github.com/coreos/dex/issues/1001

nabokihms:
The option is present in the specific Google connector and has been removed from the OIDC connector recently.
I think the initial problem was solved.

PR #974, opened 2017-06-21
Google hosted domain support

roguePanda:
See #959 - Google's OIDC implementation supports limiting access to a certain Google Apps / G Suite domain. This lets dex users ensure that only people in their organization can log in, and not just anyone with a Google account.

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -33,6 +33,7 @@ type Config struct {
 
 	Scopes []string `json:"scopes"` // defaults to "profile" and "email"
 
+	HostedDomain string `json:"hostedDomain"`
 }
 
 // Domains that don't support basic auth. golang.org/x/oauth2 has an internal
@@ -110,8 +111,9 @@ func (c *Config) Open(logger logrus.FieldLogger) (conn connector.Connector, err 
 		verifier: provider.Verifier(
 			&oidc.Config{ClientID: clientID},
 		),
-		logger: logger,
-		cancel: cancel,
+		logger:       logger,
+		cancel:       cancel,
+		hostedDomain: c.HostedDomain,
 	}, nil
 }
 
@@ -127,6 +129,7 @@ type oidcConnector struct {
 	ctx          context.Context
 	cancel       context.CancelFunc
 	logger       logrus.FieldLogger
+	hostedDomain string
 }
 
 func (c *oidcConnector) Close() error {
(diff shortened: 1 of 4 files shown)


Steniaz:
Tests fail because of a lint suggestion: 
/home/travis/gopath/src/github.com/coreos/dex/connector/oidc/oidc.go:147:9: if block ends with a return statement, so drop this else and outdent its block

connector/oidc/oidc.go
```diff
@@ -33,6 +33,7 @@ type Config struct {
 
 	Scopes []string `json:"scopes"` // defaults to "profile" and "email"
 
+	HostedDomain string `json:"hostedDomain"`
```
rithujohn191: Please add a comment describing what this field is and highlighting that it is optional

rithujohn191:
Please also add an explanation of the new field [here](https://github.com/coreos/dex/blob/master/Documentation/oidc-connector.md). If possible please do include unit tests in [oidc tests](https://github.com/coreos/dex/blob/master/connector/oidc/oidc_test.go). You can refer the [ldap](https://github.com/coreos/dex/blob/master/connector/ldap/ldap_test.go) or [saml](https://github.com/coreos/dex/blob/master/connector/saml/saml_test.go) tests for reference.

connector/oidc/oidc.go
melignus: It might be nice to support an array of domains in the configuration and then match on "at least one" in HandleCallback?

roguePanda:
I added a test for `LoginURL`, but `HandleCallback` makes a request to the issuer and I couldn't find a way to mock that.

stijndehaes:
Is there anything I can do to help this get merged? I am really in need of this functionality.

connector/oidc/oidc.go
```diff
@@ -176,11 +189,26 @@ func (c *oidcConnector) HandleCallback(s connector.Scopes, r *http.Request) (ide
...
+		HostedDomain  string `json:"hd"`
 	}
 	if err := idToken.Claims(&claims); err != nil {
 		return identity, fmt.Errorf("oidc: failed to decode claims: %v", err)
 	}
 
+	if len(c.hostedDomains) > 0 {
+		found := false
+		for _, domain := range c.hostedDomains {
+			if claims.HostedDomain != domain {
```
alonl: shouldn't it be `if claims.HostedDomain == domain` ?
@roguePanda @rithujohn191

connector/oidc/oidc.go
```diff
@@ -33,6 +33,9 @@ type Config struct {
 
 	Scopes []string `json:"scopes"` // defaults to "profile" and "email"
 
+	// Optional list of whitelisted domains when using Google
+	// If this field is nonempty, only users from a listed domain will be allowed to log in
+	HostedDomains []string `json:"hostedDomain"`
```
alonl: should be `json:"hostedDomains"` (domain**s**)

examples/config-dev.yaml
```diff
@@ -61,6 +61,7 @@ connectors:
 #     clientID: $GOOGLE_CLIENT_ID
 #     clientSecret: $GOOGLE_CLIENT_SECRET
 #     redirectURI: http://127.0.0.1:5556/dex/callback
+#     hostedDomain: $GOOGLE_HOSTED_DOMAIN
```
alonl: I'm afraid this doesn't match the code. If I understand it correctly, it should be:
```
hostedDomains:
  - $GOOGLE_HOSTED_DOMAIN
```

PR #985, opened 2017-07-07
bugfix: avoid generating an invalid ID attribute in SAML's AuthenRequest element

kkohtaka:
Our team is going to use Azure AD as an IdP with SAML 2.0 connector. However, when the connector sends an AuthnRequest to Azure AD's endpoint, Azure AD often responds with an error message such as `The request is not a valid Saml2 protocol message`.

After a brief investigation, I found that the connector often generates an AuthnRequest whose ID attribute starts with a digit though such an ID attribute is forbidden by the XML specification. The ID attribute is originally generated by `storage.NewID()` and this PR will change the behavior of the function to avoid generating an ID which starts with a digit.

Refs: https://www.w3.org/TR/REC-xml/#sec-attribute-types

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -30,8 +30,8 @@ func NewID() string {
 	if _, err := io.ReadFull(rand.Reader, buff); err != nil {
 		panic(err)
 	}
-	// Trim padding
-	return strings.TrimRight(encoding.EncodeToString(buff), "=")
+	// Avoid the identifier to begin with number and trim padding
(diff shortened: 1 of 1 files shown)


rithujohn191:
Thank you for the change.

kkohtaka:
Thanks for reviewing and merging. Do you have a specific schedule to release a next version?

rithujohn191:
@kkohtaka Latest release is available at https://github.com/coreos/dex/releases/tag/v2.5.0

PR #989, opened 2017-07-11
Documentation: add a group query example for the ldap connector.

rithujohn191:
A lot of people seem to be confused about how they should implement group specific queries. Thought it would be nice to have this as part of our docs.

### DIFF
--- a/Documentation/ldap-connector.md
+++ b/Documentation/ldap-connector.md
@@ -191,6 +191,32 @@ groupSearch:
   # Unique name of the group.
   nameAttr: cn
 ```
+To extract group specific information the `DN` can be used in the `userAttr` field.
+
+```
+# Top level object example.coma in LDIF file.
+dn: dc=example,dc=com
+objectClass: top
+objectClass: dcObject
+objectClass: organization
+dc: example
+```
+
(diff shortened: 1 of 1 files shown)


rithujohn191:
@ericchiang do you want me to replace `user DN` with just `DN`?

rithujohn191:
Added an LDIF object example to clarify "DN". Also moved the example to the config section.

Documentation/ldap-connector.md
```diff
@@ -191,6 +191,32 @@ groupSearch:
...
+
+```yaml
+groupSearch:
+  # BaseDN to start the search from. It will translate to the query
+  # "(&(objectClass=group)(member=<user DN>))".
+  baseDN: cn=groups,cn=compat,dc=example,dc=com
+  # Optional filter to apply when searching the directory.
+  filter: "(objectClass=group)"
+
+  userAttr: dc=example,dc=com # Use "DN" here not "uid"
```
ericchiang: This needs to be

```
userAttr: DN
```

rithujohn191:
@ericchiang any additional change suggestions?

PR #997, opened 2017-07-19
config: extract config from main to separate package

brancz:
In order for external applications to be able to import the config definition to parse and marshal the configuration, the config needs to be moved to a separate package, as main packages are not importable.

@ericchiang @rithujohn191

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -20,6 +20,7 @@ import (
 	"google.golang.org/grpc/credentials"
 
 	"github.com/coreos/dex/api"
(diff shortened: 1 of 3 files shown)


brancz:
Actually I was a bit too quick, I don't think this will work out at I have to enable CGO enabled builds for this to work (because of the sqlite dependency of the storage/sql package).

ericchiang:
@brancz I had some additional thoughts of config format changes we might want to make while doing this here: https://github.com/coreos/dex/issues/968

TL;DR - json.RawMessage kinda sucks

PR #998, opened 2017-07-19
server/rotation.go: Fix key rotation with multiple dex instances.

rithujohn191:
This fixes https://github.com/coreos/dex/issues/976.

It also ensures that it displays "keys already rotated" as an info log message instead of an error message.

### DIFF
--- a/server/rotation.go
+++ b/server/rotation.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"time"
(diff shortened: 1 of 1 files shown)


server/rotation.go
```diff
@@ -15,6 +16,8 @@ import (
 	"github.com/coreos/dex/storage"
 )
 
+var errAlreadyRotated = errors.New("key already rotated by another instance of Dex")
```
ericchiang: nit:

s/key/keys/g
s/another instance of Dex/another server instance/g

robszumski:
From the changes it looks like this was cosmetic not an actual bug? Is that true?

ericchiang:
@robszumski returning `nil` instead of an error was causing the update to write an empty keys object to the backing storage. This was indeed a pretty bad bug if two instances of dex were attempting to rotate keys within a second of each other.

rithujohn191:
@robszumski Adding to that its necessary to have more than one instance of Dex to reproduce this bug. Its not an issue if the user has only one replica of Dex

PR #1002, opened 2017-07-25
*: {S->s}irupsen/logrus

estroz:
Updated `logrus` package to take advantage of [log hooks](https://github.com/sirupsen/logrus#testing) in testing.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -74,7 +74,7 @@ import:
   version: c97913dcbd76de40b051a9b4cd827f7eaeb7a868
 - package: golang.org/x/oauth2
   version: 08c8d727d2392d18286f9f88ad775ad98f09ab33
-  subpackages: []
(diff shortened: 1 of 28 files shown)


glide.yaml
```diff
@@ -131,10 +131,14 @@ import:
   version: 411e09b969b1170a9f0c467558eb4c4c110d9c77
 
 # Structured logging
-- package: github.com/Sirupsen/logrus
-  version: v0.11.0
-- package: golang.org/x/sys/unix
+- package: github.com/sirupsen/logrus
+  version: v1.0.2
+  subpackages:
+  - hooks/test
```
ericchiang: Do we actually import this new package?

estroz: In #1003 yes. Should I remove here and add in that PR?

PR #1003, opened 2017-07-25
Log large bcrypt costs, error if password-hash comparison takes > 10s

### LINKED ISSUE #962: warning for bcrypt hashing cost
This refers to bcrypt hashes provided for static password login via the local connector.

Bcrypt password hashes generated with too many rounds can cause the `auth/local` call to take many seconds, an even timeout. As of now, there's no indication as to what the reason for the timeout is.

It would be nice if dex detected and printed a warning to logs if: 

* bcrypt hash generated with > 12 rounds is supplied for static user
* bcrypt comparison function ` CompareHashAndPassword()` taking longer than 10 seconds during a login

estroz:
server: log bcrypt cost if > 12, error on runtime > 10s

The bcrypt hashing algorithm runtime grows exponentially with cost, and might cause a timeout if the cost is too high. Notifying the user of high cost and of long running calculations will help with parameter tuning and debugging.

Tests: added test cases for both additions

Fixes #962

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -21,6 +21,9 @@ import (
 // to determine if the server supports specific features.
 const apiVersion = 2
 
+// recCost is the recommended bcrypt cost, which balances hash strength and time
+const recCost = 12
+
 // NewAPI returns a server which implements the gRPC API interface.
 func NewAPI(s storage.Storage, logger logrus.FieldLogger) api.DexServer {
 	return dexAPI{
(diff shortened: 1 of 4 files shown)


estroz:
ok to test

server/api.go
```diff
@@ -100,9 +103,13 @@ func (d dexAPI) CreatePassword(ctx context.Context, req *api.CreatePasswordReq)
 		return nil, errors.New("no user ID supplied")
 	}
 	if req.Password.Hash != nil {
-		if err := checkCost(req.Password.Hash); err != nil {
+		cost, err := checkCost(req.Password.Hash)
+		if err != nil {
 			return nil, err
 		}
+		if cost > recCost {
+			d.logger.Warnln("password encryption might take awhile; bcrypt cost was", cost)
```
rithujohn191: Could you change this message to something like "Bcrypt cost = %d, password encryption might timeout. Recommended bcrypt cost is 12."

server/api.go
```diff
@@ -133,9 +140,13 @@ func (d dexAPI) UpdatePassword(ctx context.Context, req *api.UpdatePasswordReq)
 	}
 
 	if req.NewHash != nil {
-		if err := checkCost(req.NewHash); err != nil {
+		cost, err := checkCost(req.NewHash)
+		if err != nil {
 			return nil, err
 		}
+		if cost > recCost {
+			d.logger.Warnln("password encryption might take awhile; bcrypt cost was", cost)
```
rithujohn191: same as above

server/api_test.go
```diff
@@ -119,6 +119,63 @@ func TestPassword(t *testing.T) {
...
+			if !tc.wantErr {
+				t.Errorf("%s: %s", tc.name, err)
+			}
+			continue
+		}
+
+		if tc.wantErr {
+			t.Errorf("%s: expected err", tc.name)
+			continue
+		}
```
rithujohn191: Could you also add a test case with a bcrypt hash with a really high cost (much more than 12)

server/api_test.go
```diff
@@ -119,6 +119,63 @@ func TestPassword(t *testing.T) {
...
+		name         string
+		inputHash    []byte
+		expectedCost int
+		wantErr      bool
+	}{
+		{
+			name: "valid cost",
+			// bcrypt hash of the value "test1" with cost 12
+			inputHash:    []byte("$2a$12$M2Ot95Qty1MuQdubh1acWOiYadJDzeVg3ve4n5b.dgcgPdjCseKx2"),
+			expectedCost: recCost,
```
rithujohn191: Shouldn't wantErr=false here?

PR #1008, opened 2017-08-01
connector/github: prevent users not in org from authenticating

### LINKED ISSUE #920: connectors/github: Authentication via GitHub allows users outside of Org to login
Only allow users of the configured organization should be allowed to authenticate / login. This is not the case now.

estroz:
connector/github: check if user is a member of an organization, and if not return an error that can be type-checked against in callback handler

server: handle case where user is not an organization member by returning an appropriate error message

Fixes #920

### DIFF
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -214,6 +214,14 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
 	}
 
 	if s.Groups && c.org != "" {
+		inOrg, err := c.userInOrg(ctx, client, c.org)
+		if err != nil {
+			return identity, err
+		}
+		// User was not in org, return related error
+		if !inOrg {
+			return identity, fmt.Errorf("github: user is not an organization member")
+		}
 		groups, err := c.teams(ctx, client, c.org)
 		if err != nil {
 			return identity, fmt.Errorf("github: get teams: %v", err)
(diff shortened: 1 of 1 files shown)


connector/github/github.go
```diff
@@ -176,6 +176,16 @@ func newHTTPClient(rootCA string) (*http.Client, error) {
 	}, nil
 }
 
+// ErrForceOrgAddRequest is returned when a user not part of an org requests
+// access to an app requiring org membership
+type ErrForceOrgAddRequest struct {
```
ericchiang: doesn't need to be exported

connector/github/github.go
```diff
@@ -372,3 +390,68 @@ func (c *githubConnector) teams(ctx context.Context, client *http.Client, org st
...
+
+		// https://developer.github.com/v3/orgs/#list-your-organizations
+		var orgs []struct {
+			Name string `json:"login"`
+		}
+		if err := json.NewDecoder(resp.Body).Decode(&orgs); err != nil {
+			return false, fmt.Errorf("github: unmarshal groups: %v", err)
+		}
+
+		fmt.Println("userInOrg: orgs:", orgs)
```
ericchiang: need to fix this.

server/handlers.go
```diff
@@ -375,11 +376,17 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
 	}
 
 	if err != nil {
+		errMsg, errCode := "Failed to return user's identity.", http.StatusInternalServerError
+		// Prevent users not in an organization from logging into an application
+		if _, ok := err.(github.ErrForceOrgAddRequest); ok {
+			errMsg, errCode = "User is not an organization member.", http.StatusForbidden
```
ericchiang: I wouldn't hard code this case. Other connectors, e.g. LDAP, will want to differentiate between internal errors and forbidden. If we're going to solve this we should do it in a general way. Probably out of scope for this PR.

ericchiang:
Does https://github.com/coreos/dex/pull/1013 cover this?

estroz:
@ericchiang yup, I left a note in #1013 that this is obsolete. I'll close now.

PR #1010, opened 2017-08-01
connector/gitlab: fix default scope and API URL structure

### LINKED ISSUE #1009: Gitlab: scopes array is hard-coded
Scopes in the Gitlab connector are [hard-coded](https://github.com/coreos/dex/blob/master/connector/gitlab/gitlab.go#L81) to give `api` level access only. The `api` scope is the most [permissive](https://gitlab.com/gitlab-org/gitlab-ce/issues/31740) which might not be desirable in some cases. The default scope should be the least permissive scope `read_user`; the application should have to specify full API access themselves.

Furthermore, both scope keys at the top of [gitlab.go](https://github.com/coreos/dex/blob/master/connector/gitlab/gitlab.go#L20-L21) do not apply to the Gitlab OAuth API. They seem to be copies of scope keys from the Github connector.

estroz:
connector/gitlab: removed hard-coded default scope and fixed API URL structure

Documentation: updated connector field examples, added enterprise instructions.

Set the default scope to 'read_user' instead of providing full API access via the 'api' scope. Only grant full API access if a user sets the 'Groups' scope. Also refactored API URL structure in both gitlabConnector and query functions.

Made gitlab connector config more like githubs': enterprise support, group support

Tests: manually tested with `read_user` and `api` scopes by changing both example-app scopes, and GitLab OAuth app authorization levels in the GitLab OAuth app console.

Fixes #1009

### DIFF
--- a/Documentation/gitlab-connector.md
+++ b/Documentation/gitlab-connector.md
@@ -20,10 +20,42 @@ connectors:
     # Required field for connector name.
     name: GitLab
     config:
-      # optional, default = https://www.gitlab.com 
-      baseURL: https://www.gitlab.com
       # Credentials can be string literals or pulled from the environment.  
-      clientID: $GITLAB_APPLICATION_ID 
+      clientID: $GITLAB_APPLICATION_ID
       clientSecret: $GITLAB_CLIENT_SECRET
       redirectURI: http://127.0.0.1:5556/dex/callback
+      # Optional group, communicate through the "groups" scope.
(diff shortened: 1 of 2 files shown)


ericchiang:
Have you tested this?

estroz:
@ericchiang not yet. Got sidetracked, will do before merge.

rithujohn191:
Please split this out into separate PRs.

Issue #975, opened 2017-06-21
Add detailed connector tests


rithujohn191:
We do not have tests for github and gitlab connectors. We also need to add more unit tests to the existing oidc connector tests.

PR #1011, opened 2017-08-02
connector/github: test LoginURL

estroz:
Test case addressing the GitHub connector's `LoginURL()` function.

Relates to #975

### DIFF
--- a/connector/github/github_test.go
+++ b/connector/github/github_test.go
@@ -1,0 +1,93 @@
+package github
+
+import (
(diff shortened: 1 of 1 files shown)


ericchiang:
Why is this test useful? It looks like it's just another way to express this logic.

rithujohn191:
Going to close this for now. We need to add more tests to verify functionality of the GitHub connector to make this useful.

PR #1013, opened 2017-08-03
connector/github: multiple orgs, query by teams

estroz:
connector/github: support for multiple orgs and teams within orgs; allow filtering by both orgs and teams

Documentation: examples of GitHub `orgs` field with multiple orgs and org with teams

The GitHub connector should give more fine-grained control over which organizations and teams have access to an application. Restricting authorization to (multiple) `teams` within (multiple) `orgs` by specifying them in the `connector` config section gives dex users such control.

Tests: set up organizations with/out teams and a test user; went through scenarios listed in this [function description](https://github.com/estroz/dex/blob/6cab4ada83802b80eb21e3fdcb3123a778d1c196/connector/github/github.go#L286-L288)

Note: this PR obsoletes #1008 because similar errors are returned if user is not a org member

### DIFF
--- a/Documentation/github-connector.md
+++ b/Documentation/github-connector.md
@@ -9,7 +9,6 @@ When a client redeems a refresh token through dex, dex will re-query GitHub to u
 ## Caveats
 
 * Please note that in order for a user to be authenticated via GitHub, the user needs to mark their email id as public on GitHub. This will enable the API to return the user's email to Dex.
-* Currently, authentication via GitHub allows users outside of the `Org` specified in the connector to login. This is being tracked by [issue #920][issue-920].
 
 ## Configuration
 
@@ -29,11 +28,29 @@ connectors:
     clientID: $GITHUB_CLIENT_ID
     clientSecret: $GITHUB_CLIENT_SECRET
     redirectURI: http://127.0.0.1:5556/dex/callback
-    # Optional organization to pull teams from, communicate through the
-    # "groups" scope.
+    # Optional organizations and teams, communicated through the "groups" scope.
     #
     # NOTE: This is an EXPERIMENTAL config option and will likely change.
-    org: my-oranization
+    #
+    # Legacy 'org' field. 'org' and 'orgs' cannot be used simultaneously. A user
+    # MUST be a member of the following org to authenticate with dex.
+    # org: my-organization
+    #
+    # Dex queries the following organizations for group information if the
+    # "groups" scope is provided. Group claims are formatted as "(org):(team)".
+    # For example if a user is part of the "engineering" team of the "coreos"
+    # org, the group claim would include "coreos:engineering".
+    #
+    # A user MUST be a member of at least one of the following orgs to
+    # authenticate with dex.
+    orgs:
+    - name: my-organization
+      # Include all teams as claims.
+    - name: my-organization-with-teams
+      # A white list of teams. Only include group claims for these teams.
+      teams:
+      - read-team
+      - blue-team
 ```
 
 ## GitHub Enterprise
@@ -54,12 +71,29 @@ connectors:
     clientID: $GITHUB_CLIENT_ID
     clientSecret: $GITHUB_CLIENT_SECRET
     redirectURI: http://127.0.0.1:5556/dex/callback
-    # Optional organization to pull teams from, communicate through the
-    # "groups" scope.
+    # Optional organizations and teams, communicated through the "groups" scope.
     #
     # NOTE: This is an EXPERIMENTAL config option and will likely change.
-    org: my-oranization
-
+    #
+    # Legacy 'org' field. 'org' and 'orgs' cannot be used simultaneously. A user
+    # MUST be a member of the following org to authenticate with dex.
+    # org: my-organization
+    #
+    # Dex queries the following organizations for group information if the
+    # "groups" scope is provided. Group claims are formatted as "(org):(team)".
+    # For example if a user is part of the "engineering" team of the "coreos"
+    # org, the group claim would include "coreos:engineering".
+    #
+    # A user MUST be a member of at least one of the following orgs to
+    # authenticate with dex.
+    orgs:
+    - name: my-organization
+      # Include all teams as claims.
+    - name: my-organization-with-teams
+      # A white list of teams. Only include group claims for these teams.
+      teams:
+      - read-team
+      - blue-team
     # Required ONLY for GitHub Enterprise.
     # This is the Hostname of the GitHub Enterprise account listed on the
     # management console. Ensure this domain is routable on your network.
@@ -70,4 +104,3 @@ connectors:
 ```
 
 [github-oauth2]: https://github.com/settings/applications/new
-[issue-920]: https://github.com/coreos/dex/issues/920
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -35,15 +35,40 @@ type Config struct {
 	ClientSecret string `json:"clientSecret"`
 	RedirectURI  string `json:"redirectURI"`
 	Org          string `json:"org"`
+	Orgs         []Org  `json:"orgs"`
 	HostName     string `json:"hostName"`
 	RootCA       string `json:"rootCA"`
 }
 
+// Org holds org-team filters, in which teams are optional.
+type Org struct {
+
+	// Organization name in github (not slug, full name). Only users in this github
+	// organization can authenticate.
+	Name string `json:"name"`
+
+	// Names of teams in a github organization. A user will be able to
+	// authenticate if they are members of at least one of these teams. Users
+	// in the organization can authenticate if this field is omitted from the
+	// config file.
+	Teams []string `json:"teams,omitempty"`
+}
+
 // Open returns a strategy for logging in through GitHub.
 func (c *Config) Open(logger logrus.FieldLogger) (connector.Connector, error) {
+
+	if c.Org != "" {
+		// Return error if both 'org' and 'orgs' fields are used.
+		if len(c.Orgs) > 0 {
+			return nil, errors.New("github: cannot use both 'org' and 'orgs' fields simultaneously")
+		}
+		logger.Warnln("github: legacy field 'org' being used. Switch to the newer 'orgs' field structure")
+	}
+
 	g := githubConnector{
 		redirectURI:  c.RedirectURI,
 		org:          c.Org,
+		orgs:         c.Orgs,
 		clientID:     c.ClientID,
 		clientSecret: c.ClientSecret,
 		apiURL:       apiURL,
@@ -89,6 +114,7 @@ var (
 type githubConnector struct {
 	redirectURI  string
 	org          string
+	orgs         []Org
 	clientID     string
 	clientSecret string
 	logger       logrus.FieldLogger
@@ -213,10 +239,23 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
 		EmailVerified: true,
 	}
 
-	if s.Groups && c.org != "" {
-		groups, err := c.teams(ctx, client, c.org)
-		if err != nil {
-			return identity, fmt.Errorf("github: get teams: %v", err)
+	if s.Groups {
+		var groups []string
+		if len(c.orgs) > 0 {
+			if groups, err = c.listGroups(ctx, client, username); err != nil {
+				return identity, err
+			}
+		} else if c.org != "" {
+			inOrg, err := c.userInOrg(ctx, client, username, c.org)
+			if err != nil {
+				return identity, err
+			}
+			if !inOrg {
+				return identity, fmt.Errorf("github: user %q not a member of org %q", username, c.org)
+			}
+			if groups, err = c.teams(ctx, client, c.org); err != nil {
+				return identity, fmt.Errorf("github: get teams: %v", err)
+			}
 		}
 		identity.Groups = groups
 	}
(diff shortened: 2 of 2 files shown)


connector/github/github.go
```diff
@@ -34,16 +34,22 @@ type Config struct {
 	ClientID     string `json:"clientID"`
 	ClientSecret string `json:"clientSecret"`
 	RedirectURI  string `json:"redirectURI"`
-	Org          string `json:"org"`
```
ericchiang: We should keep this field and behavior around to not break existing users.

connector/github/github.go
```diff
@@ -34,16 +34,22 @@ type Config struct {
...
+	Orgs         []Org  `json:"orgs"`
 	HostName     string `json:"hostName"`
 	RootCA       string `json:"rootCA"`
 }
 
+// Org holds org-team filters, in which teams are optional
```
ericchiang: Need to explain the impact of providing the teams field. Also the logic about what teams or orgs a user has to belong to to login.

connector/github/github.go
```diff
@@ -233,37 +243,82 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
+}
 
-	if s.Groups && c.org != "" {
-		groups, err := c.teams(ctx, client, c.org)
+// isUserAuthorized enforces org and team constraints on user authorization
+// Cases in which user is authorized:
+// 	N orgs, no teams: user is member of at least 1 org
+// 	N orgs, M teams per org: user is member of any team from at least 1 org
+// 	N-1 orgs, M teams per org, 1 org with no teams: user is member of any team from at least 1 org, or member of org with no teams
+func (c *githubConnector) isUserAuthorized(ctx context.Context, client *http.Client, userName string) (groups []string, isAuthorized bool, err error) {
```
ericchiang: Method name could probably be more specific. Maybe just `listGroups`

Also two orgs can have teams with the same name. We probably need to come up with some naming schema for the groups. e.g. `( org ):( team )` `"coreos:engineers"`

connector/github/github.go
```diff
@@ -303,11 +358,82 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
+	defer resp.Body.Close()
+
+	err = nil
+	switch resp.StatusCode {
+	case http.StatusNoContent, http.StatusFound, http.StatusNotFound:
+	default:
+		err = fmt.Errorf("unexpected return status (code %d): %q", resp.StatusCode, resp.Status)
+	}
+
+	// 204 if user is a member
```
ericchiang: Ha. Fun API.

connector/github/github.go
```diff
@@ -303,11 +358,82 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
+	req = req.WithContext(ctx)
+	resp, err := client.Do(req)
+	if err != nil {
+		return false, fmt.Errorf("github: get teams: %v", err)
+	}
+	defer resp.Body.Close()
+
+	err = nil
+	switch resp.StatusCode {
+	case http.StatusNoContent, http.StatusFound, http.StatusNotFound:
```
ericchiang: Probably want to map these to a message about what they represent https://developer.github.com/v3/orgs/members/#check-membership

connector/github/github.go
```diff
@@ -303,11 +358,82 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
+		return false, fmt.Errorf("github: new req: %v", err)
+	}
+	req = req.WithContext(ctx)
+	resp, err := client.Do(req)
+	if err != nil {
+		return false, fmt.Errorf("github: get teams: %v", err)
+	}
+	defer resp.Body.Close()
+
+	err = nil
```
ericchiang: err is already nil

connector/github/github.go
```diff
@@ -233,37 +243,82 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
+		if !inOrg {
+			c.logger.Debugf("github: user %q not in org %q", userName, org.Name)
+			continue
+		}
+		if len(org.Teams) == 0 {
+			// User is in at least in one org with no team restrictions
+			isAuthorized = true
+			groups = append(groups, teams...)
+			c.logger.Debugf("github: user %q in org %q", userName, org.Name)
+			continue
```
ericchiang: Really confused by this `continue` and the one following it. Not clear what you're trying to do here.

connector/github/github.go
```diff
@@ -303,11 +358,82 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
+	userTeams, err := c.teams(ctx, client, org.Name)
+	if err != nil {
+		return nil, true, err
+	}
+	// Return all teams user is a member of if no team restrictions
+	if len(org.Teams) == 0 {
+		return userTeams, true, nil
+	}
+
+	wantedTeams := make(map[string]struct{})
```
ericchiang: `wantedTeams` is probably a bad name here. Can probably just write this as a method

```
func filterTeams(whiteList, userTeams []string) []string
```

connector/github/github.go
```diff
@@ -303,11 +358,82 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
 
+// userInOrgTeams first queries the GitHub API for a users' org membership, and
+// if user is an org member, queries the API for team membership within the org.
+// userInOrgTeams will return false if user is not in the org, and an empty
+// slice if user is in the org but not any teams in the org
+//
+// The HTTP passed client is expected to be constructed by the golang.org/x/oauth2 package,
+// which inserts a bearer token as part of the request.
+func (c *githubConnector) userInOrgTeams(ctx context.Context, client *http.Client, userName string, org Org) ([]string, bool, error) {
+	if inOrg, err := c.userInOrg(ctx, client, userName, org.Name); err != nil || !inOrg {
```
ericchiang: Can this logic just be done in `isUserAuthorized`?

connector/github/github.go
estroz: Status(Not)Found mean the same thing in our context (requester == user), so I'll return a 'user not in org' error

connector/github/github.go
ericchiang: You've still removed the feature.

connector/github/github.go
```diff
@@ -233,37 +261,121 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
+			c.logger.Errorf("user in org: %s", err)
+			continue
+		}
+		if !inOrg {
+			c.logger.Debugf("github: user %q not in org %q", userName, org.Name)
+			continue
+		}
+
+		if teams, err = c.teams(ctx, client, org.Name); err != nil {
+			c.logger.Errorf("teams: %s", err)
```
ericchiang: Just return this error immediately.

connector/github/github.go
```diff
@@ -233,37 +261,121 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
+func (c *githubConnector) listGroups(ctx context.Context, client *http.Client, userName string) ([]string, error) {
+	var (
+		inOrg         bool
+		teams, groups []string
+		err           error
+	)
+	for _, org := range c.orgs {
+		inOrg, err = c.userInOrg(ctx, client, userName, org.Name)
+		if err != nil {
+			c.logger.Errorf("user in org: %s", err)
```
ericchiang: Just return this error immediately.

connector/github/github.go
```diff
@@ -233,37 +261,121 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
+
+		if teams, err = c.teams(ctx, client, org.Name); err != nil {
+			c.logger.Errorf("teams: %s", err)
+			continue
+		}
+
+		// User is in at least in one org. No team restrictions are imposed from
+		// config. We add all the teams in the organization the user is a member of
+		// to 'groups' to preserve legacy behavior.
+		if len(org.Teams) == 0 {
```
ericchiang:
```
if len(org.Teams) > 0 {
    teams = filterTeams(teams, org.Teams)
}

groups = append(groups, joinOrgToTeams(org.Name, teams)...)
```

No need to have these two `continue` statements.

connector/github/github.go
```diff
@@ -233,37 +261,121 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
+		groups = append(groups, joinOrgToTeams(org.Name, teams)...)
+		c.logger.Debugf("github: user %q in org %q and teams %s", userName, org.Name, teams)
+	}
+
+	return groups, nil
+}
+
+// Orgs might have the same team names. We append the orgName to team name
+// delimited by a colon, "org:team", to make team names unique across orgs.
+func joinOrgToTeams(orgName string, teamNames []string) []string {
```
ericchiang: doesn't need to be it's own method.

connector/github/github.go
estroz: How about something like:
```go
orgPrefix := org.Name + ":"
if len(org.Teams) > 0 {
  for _, teamName := range filterTeams(teams, org.Teams) {
    groups = append(groups, orgPrefix+teamName)
  }
} else {
  groups = append(groups, orgPrefix)
}
```
If teams are specified in the config file, then check if the user is a team member and add those teams to groups. Otherwise, use an org-level group `"orgName:"` to indicate that the user is an org member.

connector/github/github.go
```diff
@@ -313,55 +313,36 @@ func (c *githubConnector) listGroups(ctx context.Context, client *http.Client, u
...
+		// i.e. "org:team", to make team names unique across orgs.
+		orgPrefix := org.Name + ":"
+		// User is in at least in one org. If team restrictions are present in config,
+		// ensure user is a member of at least one team.
+		if len(org.Teams) > 0 {
+			for _, teamName := range filterTeams(teams, org.Teams) {
+				groups = append(groups, orgPrefix+teamName)
+				c.logger.Debugf("github: user %q in org %q team %q", userName, org.Name, teamName)
+			}
+		} else { // No team restrictions. Append "org:" to groups
```
ericchiang: I really dislike this. Even if we're going to do this, we should always add the "(org):" or even just "(org)"

connector/github/github.go
ericchiang: @estroz btw this behavior _has_ changed. You'll start prefixing "(org):" to any group.

estroz: Adding legacy logic now.

Documentation/github-connector.md
```diff
@@ -29,11 +29,20 @@ connectors:
...
+    #
+    # Legacy 'org' field. 'org' and 'orgs' cannot be used simultaneously.
+    # org: my-organization
+    #
+    orgs:
+      - name: my-organization-with-teams
+        teams:
+          - read-team
+          - blue-team
+      - name: my-organization-no-teams
```
ericchiang: Need to document why these two are different.

Documentation/github-connector.md
```diff
@@ -29,11 +29,20 @@ connectors:
     clientID: $GITHUB_CLIENT_ID
     clientSecret: $GITHUB_CLIENT_SECRET
     redirectURI: http://127.0.0.1:5556/dex/callback
-    # Optional organization to pull teams from, communicate through the
-    # "groups" scope.
+    # Optional organizations and teams, communicated through the
+    # "groups" scope. The 'org' field is kept for legacy users.
```
ericchiang: Lots of "legacy" references. I'm starting to feel old :)

We can probably just drop these docs.

Documentation/github-connector.md
```diff
@@ -29,11 +28,26 @@ connectors:
...
     # "groups" scope.
     #
     # NOTE: This is an EXPERIMENTAL config option and will likely change.
-    org: my-oranization
+    #
+    # Legacy 'org' field. 'org' and 'orgs' cannot be used simultaneously. Use
+    # this field to restrict access to the specified GitHub org.
+    # org: my-organization
+    #
+    # New 'orgs' field. 'orgs' supports organizations with or without teams.
```
ericchiang: Few nits.

* This field is 'new' in the release notes, for docs it just is.
* "restrict access" is probably the wrong phrase.
* Let's just remove the `NOTE: ...` line and `Legacy 'org' field...` block.

Maybe something like this:

```yaml
# Query the following organizations for group information if the "groups"
# scope is provided. Group claims are formatted as "(org):(team)". For
# example if a user is part of the "engineering" team of the "coreos" org,
# the group claim would include "coreos:engineering"
#
# A user MUST be a member of at least one of the following orgs to
# authenticate with dex.
orgs:
- name: my-org
  # Include all teams as claims.
- name: my-org-with-teams
  # A white list of teams. Only include group claims for these teams.
  teams:
  - read-team
  - blue-team
```

ericchiang:
@estroz lgtm. Have you tested this?

estroz:
@ericchiang yes, although I'm going through the process once more to be sure.

connector/github/github.go
```diff
@@ -233,37 +278,101 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
+
+// listGroups enforces org and team constraints on user authorization
+// Cases in which user is authorized:
+// 	N orgs, no teams: user is member of at least 1 org
+// 	N orgs, M teams per org: user is member of any team from at least 1 org
+// 	N-1 orgs, M teams per org, 1 org with no teams: user is member of any team
+// from at least 1 org, or member of org with no teams
+func (c *githubConnector) listGroups(ctx context.Context, client *http.Client, userName string) ([]string, bool, error) {
+	var (
+		inOrg, inOrgNoTeams bool
```
ericchiang: inOrgNoTeams is never set. It's always false.

connector/github/github.go
```diff
@@ -233,37 +278,101 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
+		teams, groups       []string
```
ericchiang: `inOrg`, `teams`, and `err` do not need to be defined outside the for loop. Just define them as used in the for loop,

dex-org-user:
@ericchiang tested with both `orgs` and `org` fields and made some updates. works as expected in all scenarios described by docs.

estroz:
@ericchiang tested with both `orgs` and `org` fields and made some updates. works as expected in all scenarios described by docs.

connector/github/github.go
```diff
@@ -233,37 +272,108 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
+	return identity, nil
+}
+
+// listGroups enforces org and team constraints on user authorization
+// Cases in which user is authorized:
+// 	N orgs, no teams: user is member of at least 1 org
+// 	N orgs, M teams per org: user is member of any team from at least 1 org
+// 	N-1 orgs, M teams per org, 1 org with no teams: user is member of any team
+// from at least 1 org, or member of org with no teams
+func (c *githubConnector) listGroups(ctx context.Context, client *http.Client, userName string) (groups []string, err error) {
```
ericchiang: If a user isn't in any of the orgs, this will still return `[]string{}, nil`

estroz: Good catch. Fixing now.

estroz: Fixed and tested.

PR #1018, opened 2017-08-07
connector/github: enable private, primary emails

### LINKED ISSUE #1017: GitHub: support email when the email isn't public
Per @simonbyrne's comment https://github.com/coreos/dex/pull/972#issuecomment-320718897

https://developer.github.com/v3/users/emails/

estroz:
connector/github: `user` email field will be populated even if user has a private (primary) email

Documentation: removed private emails caveats section

Tests: tested manually using a user with a private primary email.

Fixes #1017

### DIFF
--- a/Documentation/github-connector.md
+++ b/Documentation/github-connector.md
@@ -6,10 +6,6 @@ One of the login options for dex uses the GitHub OAuth2 flow to identify the end
 
 When a client redeems a refresh token through dex, dex will re-query GitHub to update user information in the ID Token. To do this, __dex stores a readonly GitHub access token in its backing datastore.__ Users that reject dex's access through GitHub will also revoke all dex clients which authenticated them through GitHub.
 
-## Caveats
-
-* Please note that in order for a user to be authenticated via GitHub, the user needs to mark their email id as public on GitHub. This will enable the API to return the user's email to Dex.
-
 ## Configuration
 
 Register a new application with [GitHub][github-oauth2] ensuring the callback URL is `(dex issuer)/callback`. For example if dex is listening at the non-root path `https://auth.example.com/dex` the callback would be `https://auth.example.com/dex/callback`.
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -29,6 +29,11 @@ const (
 	scopeOrgs  = "read:org"
 )
 
+// Pagination URL patterns
+// https://developer.github.com/v3/#pagination
+var reNext = regexp.MustCompile("<([^>]+)>; rel=\"next\"")
+var reLast = regexp.MustCompile("<([^>]+)>; rel=\"last\"")
+
 // Config holds configuration options for github logins.
 type Config struct {
 	ClientID     string `json:"clientID"`
@@ -380,6 +385,67 @@ func filterTeams(userTeams, configTeams []string) (teams []string) {
 	return
 }
 
+// get creates a "GET `apiURL`" request with context, sends the request using
+// the client, and decodes the resulting response body into v. A pagination URL
+// is returned if one exists. Any errors encountered when building requests,
+// sending requests, and reading and decoding response data are returned.
+func get(ctx context.Context, client *http.Client, apiURL string, v interface{}) (string, error) {
+	req, err := http.NewRequest("GET", apiURL, nil)
+	if err != nil {
+		return "", fmt.Errorf("github: new req: %v", err)
+	}
+	req = req.WithContext(ctx)
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("github: get URL %v", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return "", fmt.Errorf("github: read body: %v", err)
+		}
+		return "", fmt.Errorf("%s: %s", resp.Status, body)
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(v); err != nil {
+		return "", fmt.Errorf("failed to decode response: %v", err)
+	}
+
+	return getPagination(apiURL, resp), nil
+}
+
+// getPagination checks the "Link" header field for "next" or "last" pagination
+// URLs, and returns true only if a "next" URL is found. The next pages' URL is
+// returned if a "next" URL is found. apiURL is returned if apiURL equals the
+// "last" URL or no "next" or "last" URL are found.
+//
+// https://developer.github.com/v3/#pagination
+func getPagination(apiURL string, resp *http.Response) string {
+	if resp == nil {
+		return ""
+	}
+
+	links := resp.Header.Get("Link")
+	if len(reLast.FindStringSubmatch(links)) > 1 {
+		lastPageURL := reLast.FindStringSubmatch(links)[1]
+		if apiURL == lastPageURL {
+			return ""
+		}
+	} else {
+		return ""
+	}
+
+	if len(reNext.FindStringSubmatch(links)) > 1 {
+		return reNext.FindStringSubmatch(links)[1]
+	}
+
+	return ""
+}
+
+// user holds GitHub user information (relevant to dex) as defined by
+// https://developer.github.com/v3/users/#response-with-public-profile-information
 type user struct {
 	Name  string `json:"name"`
 	Login string `json:"login"`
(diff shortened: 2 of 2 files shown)


connector/github/github.go
```diff
@@ -411,12 +416,89 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
+// userEmail queries the GitHub API for a users' email information using the
+// provided client. Only returns the users' verified, primary email (private or
+// public).
+//
+// The HTTP client is expected to be constructed by the golang.org/x/oauth2 package,
+// which inserts a bearer token as part of the request.
+func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (string, error) {
+
+	// https://developer.github.com/v3/#pagination
+	reNext := regexp.MustCompile("<(.*)>; rel=\"next\"")
```
ericchiang: We should define these statically at the package level. Particularly since they're just copied from above.

connector/github/github.go
```diff
@@ -411,12 +416,89 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
 		return u, fmt.Errorf("%s: %s", resp.Status, body)
 	}
 
-	if err := json.NewDecoder(resp.Body).Decode(&u); err != nil {
+	if err = json.NewDecoder(resp.Body).Decode(&u); err != nil {
 		return u, fmt.Errorf("failed to decode response: %v", err)
 	}
+
+	if u.Email == "" {
+		if u.Email, err = c.userEmail(ctx, client); err != nil {
```
ericchiang: Add a comment here as well explaining why we make this call.

connector/github/github.go
```diff
@@ -411,12 +416,89 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
+
+	// https://developer.github.com/v3/#pagination
+	reNext := regexp.MustCompile("<(.*)>; rel=\"next\"")
+	reLast := regexp.MustCompile("<(.*)>; rel=\"last\"")
+	// https://developer.github.com/v3/users/emails/#list-email-addresses-for-a-user
+	apiURL := c.apiURL + "/user/emails"
+
+	var emails []userEmail
+	for {
+		req, err := http.NewRequest("GET", apiURL, nil)
```
ericchiang: Split this GET, handle bad response, deserialize response body into a helper and share it with the method above.

```go
func get(ctx context.Context, client *http.Client, apiURL, path string, resp interface{})
``` 

This method then becomes something like

```go
    var emails []userEmail
    if err := get(ctx, client, apiURL, "/user/emails", &emails); err != nil {
    }
```

connector/github/github.go
```diff
@@ -411,12 +416,89 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
+				return "", fmt.Errorf("github: read body: %v", err)
+			}
+			return "", fmt.Errorf("%s: %s", resp.Status, body)
+		}
+
+		if err := json.NewDecoder(resp.Body).Decode(&emails); err != nil {
+			return "", fmt.Errorf("failed to decode response: %v", err)
+		}
+
+		for _, email := range emails {
```
ericchiang: What's the ordering here? If the user has multiple emails how do we determine which one we choose? Does GitHub rank these in some way?

If GitHub's doesn't guarantee an ordering we should at least short them. But I wonder if there's a better way.

cc @simonbyrne

connector/github/github.go
```diff
@@ -411,12 +416,89 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
+
+	if u.Email == "" {
+		if u.Email, err = c.userEmail(ctx, client); err != nil {
+			return u, err
+		}
+	}
 	return u, nil
 }
 
+type userEmail struct {
```
ericchiang: docstring. need to explain what this is used for.

connector/github/github.go
estroz: You don't think the users' primary email is adequate?

ericchiang: Ah, missed that bit. Yeah that works for me.

Documentation/github-connector.md
```diff
@@ -8,7 +8,7 @@ When a client redeems a refresh token through dex, dex will re-query GitHub to u
 
 ## Caveats
 
-* Please note that in order for a user to be authenticated via GitHub, the user needs to mark their email id as public on GitHub. This will enable the API to return the user's email to Dex.
+* A user added to an organization __after__ they have granted an application access their GitHub information via dex will not be redirected to a 'request organization access' page. That user should log into their GitHub account, go to Settings -> Authorized OAuth Apps, click 'Revoke' to revoke the application's grant, and restart the dex login process. Dex will issue a new grant to the application with proper organization permissions. This issue arises because GitHub's OAuth redirect URI does not return a 'request organization access' page.
```
rithujohn191: I don't think this doc's change should go in as a part of this PR. The title of the PR and the above documentation don't have any connection. It can be confusing. For this PR just remove the line about the email id of the user having to be public

connector/github/github.go
```diff
@@ -411,12 +416,89 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
 		return u, fmt.Errorf("%s: %s", resp.Status, body)
 	}
 
-	if err := json.NewDecoder(resp.Body).Decode(&u); err != nil {
+	if err = json.NewDecoder(resp.Body).Decode(&u); err != nil {
```
rithujohn191: nit: using `err :=` in an if statement ensure that the variable is used only in that if statement. Did you remove the initialization for any particular reason?

connector/github/github.go
```diff
@@ -411,12 +416,89 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user,
...
+		if err := json.NewDecoder(resp.Body).Decode(&emails); err != nil {
+			return "", fmt.Errorf("failed to decode response: %v", err)
+		}
+
+		for _, email := range emails {
+			if email.Verified && email.Primary {
+				return email.Email, nil
+			}
+		}
+		links := resp.Header.Get("Link")
```
rithujohn191: Could you also add a comment here explaining why you are doing this get

connector/github/github.go
estroz: I changed this because my linter yells at me for shadowing `err` :sweat_smile: . `err` will be nil after this statement if not caught by the condition anyways, so I guess it's just a matter of style. Will change back.

connector/github/github.go
estroz: Still needed after refactor?

connector/github/github.go
```diff
@@ -380,43 +385,144 @@ func filterTeams(userTeams, configTeams []string) (teams []string) {
...
+// public).
+//
+// The HTTP client is expected to be constructed by the golang.org/x/oauth2 package,
+// which inserts a bearer token as part of the request.
+func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (string, error) {
+
+	// https://developer.github.com/v3/users/emails/#list-email-addresses-for-a-user
+	apiURL, resourcePath := c.apiURL, "/user/emails"
+
+	var (
```
ericchiang: You're overusing this pattern. Probably better to just define things when they're actually used. e.g. within the for loop.

connector/github/github.go
```diff
@@ -380,43 +385,144 @@ func filterTeams(userTeams, configTeams []string) (teams []string) {
...
-// client is expected to be constructed by the golang.org/x/oauth2 package, which inserts
-// a bearer token as part of the request.
-func (c *githubConnector) user(ctx context.Context, client *http.Client) (user, error) {
-	var u user
-	req, err := http.NewRequest("GET", c.apiURL+"/user", nil)
+// get creates a "GET `apiURL + path`" request with context, sends the request
+// using client, and decodes the resulting response body into v. The full HTTP
+// response object and any errors encountered are returned to facilitate post-
+// processing and error handling.
+func get(ctx context.Context, client *http.Client, apiURL, path string, v interface{}) (*http.Response, error) {
```
ericchiang: Why is this returning the response?

estroz: To check for [pagination links](https://github.com/estroz/dex/blob/763d087c42d08586cf7ef744ddf3a1476b6fb9f9/connector/github/github.go#L424)

connector/github/github.go
estroz: I'll move resource path strings into `get()` calls

connector/github/github.go
```diff
@@ -380,43 +385,139 @@ func filterTeams(userTeams, configTeams []string) (teams []string) {
...
+
+// getPagination checks the "Link" header field for "next" or "last" pagination
+// URLs, and returns true only if a "next" URL is found. The next pages' URL is
+// returned if a "next" URL is found. apiURL is returned if apiURL equals the
+// "last" URL or no "next" or "last" URL are found.
+//
+// https://developer.github.com/v3/#pagination
+func getPagination(apiURL string, resp *http.Response) (string, bool) {
+	if resp == nil {
+		return apiURL, false
```
ericchiang: nit: just return `"", false`

connector/github/github.go
ericchiang: Maybe do the pagination logic here and return something like `(nextPath string, err error)`?

And if the `nextPath != ""` there's another page.

connector/github/github.go
```diff
@@ -380,43 +385,139 @@ func filterTeams(userTeams, configTeams []string) (teams []string) {
...
+	ID    int    `json:"id"`
+	Email string `json:"email"`
+}
+
+// user queries the GitHub API for profile information using the provided client.
+//
+// The HTTP client is expected to be constructed by the golang.org/x/oauth2 package,
+// which inserts a bearer token as part of the request.
+func (c *githubConnector) user(ctx context.Context, client *http.Client) (user, error) {
+	var (
```
ericchiang: nit. Don't do this. Just do

```
var u user
if _, err := get(...); err != nil {

}
```

connector/github/github.go
```diff
@@ -380,43 +385,139 @@ func filterTeams(userTeams, configTeams []string) (teams []string) {
...
+
+// userEmail queries the GitHub API for a users' email information using the
+// provided client. Only returns the users' verified, primary email (private or
+// public).
+//
+// The HTTP client is expected to be constructed by the golang.org/x/oauth2 package,
+// which inserts a bearer token as part of the request.
+func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (string, error) {
+	apiURL := c.apiURL
+	var (
```
ericchiang: Same here. Don't define these at the beginning of the function.

connector/github/github.go
```diff
@@ -452,49 +553,30 @@ func (c *githubConnector) userInOrg(ctx context.Context, client *http.Client, us
...
-
-	groups := []string{}
-
-	// https://developer.github.com/v3/#pagination
-	reNext := regexp.MustCompile("<(.*)>; rel=\"next\"")
-	reLast := regexp.MustCompile("<(.*)>; rel=\"last\"")
-	apiURL := c.apiURL + "/user/teams"
-
+	apiURL, groups := c.apiURL, []string{}
+	var (
```
ericchiang: Same

connector/github/github.go
ericchiang: To explain a bit, it's not clear from the function method that the `*http.Response` should only be used to determine pagination information. Something like the above suggestion of changing the return values makes it more explicit.

connector/github/github.go
```diff
@@ -452,49 +551,29 @@ func (c *githubConnector) userInOrg(ctx context.Context, client *http.Client, us
...
-			} `json:"organization"`
-		}
-		if err := json.NewDecoder(resp.Body).Decode(&teams); err != nil {
-			return nil, fmt.Errorf("github: unmarshal groups: %v", err)
+		// https://developer.github.com/v3/orgs/teams/#list-user-teams
+		var (
+			teams []team
+			err   error
+		)
+		if apiURL, err = get(ctx, client, apiURL, "/user/teams", &teams); err != nil {
```
ericchiang: This logic looks wrong. If the pagination apiURL is `https://api.github.com/user/repos?page=3&per_page=100` then the next call will call `https://api.github.com/user/repos?page=3&per_page=100/user/teams`

estroz: Fixing.

estroz:
@ericchiang found a bug in the original code where the paginated links found in `Links` headers are structured like:
```
<https://example.github.com/user/emails?page=3&per_page=100>; rel="next",
<https://example.github.com/user/emails?page=4&per_page=100>; rel="last"
```
would be matched as (note the lack of opening/closing angled brackets):
```
https://example.github.com/user/emails?page=3&per_page=100>; rel="next",
<https://example.github.com/user/emails?page=4&per_page=100
```
because `reNext` and `reLast` were matching too greedily.
Switching out [`"<(.*)>; rel=\"next\""`](https://github.com/estroz/dex/blob/e2da50561227aebdb42e4ca3ebded4e9140fb6f9/connector/github/github.go#L34-L35) for `"<([^>]+)>; rel=\"next\""`.

ericchiang:
> found a bug in the original code where the paginated links found in Links headers are structured like:

not super surprised...

Thanks. If you've tested this feel free to merge.

PR #1019, opened 2017-08-08
Documentation: github org redirect caveat

estroz:
This PR updates GitHub connector docs to notify users of our current workaround relating to #920. Part of the issue, users outside of the organization being able to log into dex-authenticated apps, was solved by https://github.com/coreos/dex/pull/1013. 

Test cases for the above PR:
1. User not in a config org: fails
1. User in a config org but not in a config team: fails
1. User in a config org with no config teams: passes
1. User in a config org with no config teams, but already granted the app permission: **fails** <- issue
1. User in a config org and in a config team: passes

### DIFF
--- a/Documentation/github-connector.md
+++ b/Documentation/github-connector.md
@@ -6,6 +6,11 @@ One of the login options for dex uses the GitHub OAuth2 flow to identify the end
 
 When a client redeems a refresh token through dex, dex will re-query GitHub to update user information in the ID Token. To do this, __dex stores a readonly GitHub access token in its backing datastore.__ Users that reject dex's access through GitHub will also revoke all dex clients which authenticated them through GitHub.
 
+## Caveats
+
+* The GitHub API calls dex uses requires a user to have their organization membership visibility set to public. This can be done on the ["request access from org" page](github-request-org-access) which GitHub will skip if the client app is an existing authorized application. The current workaround is as follows: the user should log into their GitHub account, go to Settings -> Authorized OAuth Apps, click 'Revoke' to revoke the application's grant, and restart the dex login process. This will force the "request access from org" page to be shown, allowing the user to request that the organization owner make their membership public.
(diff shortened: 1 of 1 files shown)


ericchiang:
> User in an org but not in any teams: fails

This should be, passes, right?

Also this diff conflicts with https://github.com/coreos/dex/pull/1018. Let's get that merged first.

Documentation/github-connector.md
```diff
@@ -8,7 +8,8 @@ When a client redeems a refresh token through dex, dex will re-query GitHub to u
 
 ## Caveats
 
-* Please note that in order for a user to be authenticated via GitHub, the user needs to mark their email id as public on GitHub. This will enable the API to return the user's email to Dex.
+* A user added to a GitHub organization __after__ they have granted an application access their GitHub information via dex will not be redirected to a 'request organization access' page. This issue arises because GitHub's OAuth redirect URI does not return a 'request organization access' page. The current workaround is as follows: the user should log into their GitHub account, go to Settings -> Authorized OAuth Apps, click 'Revoke' to revoke the application's grant, and restart the dex login process. Dex will issue a new grant to the application with proper organization permissions.
```
rithujohn191: they have granted an application access <to> their GitHub....

rithujohn191: I think you should add an introductory line stating that "the github api call requires the user to have their org membership to be public. this can be done by requesting an org for access....." something along those line

rithujohn191: Also the last line "Dex will issue a new grant to the application with proper organization permissions." is not correct. Once the app is revoked, when the user tries to login via GitHub....Github explicitly asks the user to grant permission...no permissions are automatically assigned.

estroz:
@ericchiang updated test case descriptions to be more specific. If a user is in an org but not in any of the teams listed in the config, they should not get access.

Issue #719, opened 2016-12-02
connector/ldap: log as much meaningful information as possible when connecting


sym3tri:
To help users debug bad LDAP configurations Dex should log as much info as possible.

ericchiang:
Yes definitely, but to a certain degree, I don't see us being able to provide more information without providing the option for leveled logging (https://github.com/coreos/dex/issues/677). Log as much meaningful information as possible is pretty much the definition of a "debug" flag.

sym3tri:
definitely makes sense to block this on #677

PR #1021, opened 2017-08-11
*: add log events for login, LDAP queries, and SAML responses

ericchiang:
Add some log events to help debug common mis-configuration issues.

Will do a login through SAML and LDAP providers to show what these logs look like.

### DIFF
--- a/connector/saml/saml.go
+++ b/connector/saml/saml.go
@@ -366,6 +366,11 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse, inResponseTo str
 		return ident, fmt.Errorf("response did not contain a AttributeStatement")
 	}
 
+	// Log the actual attributes we got back from the server. This helps debug
+	// configuration errors on the server side, where the SAML server doesn't
+	// send us the correct attributes.
+	p.logger.Infof("parsed and verified saml response attributes %s", attributes)
+
 	// Grab the email.
 	if ident.Email, _ = attributes.get(p.emailAttr); ident.Email == "" {
 		return ident, fmt.Errorf("no attribute with name %q: %s", p.emailAttr, attributes.names())
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -390,6 +390,8 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
 	http.Redirect(w, r, redirectURL, http.StatusSeeOther)
 }
 
+// finalizeLogin associates the user's identity with the current AuthRequest, then returns
+// the approval page's path.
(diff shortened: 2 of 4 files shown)


server/handlers.go
```diff
@@ -408,6 +410,15 @@ func (s *Server) finalizeLogin(identity connector.Identity, authReq storage.Auth
 	if err := s.storage.UpdateAuthRequest(authReq.ID, updater); err != nil {
 		return "", fmt.Errorf("failed to update auth request: %v", err)
 	}
+
+	email := claims.Email
+	if !claims.EmailVerified {
+		email = email + " (unverified)"
+	}
+
+	s.logger.Infof("login: connector %q, username=%q, email=%q, groups=%q",
```
rithujohn191: Maybe we could change this to "login successful: connector....." just so that it will be clear that it went through correctly

connector/saml/saml.go
```diff
@@ -366,6 +366,8 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse, inResponseTo str
 		return ident, fmt.Errorf("response did not contain a AttributeStatement")
 	}
 
+	p.logger.Infof("parsed and verified saml response attributes %s", attributes)
```
rithujohn191: Does this just display all the attributes together? Could we display this as emailAttr:%s, usernameAttr:%s and groupsAttr:%s?

ericchiang: The intent here is to spit out all the attributes that the server actually sent us. Helps us identify "you didn't configure your SAML provider to send us the correct set of information."

If the admin has done the correct mapping the attributes will show ups in the "login successful:" entry you commented on above anyway.

ericchiang: Maybe this shows that I should add a comment :)

rithujohn191: Would help :) attributes are usually what always gets misconfigured :(

ericchiang:
This is what an LDAP login now looks like

```
time="2017-08-11T18:26:04Z" level=info msg="performing ldap search ou=People,dc=example,dc=org sub (&(objectClass=person)(mail=janedoe@example.com))" connector=OpenLDAP 
time="2017-08-11T18:26:04Z" level=info msg="username \"janedoe@example.com\" mapped to entry cn=jane,ou=People,dc=example,dc=org" connector=OpenLDAP 
time="2017-08-11T18:26:04Z" level=info msg="performing ldap search ou=Groups,dc=example,dc=org sub (&(objectClass=groupOfNames)(member=cn=jane,ou=People,dc=example,dc=org))" connector=OpenLDAP 
time="2017-08-11T18:26:04Z" level=info msg="login: connector \"ldap\", username=\"jane\", email=\"janedoe@example.com\", groups=[\"admins\" \"developers\"]" 
```

estroz:
Fixes #719

PR #1026, opened 2017-08-15
added possibility to choose connector in request from client

sdrozdkov:
Perhaps this will be useful to someone else.
I use this little hack for different groups filter (LDAP) for different client applications (dex used as OIDC for internal SSO)

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -167,6 +167,11 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if authReq.ConnectorID != "" {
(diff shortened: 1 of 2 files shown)


ericchiang:
We've consider this in the past, but this has a lot of implications and questions around guarantees. Should it be a scope? Client configuration? Should a user be able to change this value? Do we have to guarantee that the user can't use other connectors? Can an end user see the other connectors? Would a client be surprised if they used this feature and a user logged in through a different connector?

Going to close. I think a change like this deserves some discussions around use cases that I'd rather see in a issue. After we figure out that we can review the implementation.

PR #1031, opened 2017-08-17
Documentation: fix redirect caveat description

estroz:
Part of the caveat currently described by GitHub connector documentation is no longer an issue, but the part relating to re-authorization of an application after being added to an organization is still valid.

### DIFF
--- a/Documentation/github-connector.md
+++ b/Documentation/github-connector.md
@@ -8,8 +8,7 @@ When a client redeems a refresh token through dex, dex will re-query GitHub to u
 
 ## Caveats
 
-* The GitHub API calls dex uses requires a user to have their organization membership visibility set to public. This can be done on the ["request access from org" page][github-request-org-access] which GitHub will skip if the client app is an existing authorized application. The current workaround is as follows: the user should log into their GitHub account, go to Settings -> Authorized OAuth Apps, click 'Revoke' to revoke the application's grant, and restart the dex login process. This will force the "request access from org" page to be shown, allowing the user to request that the organization owner make their membership public.
-    * Note: GitHub [organizations][github-orgs] are different from GitHub [teams][github-teams]
+* A user must explicitly [request][github-request-org-access] an [organization][github-orgs] give dex [resource access][github-approve-org-access]. Dex will not have the correct permissions to determine if the user is in that organization otherwise, and the user will not be able to log in. This request mechanism is a feature of the GitHub API.
 
 ## Configuration
 
@@ -50,7 +49,7 @@ connectors:
     - name: my-organization-with-teams
       # A white list of teams. Only include group claims for these teams.
       teams:
-      - read-team
+      - red-team
       - blue-team
 ```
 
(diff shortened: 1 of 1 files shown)


Documentation/github-connector.md
```diff
@@ -8,7 +8,7 @@ When a client redeems a refresh token through dex, dex will re-query GitHub to u
...
+* A user added to a GitHub organization __after__ granting an application access to their GitHub information may see a "Failed to return user's identity" response when using dex. This issue arises because GitHub's OAuth redirect URI does not redirect users to a "request organization access" page, and only occurs when the application owner has populated either the `orgs` or `org` field in their configuration file. The current workaround is as follows: the user should log into their GitHub account, go to Settings -> Authorized OAuth Apps, click "Revoke" to revoke the application's grant, and restart the dex login process. This will force the "request access from org" page to be shown, allowing the user to request that an organization owner permit access by the application to their organization.
```
ericchiang: This needs to be way shorter and describe the access issues.

Documentation/github-connector.md
```diff
@@ -8,8 +8,9 @@ When a client redeems a refresh token through dex, dex will re-query GitHub to u
 
 ## Caveats
 
-* The GitHub API calls dex uses requires a user to have their organization membership visibility set to public. This can be done on the ["request access from org" page][github-request-org-access] which GitHub will skip if the client app is an existing authorized application. The current workaround is as follows: the user should log into their GitHub account, go to Settings -> Authorized OAuth Apps, click 'Revoke' to revoke the application's grant, and restart the dex login process. This will force the "request access from org" page to be shown, allowing the user to request that the organization owner make their membership public.
-    * Note: GitHub [organizations][github-orgs] are different from GitHub [teams][github-teams]
+* A user must explicitly [request][github-request-org-access] that an organization give an application [access to that organizations' data][github-approve-org-access]. Dex will not have the correct permissions to determine if the user is in that organization otherwise, and the user will not be able to log in. This request mechanism is a feature of the GitHub API.
```
rithujohn191: Please abbreviate the first line "A user must explicitly request an organization to give resource access to Dex".

Documentation/github-connector.md
```diff
@@ -8,8 +8,9 @@ When a client redeems a refresh token through dex, dex will re-query GitHub to u
...
-    * Note: GitHub [organizations][github-orgs] are different from GitHub [teams][github-teams]
+* A user must explicitly [request][github-request-org-access] an organization give dex [resource access][github-approve-org-access]. Dex will not have the correct permissions to determine if the user is in that organization otherwise, and the user will not be able to log in. This request mechanism is a feature of the GitHub API.
+* A user added to a GitHub organization __after__ granting an application access to their GitHub information may see a "Failed to return user's identity" response when using dex. The current workaround is as follows: the user should log into GitHub, go to Settings -> Authorized OAuth Apps, click "Revoke" to revoke the application's grant, and restart the dex login process. The user should see a "request access" page and allow the user to request access to an organization's data.
```
ericchiang: I'd almost vote we just drop this. It's important when testing, but practically it's not for real use cases.

estroz: If you think it's unnecessary here then I can remove it. @rithujohn191 thoughts?

rithujohn191: The use case I was thinking about was users would add Dex as an authorized app and then add users to orgs and since we don't explicitly refresh this info today, the user would be not be successfully authorized. Maybe we can call that out somewhere else or in simpler words. I can see how this can be confusing

PR #1036, opened 2017-08-17
connector/github: debug->info logging, informative userInOrg msg

estroz:
connector/github: switch debug -> info log levels and add a more informative negative userInOrg log message

Info logs are more useful as they're always created. More informative userInOrg log message will help users troubleshoot login issues.

### DIFF
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -350,9 +350,8 @@ func (c *githubConnector) listGroups(ctx context.Context, client *http.Client, u
 		// 'teams' list in config.
 		if len(org.Teams) == 0 {
 			inOrgNoTeams = true
-			c.logger.Debugf("github: user %q in org %q", userName, org.Name)
 		} else if teams = filterTeams(teams, org.Teams); len(teams) == 0 {
-			c.logger.Debugf("github: user %q in org %q but no teams", userName, org.Name)
+			c.logger.Infof("github: user %q in org %q but no teams", userName, org.Name)
 		}
 
 		// Orgs might have the same team names. We append orgPrefix to team name,
@@ -360,7 +359,6 @@ func (c *githubConnector) listGroups(ctx context.Context, client *http.Client, u
 		orgPrefix := org.Name + ":"
 		for _, teamName := range teams {
 			groups = append(groups, orgPrefix+teamName)
-			c.logger.Debugf("github: user %q in org %q team %q", userName, org.Name, teamName)
 		}
 	}
 	if inOrgNoTeams || len(groups) > 0 {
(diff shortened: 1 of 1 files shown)


connector/github/github.go
```diff
@@ -350,17 +350,17 @@ func (c *githubConnector) listGroups(ctx context.Context, client *http.Client, u
 		// 'teams' list in config.
 		if len(org.Teams) == 0 {
 			inOrgNoTeams = true
-			c.logger.Debugf("github: user %q in org %q", userName, org.Name)
+			c.logger.Infof("github: user %q in org %q", userName, org.Name)
```
ericchiang: I don't actually think we want to log this info. It's going to spam the logs and we'll already get the teams that a user is apart of based on other log messages

connector/github/github.go
```diff
@@ -542,7 +542,7 @@ func (c *githubConnector) userInOrg(ctx context.Context, client *http.Client, us
 	switch resp.StatusCode {
 	case http.StatusNoContent:
 	case http.StatusFound, http.StatusNotFound:
-		c.logger.Debugf("github: user %q not in org %q", userName, orgName)
+		c.logger.Infof("github: user %q not in org %q or application not authorized to read org data", userName, orgName)
```
ericchiang: I think this is probably the only one we want to log by default, right?

connector/github/github.go
estroz: sgtm. will remove

connector/github/github.go
estroz: Yeah, it's the most relevant log message in here. I'll remove other log statements

connector/github/github.go
```diff
@@ -350,17 +350,17 @@ func (c *githubConnector) listGroups(ctx context.Context, client *http.Client, u
 		// 'teams' list in config.
 		if len(org.Teams) == 0 {
 			inOrgNoTeams = true
-			c.logger.Debugf("github: user %q in org %q", userName, org.Name)
+			c.logger.Infof("github: user %q in org %q", userName, org.Name)
 		} else if teams = filterTeams(teams, org.Teams); len(teams) == 0 {
-			c.logger.Debugf("github: user %q in org %q but no teams", userName, org.Name)
+			c.logger.Infof("github: user %q in org %q but no teams", userName, org.Name)
```
estroz: @ericchiang we might want to keep this. it logs info about a situation downstream to [the user not in an org](https://github.com/estroz/dex/blob/c4c11bdf340da5edf7114c8c1681b5c1f7487ad6/connector/github/github.go#L545) that could be relevant to someone looking at logs.

ericchiang: seems reasonable

PR #1038, opened 2017-08-17
When connecting to GitHub Enterprise, force email verified field to true

ChienHuey:
This change being made to account for GitHub Enterprise not supporting verified emails. If connecting to GitHub Enterprise (as indicated by hostName field being populated), the email.Verified field is set to true regardless of what comes back from the GitHub API.

### DIFF
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -503,6 +503,11 @@ func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (s
 		}
 
 		for _, email := range emails {
+			// if GitHub Enterprise, set email.Verified to true
+			if c.hostName != "" {
+				email.Verified = true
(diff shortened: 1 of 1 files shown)


ericchiang:
> ... GitHub Enterprise not supporting verified emails.

Are there some corresponding docs for this?

ChienHuey:
Found this when troubleshooting the issue https://circleci.com/enterprise/changelog/#1-47-1. CircleCI claims that they were advised not to check for verified emails. Also have a support case open with GitHub where I'm asking them to confirm this. Will update here when I get a response.

ericchiang:
:+1: thanks. can you add that link to the code comment and expand that comment to explain why we don't care about the verified status for enterprise?

PR #1039, opened 2017-08-17
connector/github: fix groups scope check when 'orgs' is populated

estroz:
connector/github: fix issue of skipped org membership check when client doesn't specify `groups` scope while using `orgs` config field.

Fix a bug where checking if a user is an organization member would be skipped if a client didn't provide the "groups" scope.

Revert behavior of "org" to not check if a user is an organization member. Admins that need to require a user be an organization member to log in should use the newly added "orgs" field.

Tests: went through tests listed in #1019 for `orgs`, and tested `org` to make sure users outside that org can log in (old behavior)

### DIFF
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -24,9 +24,12 @@ import (
 )
 
 const (
-	apiURL     = "https://api.github.com"
+	apiURL = "https://api.github.com"
+	// GitHub requires this scope to access '/user' and '/user/emails' API endpoints.
 	scopeEmail = "user:email"
-	scopeOrgs  = "read:org"
+	// GitHub requires this scope to access '/user/teams' and '/orgs' API endpoints
+	// which are used when a client includes the 'groups' scope.
+	scopeOrgs = "read:org"
 )
 
 // Pagination URL patterns
@@ -133,11 +136,18 @@ type githubConnector struct {
 	httpClient *http.Client
 }
 
+// groupsRequired returns whether dex requires GitHub's 'read:org' scope. Dex
+// needs 'read:org' if 'orgs' or 'org' fields are populated in a config file.
+// Clients can require 'groups' scope without setting 'orgs'/'org'.
+func (c *githubConnector) groupsRequired(groupScope bool) bool {
+	return len(c.orgs) > 0 || c.org != "" || groupScope
+}
+
 func (c *githubConnector) oauth2Config(scopes connector.Scopes) *oauth2.Config {
-	// The GitHub API requires requires read:org scope to ensure users are members
-	// of orgs and teams provided in configs.
+	// 'read:org' scope is required by the GitHub API, and thus for dex to ensure
+	// a user is a member of orgs and teams provided in configs.
 	githubScopes := []string{scopeEmail}
-	if len(c.orgs) > 0 || c.org != "" {
+	if c.groupsRequired(scopes.Groups) {
 		githubScopes = append(githubScopes, scopeOrgs)
 	}
 
@@ -206,22 +216,6 @@ func newHTTPClient(rootCA string) (*http.Client, error) {
 	}, nil
 }
 
-// getGroups retrieves GitHub orgs and teams a user is in, if any.
-func (c *githubConnector) getGroups(ctx context.Context, client *http.Client, groupScope bool, userLogin string) (groups []string, err error) {
-	// Org and team membership should always be checked if specified by config.
-	if len(c.orgs) > 0 {
-		if groups, err = c.listGroups(ctx, client, userLogin); err != nil {
-			return groups, err
-		}
-	} else if groupScope && c.org != "" {
-		// Do not check org membership here to preserve legacy behavior.
-		if groups, err = c.teams(ctx, client, c.org); err != nil {
-			return groups, fmt.Errorf("github: get teams: %v", err)
-		}
-	}
-	return groups, nil
-}
-
 func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (identity connector.Identity, err error) {
 	q := r.URL.Query()
 	if errType := q.Get("error"); errType != "" {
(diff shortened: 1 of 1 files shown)


connector/github/github.go
```diff
@@ -244,23 +244,23 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
-			if groups, err = c.teams(ctx, client, c.org); err != nil {
-				return identity, fmt.Errorf("github: get teams: %v", err)
-			}
+	// Org and team membership should always be checked if specified by config.
+	if len(c.orgs) > 0 {
+		// Error if groups scope is not provided by the client; the groups scope and
+		// orgs must be used simultaneously. API endpoints might return a 404
+		// response otherwise.
+		if !s.Groups {
+			return identity, errors.New("github: orgs are specified without 'groups' scope provided")
```
ericchiang: It's perfectly valid to provide `orgs` without the groups scope

connector/github/github.go
```diff
@@ -244,23 +244,23 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
+		}
+		groups, err := c.listGroups(ctx, client, user.Login)
+		if err != nil {
+			return identity, err
+		}
+		identity.Groups = groups
+	} else if s.Groups && c.org != "" {
+		groups, err := c.teams(ctx, client, c.org)
```
ericchiang: Add a comment that we don't check org membership here to preserve the old behavior.

connector/github/github.go
```diff
@@ -300,23 +300,23 @@ func (c *githubConnector) Refresh(ctx context.Context, s connector.Scopes, ident
...
-				return identity, err
-			}
-			if !inOrg {
-				return identity, fmt.Errorf("github: user %q not a member of org %q", user.Login, c.org)
-			}
-			if groups, err = c.teams(ctx, client, c.org); err != nil {
-				return identity, fmt.Errorf("github: get teams: %v", err)
-			}
+	// Org and team membership should always be checked if specified by config.
+	if len(c.orgs) > 0 {
```
ericchiang: This is the exact same behavior as above. Can you split it into a helper?

connector/github/github.go
estroz: But API requests won't succeed in this case because we don't set the [`read:org` scope](https://github.com/coreos/dex/blob/master/connector/github/github.go#L138-L141).

ericchiang: We should just always include that value then

connector/github/github.go
estroz: Commented in [helper](https://github.com/coreos/dex/pull/1039/commits/e92f38f38f211967ef40d23e03de569b7a30f01e#diff-de977d377b738d6060529d91926d264bR217).

connector/github/github.go
estroz: Done

connector/github/github.go
```diff
@@ -134,15 +134,14 @@ type githubConnector struct {
...
 func (c *githubConnector) oauth2Config(scopes connector.Scopes) *oauth2.Config {
-	var githubScopes []string
-	if scopes.Groups {
-		githubScopes = []string{scopeEmail, scopeOrgs}
-	} else {
-		githubScopes = []string{scopeEmail}
+	// The GitHub API requires requires read:org scope to ensure users are members
+	// of orgs and teams provided in configs.
+	githubScopes := []string{scopeEmail}
+	if len(c.orgs) > 0 || c.org != "" {
```
estroz: Instead of relying on `groups` scope, use `orgs`/`org` to determine whether `read:org` scope is needed. If neither are populated in the config, we only make `/user` and `/user/emails` API calls, which require `user:email` scope.

estroz:
Re-tested, looks good

ericchiang:
@estroz can you clean up the PR description?

> Fix a bug where org membership checks would be skipped if a client didn't provide the "groups" scope.
>
> Revert behavior of "org" to not do a membership check. Admins that need to require a user must be a
> member of an organization should use the newly added "orgs" field.

estroz:
@ericchiang done. I re-articulated "membership" because that references different a API resource.

connector/github/github.go
```diff
@@ -133,16 +133,22 @@ type githubConnector struct {
 	httpClient *http.Client
 }
 
+// groupsRequired returns whether dex must access GitHub "groups" (org and team)
+// info. If 'orgs' is populated this method returns true regardless of client
+// scope values. However groups scope must be set with 'org' to maintain old behavior.
```
rithujohn191: Could you add a little reasoning why we "set it true regardless of client scope values"? *since the GitHIb API requires it*

connector/github/github.go
```diff
@@ -207,6 +216,16 @@ func newHTTPClient(rootCA string) (*http.Client, error) {
 	}, nil
 }
 
+// getGroups retrieves GitHub orgs and teams a user is in, if any.
+func (c *githubConnector) getGroups(ctx context.Context, client *http.Client, groupScope bool, userLogin string) ([]string, error) {
+	if len(c.orgs) > 0 {
+		return c.listGroups(ctx, client, userLogin)
+	} else if c.org != "" {
+		return c.teams(ctx, client, c.org)
```
ericchiang: s/teams/teamsForOrg/g

connector/github/github.go
```diff
@@ -207,6 +216,16 @@ func newHTTPClient(rootCA string) (*http.Client, error) {
 	}, nil
 }
 
+// getGroups retrieves GitHub orgs and teams a user is in, if any.
+func (c *githubConnector) getGroups(ctx context.Context, client *http.Client, groupScope bool, userLogin string) ([]string, error) {
+	if len(c.orgs) > 0 {
+		return c.listGroups(ctx, client, userLogin)
```
ericchiang: s/listGroups/groupForOrgs/s

Issue #1030, opened 2017-08-16
TestLoginTimeout takes 20 seconds to run


ericchiang:
The rest take 3. We should really fix that.

ericchiang:
Some thoughts:

* The test shouldn't be tied to a literal value `10*time.Second` embedded in the server.
* We should probably just reject bcrypt values that are too high. If we compute something with a cost of 20, it'll still eat up CPU in the other goroutine even if we exit early to a user.
* We should generally timeout for long HTTP requests. There might be a more general solution like some server middleware.

cc @estroz

estroz:
@ericchiang closing (fixed by #1040) and creating an issue specifically addressing generic request timeouts.

PR #1040, opened 2017-08-17
server: set sane bcrypt cost upper bound

estroz:
api.go: `checkCost` only returns an error if hash cost is too high or low (or if `Cost` fails)
server.go: remove timeout logic and reverte back to a hash comparison, with an addition cost check
server_test.go: `MinCost` will cause `Login` to error now, so use `DefaultCost` instead.

Instead of timing out after 10 seconds of hashing, dex should not permit a bcrypt hash cost > 16. The below benchmarks show that cost > 16 exponentially grows to times that a user would notice when logging in, considering that dex will likely be running on a server (much beefier than my machine).

Tests for login timeouts have been removed.

Applies to #1030. Middleware for generically timing out requests will be implemented soon.

Benchmarks (Intel(R) Core(TM) i7-6500U CPU @ 2.50GHz):

Cost | Time (sec)
:---: | :---:
10 | 0.073507
11 | 0.151875
12 | 0.301854
13 | 0.596907
14 | 1.151219
15 | 2.365981
**16** | **4.941790**
17 | 9.467757

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -288,25 +288,14 @@ func (db passwordDB) Login(ctx context.Context, s connector.Scopes, email, passw
 		}
 		return connector.Identity{}, false, nil
 	}
-
-	// Return an error if password-hash comparison takes longer than 10 seconds
-	errCh := make(chan error, 1)
-	go func() {
-		errCh <- bcrypt.CompareHashAndPassword(p.Hash, []byte(password))
-	}()
-	select {
-	case err = <-errCh:
-		if err != nil {
-			return connector.Identity{}, false, nil
-		}
-	case <-time.After(time.Second * 10):
-		var cost int
-		if cost, err = bcrypt.Cost(p.Hash); err == nil {
-			err = fmt.Errorf("password-hash comparison timeout: your bcrypt cost = %d, recommended cost = %d", cost, recCost)
-		}
+	// This check prevents dex users from logging in using static passwords
+	// configured with hash costs that are too high or low.
+	if err := checkCost(p.Hash); err != nil {
 		return connector.Identity{}, false, err
 	}
-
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -288,25 +288,9 @@ func (db passwordDB) Login(ctx context.Context, s connector.Scopes, email, passw
...
-		if err != nil {
-			return connector.Identity{}, false, nil
-		}
-	case <-time.After(time.Second * 10):
-		var cost int
-		if cost, err = bcrypt.Cost(p.Hash); err == nil {
-			err = fmt.Errorf("password-hash comparison timeout: your bcrypt cost = %d, recommended cost = %d", cost, recCost)
-		}
-		return connector.Identity{}, false, err
+	if err := bcrypt.CompareHashAndPassword(p.Hash, []byte(password)); err != nil {
```
ericchiang: We probably want to do the checkCost check here too. If a user provides a static password with a large bcrypt cost, we cant to error out early.

server/server.go
```diff
@@ -288,25 +288,12 @@ func (db passwordDB) Login(ctx context.Context, s connector.Scopes, email, passw
...
-	case err = <-errCh:
-		if err != nil {
-			return connector.Identity{}, false, nil
-		}
-	case <-time.After(time.Second * 10):
-		var cost int
-		if cost, err = bcrypt.Cost(p.Hash); err == nil {
-			err = fmt.Errorf("password-hash comparison timeout: your bcrypt cost = %d, recommended cost = %d", cost, recCost)
-		}
+	if err := checkCost(p.Hash); err != nil {
```
ericchiang: Add a comment here explaining why we're doing this.

Issue #1042, opened 2017-08-22
SAML: no configuration option to compress SAMLRequest


easeway:
Some identity provider requires SAMLRequest to be compressed and then base64 encoded. There's no configuration option to compress SAMLRequest and dex won't work with these identity providers.

ericchiang:
Dex supports the SAML POST binding, which doesn't use compression

> Messages are encoded for use with this binding by encoding the XML into an HTML form control and are
transmitted using the HTTP POST method. A SAML protocol message is form-encoded by applying the
base-64 encoding rules to the XML representation of the message and placing the result in a hidden form
control within a form as defined by [HTML401] Section 17. The HTML document MUST adhere to the
XHTML specification, [XHTML]. The base64-encoded value MAY be line-wrapped at a reasonable length
in accordance with common practice.

_3.5.4 Message Encoding_ - https://docs.oasis-open.org/security/saml/v2.0/saml-bindings-2.0-os.pdf

The HTTP redirect binding does use encryption (section 3.4.4), but that binding isn't used by dex. 

This behavior was actually addressed in https://github.com/coreos/dex/pull/785

From my reading, compliant providers shouldn't enforce compression (or even support it) of the SAMLRequest on the HTTP POST binding. Though they may for the HTTP redirect binding.

Which provider are you using?

easeway:
vSphere SSO. It offers HTTP redirect binding. However it also accepts POST but requires compression on SAMLRequest.

easeway:
I have PR proposed: https://github.com/coreos/dex/pull/1044

Without the change, dex won't work with vSphere SSO using SAML2.0.

easeway:
According to SAML standard, how can we use HTTP redirect binding with dex? It seems HTTP redirect binding is the only option in vSphere SSO.

ericchiang:
We would have to add the HTTP redirect binding, though that's not unreasonable.

easeway:
I can help on that. I have a PR working in progress to support HTTP redirect binding.

ericchiang:
@easeway perfect, please feel free to send the PR and tag me for review.

I can also do this if that's easier.

srenatus:
I wasn't aware of this issue, but there's another recent attempt to fix this here: #1175

PR #1044, opened 2017-08-22
Deflate SAMLRequest (Issue #1042)

easeway:
Some identity provider requires SAMLRequest to be
deflated. Add the "compress" option in config to
deflate the request.


ericchiang:
Closing per https://github.com/coreos/dex/issues/1042

The fix should be adding the redirect binding.

PR #1045, opened 2017-08-23
support SAML HTTP redirect binding

easeway:
Solve issue: https://github.com/coreos/dex/issues/1042


easeway:
@ericchiang Please take a look...

server/handlers.go
```diff
@@ -254,30 +254,47 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
...
+					    <input type="hidden" name="SAMLRequest" value="%s" />
+					    <input type="hidden" name="RelayState" value="%s" />
+				    </form>
+					<script>
+					    document.forms[0].submit();
+					</script>
+				  </body>
+				  </html>`, ssoURL, value, authReqID)
+			} else {
+				// use HTTP redirect binding
```
ericchiang: We should just return an actual redirect here, not render a page. E.g.

`http.Redirect(w, r, ...)`

Documentation/saml-connector.md
```diff
@@ -86,6 +86,12 @@ connectors:
     #     urn:oasis:names:tc:SAML:2.0:nameid-format:persistent
     #
     nameIDPolicyFormat: persistent
+
+    # Optional: Use HTTP redirect binding instead of HTTP-POST binding
+    #
+    # Some identity provider only supports HTTP redirect binding
+    #
+    # redirectBinding: true
```
ericchiang: this should probably be something like

```
binding: redirect # defaults to post
```

in case we want to support more bindings in the future.

server/handlers.go
```diff
@@ -254,30 +254,47 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
 				s.logger.Errorf("Server template error: %v", err)
 			}
 		case connector.SAMLConnector:
-			action, value, err := conn.POSTData(scopes, authReqID)
+			ssoURL, value, err := conn.POSTData(scopes, authReqID)
```
ericchiang: Need to rename POSTData or break it into a separate method.

connector/saml/saml.go
```diff
@@ -263,6 +272,28 @@ func (p *provider) POSTData(s connector.Scopes, id string) (action, value string
 		return "", "", fmt.Errorf("marshal authn request: %v", err)
 	}
 
+	// Compress the request if redirect-binding is used
+	if p.redirectBinding {
+		var buf bytes.Buffer
+		zw, e := flate.NewWriter(&buf, flate.DefaultCompression)
+		if e == nil {
```
ericchiang: This block shouldn't keep going if err != nil

connector/saml/saml.go
```diff
@@ -263,6 +272,28 @@ func (p *provider) POSTData(s connector.Scopes, id string) (action, value string
...
+		}
+		if e == nil {
+			e = zw.Flush()
+		}
+		zw.Close()
+		if e != nil {
+			return "", "", fmt.Errorf("deflate request: %v", e)
+		}
+
+		redirectURL := fmt.Sprintf("%s?SAMLRequest=%s&RelayState=%s",
```
ericchiang: use url.URL and an actual url.Values object https://golang.org/pkg/net/url/#Values

connector/saml/saml.go
```diff
@@ -263,6 +272,28 @@ func (p *provider) POSTData(s connector.Scopes, id string) (action, value string
...
+			p.ssoURL,
+			url.QueryEscape(base64.StdEncoding.EncodeToString(buf.Bytes())),
+			url.QueryEscape(id))
+		return redirectURL, "", nil
```
ericchiang: This method now returns completely different kind of values depending on how it's configured. The URL creation and HTML creation should be done in the same place.

Documentation/saml-connector.md
easeway: gotcha

connector/saml/saml.go
easeway: here `e == nil` and `err != nil` is handled in one place below.

I use `e == nil` because the sequence will go through 3 steps: NewWriter, Write and Flush. Each may fail and should bail out immediately. So use `if e == nil { do...}` will make the code simpler, and handle the error in one place.

connector/saml/saml.go
easeway: OK. I will construct a URL.

connector/saml/saml.go
easeway: I will update the interface according to your comments below.

server/handlers.go
easeway: I think it's better to pass http.ResponseWriter and http.Request into provider to reduce provider specific code in server/handlers.go. Eventually we can eliminate this `switch conn := conn.Connector.(type)`

My proposal is to define a common Connector interface including a method:

HandleLogin(w http.ResponseWriter, r *http.Request)

And leave to the Connector implementation for all provider specific logic.

In the mean time, I will do the following change:

Rename PostData to AuthnRequest(w http.ResponseWriter, r *http.Request) and perform the logic there.

server/handlers.go
easeway: Yes.

server/handlers.go
ericchiang: I'm actually against that. The server package controls things like html error responses and I'd rather all of that logic continue to live there. Not have any connectors refer to the net/http package directly.

easeway:
@ericchiang PR updated. Please take a look when you have time.

server/handlers.go
```diff
@@ -254,30 +254,41 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
...
+					    <input type="hidden" name="RelayState" value="%s" />
+				    </form>
+					<script>
+					    document.forms[0].submit();
+					</script>
+				  </body>
+				  </html>`, ssoURL, value, authReqID)
+			case connector.SAMLBindingRedirect:
+				query := make(url.Values)
+				query["SAMLRequest"] = []string{value}
```
ericchiang: nit: this is just `query.Set("SAMLRequest", value)`

connector/connector.go
```diff
@@ -65,18 +65,24 @@ type CallbackConnector interface {
 	HandleCallback(s Scopes, r *http.Request) (identity Identity, err error)
 }
 
+// SAMLConnector bindings
+const (
+	SAMLBindingPOST     = "post"
```
ericchiang: nit: this should probably be httpPost and httpRedirect, right? Since the other ones are SOAP based.

easeway:
The handler doesn't need any change. When specifying "binding: redirect", it only specifies how dex opens login page on identity provider. The way identity provider calls back to dex can use a different http method, which is defined by the trusted service callback mechanism in identity provider (the pre-registered trusted service provider definition).

If identity provider respect AssertionConsumerServiceURL in AuthnRequest, it will use ProtocolBinding to send back response. Currently in dex, the ProtocolBinding is always POST.

I have verified the implement against vSphere 6.5

server/handlers.go
easeway: yes

connector/connector.go
easeway: Do you mean the const name should be "SAMLHTTPPOST" or the value should be "httpPost"? I'm using the same value from config option. I'm OK if we decide to use "http-post", "http-redirect" in values of config option "binding". But I'd like to use the same value internally in code.

ericchiang: I meant the "post" and "redirect" values.

Actually, I think the way it currently is is fine. Let's keep it as is.

ericchiang:
@easeway ah I see that you can actually mix bindings.

It's good that vSphere works with this setup, but what about providers that don't support the POST binding for the SAMLResponse? Do we need explicit `requestBinding` and `responseBinding` values?

easeway:
@ericchiang For providers don't support POST binding in response, we need to do further fixes. We should make it configurable, that when constructing AuthnRequest, we can specify ProtocolBinding from the configuration.

What about making a separate PR for that?

easeway:
@ericchiang I created a separate PR #1048 to support response redirect binding. However vSphere doesn't support redirect binding on response, so I can't verify that, and that's why I make it a separate PR.

Would you approve this PR if you think the change is good here?

ericchiang:
> Would you approve this PR if you think the change is good here?

No we should just move to #1048. Since it's all the same feature I'd rather we design the full thing instead of doing it piecemeal.

easeway:
@ericchiang All right. I will collapse the changes in #1048 into this PR and close #1048. And I make sure the features except response redirect binding is verified.

easeway:
@ericchiang Updated. I separate the config for requestBinding and responseBinding and updated the handler. Please take a look.

connector/saml/saml.go
```diff
@@ -72,6 +75,18 @@ func init() {
 	}
 }
 
+// normalizeBinding verifies binding from config and normalize the string
+func normalizeBinding(binding string) (string, error) {
+	binding = strings.ToLower(strings.TrimSpace(binding))
```
ericchiang: I'd avoid any normalization. Keeps the configuration explicit.

connector/saml/saml.go
```diff
@@ -72,6 +75,18 @@ func init() {
...
+	if binding == "" {
+		return connector.SAMLBindingPOST, nil
+	}
+	if binding != connector.SAMLBindingPOST && binding != connector.SAMLBindingRedirect {
```
ericchiang: probably just use a globally defined map + a lookup

connector/saml/saml.go
```diff
@@ -167,6 +186,14 @@ func (c *Config) openConnector(logger logrus.FieldLogger) (*provider, error) {
 		nameIDPolicyFormat: c.NameIDPolicyFormat,
 	}
 
+	var err error
+	if p.requestBinding, err = normalizeBinding(c.RequestBinding); err != nil {
+		return nil, err
+	}
+	if p.responseBinding, err = normalizeBinding(c.ResponseBinding); err != nil {
```
ericchiang: the response binding should probably default to whatever the request binding is, right?

connector/saml/saml.go
```diff
@@ -252,6 +281,11 @@ func (p *provider) POSTData(s connector.Scopes, id string) (action, value string
 		},
 		AssertionConsumerServiceURL: p.redirectURI,
 	}
+
+	if p.responseBinding == connector.SAMLBindingRedirect {
+		r.ProtocolBinding = bindingRedirect
```
ericchiang: Do we not set this if it's a POST?

connector/saml/saml.go
```diff
@@ -598,3 +647,34 @@ func before(now, notBefore time.Time) bool {
 func after(now, notOnOrAfter time.Time) bool {
 	return now.After(notOnOrAfter.Add(allowedClockDrift))
 }
+
+// deflate compresses the data
+func deflate(data []byte) ([]byte, error) {
```
ericchiang: nit: just call this "compressRequest" and "decompressRequest"

server/handlers.go
```diff
@@ -356,19 +373,25 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
...
+			response = r.URL.Query().Get("SAMLResponse")
+		case http.MethodPost:
+			response = r.PostFormValue("SAMLResponse")
+		default:
+			s.logger.Errorf("Invalid method responds to SAML callback")
 			s.renderError(w, http.StatusBadRequest, "Invalid request")
 			return
 		}
-		identity, err = conn.HandlePOST(parseScopes(authReq.Scopes), r.PostFormValue("SAMLResponse"), authReq.ID)
+		identity, err = conn.HandleResponse(parseScopes(authReq.Scopes), response, authReq.ID)
```
ericchiang: So in this case I could request `responseBinding: post`, get an HTTP redirect response, and we'd still process it even though we've explicitly stated that we don't.

I think we need to enforce that only the binding that the connector is configured with is actually used.

server/handlers.go
```diff
@@ -356,19 +373,25 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
...
 			return
 		}
 		identity, err = conn.HandleCallback(parseScopes(authReq.Scopes), r)
 	case connector.SAMLConnector:
-		if r.Method != "POST" {
-			s.logger.Errorf("OAuth2 request mapped to SAML connector")
+		var response string
+		switch r.Method {
+		case http.MethodGet:
+			response = r.URL.Query().Get("SAMLResponse")
```
ericchiang: With this logic I could use "state" to look up the authID and never actually inspect the "RelayState" even though it's required for this response. This seems wrong.

connector/saml/saml.go
easeway: OK. I will remove this ToLower line.

connector/saml/saml.go
easeway: only two string const comparisons. I'm trying to avoid a global var. Anyway, I will add a global map.

connector/saml/saml.go
easeway: not sure. In most cases, responseBinding will be POST. Even requestBinding is redirect. Would it be simpler to define fixed default values in config?

ericchiang: > Would it be simpler to define fixed default values in config?

Works for me.

connector/saml/saml.go
easeway: It's already set to POST in a few lines above when constructing the authnRequest.

server/handlers.go
easeway: handleConnectorCallback is shared by multiple providers. OAuth2 uses "state" while SAML uses "RelayState". So if it's SAML, there's no "state", and we should use "RelayState".

server/handlers.go
easeway: You are right. I will fix this.

connector/saml/saml.go
easeway: Yes

server/handlers.go
easeway: The problem is at this layer, it doesn't have the knowledge about bindings. Anyway, let me figure it out...

server/handlers.go
```diff
@@ -319,12 +330,18 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
 func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request) {
 	var authID string
 	switch r.Method {
-	case "GET": // OAuth2 callback
-		if authID = r.URL.Query().Get("state"); authID == "" {
+	case http.MethodGet: // OAuth2 or SAML HTTP-Redirect callback
+		// try OAuth2
+		authID = r.URL.Query().Get("state")
```
ericchiang: The SAML connector can still use the value set by "state" here. We need to enforce:

```
var usedRelayState bool
authID = r.URL.Query().Get("state")
if authID == "" {
    usedRelayState = true
    authID = r.URL.Query.Get("RelayState")
}
```  

Then verify this value later.

```
case connector.CallbackConnector:
    if usedRelayState {
        return nil, fmt.Errorf("no state parameter provided")
    }
case connector.SAMLConnector:
    if !userRelayState {
        return nil, fmt.Errorf("invalid url query value 'state' provided")
    }

```

connector/saml/saml.go
```diff
@@ -72,6 +79,17 @@ func init() {
 	}
 }
 
+// verifyBinding verifies binding from config
+func verifyBinding(binding string) (string, error) {
+	if binding == "" {
```
ericchiang: nit: just set defaults in `openConnector`

ericchiang:
I need to grab an open source SAML provider and try to write up some instructions for how to test this before merging. We want to make sure we can debug future changes to this binding.

server/handlers.go
easeway: understand!

connector/saml/saml.go
easeway: OK

ericchiang:
For all my efforts I'm having trouble finding a SAML provider to test this with. @easeway any suggestions?

easeway:
I've verified with vSphere with both POST and redirect binding for request, and POST binding for response. I believe okta works with POST binding for both request and response (as in original code). So we are good with bindings for requests. Unfortunately, I didn't find a provider which supports redirect binding for response. I can continue looking for a provider which allows a services to be registered with a redirect binding for response.

PR #1046, opened 2017-08-23
storage/static.go: storage backend should not explicitly lower-case email ids.

rithujohn191:


storage/static.go
```diff
@@ -91,8 +91,9 @@ type staticPasswordsStorage struct {
 func WithStaticPasswords(s Storage, staticPasswords []Password) Storage {
 	passwordsByEmail := make(map[string]Password, len(staticPasswords))
 	for _, p := range staticPasswords {
-		p.Email = strings.ToLower(p.Email)
-		passwordsByEmail[p.Email] = p
+		//Enable case insensitive email comparison.
+		lowerEmail = strings.ToLower(p.Email)
```
ericchiang: :=

rithujohn191:
added a test and documentation

storage/memory/static_test.go
```diff
@@ -173,10 +174,14 @@ func TestStaticPasswords(t *testing.T) {
...
 		},
 		{
-			name: "create password",
+			name: "create passwords",
 			action: func() error {
-				return s.CreatePassword(p3)
+				if err := s.CreatePassword(p3); err != nil {
+					return err
+				}
+				return s.CreatePassword(p4)
```
ericchiang: Need to read this value back an ensure the emails match.

storage/static.go
```diff
@@ -91,8 +92,9 @@ type staticPasswordsStorage struct {
 func WithStaticPasswords(s Storage, staticPasswords []Password) Storage {
 	passwordsByEmail := make(map[string]Password, len(staticPasswords))
 	for _, p := range staticPasswords {
-		p.Email = strings.ToLower(p.Email)
-		passwordsByEmail[p.Email] = p
+		//Enable case insensitive email comparison.
+		lowerEmail := strings.ToLower(p.Email)
+		passwordsByEmail[lowerEmail] = p
```
kans: Could we log or something if the `lowered` email as already in the map?

ericchiang: We would just return an error here. @rithujohn191 what do you think?

storage/memory/static_test.go
rithujohn191: This test makes sure that "Spam@example.com" cannot be created, so it makes sure an error is returned

rithujohn191: do you want me to check if "spam@example.com" gets created with the all lower case?

ericchiang: Something like

```
p, err := s.GetPassword(strings.ToLower(p4.Email))

p.Email == p4.Email
```

storage/static.go
ericchiang: Actually I only think we can log this error. Erroring out would be backward incompatible, if sensible :(

rithujohn191: So if we don't error out we essentially allow the user to overwrite the map. For the following ConfigMap:
```
staticPasswords:
- email: "admin@example.com"
  # bcrypt hash of the string "password"
  username: "admin"
- email: "Admin@example.com"
  # bcrypt hash of the string "password"
  username: "Admin"
```

Only the second password gets registered. And then if the user tries to login with the first email it still succeed but returns the wrong email id in the claims:

```
"email": "Admin@example.com",
```

That's not desirable behavior :(

ericchiang: @rithujohn191 right, but prior to this that config would not have errored, if we error now we might break someone during a dex upgrade.

rithujohn191: logged the error

storage/memory/static_test.go
rithujohn191: done

rithujohn191:
working on adding the logger

rithujohn191:
Done :)

PR #1048, opened 2017-08-24
Resp redirect binding

easeway:
Update SAML connector config to distinguish protocol bindings for request and response in case the identity provider has specific requirements on these.


easeway:
Changes collapsed to #1042

PR #1058, opened 2017-09-08
Add LDAP parent groups search (For permissions to groups and its subgroups)

paroque28:
This allows group search to add parent groups, used when you want to add permissions to a group and its subgroups. 
Now, with the flag "inheritance", the group search adds all the upper groups of certain user. Allowing to have permissions for groups and sub groups.

This first abstracts all the query of groups and use it to reuse code. The method is called queryGroups() and it receives only the DN/uid of an object and returns its groups.
Then I created an iterative loop to look for parent groups of the current group's list. The algorithm skips if the group is already added, preventing infinite loops.
The search loop ends when no more parents are found.

I attached an image with inheritance enabled and one with out
inheritance enabled
inheritance disabled


paroque28:
@ericchiang can you check this?

ericchiang:
Hi @paroque28 thanks for the feature request. We normally require some conversation about what problem you're trying to solve before submitting a PR and doing a code review.

Can you please post an example LDAP schema and query that this new feature enables? The concept of "parent group" is extremely general and can be implemented by LDAP in many different ways. Why can't this be done with the existing code? What LDAP providers implement this?

Also, all large changes to the LDAP code in dex require an integration tests (but after we've agreed to accept the feature) https://github.com/coreos/dex/blob/master/connector/ldap/ldap_test.go#L50

paroque28:
Hi @ericchiang I added Integration testing to this feature:
https://github.com/paroque28/dex/blob/master/connector/ldap/ldap_test.go#L213
It passed :):
"--- PASS: TestGroupInheritanceQuery (0.85s)"


The LDAP schema is described in the test but I made an image for convenience 
ldap

The need for this feature is seen when you want to know if "Jane" (for example) is a member of IT. With actual implementation, you can't know. Therefore you can't, for example, allow all IT members to access certain namespace on Tectonic.

paroque28:
With that LDAP schema and inheritance flag enabled, Jane would return groups=["team-one", "team-two", "IT", "developers"] instead of actual response groups=["team-one", "team-two"]. As described here https://github.com/paroque28/dex/blob/master/connector/ldap/ldap_test.go#L298

paroque28:
@ericchiang Fixed

connector/ldap/ldap.go
```diff
@@ -496,52 +503,107 @@ func (c *ldapConnector) groups(ctx context.Context, user ldap.Entry) ([]string,
...
-			groups = append(groups, resp.Entries...)
-			return nil
-		}); err != nil {
-			return nil, err
 		}
-		if !gotGroups {
-			// TODO(ericchiang): Is this going to spam the logs?
-			c.logger.Errorf("ldap: groups search with filter %q returned no groups", filter)
+		// If there's no remaining levels -> exit loop
+		if len(nextLevelGroups) == 0 {
```
ericchiang: What prevents this from looping infinitely?

Documentation/ldap-connector.md
```diff
@@ -125,6 +125,9 @@ connectors:
 
       # Represents group name.
       nameAttr: name
+
+      #Add also parents of groups
+      inheritance: true
```
ericchiang: This doesn't actually define how groups are determined to be a child of another group.

connector/ldap/ldap.go
```diff
@@ -496,52 +503,107 @@ func (c *ldapConnector) groups(ctx context.Context, user ldap.Entry) ([]string,
...
-			return nil, fmt.Errorf("ldap: group entity %q missing required attribute %q",
-				group.DN, c.GroupSearch.NameAttr)
-		}
+}
 
-		groupNames = append(groupNames, name)
+//  Abrstraction query of Groups and reuse code
+func (c *ldapConnector) queryGroups(ctx context.Context, dn string) ([]*ldap.Entry, string, error) {
+	var groups []*ldap.Entry
+	filter := fmt.Sprintf("(%s=%s)", c.GroupSearch.GroupAttr, ldap.EscapeFilter(dn))
```
ericchiang: I don't think we can assume that the attribute used for determining membership of a user will be the same for determining the membership of the group.

connector/ldap/ldap.go
```diff
@@ -50,6 +50,11 @@ import (
 //         groupAttr: member
 //         nameAttr: name
 //
+//         #Used when the filter member:1.2.840.113556.1.4.1941: timesout
```
ericchiang: About how many groups does it take for this query to timeout?

connector/ldap/ldap.go
```diff
@@ -50,6 +50,11 @@ import (
...
+//         #Better implementation of recursive search
```
ericchiang: Maybe we can refer to this as "recursive" instead of "inheritance"

connector/ldap/ldap.go
paroque28: Usually, on ActiveDirectory GroupAttr is  "member". 
I find it wrong to have a LDAP setup where for example I have the attribute "member" for users and some other word for groups. It can happen of course but I find more confusing for most users to have two different attributes being the same most of the cases. 

For example:
It can happen that someone instead of assigning 
```yaml
groupAttr:"member"
recursionAttr:"member"
```
assigns
```yaml
groupAttr:"member"
recursionAttr:"memberOf"
``` 
by thinking the attributes make different things. But is up to you to make the decision.
What do you prefer?

connector/ldap/ldap.go
paroque28: Done

Documentation/ldap-connector.md
paroque28: Fixed

connector/ldap/ldap.go
paroque28: I don't really know but I tried this and it timed out after one minute. I find it unacceptable to have queries that last longer than 20 seconds (Since it's just a login).
With this implementation, the duration of login reduced in just a couple of seconds

connector/ldap/ldap.go
paroque28: The infinite loop is prevented here: https://github.com/paroque28/dex/blob/master/connector/ldap/ldap.go#L542
I'm going to add integration testing for this

paroque28:
I added an infinite loop test using this LDAP schema 
ldap_test

paroque28:
Any update @ericchiang ?

ericchiang:
@paroque28 I'm still not convinced that the current proposal expresses how groups can be members of other groups in a general enough way. This works well for how AD does it, but what about schema where the CN is used as the attribute value instead of the DN? Can the attribute name be different for user membership instead of group membership? I don't think the current proposed change coves these scenarios.

paroque28:
Hi @ericchiang , I added an attribute called "recursionGroupAttr" (I didn't know how to name it). I think this solves what you said before.

connector/ldap/ldap.go
```diff
@@ -496,52 +504,111 @@ func (c *ldapConnector) groups(ctx context.Context, user ldap.Entry) ([]string,
...
+			if exit {
+				continue
+			}
+
+			groupNames = append(groupNames, name)
+			if c.GroupSearch.Recursive {
+				if c.GroupSearch.RecursionGroupAttr == "" {
+					return nil, fmt.Errorf("ldap: recursionGroupAttr attribute is not set but recursive search is enabled")
+				}
+				if obtainedGroups, _, err := c.queryGroups(ctx, c.GroupSearch.RecursionGroupAttr, group.DN); err == nil {
```
paroque28: This is where the new attribute is used

paroque28:
Hi @ericchiang I was evaluating Docker Enterprise today and this  is the feature I am asking for (read the description of iteration/ recursive search)
docker

paroque28:
Hi @rithujohn191,
I was wondering if there any reason this cannot be approved.
If there, tell me and I can fix it.
Thanks

srenatus:
This is an interesting feature. Is this still being pursued? (I hope so 🤞)
😄

srenatus:
@paroque28 Two notes/questions that might be useful to you, as this seems to be Active Directory related:

1. Have you considered using the `microsoft` connector that [has recently been merged](https://github.com/coreos/dex/pull/1131)? I've verified that it will return nested groups just fine.
2. If you depend on using LDAP to talk to AD, can you try [setting `groupAttr`](https://github.com/coreos/dex/blob/861d4ae4475a75885339e11cefd718df2cb2e2a9/connector/ldap/ldap.go#L119-L126) to `member:1.2.840.113556.1.4.1941:`? [According to this document](https://social.technet.microsoft.com/wiki/contents/articles/5392.active-directory-ldap-syntax-filters.aspx), this should return
    > All groups specified user belongs to,
including due to group nesting

paroque28:
Hi, @srenatus thank you for your comment. The Microsoft connector sounds excellent, I'll put an eye on that. 
And regarding the groupAttr option, unfortunately, I cannot use that since it is a highly inefficient algorithm and my queries times out.

Thank you again

paroque28:
@srenatus I cannot use Microsoft connector since my LDAP server is in a secured network

thomasem:
This came up for my team recently as we're using Dex to integrate with our FreeIPA backend where `memberOf` is an attribute on the user with parent groups resolved. I was wondering what the going thoughts were on taking that approach instead of querying necessarily from the group perspective?

vavdoshka:
hello guys, is this feature somehow supported or just abandoned?

vavdoshka:
in case if someone stumbled upon here to find the nested group lookup solution: 

thanks, @srenatus for the hint!

```
              groupSearch:
                baseDN: OU=ACME,DC=noway,DC=com
                filter: "(objectClass=group)"
                userAttr: DN
                groupAttr: "member:1.2.840.113556.1.4.1941:"
                nameAttr: cn
```

lance-plusai:
> in case if someone stumbled upon here to find the nested group lookup solution:
> 
> thanks, @srenatus for the hint!
> 
> ```
> groupSearch:
>   baseDN: OU=ACME,DC=noway,DC=com
>   filter: "(objectClass=group)"
>   userAttr: DN
>   groupAttr: "member:1.2.840.113556.1.4.1941:"
>   nameAttr: cn
> ```

Saved my life

paroque28:
Abandoned

anuviswa:
"member:1.2.840.113556.1.4.1941:" this seems an AD specific solution. Can we reconsider this PR? Is there a reason this PR has been abandoned?

PR #1063, opened 2017-09-13
Updates coreos themes and icons for various providers

amrutac:


ericchiang:
It looks like `web/themes/coreos/favicon.png` is a tectonic logo. Shouldn't it be a coreos logo?

amrutac:
Good catch, updated.

PR #1068, opened 2017-09-18
Replace docker-build script with multi-stage build

furuholm:
Uses docker multi stage builds [1] instead docker-build script. The resulting image should be identical to running scripts/docker-build and docker build separately.

[1] https://docs.docker.com/engine/userguide/eng-image/multistage-build/


furuholm:
Fixes #1074

.dockerignore
```diff
@@ -1,3 +1,14 @@
 *
```
ericchiang: We actually don't need to ignore anything now, right?

furuholm: Not really.

furuholm:
No changes to the makefile are required (AFAIK)

.dockerignore
furuholm: I just tried to follow the convention of the current dockerfile. Figured you maybe had reasons for wanting to ignore everything but the necessary files.

ericchiang:
This PR removes a script referenced in the Makefile https://github.com/coreos/dex/blob/64c97c73080bfc951d5f01906257a633bd164909/Makefile#L61-L63

I'd expect changes to that Makefile as part of the PR.

.dockerignore
ericchiang: Can we just remove the `.dockerignore` file then? It'd be weird to have to always update this file every time we add a new top level directory.

furuholm: Will do

furuholm:
My mistake! I'll go through the makefile and make sure it is up to date.

furuholm:
Makefile updated. Did not remove the docker ignore file though. Left the bin directory in there as it gets quite large.

ericchiang:
This worked for me locally. Thanks for the change! Can you please squash your commits then we can merge?

furuholm:
Great! Rebased and squashed.

PR #1079, opened 2017-09-26
Groups support for static passwords

dpacierpnik:
We would like to configure groups for statically configured users. Currently it is not possible as static passwords do not support this attribute.
This pull request adds support for groups attribute in static passwords.
Additionally I added go_import_path in travis.yaml because I was not able to run tests for my fork in Travis. With go_import_path Travis should work for your project as well as for forks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 bin
 dist
 _output
+config
+examples/k8s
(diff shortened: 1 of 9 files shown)


ericchiang:
Closing since we prefer discussing this in issues instead of code reviews.

PR #1082, opened 2017-09-26
Possibility to select connector in authorization request.

pbochynski:
Right now a client is not able to select connector that should be used to authenticate a user. This pull request adds connector query param to auth endpoint like it is done in other federated ID providers. Example:
https://auth0.com/docs/api/authentication#social
See connection parameter.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -174,6 +174,22 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if authReq.Connector != "" {
+		filtered := connectors[:0]
+		for _, c := range connectors {
+			if c.ID == authReq.Connector {
+				filtered = append(filtered, c)
+			}
+		}
+		if len(filtered) == 0 {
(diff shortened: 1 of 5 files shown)


ericchiang:
Closing since we try to not use PRs to discuss feature requests.

Had some concerns last time this was brought up though https://github.com/coreos/dex/pull/1026#issuecomment-322573698

> We've consider this in the past, but this has a lot of implications and questions around guarantees. Should it be a scope? Client configuration? Should a user be able to change this value? Do we have to guarantee that the user can't use other connectors? Can an end user see the other connectors? Would a client be surprised if they used this feature and a user logged in through a different connector?

server/handlers.go
```diff
@@ -174,6 +174,22 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if authReq.Connector != "" {
+		filtered := connectors[:0]
```
dpacierpnik: You should not create slice that way. When you will modify filtered, you will modify also connectors ;)

PR #1083, opened 2017-09-26
Possibility to select a connector in the authorization request.

pbochynski:
If multiple connectors are defined DEX displays the page with the list of options. This pull request makes it possible to select connector in the authorization request and skip that step. It is similar to the connection parameter in Auth0:
https://auth0.com/docs/api/authentication#social


pbochynski:
Created by mistake. Please ignore.

PR #1085, opened 2017-09-26
add client request timeout

rphillips:
does the http.Client ever get a timeout attached to it for requests that timeout? Or would the server timeout the request?

### DIFF
--- a/storage/kubernetes/client.go
+++ b/storage/kubernetes/client.go
@@ -333,7 +333,10 @@ func newClient(cluster k8sapi.Cluster, user k8sapi.AuthInfo, namespace string, l
 
 	logger.Infof("kubernetes client apiVersion = %s", apiVersion)
 	return &client{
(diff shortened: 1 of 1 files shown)


storage/kubernetes/client.go
```diff
@@ -333,7 +333,10 @@ func newClient(cluster k8sapi.Cluster, user k8sapi.AuthInfo, namespace string, l
 
 	logger.Infof("kubernetes client apiVersion = %s", apiVersion)
 	return &client{
-		client:     &http.Client{Transport: t},
+		client: &http.Client{
+			Transport: t,
+			Timeout:   30 * time.Second,
```
ericchiang: can we make this shorter? e.g. 15 seconds. There's no reason any of these requests should take 30 seconds.

ericchiang:
> Or would the server timeout the request?

we definitely don't want to depend on the server to time it out even if it will.

PR #1086, opened 2017-09-26
storage/kubernetes: Log before registering custom resources

chancez:
Logging before attempting to make any connection to Kubernetes is useful when the connection hangs and dex is killed before it can log any errors.


ericchiang:
@chancez in the future please use a fork when working on branches.

PR #1088, opened 2017-09-27
Cross clients improvement - requesting client ID always added to the audience claim

dpacierpnik:
See https://github.com/coreos/dex/issues/1087

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -812,6 +812,127 @@ func TestCrossClientScopes(t *testing.T) {
 	}
 }
 
+func TestCrossClientScopesWithAzpInAudienceByDefault(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	httpServer, s := newTestServer(ctx, t, func(c *Config) {
+		c.Issuer = c.Issuer + "/non-root-path"
+	})
+	defer httpServer.Close()
+
+	p, err := oidc.NewProvider(ctx, httpServer.URL)
+	if err != nil {
+		t.Fatalf("failed to get provider: %v", err)
+	}
+
+	var (
+		reqDump, respDump []byte
+		gotCode           bool
+		state             = "a_state"
+	)
+	defer func() {
+		if !gotCode {
+			t.Errorf("never got a code in callback\n%s\n%s", reqDump, respDump)
+		}
+	}()
+
+	testClientID := "testclient"
+	peerID := "peer"
+
+	var oauth2Config *oauth2.Config
+	oauth2Server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/callback" {
+			q := r.URL.Query()
+			if errType := q.Get("error"); errType != "" {
+				if desc := q.Get("error_description"); desc != "" {
+					t.Errorf("got error from server %s: %s", errType, desc)
+				} else {
+					t.Errorf("got error from server %s", errType)
+				}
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: go
+go_import_path: github.com/coreos/dex
```
ericchiang: please remove this, it doesn't have anything to do with the change.

server/server_test.go
```diff
@@ -21,12 +21,12 @@ import (
 	"testing"
 	"time"
 
-	oidc "github.com/coreos/go-oidc"
+	"github.com/coreos/go-oidc"
```
ericchiang: please don't rewrite these imports.

server/server_test.go
```diff
@@ -787,31 +791,105 @@ func TestCrossClientScopes(t *testing.T) {
...
 }
 
+type crossClientsFixture struct {
+	provider     *oidc.Provider
+	client       *storage.Client
+	peer         *storage.Client
+	oauth2Server *httptest.Server
+}
+
+type crossClientsTestDeferred struct {
```
ericchiang: What are these new types? I'm confused by the naming here. Can you add some explanations or not refactor the tests?

server/server_test.go
```diff
@@ -690,10 +690,14 @@ func TestOAuth2ImplicitFlow(t *testing.T) {
 	}
 }
 
-func TestCrossClientScopes(t *testing.T) {
+func setupCrossClientsFixture(t *testing.T, crossClientsTest func(*crossClientsFixture, *crossClientsTestDeferred) (reqDump, respDump []byte)) {
```
ericchiang: What does `crossClientsTest` do? What should it perform? Why does it return `reqDump` and `respDump`? Can we just avoid this re-factor in this PR?

server/server_test.go
```diff
@@ -787,31 +791,105 @@ func TestCrossClientScopes(t *testing.T) {
...
 	}
+	fixture.peer = &peer
 
-	oauth2Config = &oauth2.Config{
-		ClientID:     client.ID,
-		ClientSecret: client.Secret,
-		Endpoint:     p.Endpoint(),
-		Scopes: []string{
-			oidc.ScopeOpenID, "profile", "email",
-			"audience:server:client_id:" + client.ID,
```
ericchiang: Why don't we just remove this claim for the test?

ericchiang:
Thanks. Overall I think I'd rather avoid doing any test refactoring in this PR. They need refactoring but that's a bigger job than just changing the claim behavior.

Can you squash your commits before you push again?

server/server_test.go
dpacierpnik: I wanted to ensure backward-compatibility so I left an old test (if clientID is explicitly requested, issued ID token should still contain only one occurrence of clientID in audience claim)

server/server_test.go
dpacierpnik: I introduced new types because there were cyclic dependencies in this code, and I wanted to somehow refactor it. But I will remove this refactoring as you suggested in overall comment, and I will just duplicate this test for the new case.

.travis.yml
dpacierpnik: Removed. I added go_import_path in travis.yaml because I was not able to run tests for my fork in Travis. With go_import_path Travis should work for your project as well as for all forks

server/server_test.go
dpacierpnik: Fixed

ericchiang:
lgtm! Please squash

dpacierpnik:
Squashed

PR #1090, opened 2017-09-28
connector/gitlab: Fix regexp in Link parser

lsjostro:
Fix regression in Link parser. 
.* is to ambiguous.

### DIFF
--- a/connector/gitlab/gitlab.go
+++ b/connector/gitlab/gitlab.go
@@ -22,6 +22,11 @@ const (
 	scopeAPI  = "api"
 )
 
+var (
+	reNext = regexp.MustCompile("<([^>]+)>; rel=\"next\"")
+	reLast = regexp.MustCompile("<([^>]+)>; rel=\"last\"")
+)
+
 // Config holds configuration options for gilab logins.
 type Config struct {
 	BaseURL      string `json:"baseURL"`
@@ -236,9 +241,6 @@ func (c *gitlabConnector) groups(ctx context.Context, client *http.Client) ([]st
 
 	apiURL := c.baseURL + "/api/v4/groups"
 
-	reNext := regexp.MustCompile("<(.*)>; rel=\"next\"")
-	reLast := regexp.MustCompile("<(.*)>; rel=\"last\"")
-
 	groups := []string{}
 	var gitlabGroups []gitlabGroup
 	for {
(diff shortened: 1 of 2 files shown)


ericchiang:
Can you add a test?

lsjostro:
Added test

connector/gitlab/gitlab_test.go
```diff
@@ -0,0 +1,29 @@
+package gitlab
+
+import (
+	"testing"
+)
+
+func TestNextURL(t *testing.T) {
+	linkHeader := "<https://gitlab.com/api/v4/groups?page=3&per_page=20>; rel=\"next\", " +
```
ericchiang: nit: we generally prefer table driven tests https://github.com/golang/go/wiki/TableDrivenTests

If you want to add that here that works, if not I'm fine with the test as is.

connector/gitlab/gitlab.go
```diff
@@ -272,22 +269,31 @@ func (c *gitlabConnector) groups(ctx context.Context, client *http.Client) ([]st
...
+	}
+	return groups, nil
+}
 
-		if len(reNext.FindStringSubmatch(link)) > 1 {
-			apiURL = reNext.FindStringSubmatch(link)[1]
-		} else {
-			break
+func nextURL(url string, link string) string {
+	reNext := regexp.MustCompile("<([^>]+)>; rel=\"next\"")
```
ericchiang: nit: move these to global vars so we don't re-compile them every time the function is called

```
var (
    reNext = regexp.MustCompile("...")
    reLast = regexp.MustCompile("...")
)

func nextURL() {
    // .. use reNext and reLast
}
```

connector/gitlab/gitlab.go
```diff
@@ -272,22 +269,31 @@ func (c *gitlabConnector) groups(ctx context.Context, client *http.Client) ([]st
...
+		if url == lastPageURL {
+			return ""
 		}
+	} else {
+		return ""
+	}
 
+	if len(reNext.FindStringSubmatch(link)) > 1 {
+		return reNext.FindStringSubmatch(link)[1]
+	} else {
```
ericchiang: go lint picked this up as fishy

```
/home/travis/gopath/src/github.com/coreos/dex/connector/gitlab/gitlab.go:296:9: if block ends with a return statement, so drop this else and outdent its block
Found 1 lint suggestions; failing.
```

connector/gitlab/gitlab_test.go
lsjostro: I'll fix! Thanks!

connector/gitlab/gitlab.go
lsjostro: Will do!

lsjostro:
squashed and fixed the above comments

lsjostro:
Awesome! Thanks! Looking forward to a patch release soon then. 
I have it hot patched in prod currently.

lsjostro:
@ericchiang when can we expect this merged?

ericchiang:
@lsjostro sorry was out of the office.

PR #1092, opened 2017-10-05
travis.yml: bump golang version

rithujohn191:


.travis.yml
```diff
@@ -4,7 +4,8 @@ sudo: required
 
 go:
   - 1.7.5
```
diegs: Can we drop this one, since it's vulnerable?

.travis.yml
```diff
@@ -4,7 +4,8 @@ sudo: required
...
-  - 1.8.3
+  - 1.8.4
```
diegs: Perhaps just specify `1.8` to pick up latest?

.travis.yml
```diff
@@ -4,7 +4,8 @@ sudo: required
...
+  - 1.9.1
```
diegs: Same here

rithujohn191:
addressed feedback

PR #1093, opened 2017-10-05
Unmarshal `WebConfig` settings

ereslibre:
When `frontend` is provided in the configuration, allow Dex to
unmarshal it, so their set properties are effective.

```
frontend:
  dir: /some/dir
  theme: my-theme
```


ericchiang:
This should just work anyway. "dir" should map to a field called "Dir". Can you add a test to show what you're fixing?

ereslibre:
You are right, my bad. Closing this PR. Thanks!

PR #1096, opened 2017-10-09
connector/ldap: add test for InsecureSkipVerify option

ericchiang:
updates https://github.com/coreos/dex/issues/907


ericchiang:
cc @rithujohn191 this only adds a test

PR #1100, opened 2017-10-21
Implement the “external” connector (for Apache2 mod_auth etc.)

stapelberg:
In Debian, we’re currently investigating whether we can use dex as part of https://sso.debian.org/ (see also https://wiki.debian.org/DebianSingleSignOn).

In the current setup, users authenticate against Apache2, and the web application behind sso.debian.org issues certificates based on the `REMOTE_USER` request environment variable.

Hence, to keep (especially user-visible) changes at a minimum, we’d like to keep the setup the same regarding authentication, just replace the certificates part with dex as an OIDC provider (for GitLab and others).

The “external” connector in this pull request accomplishes that; see also the included documentation.

Please consider merging this pull request. Thanks!

### DIFF
--- a/Documentation/authproxy.md
+++ b/Documentation/authproxy.md
@@ -1,0 +1,87 @@
+# External authentication
+
+## Overview
+
+The authproxy connector returns identities based on authentication which your
+front-end web server performs.
+
+The connector does not support refresh tokens or groups at this point.
+
+## Configuration
+
+The following is an example config file that can be used by the external
+connector to authenticate a user.
+
+```yaml
+connectors:
+- type: authproxy
+  id: myBasicAuth
+  name: HTTP Basic Auth
+```
+
+The authproxy connector assumes that you configured your front-end web server
+such that it performs authentication for the `/dex/callback/myBasicAuth`
+location and provides the result in the X-Remote-User HTTP header. The following
+configuration will work for Apache 2.4.10+:
+
+```
+<Location /dex/callback/myBasicAuth>
+    AuthType Basic
+    AuthName "db.debian.org webPassword"
+    AuthBasicProvider file
+    AuthUserFile "/etc/apache2/debian-web-pw.htpasswd"
+    Require valid-user
+
+    # Defense in depth: clear the Authorization header so that
+    # Debian Web Passwords never even reach dex.
+    RequestHeader unset Authorization
+
+    # Requires Apache 2.4.10+
+    RequestHeader set X-Remote-User expr=%{REMOTE_USER}@debian.org
+
+    ProxyPass "http://localhost:5556/dex/callback/myBasicAuth"
+    ProxyPassReverse "http://localhost:5556/dex/callback/myBasicAuth"
+</Location>
+```
+
+## Full Apache2 setup
+
+After installing your Linux distribution’s Apache2 package, place the following
+virtual host configuration in e.g. `/etc/apache2/sites-available/sso.conf`:
+
+```
+<VirtualHost sso.example.net>
+    ServerName sso.example.net
+
+    ServerAdmin webmaster@localhost
+    DocumentRoot /var/www/html
+
+    ErrorLog ${APACHE_LOG_DIR}/error.log
+    CustomLog ${APACHE_LOG_DIR}/access.log combined
+
+    <Location /dex/>
+        ProxyPass "http://localhost:5556/dex/"
+        ProxyPassReverse "http://localhost:5556/dex/"
+    </Location>
+
+    <Location /dex/callback/myBasicAuth>
+        AuthType Basic
+        AuthName "db.debian.org webPassword"
+        AuthBasicProvider file
+        AuthUserFile "/etc/apache2/debian-web-pw.htpasswd"
+        Require valid-user
+
+        # Defense in depth: clear the Authorization header so that
+        # Debian Web Passwords never even reach dex.
+        RequestHeader unset Authorization
+
+        # Requires Apache 2.4.10+
+        RequestHeader set X-Remote-User expr=%{REMOTE_USER}@debian.org
+
+        ProxyPass "http://localhost:5556/dex/callback/myBasicAuth"
+        ProxyPassReverse "http://localhost:5556/dex/callback/myBasicAuth"
+    </Location>
+</VirtualHost>
+```
+
+Then, enable it using `a2ensite sso.conf`, followed by a restart of Apache2.
--- a/server/server.go
+++ b/server/server.go
@@ -19,6 +19,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/coreos/dex/connector"
+	"github.com/coreos/dex/connector/authproxy"
 	"github.com/coreos/dex/connector/github"
 	"github.com/coreos/dex/connector/gitlab"
 	"github.com/coreos/dex/connector/ldap"
@@ -240,6 +241,9 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
 	handleFunc("/auth", s.handleAuthorization)
 	handleFunc("/auth/{connector}", s.handleConnectorLogin)
 	handleFunc("/callback", s.handleConnectorCallback)
+	// For easier connector-specific web server configuration, e.g. for the
+	// "authproxy" connector.
+	handleFunc("/callback/{connector}", s.handleConnectorCallback)
 	handleFunc("/approval", s.handleApproval)
 	handleFunc("/healthz", s.handleHealth)
 	handlePrefix("/static", static)
@@ -381,7 +385,7 @@ func (s *Server) startGarbageCollection(ctx context.Context, frequency time.Dura
 
 // ConnectorConfig is a configuration that can open a connector.
 type ConnectorConfig interface {
-	Open(logrus.FieldLogger) (connector.Connector, error)
+	Open(id string, logger logrus.FieldLogger) (connector.Connector, error)
 }
 
 // ConnectorsConfig variable provides an easy way to return a config struct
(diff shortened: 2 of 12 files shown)


ericchiang:
Yeah does look interesting.

Just trying to understand the use cases. You run a proxy, then if the user ever tries to hit `(dex URL)/callback/external` you log the user in, setting the `X-Remote-User` header on the way back?

What's the UX look like for that? How would another provider other than Apache2 mod_auth use this? Do you have examples?

stapelberg:
> Just trying to understand the use cases. You run a proxy, then if the user ever tries to hit (dex URL)/callback/external you log the user in, setting the X-Remote-User header on the way back?

This is correct.

> What's the UX look like for that?

This depends on the specific authentication method you have configured in Apache2. Common cases include a modal password prompt (e.g. https://user-images.githubusercontent.com/59393/26989841-371d98c2-4d1a-11e7-8182-6901d76081b5.png), a certificate prompt, some sort of smart-card authentication, etc.

> How would another provider other than Apache2 mod_auth use this? Do you have examples?

Have a look at https://httpd.apache.org/docs/current/howto/auth.html for an overview of how to configure authentication in apache.

Also note that this connector is not specific to apache, it’s just that we’re using apache in Debian. I could also imagine using this connector to integrate with proprietary web servers / appliances in company intranets.

server/server.go
```diff
@@ -240,6 +241,9 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
 	handleFunc("/auth", s.handleAuthorization)
 	handleFunc("/auth/{connector}", s.handleConnectorLogin)
 	handleFunc("/callback", s.handleConnectorCallback)
+	// For easier connector-specific web server configuration, e.g. for the
+	// "external" connector.
+	handleFunc("/callback/{connector}", s.handleConnectorCallback)
```
ericchiang: This needs to be the ID of the connector not the type, all connectors can be configured multiple times. I can configure two LDAP connectors with IDs `activedirectory` and `freeipa`.

It also needs enforce that `handleConnectorCallback` only allows a session using whatever value `{connector}` is. If it's `/callback/okta` that the user actually logging in through `okta`.

connector/external/external.go
```diff
@@ -0,0 +1,59 @@
...
+		return connector.Identity{}, fmt.Errorf("required HTTP header X-Remote-User is not set")
+	}
+	return connector.Identity{
+		Email:         remoteUser,
+		EmailVerified: true,
+	}, nil
+}
+
+// Refresh updates the identity during a refresh token request.
+func (m *Callback) Refresh(ctx context.Context, s connector.Scopes, identity connector.Identity) (connector.Identity, error) {
```
ericchiang: Since this can't be implemented, drop the method. We did this in the OpenID Connect connector because refreshes are hard, but regret it. Better to just not support refresh tokens.

connector/external/external.go
```diff
@@ -0,0 +1,59 @@
...
+	return u.String(), nil
+}
+
+// HandleCallback parses the request and returns the user's identity
+func (m *Callback) HandleCallback(s connector.Scopes, r *http.Request) (connector.Identity, error) {
+	remoteUser := r.Header.Get("X-Remote-User")
+	if remoteUser == "" {
+		return connector.Identity{}, fmt.Errorf("required HTTP header X-Remote-User is not set")
+	}
+	return connector.Identity{
```
ericchiang: Groups as well?

connector/external/external.go
```diff
@@ -0,0 +1,59 @@
...
+	Logger logrus.FieldLogger
+}
+
+// LoginURL returns the URL to redirect the user to login with.
+func (m *Callback) LoginURL(s connector.Scopes, callbackURL, state string) (string, error) {
+	u, err := url.Parse(callbackURL)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse callbackURL %q: %v", callbackURL, err)
+	}
+	u.Path = u.Path + "/external"
```
ericchiang: nit, this should probably be passed to the config. Feel free to leave a TODO if it's not obvious.

connector/external/external.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Open returns an authentication strategy which requires no user interaction.
+func (c *Config) Open(logger logrus.FieldLogger) (connector.Connector, error) {
+	return &Callback{Logger: logger}, nil
+}
+
+// Callback is a connector which returns an identity with the HTTP header
+// X-Remote-User as verified email.
+type Callback struct {
+	Logger logrus.FieldLogger
```
ericchiang: nit: callback and logger should be lowercase so they're unexported. We only export things we explicitly have a reason to share.

stapelberg: Done.

connector/external/external.go
stapelberg: Done. On second thought, though: shouldn’t this be derived from the connector ID? I couldn’t see whether connectors have access to the ID they’re configured as.

connector/external/external.go
stapelberg: How can I derive any groups from X-Remote-User header, which is just an email address?

connector/external/external.go
stapelberg: Done.

server/server.go
stapelberg: > This needs to be the ID of the connector not the type, all connectors can be configured multiple times. I can configure two LDAP connectors with IDs activedirectory and freeipa.

Yes. The `/auth/{connector}` route assigns `connID := mux.Vars(r)["connector"]`, which is why I chose `/callback/{connector}` to signify the connector ID. Is that incorrect? Should “connector” be changed to “id”, here and above?

> It also needs enforce that handleConnectorCallback only allows a session using whatever value {connector} is. If it's /callback/okta that the user actually logging in through okta.

Done.

connector/external/external.go
```diff
@@ -0,0 +1,59 @@
...
+	"net/url"
+
+	"github.com/coreos/dex/connector"
+	"github.com/sirupsen/logrus"
+)
+
+// Config holds the configuration parameters for a connector which returns an
+// identity with the HTTP header X-Remote-User as verified email.
+type Config struct {
+	PathSuffix string `json:"pathSuffix"`
```
ericchiang: I mean, this should be passed by the server since it know the URL this connector will operate at.

connector/external/external.go
ericchiang: Kubernetes allows additional headers that represent groups. https://kubernetes.io/docs/admin/authentication/#authenticating-proxy

Fine with not addressing this now, but it limits how useful this will be to other services.

connector/external/external.go
ericchiang: Maybe let connectors know what their ID is?

```
func (c *Config) Open(id string, logger logrus.FieldLogger) (connector.Connector, error)
```

stapelberg: Done.

connector/external/external.go
stapelberg: Ah, thanks for the pointer. Added a TODO to the code, so that someone can actually verify it works with Kubernetes when they implement it :).

ericchiang:
Code looks fine. Few final requests:

* "external" doesn't feel like the correct name. It seems like it should be "authenticatingproxy" or something like that. (open to shorter names)
* A working example in tree that we can use to test this. See our LDAP setup for a good example https://github.com/coreos/dex/blob/master/Documentation/ldap-connector.md#getting-started
* Docs should call out limitations. This mode doesn't support refresh tokens or groups.

stapelberg:
> "external" doesn't feel like the correct name. It seems like it should be "authenticatingproxy" or something like that. (open to shorter names)

How about “authproxy” as a shorter version?

> A working example in tree that we can use to test this. See our LDAP setup for a good example https://github.com/coreos/dex/blob/master/Documentation/ldap-connector.md#getting-started

The docs already contain an example Apache config. Are you asking me to add the required config files to actually start a full Apache server, or are you asking me to add a unit test?

> Docs should call out limitations. This mode doesn't support refresh tokens or groups.

Will do.

ericchiang:
> How about “authproxy” as a shorter version?

Works for me

> The docs already contain an example Apache config. Are you asking me to add the required config files to actually start a full Apache server, or are you asking me to add a unit test?

A more complete example would be preferable. I've never run an Apache server, for instance. How would I test this? Even links to a basic "getting started" guide would be helpful for future dex maintainers.

stapelberg:
Got it. I’ll update the pull requests in a few hours.

stapelberg:
All done. Please take another look.

Documentation/authproxy.md
```diff
@@ -0,0 +1,87 @@
...
+        Require valid-user
+
+        # Defense in depth: clear the Authorization header so that
+        # Debian Web Passwords never even reach dex.
+        RequestHeader unset Authorization
+
+        # Requires Apache 2.4.10+
+        RequestHeader set X-Remote-User expr=%{REMOTE_USER}@debian.org
+
+        ProxyPass "http://localhost:5556/dex/callback/myBasicAuth"
```
ericchiang: One concern I just realized. The user can still hit the `http://localhost:5556/dex/callback` URL to login against the `authproxy` connector. Is this protected from the user setting X-Remote-User?

This seems dangerous. I'd like to add the following recommendation:

```
If this connect is used, it's recommended to use this as a sole connector.
```

I can't really imagine a safe way to allow multiple connectors while this is set...

ericchiang: cc @stapelberg

stapelberg: Good catch. Let me send a PR which strips X-Remote-User for all dex URLs aside from the correct one.

ericchiang: :+1: awesome, thanks

PR #1101, opened 2017-10-24
connector: implement LinkedIn connector

pborzenkov:
connector/linkedin implements authorization strategy via LinkedIn's
OAuth2 endpoint + profile API.

It doesn't implement RefreshConnector as LinkedIn doesn't provide any
refresh token at all (https://developer.linkedin.com/docs/oauth2, Step 5
— Refresh your Access Tokens) and recommends ordinary AuthCode exchange
flow when token refresh is required.


connector/linkedin/linkedin.go
```diff
@@ -0,0 +1,181 @@
...
+	ctx := r.Context()
+	token, err := c.oauth2Config.Exchange(ctx, q.Get("code"))
+	if err != nil {
+		return identity, fmt.Errorf("linkedin: get token: %v", err)
+	}
+
+	client := c.oauth2Config.Client(ctx, token)
+	profile, err := c.profile(ctx, client)
+	if err != nil {
+		return identity, fmt.Errorf("github: get profile: %v", err)
```
ericchiang: nit: "github"

connector/linkedin/linkedin.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	identity = connector.Identity{
+		UserID:        profile.ID,
+		Username:      profile.FirstName + " " + profile.LastName,
```
ericchiang: nit: we should be smarter about this value. e.g. ensure that the FirstName and LastName are actually set.

connector/linkedin/linkedin.go
```diff
@@ -0,0 +1,181 @@
...
+		Email:         profile.Email,
```
ericchiang: Please error if profile.Email isn't set.

connector/linkedin/linkedin.go
```diff
@@ -0,0 +1,181 @@
...
+type profile struct {
+	ID        string `json:"id"`
+	FirstName string `json:"firstName"`
+	LastName  string `json:"lastName"`
+	Email     string `json:"emailAddress"`
+}
+
+func (c *linkedInConnector) profile(ctx context.Context, client *http.Client) (p profile, err error) {
+	// https://developer.linkedin.com/docs/fields/basic-profile
+	req, err := http.NewRequest("GET", apiURL+"/people/~:(id,first-name,last-name,email-address)", nil)
```
ericchiang: this is a really weird URL. ha

connector/linkedin/linkedin.go
pborzenkov: That's seems like an impossible condition. This is the main email that was used for registration and we always ask permission for r_emailaddress claim.
I'd add the check just to be sure anyway.

connector/linkedin/linkedin.go
pborzenkov: ok

connector/linkedin/linkedin.go
pborzenkov: Will fix.

ericchiang:
This conflicts with some additions in https://github.com/coreos/dex/pull/1100

Mind rebasing?

PR #1105, opened 2017-10-27
Add support for more OOB templating

kfox1111:
This PR adds support for templating the OOB support per client, with custom
response headers, and template values. For example, this can be used to
allow the authcode to be returned in a downloadable json file with per
client contact url like so in the config:
  staticClients:
 - name: 'Example App'
    oobHeaders:
    - Content-Type: application/octet-stream
    - Content-Disposition: attachment; filename="bootstrap.json"
    oobTemplate: bootstrap.json
    oobValues:
      server: http://foo.example.com
and template web/templates/bootstrap.json:
{
  "server": "{{ .Values.server }}",
  "authCode": "{{ .Code }}"
}

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -525,7 +525,18 @@ func (s *Server) sendCodeResponse(w http.ResponseWriter, r *http.Request, authRe
 			// Implicit and hybrid flows that try to use the OOB redirect URI are
 			// rejected earlier. If we got here we're using the code flow.
 			if authReq.RedirectURI == redirectURIOOB {
-				if err := s.templates.oob(w, code.ID); err != nil {
+				client, err := s.storage.GetClient(authReq.ClientID)
+				if err != nil {
+					s.logger.Errorf("Failed to get client %q: %v", authReq.ClientID, err)
+					s.renderError(w, http.StatusInternalServerError, "Failed to retrieve client.")
+					return
+				}
+		        	for _, headers := range client.OOBHeaders {
+		        		for k, v := range headers {
+						w.Header().Add(k, v)
+					}
+				}
+				if err := s.templates.oob(w, code.ID, client.OOBTemplate, client.OOBValues); err != nil {
 					s.logger.Errorf("Server template error: %v", err)
 				}
 				return
--- a/server/server.go
+++ b/server/server.go
@@ -173,7 +173,7 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
 		theme:     c.Web.Theme,
 	}
 
-	static, theme, tmpls, err := loadWebConfig(web)
+	static, theme, tmpls, err := loadWebConfig(web, c.Storage)
 	if err != nil {
 		return nil, fmt.Errorf("server: failed to load web static: %v", err)
 	}
--- a/server/templates.go
+++ b/server/templates.go
@@ -2,7 +2,7 @@ package server
 
 import (
 	"fmt"
-	"html/template"
+	"text/template"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+
+        "github.com/coreos/dex/storage"
 )
 
 const (
@@ -34,6 +36,7 @@ type templates struct {
 	passwordTmpl *template.Template
 	oobTmpl      *template.Template
 	errorTmpl    *template.Template
+	oobTmpls     map[string]*template.Template
 }
 
 type webConfig struct {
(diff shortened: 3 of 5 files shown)


kfox1111:
Not sure whats wrong here. I didn't change version.go, but its complaining:
```
?   	github.com/coreos/dex/version	[no test files]
```
Any ideas?

kfox1111:
@ericchiang

ericchiang:
To the best of my understanding, the limited UI is part of the security of this feature. Forcing a user to copy and paste a code instead of doing automatic redirects where they could be tricked into visiting links that log them into something. Though I could be wrong, since we copied this from a similar Google OAuth2 flow.

I'd rather see us invest effort in things like https://github.com/coreos/dex/issues/911 than trying to making the oob flow customizable. At least the device flow has specs and recommendations, whereas anything that drifts from Google's oob flow might mean we're on our own with the security implications.

kfox1111:
all its doing is letting you basically download the auth code as a file, rather then requiring the user to copy/paste it into a program/file. I don't think thats directly chainable to "run something with the code"? Maybe I'm mistaken.

It should have the same security properties as the html page I think. Its kicked off the same exact way one and isn't directly usable by ajax to retrieve the contents of the file. (At least no more so then the web page)

With this kind of feature, I think you could put up a template for a kubeconf for your cluster, and it could download a kubeconf for your cluster, preconfigured with an access code, that the ps in https://github.com/kubernetes/kubernetes/pull/54421 could use it.

The main benifit to it is nothing but kubeactl itself would ever be able to see a refresh code for the user, and that client must have it for it to work properly if it is to refresh id tokens. refresh tokens are the most sensitive thing in the system so protection of them is the most important tradeoff I think? It can also be used without the offline scope and an id only token access code can be given and it will still work. the same flow would work either way for those that have plugins that support refresh tokens  (ldap) and those that dont (saml).

Something like this use case:
1. user logs into kube-dashboard via dex
2. user does some dashboardy things.... then decides they want to use the cli.
3. user click on "download kubeconf". Its really just a link to something like: 
https://xxx.xxx.xxx.xxx:5556/auth?approval_prompt=force&client_id=kubernetes&redirect_uri=urn:ietf:wg:oauth:2.0:oob&response_type=code&scope=openid+email+profile+offline_access+groups
4. the user authenticates with dex (safer, as you don't want someone walking away from their keyboard and some other user being able to fetch a refresh token)
5. download of kubeconf happens.
6. user can use kubectl against their downloaded kubeconf. If user wants to use a kubectl on a remote machine, they scp the kubeconf over first. no local web server is needed. should work in a multiuser environment safely.

With this pr, the pr at https://github.com/kubernetes/kubernetes/pull/54421 and a small change to kube-dashboard, this whole flow works.

What do you think?

kfox1111:
As for the device flow route, I don't think there is any reference to supporting the device flow in the OIDC spec (OAuth2 only)? Is that really a valid option?

ericchiang:
As I stated in that kubernetes issue I think the proposal is a weird use of an OAuth2 code. Inventing expedited flows and having clients ignore "state" parameters is not something I want to encourage, and dex shouldn't enable it.

Please write a proper OAuth2 client that talks to dex and does this rendering, while adhering to best OAuth2 practices.

ericchiang:
> As for the device flow route, I don't think there is any reference to supporting the device flow in the OIDC spec (OAuth2 only)? Is that really a valid option?

The OIDC spec doesn't cover things like the oob flow either. It only covers the authorization code, implicit, and hybrid flows. However if the flows produce a token response, we're generally open to adding an id_token to that response in dex.

kfox1111:
I've never seen anyone describe how to tie a private and public client together and get a refresh token for the public client (needed if you were to have a web based client generate a kubeconf with a valid refresh token in it), or securely handle having a public client that was safe to run on a multiuser machine. Do you know how to do those things?

If not, this is so far the best thing I've been able to come up with so far. it works with the oob flow and if you can't trust your oidc provider to render the template I'm not sure you should be trusting it to manage refresh tokens either? Is there an attack vector I missed?

My understanding was state was optional? Is there a use case here that I'm missing?

Thanks,
Kevin

kfox1111:
So, how do we make progress on using dex for cli tokens? I really would like to encourage dex's adoption in new communities such as the helm community, but since a lot of their users can only use the cli at the moment, lack of cli support is crippling adoption and my ability to advocate it.

ericchiang:
> So, how do we make progress on using dex for cli tokens?

In tectonic we render kubeconfigs from a client app. I would expect other components to do that as well.

Or again, adding features like https://github.com/coreos/dex/issues/911

PR #1106, opened 2017-10-27
Added User Info endoint for oidc connectors

fjbsantiago:
User Info endpoint.
Currently only works with OIDC connector by adding a **userInfo** configuration entry.

Other connectors can make use of the endpoint by implementing the GetUserInfo interface.
Upon login, the connector can store any relevant context information in Identity.connectorData before returning.
This data will be passed back later on as an parameter for the GetUserInfo method.


fjbsantiago:
Will fix errors.

PR #1107, opened 2017-10-27
Added User Info endpoint

fjbsantiago:
Currently only works with OIDC connector by adding a userInfo configuration entry, but ready to be implemented for other connectors.

Other connectors can make use of the endpoint by implementing the GetUserInfo interface.
Upon login, the connector can store any relevant context information in Identity.connectorData before returning.
This data will be passed back later on as an parameter for the GetUserInfo method.

### DIFF
--- a/Documentation/oidc-connector.md
+++ b/Documentation/oidc-connector.md
@@ -42,6 +42,12 @@ connectors:
     # following field.
     #
     # basicAuthUnsupported: true
+
+    # Some clients require the possibility to get further user details provided
+    # by via the UserInfo endpoint.
+    #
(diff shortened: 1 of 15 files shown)


ericchiang:
Per my original proposal, this shouldn't require storing additional data https://github.com/coreos/dex/issues/376#issuecomment-271725208

I'm really sorry, but I'm going to close this. An implementation that only works with an OpenID Connect connector isn't going to get merged, and I want to be forthright about our limited bandwidth to review and get other contributor's PRs into shape.

We're trying get some internal sprint time soon to work on this.

PR #1108, opened 2017-10-27
Add etcd backed storage

dqminh:
This adds a storage implementation backed by etcd. This should be useful mostly in cases where we dont want to depend on a separate SQL cluster ( and etcd is easier to operate ), or if we dont want to incur some overhead of using kubernetes CRD.

There are no documentation yet ! Once this looks good i can try to add them.


storage/etcd/config.go
```diff
@@ -0,0 +1,82 @@
...
+	"time"
+
+	"github.com/coreos/dex/storage"
+	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/pkg/transport"
+	"github.com/sirupsen/logrus"
+)
+
+type EtcdSSL struct {
+	ServerName string `json:"server_name" yaml:"server_name"`
```
ericchiang: nit: we use camel case for struct tags. `json:"serverName"`

storage/etcd/config.go
```diff
@@ -0,0 +1,82 @@
...
+	CAFile     string `json:"ca_file" yaml:"ca_file"`
+	KeyFile    string `json:"key_file" yaml:"key_file"`
+	CertFile   string `json:"cert_file" yaml:"cert_file"`
+}
+
+type Etcd struct {
```
ericchiang: Allow a prefix

storage/etcd/config.go
```diff
@@ -0,0 +1,82 @@
...
+	Endpoints        []string `json:"endpoints" yaml:"endpoints"`
+	DialTimeout      int      `json:"dial_timeout" yaml:"dial_timeout"`
```
ericchiang: We generally don't make everything configurable just because the client library does. Can we set DialTimeout and AutoSyncInterval to reasonable defaults and if there are compelling reasons to expose them later we can?

storage/etcd/config.go
```diff
@@ -0,0 +1,82 @@
...
+	AutoSyncInterval int      `json:"auto_sync_interval" yaml:"auto_sync_interval"`
+	Username         string   `json:"username" yaml:"username"`
+	Password         string   `json:"password" yaml:"password"`
+	SSL              EtcdSSL  `json:"ssl" yaml:"ssl"`
+	RejectOldCluster bool     `json:"reject-old-cluster" yaml:"reject-old-cluster"`
```
ericchiang: What does this do?

storage/etcd/etcd.go
```diff
@@ -0,0 +1,472 @@
...
+	return toStorageRefreshToken(token), nil
+}
+
+func (c *conn) UpdateRefreshToken(id string, updater func(old storage.RefreshToken) (storage.RefreshToken, error)) error {
+	key := refreshTokenPrefix + id
+	ctx := context.TODO()
+	return c.txnUpdate(ctx, key, func(currentValue []byte) ([]byte, error) {
+		var current RefreshToken
+		if len(currentValue) > 0 {
+			if err := json.Unmarshal([]byte(currentValue), &current); err != nil {
```
ericchiang: Any thoughts on using a faster serialization format like protobuf? We use JSON for CRDs because CRDs only support JSON

storage/etcd/standup.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+
+if [ "$EUID" -ne 0 ]
```
ericchiang: Admittedly we've used these scripts less and less. I think instructions for downloading the binary releases would suffice https://github.com/coreos/etcd/releases

storage/etcd/etcd.go
```diff
@@ -0,0 +1,472 @@
...
+	if err != nil {
+		return err
+	}
+	if !res.Succeeded {
+		return storage.ErrAlreadyExists
+	}
+	return nil
+}
+
+func (c *conn) txnUpdate(ctx context.Context, key string, update func(current []byte) ([]byte, error)) error {
```
ericchiang: cc @xiang90 mind taking a look at this?

storage/etcd/etcd_test.go
```diff
@@ -0,0 +1,91 @@
...
+func TestEtcd(t *testing.T) {
+	testEtcdEnv := "DEX_ETCD_ENDPOINTS"
+	endpoints := os.Getenv(testEtcdEnv)
+	if endpoints == "" {
+		t.Skipf("test environment variable %q not set, skipping", testEtcdEnv)
+		return
+	}
+	newStorage := func() storage.Storage {
+		s := &Etcd{
+			Endpoints: []string{endpoints},
```
ericchiang: Weird that we call the var ENDPOINTS but only allow one.

ericchiang:
Since adding a new storage is a big overhead to all future developers, if this is committed we _need_ to be running these tests in travis.

Also just to address your commit message

> - kubernetes is not available yet, or if kubernetes depends on dex
to perform authentication and the operator would like to remove any
circular dependency if possible.

We intentionally build the OpenID Connect authenticator in Kubernetes to operate fine if dex isn't running. While the dependency looks circular, our experience of running dex on Tectonic would imply that that's a perfectly fine setup.

storage/etcd/config.go
gyuho: Why don't we just use `transport.TLSInfo` https://github.com/coreos/etcd/blob/master/pkg/transport/listener.go#L59?

```go
tlsInfo := transport.TLSInfo{
    CertFile:      "/tmp/test-certs/test-name-1.pem",
    KeyFile:       "/tmp/test-certs/test-name-1-key.pem",
    TrustedCAFile: "/tmp/test-certs/trusted-ca.pem",
}
tlsConfig, err := tlsInfo.ClientConfig()
...
clientv3.New(clientv3.Config{
    ...
    TLS:         tlsConfig,
})
```

xiang90:
@dqminh 

Can we handle timeout better? Or there is a future plan? Right now, we simply pass in context.TODO().

storage/etcd/config.go
gyuho: Ok, nvm. It's for json, yaml tagging.

ericchiang:
@xiang90 I think exposing a context is something we have to do on the dex end, unfortunately.

storage/etcd/etcd.go
xiang90: looks right to me. 

we could do auto retry inside the txnUpdate by calling update func multiple times when there are conflicts (maybe it is already wrapped?).

storage/etcd/etcd.go
```diff
@@ -0,0 +1,472 @@
...
+	txn := c.db.Txn(ctx)
+	updateResp, err := txn.
+		If(clientv3.Compare(clientv3.ModRevision(key), "=", modRev)).
+		Then(clientv3.OpPut(key, string(updatedValue))).
+		Commit()
+	if err != nil {
+		return err
+	}
+	if !updateResp.Succeeded {
+		return fmt.Errorf("failed to update key=%q: updated not by us", key)
```
xiang90: failed to update key = %q: concurrent conflicting update happened.

xiang90:
I only took a look at the txnUpdate func. Let me know if there is anything else I could help on.

storage/etcd/etcd.go
dqminh: I dont think auto-retry should be applied here. At least thats how the current test is structured ( also how sql transaction works ).
Although IIUC there are differences on the order of transaction between sql and non-sql storages now ( k8s or etcd ). In sql the first started transaction win vs the first finished transaction win in others.

storage/etcd/config.go
dqminh: Hmm what do you mean ? Like setting a etcd namespace ? I think its usually done now with the help of the grpc-proxy, isnt it ?

storage/etcd/config.go
dqminh: Hmm that works for me to. I can set them to the same default as etcdctl.

storage/etcd/config.go
dqminh:
```
// RejectOldCluster when set will refuse to create a client against an outdated cluster.
RejectOldCluster bool `json:"reject-old-cluster"`
```

as before, will also remove this knob for now.

storage/etcd/etcd_test.go
dqminh: Yah, in tests we only setup 1 endpoint so it still works :stuck_out_tongue: I was lazy to do the transformation from env vars to list of endpoints.

dqminh:
> Can we handle timeout better? Or there is a future plan? Right now, we simply pass in context.TODO().

yah i think the storage interface needs to be changed so we can pass context downwards. Right now not sure what else can we do here:
1. using context.TODO()
2. enforce a hard-code timeout for all operations.

xiang90:
> enforce a hard-code timeout for all operations.

this is what i usually do if the api layer does not expose context. it is safer. but i am not a dex expert, so i cannot speak for this exact case.

storage/etcd/etcd.go
xiang90: @dqminh understood. thanks.

storage/etcd/config.go
ericchiang: > Like setting a etcd namespace ?

Yeah an equivalent of `etcd grpc-proxy start --namespace=<prefix>`

> I think its usually done now with the help of the grpc-proxy, isnt it ?

I think it depends on your setup. But yeah we can document the grpc-proxy solution then add native namespace functionality later.

ericchiang:
A hard-coded timeout sounds fine.

storage/etcd/etcd.go
dqminh: I dont have any preference. Go's json stdlib is indeed slow-ish compared to other serialization format, but usually its not significant in practice i think.

dqminh:
added default timeout = 5s, and kv namespace support via `Etcd.Namespace` field.

storage/etcd/config.go
ericchiang: Still need struct tags

```
Endpoints []string `yaml:"endpoints"`
```

dqminh:
@ericchiang Thanks ! I will look into docs and travis change.

PR #1117, opened 2017-11-07
Email verification claim for local accounts

pborzenkov:
The pull request adds code to control 'email_verified' claim for local accounts. It is now stored in a storage and it's possible to update it (set to true) via gRPC call to Dex.


ericchiang:
@pborzenkov we closed a similar PR earlier https://github.com/coreos/dex/pull/960#issuecomment-323458881 and commented on an issue here https://github.com/coreos/dex/issues/957

I really don't want dex to add features like this. Dex shouldn't be a user database, there are better systems for that.

pborzenkov:
@ericchiang This is really sad to hear.

In this case I'm not sure how is local connector supposed to be used.
What we wanted to achieve is to restrict some functionality depending on whether a user has confirmed his/her email address or not. With third-party authentication providers we rely on them and the fact that a user has a verified email is passed to us in an ID token issued by Dex. This way each service the token is passed to can check, if needed, this claim.

But with local provider this claim is always true. Not only this is inconsistent and confusing, but the biggest problem is that we can't really known which provider is originally responsible for the token and whether we can trust the claim or not.

We've chosen Dex because it hides the complexities of dealing with different authentication methods and provides a uniform way to authenticate a user. Not having this feature for local password connector and having it for the rest of them kind of kills this uniformity.

ericchiang:
Dex has taken a really strong stance that local users are only for bootstrapping, not a full fledged feature. With that said, I think there are people actively using local users as we didn't intended, and this feature would be an improvement for them.

What we really want to avoid is building user management features like sending emails, dashboards, etc. As long as improvements are part of the existing API, I think we should start accepting them.

I'll do a review of this PR later (unless @rithujohn191 beats me to it). Thanks for sharing your use cases and sorry for taking a bit to respond.

vyshane:
I'm integrating with Dex via the gRPC API, and I also had to work around the fact that Dex wouldn't let me update the `email_verified` status.

I think that this feature is in the spirit of enabling us to keep user status in Dex in sync with our user management system, as opposed to turning Dex into a user management system.

nabokihms:
Was fixed in https://github.com/dexidp/dex/releases/tag/v2.45.0

PR #1119, opened 2017-11-08
*: Go 1.7 no longer supported and updated build image to 1.9

ericchiang:
Updates https://github.com/coreos/dex/issues/1118

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.8.3-alpine
+FROM golang:1.9.2-alpine
 
 MAINTAINER Ed Rooth <ed.rooth@coreos.com>
 MAINTAINER Lucas Servén <lucas.serven@coreos.com>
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,8 +1,4 @@
-FROM golang:1.8.3-alpine
-
-MAINTAINER Ed Rooth <ed.rooth@coreos.com>
-MAINTAINER Lucas Servén <lucas.serven@coreos.com>
-MAINTAINER Rithu John <rithu.john@coreos.com>
+FROM golang:1.9.2-alpine
```
rithujohn191: why remove the maintainers list?

ericchiang: Sorry should have called that out.

I'd like to clean up the top level MAINTAINER list soon. I don't really see the point of having this list here as well. Our container's don't need our emails in them.

Either way, I can revert this change if we'd like to keep these.

rithujohn191: Its fine to take them out here as long as we maintain the top level one and update the list there.

ericchiang: I'll revert the change from this PR for now

PR #1121, opened 2017-11-09
oauth2: don't return an access_token if token wasn't in response_type

srenatus:
Stumbling upon this part of the OIDC spec<sup>[1]</sup>,

> No Access Token is returned when the value is id_token.

I've realised that we don't comply. Now, we should. 😃 

[1]: https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthRequest


srenatus:
Also, the current logic allows *passing no nonce* when requesting `code id_token`. 🐛

srenatus:
> Does this PR now not return an access_token when the response_type is "code"? That seems wrong.

🤔 I would think so, why would it be wrong? I'm probably missing something -- if the response_type is `code` (and _not_ `code token`), should it return an `access_token`?

However, this should not affect [the actual token request](http://openid.net/specs/openid-connect/core-1_0.html#TokenRequest), I've just had a go with example-app. The authorization code flow still works.

PR #1123, opened 2017-11-09
show "back" link for password connectors

### LINKED ISSUE #1115: "go back to selection" for password template
When using Dex with multiple connectors, and one of them being the _local passwords_ connector, clicking on "Login with Email" here,

screen shot 2017-11-07 at 10 22 45

leads me to its login dialogue,

screen shot 2017-11-07 at 10 22 51

But, if I mistakenly clicked that, or I've made up my mind, there's no way to go back to the selection.

It's easy to add a ` window.history.back();` link to our template, but in situations where there's only one -- the local users -- connector, that doesn't make any sense, as there's no selection go back to.

Any ideas how to handle this best? I think we'd might have to add another field to the template [here](https://github.com/coreos/dex/blob/ccf85a726922a2fd6e4f0742d99f311dbf1fdb66/server/templates.go#L193-L197)?

srenatus:
This way, the user who has selected, say, "Log in with Email" can make up
their mind, and select a different connector instead.

However, if there's only one connector set up, none of this makes sense -- and
the link will thus not be displayed.

screen shot 2017-11-09 at 16 18 25

Fixes #1115.

### DIFF
--- a/web/templates/password.html
+++ b/web/templates/password.html
@@ -25,6 +25,11 @@
     <button tabindex="3" id="submit-login" type="submit" class="dex-btn theme-btn--primary">Login</button>
 
   </form>
+  {{ if .BackLink }}
+  <div class="theme-link-back">
+    <a class="dex-subtle-text" href="javascript:history.back()">Select another login method.</a>
+  </div>
+  {{ end }}
 </div>
 
 {{ template "footer.html" . }}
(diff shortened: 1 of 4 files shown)


web/templates/password.html
```diff
@@ -25,6 +25,11 @@ <h2 class="theme-heading">Log in to Your Account</h2>
     <button tabindex="3" id="submit-login" type="submit" class="dex-btn theme-btn--primary">Login</button>
 
   </form>
+  {{ if .BackLink }}
+  <div class="theme-link-back">
+    <a class="dex-subtle-text" href="javascript:history.back()">Select another Connector</a>
```
ericchiang: "Select another login method."

ericchiang:
One nit, otherwise lgtm

srenatus:
Fixed nit, rebased. 😃

PR #1135, opened 2017-11-27
Update slapd.sh

mpashka:


scripts/slapd.sh
```diff
@@ -13,6 +13,8 @@ command -v slapd >/dev/null 2>&1 || {
    sudo dnf -y install openldap-servers openldap-clients 
 
    sudo apt-get install slapd ldap-utils
+   
+   Please make sure to disable or setup AppArmor on ubuntu or selinux in order OpenLDAP to run.
```
ericchiang: Does OpenLDAP not ship with the correct selinux profiles? Why do you need to disable it? I can run these with selinux enabled.

mpashka: I used it on ubuntu. On ubuntu there is AppArmor enabled by default. So I killed two hours to understand why slapd can't open options file. Because by default it is allowed to open files only in some specific folder (like /etc/openldap or something lika that).

ericchiang: Instead of asking to disable this, can we add suggestions to setup AppArmor correctly?

mpashka: Well, my propose is to 'disable or setup' :). May be 'setup properly or disable' is better option.
That is very distro-specific, so I suppose there is no reason to put any instructions here. And that things (AppArmor and selinux) are usually useful in production environment. Not sure if they are used on dev computers.

mpashka: @ericchiang changed comment. Feel free to change it or extend it - I'll update PR.

mpashka:
@ericchiang Can we proceed with this? Or you suppose it is bad idea and it is better to close this PR?

scripts/slapd.sh
```diff
@@ -13,6 +13,8 @@ command -v slapd >/dev/null 2>&1 || {
    sudo dnf -y install openldap-servers openldap-clients 
 
    sudo apt-get install slapd ldap-utils
+
+   Note: Please make sure to setup properly or just disable AppArmor on ubuntu or selinux on other distros in order OpenLDAP to run corrently and be able to open configuration files.
```
ericchiang: > Note: certain OpenLDAP packages may include AppArmor or SELinux configurations which prevent actions this script takes, such as referencing config files outside of its default config directory.

PR #1144, opened 2017-12-08
handlers/connector_login: update AuthRequest irregardless of method

srenatus:
Before, you could not POST your credentials to a password-connector's
endpoint without GETing that endpoint first. While this makes sense for
browser clients; automated interactions with Dex don't need to look at
the password form to fill it in.

A symptom of that missing GET was that the POST succeeded (!) with

    login successful: connector "", username="admin", email="admin@example.com", groups=[]

Note the connector "". A subsequent call to finalizeLogin would then
fail with

    connector with ID "" not found: failed to get connector object from storage: not found

Now, the connector ID of an auth request will be updated for both GETs
and POSTs.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -224,14 +224,16 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Set the connector being used for the login.
-	updater := func(a storage.AuthRequest) (storage.AuthRequest, error) {
-		a.ConnectorID = connID
-		return a, nil
-	}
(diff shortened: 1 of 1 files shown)


srenatus:
> Does it make sense to optimize with a check before doing this? [...]

I've pushed a commit, 4a505f8. It think we can save a DB roundtrip there, indeed, for most common cases. So, yeah, it probably makes sense.

srenatus:
Can we get this merged? I'd love to kill the workaround we've got in our tests 😉

PR #1176, opened 2018-01-23
New id_provider scope that adds the connector ID and user ID to the ID token claims

vyshane:
This PR addresses #892.

It implements a new scope `id_provider` that adds a new claim `id_provider` to the ID token. The claim has the following format:

```
"id_provider": {
  "connector_id": "github",
  "user_id": "110272483197731336751"
}
```


vyshane:
@ericchiang is there anything I can do to help progress this feature?

More fundamentally, do we want this feature (in current form or another) in Dex? I'm currently working on some services that integrate with Dex and it would be helpful to know if this approach is going to be a dead end.

Documentation/custom-scopes-claims-clients.md
```diff
@@ -12,6 +12,7 @@ The following is the exhaustive list of scopes supported by dex:
 | `email` | ID token claims should include the end user's email and if that email was verified by an upstream provider. |
 | `profile` | ID token claims should include the username of the end user. |
 | `groups` | ID token claims should include a list of groups the end user is a member of. |
+| `id_provider` | ID token claims should include the connector ID and the user ID that is returned by the connector. |
```
ericchiang: `provider_id`?

Documentation/custom-scopes-claims-clients.md
```diff
@@ -22,6 +23,7 @@ Beyond the [required OpenID Connect claims][core-claims], and a handful of [stan
 | Name | Description |
 | ---- | ------------|
 | `groups` | A list of strings representing the groups a user is a member of. |
+| `id_provider` | The connector ID and the user ID assigned to the user at the provider. |
```
ericchiang: Add an example somewhere that includes the full JSON object?

Documentation/custom-scopes-claims-clients.md
vyshane: I've updated the description to say

> ID token claims should include information from the ID provider. The token will contain the connector ID and the user ID assigned at the provider.

Hopefully it makes more sense now?

Documentation/custom-scopes-claims-clients.md
vyshane: Makes sense. Done!

ericchiang:
`id_provider` is just a weird name for a claim. Even something like `connector_user_id` or `federated_id` seems better to me.

Can the scope be called `federated:id` and the claim object be called `federated_claims`? That way we could adds scopes like `federated:ldap:dn` and have a place for those claims to exist.

vyshane:
> Can the scope be called `federated:id` and the claim object be called `federated_claims`? That way we could adds scopes like ` federated:conn:ldap:dn` and have a place for those claims to exist.

@ericchiang sounds good. I'll update the PR over the weekend.

vyshane:
PR is ready for review again.

vyshane:
Done!

redbaron:
Why some claims are persisted (see Claims struct in storage/storage.go), but some are not?

PR #1200, opened 2018-03-07
Update check go major version way

carbin-gun:
the current way  just keeps one prefix number as the major number, it should be the whole number after the dot.

for now, if make executed, go 1.10 will be reported **unsupported**

### DIFF
--- a/scripts/check-go-version
+++ b/scripts/check-go-version
@@ -5,7 +5,7 @@ set -e
 VERSION=$( go version )
 
 # For development versions of Go, these will be empty.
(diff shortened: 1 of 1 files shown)


redbeard0x0a:
I ran into this problem as well:

```
[master][~/.go/src/github.com/coreos/dex]$ go version
go version go1.10 darwin/amd64
[master][~/.go/src/github.com/coreos/dex]$ make
ERROR: dex requires Go version 1.8+. Please update your Go installation: https://golang.org/dl/
make: *** [check-go-version] Error 2
[master][~/.go/src/github.com/coreos/dex]$
```

PR #1202, opened 2018-03-08
ldap: If GroupSearch.GroupAttr is "DN", treat UserAttr as group DN.

ryotarai:
In [JumpCloud LDAP](https://support.jumpcloud.com/customer/portal/articles/2439911-using-jumpcloud-s-ldap-as-a-service), a group attribute (memberOf) in a user entry is DN value, and I introduce this patch to support this case.

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -48,6 +48,8 @@ import (
 //         # Will only work if "DN" attribute does not exist in the record
 //         # userAttr: DN
 //         groupAttr: member
+//         # If userAttr value is group DN, set groupAttr to "DN"
+//         # groupAttr: DN
 //         nameAttr: name
 //
 type Config struct {
@@ -124,7 +126,8 @@ type Config struct {
 		//
 		//   (<groupAttr>=<userAttr value>)
 		//
-		UserAttr  string `json:"userAttr"`
+		UserAttr string `json:"userAttr"`
+		// If UserAttr value is DN, set GroupAttr to "DN"
 		GroupAttr string `json:"groupAttr"`
 
 		// The attribute of the group that represents its name.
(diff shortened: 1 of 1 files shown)


andrein:
@soggiest I am also having similar issues with a ldap server provided by FreeIPA. In our case, the membership info is stored in memberOf attributes on the user, and the value is the DN of the group. 

```
# andrein, users, accounts, ipa.example.net
dn: uid=andrein,cn=users,cn=accounts,dc=ipa,dc=example,dc=net
memberOf: cn=admins,cn=groups,cn=accounts,dc=ipa,dc=example,dc=net
```

I've tried the following config and it doesn't work:

```        userSearch:
          baseDN: cn=users,cn=accounts,dc=ipa,dc=example,dc=net
          filter: (objectClass=inetOrgPerson)
          username: uid
          idAttr: uid
          emailAttr: mail
        groupSearch:
          baseDN: cn=groups,cn=accounts,dc=ipa,dc=example,dc=net
          filter: (objectClass=groupOfNames)
          userMatchers:
    #      TODO(andrein): this doesn't work, find out why
          - userAttr: memberOf
            groupAttr: DN
          nameAttr: cn
```

The dex logs show the following:

```
time="2021-09-20T09:45:31Z" level=info msg="performing ldap search cn=groups,cn=accounts,dc=ipa,dc=example,dc=net sub (&(objectClass=groupOfNames)(DN=cn=admins,cn=groups,cn=accounts,dc=ipa,dc=example,dc=net))"
time="2021-09-20T09:45:31Z" level=error msg="ldap: groups search with filter \"(&(objectClass=groupOfNames)(DN=cn=admins,cn=groups,cn=accounts,dc=ipa,dc=example,dc=net))\" returned no groups"
```

I believe this pull request will fix this issue.

the-maldridge:
I would also like to see this fix applied.  In my environment its possible to get all the groups for a user with a single query, without needing to do an additional search for groups.  Since my environment is using an LDAP protocol-aware proxy, not having to keep track of additional searches would be a win.

nabokihms:
I will close this issue because the `groupSearch.GroupAttr` option became deprecated. If you still need this feature, please open a new PR. Sorry for being late with reviewing.

PR #1205, opened 2018-03-15
Port is ignored if exists

ptzianos:
Closes: https://github.com/coreos/dex/issues/1204


ericchiang:
Closing. See https://github.com/coreos/dex/issues/1204#issuecomment-374772901

PR #1206, opened 2018-03-19
Evaluate symlinks when loading templates

totahuanocotl:
to avoid considering a symlinked directory as a file.

It correctly processes a structure like:
```
templates
├── ..data  ->  ..actual
├── ..approval.html  ->  ..data/approval
└── ..actual
        └── approval.html
```
whereas before it failed with the error:
`failed to initialize server: server: failed to load web static: parse files: read web/templates/..data: is a directory`.

### DIFF
--- a/server/templates.go
+++ b/server/templates.go
@@ -126,6 +126,18 @@ func loadTemplates(c webConfig, templatesDir string) (*templates, error) {
 
 	filenames := []string{}
 	for _, file := range files {
+		if file.Mode()&os.ModeSymlink > 0 {
+			var err error
+			fp := filepath.Join(templatesDir, file.Name())
+			fp, err = filepath.EvalSymlinks(fp)
+			if err != nil {
+				return nil, fmt.Errorf("evaluate symlink: %v", err)
+			}
+			file, err = os.Stat(fp)
+			if err != nil {
+				return nil, fmt.Errorf("retrieve file info: %v", err)
+			}
+		}
 		if file.IsDir() {
 			continue
 		}
--- a/server/templates_test.go
+++ b/server/templates_test.go
@@ -1,1 +1,54 @@
 package server
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestTemplatesCheckSymlinkDirs(t *testing.T) {
+	templateDir, err := createTemplates()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	c := webConfig{}
+	_, err = loadTemplates(c, templateDir)
+
+	if err != nil {
+		t.Errorf("expected err to not have ocurred, got %q", err)
+	}
+}
+
+// createTemplates creates a directory structure containing empty templates using symlinks to folders and files, e.g.:
+// templateDir
+// ├── ..data          -> ..actual
+// ├── ..approval.html -> ..data/approval.html
+// └── ..actual
+// 	   └── approval.html
+func createTemplates() (string, error) {
+	templateDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		return "", err
+	}
+	actualDir := join(templateDir, "..actual")
+	symlinkDir := join(templateDir, "..data")
+	if err = os.Mkdir(actualDir, 0755); err != nil {
+		return "", err
+	}
+	if err = os.Symlink(actualDir, symlinkDir); err != nil {
+		return "", err
+	}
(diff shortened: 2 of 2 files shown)


totahuanocotl:
The sample structure was generated by mounting a kubernetes configMap on the /web/templates folder to override the templates.

ericchiang:
Why is this needed? I'd really strongly avoid any symlink logic if we can.

totahuanocotl:
We were trying to override the Dex default templates with custom ones mounting a configMap as a Volume using the dex image. 
Kubernetes generates a similar structure where it creates a `..data` symlink to an actual folder where the contents of the configMap are placed. 
because Dex is listing the files templates dir, and iterating over the contents to search for the template files, when it reaches the `..data` symlink it is considered a file, not a directory, and it is added to the list of templates. 
When `ParseFiles(filenames...)` is called, it fails with the `../..data: is a directory` error. 

Actual contents of the mounted configMap:
```
/web/templates # ls -alrt                                                                                                                   
total 0                                                                                                                                     
drwxr-xr-x    5 root     root            51 Mar  5 19:15 ..                                                                                 
lrwxrwxrwx    1 root     root            20 Mar 19 20:52 password.html -> ..data/password.html                                              
lrwxrwxrwx    1 root     root            15 Mar 19 20:52 oob.html -> ..data/oob.html                                                        
lrwxrwxrwx    1 root     root            17 Mar 19 20:52 login.html -> ..data/login.html                                                    
lrwxrwxrwx    1 root     root            18 Mar 19 20:52 header.html -> ..data/header.html
lrwxrwxrwx    1 root     root            18 Mar 19 20:52 footer.html -> ..data/footer.html
lrwxrwxrwx    1 root     root            17 Mar 19 20:52 error.html -> ..data/error.html
lrwxrwxrwx    1 root     root            20 Mar 19 20:52 approval.html -> ..data/approval.html                                              
lrwxrwxrwx    1 root     root            31 Mar 19 20:52 ..data -> ..2018_03_19_20_52_19.241424967
drwxr-xr-x    2 root     root           138 Mar 19 20:52 ..2018_03_19_20_52_19.241424967
```
Another solution would be to specifically search for the required templates only, removing the need to list the contests of the templates folder.

ericchiang:
> Kubernetes generates a similar structure where it creates a ..data symlink to an actual folder where the contents of the configMap are placed.

This was also the cause of one of the CVEs that went out last week.

Please just modify your dex config. You can specify something like:

```yaml
frontend:
  dir: ( your target dir )
```

totahuanocotl:
Unfortunately that value expects the root of the web folder that also includes the static and themes resources folders. 
With config maps there is no way to create that structure recursively, and it is back to pod yml meddling again.

ericchiang:
If there's a solution that doesn't involve symlinks we'd be open to a
workaround.

totahuanocotl:
@ericchiang hi again, would you consider the use `os.Stat()` to get the `FileInfo` for the files in the template dir as involving symlinks?
`ioutil.ReadDir()` uses `os.Lstat()` to get the `FileInfo`'s which does not follow the symlink, and therefore `isDir()` returns false for a symlinked dir, whereas `os.Stat().IsDir()` will properly return true for a symlinked dir.

jsravn:
I'm confused - what's wrong with symlinks? This sounds like a dex bug to me - it should handle symlinks if I choose to use them. And it seems like an easy fix without requiring the user to jump through hoops.

LongStoryMedia:
this is still an issue...

```sh
failed to initialize server: server: failed to load web static: parse files: read templates/..data: is a directory
```

given 

```yaml
frontend:
  dir: /etc/dex/dex-templates
```

and
```yaml
          volumeMounts:
            - name: config-out
              mountPath: /etc/dex
            - name: dex-web
              mountPath: /etc/dex/dex-templates
            - name: dex-static
              mountPath: /etc/dex/dex-templates/static
            - name: dex-themes
              mountPath: /etc/dex/dex-templates/themes
            - name: dex-templates
              mountPath: /etc/dex/dex-templates/templates
...
      volumes:
        - name: config-in
          configMap:
            name: config
        - name: config-out
          emptyDir: {}
        - name: dex-templates
          configMap:
            name: dex-templates
        - name: dex-static
          configMap:
            name: dex-static
        - name: dex-themes
          configMap:
            name: dex-themes
        - name: dex-web
          configMap:
            name: dex-web
```

documentation around this is woefully inadequate.

PR #1209, opened 2018-03-21
.well-known/openid-configuration: update scopes and claims

srenatus:
As per [spec](https://openid.net/specs/openid-connect-discovery-1_0.html), none of these would have to be there; but I don't see why we should omit them.

> **scopes_supported**
>    RECOMMENDED. JSON array containing a list of the OAuth 2.0 [RFC6749] scope values that this server supports. The server MUST support the openid scope value. Servers MAY choose not to advertise some supported scope values even when this parameter is used, although those defined in [OpenID.Core] SHOULD be listed, if supported.

> **claims_supported**
>    RECOMMENDED. JSON array containing a list of the Claim Names of the Claims that the OpenID Provider MAY be able to supply values for. Note that for privacy or other reasons, this might not be an exhaustive list. 

I don't really know how to account for "nested claims", i.e., how `federated_claims` works, though...

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -111,11 +111,13 @@ func (s *Server) discoveryHandler() (http.HandlerFunc, error) {
 		Keys:        s.absURL("/keys"),
 		Subjects:    []string{"public"},
 		IDTokenAlgs: []string{string(jose.RS256)},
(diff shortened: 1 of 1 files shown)


nabokihms:
The list of supported claims and listed was updated several times after this PR was introduced. I'm going to close this PR due to its age.

Issue #1212, opened 2018-03-30
Keepalive may result in client callback being invoked on dex.


joshua-rutherford:
When dex and a client are behind the same load balancer and that load balancer is performing TLS passthrough based upon SNI the client callback will be invoked on dex instead of the client since the IP address of dex and the client will be the same and the browser already has a connection open to that IP address (only the connection is with dex, not the client).  

Waiting thirty seconds or more prior to clicking the dex callback results in the keepalive timing out and thus the client callback is invoked.

I propose allowing keepalive behavior to be set by configuration (i.e., enabled or disabled with enabled as default).  This will obviously have performance impacts in some cases but not enabling this means this particular deployment topology is not possible at all.

PR #1221, opened 2018-04-18
Allow keep alives to be disabled

joshua-rutherford:
This just allows users to toggle keepalives for the REST endpoints while maintaining backwards compatibility.  This is useful in the case described in #1212.


joshua-rutherford:
@ericchiang, any chance this could get a review.  If this is not something you are interested in supporting I'm happy maintain a fork but would like to know.  Thanks.

ericchiang:
Disabling keep alives seems like a big trade-off. What do other
applications do for TCP pass through? How bad is it to have all traffic
from one user go to a single dex instance?

PR #1224, opened 2018-04-25
Ensure GitLab connector uses fullpath for subgroups

crashdump:
Currently, the Gitlab connector pulls all of the groups and subgroups into a flat list without any concept of hierarchy – which is a potential security issue in the case of GitLab where people can create top level groups/project and sub-projects with non-unique names.

This PR fixes this behaviour by prefixing all of the subgroups with the parent name.

**Gitlab group scenario**

```
 +- foo
 |   +- bar
 +- bar
```

```
GET /api/v4/groups
[
  {
    "name": "Foo 1 Parent Group",
    "path": "foo1",
    "full_path": "foo1",
  },
  {
    "name": "Bar Subgroup",
    "path": "bar",
    "full_path": "foo/bar",
  },
  {
    "name": "Bar Parent Group",
    "path": "bar",
    "full_path": "bar",
  }
] 
```

**Dex's response to the authenticating application**

Before
```
{ "groups": [
     "foo"
     "foo"
     "bar"
 ]}
```

After
```
{ "groups": [
     "foo"
     "foo-bar"
     "bar"
 ]}
```

### DIFF
--- a/connector/gitlab/gitlab.go
+++ b/connector/gitlab/gitlab.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"regexp"
 	"strconv"
+	"strings"
 
 	"github.com/coreos/dex/connector"
 	"github.com/sirupsen/logrus"
@@ -47,7 +48,7 @@ type gitlabUser struct {
 type gitlabGroup struct {
 	ID   int
 	Name string
-	Path string
+	Path string `json:"full_path"`
 }
 
 // Open returns a strategy for logging in through GitLab.
@@ -269,7 +270,7 @@ func (c *gitlabConnector) groups(ctx context.Context, client *http.Client) ([]st
 		}
 
 		for _, group := range gitlabGroups {
-			groups = append(groups, group.Name)
+			groups = append(groups, strings.Replace(group.Path, "/", "-", -1))
 		}
 
 		link := resp.Header.Get("Link")


connector/gitlab/gitlab.go
```diff
@@ -269,7 +270,7 @@ func (c *gitlabConnector) groups(ctx context.Context, client *http.Client) ([]st
 		}
 
 		for _, group := range gitlabGroups {
-			groups = append(groups, group.Name)
+			groups = append(groups, strings.Replace(group.Path, "/", "-", -1))
```
davidarcher: Any reason to change the separator character from slash to dash?

I suspect there is a similar fix happening on the GitLab side too so it would be good to be aligned to however GitLab is going to fix it to expose the groups directly via oidc in the future.

EDIT: Here's the GitLab MR that went in recently to add groups to the oidc claims: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/16929/diffs#c763217d83955db9b5d8edeaf64c1fa43c8ef54b_83_95 -- they're using the group path as-is with the slash separator so I think we should do the same here.

kskewes:
Hi @crashdump, thanks for your work on this. 
Are you interested in updating to use slash like GitLab instead of hyphen? 
Then I think we should try get this merged in. 
Lots of forks are using this with that change.

nabokihms:
The problem was fixed in https://github.com/dexidp/dex/pull/1351
Thanks to everyone who worked on this issue. If you still have problems, please open a new issue or discussion 🙏

PR #1251, opened 2018-06-19
connector/ldap: treat attributes as case-insensitive all the way

srenatus:
LDAP itself doesn't match attributes case-sensitively, so a query for

    (fOO=bar)

will return all the entries having `foo: bar`, regardless of how the
attribute name is defined in the schema.

However, the result entries we get will have the case according to what
the server returned. So, assuming the schema calls it `foo`, this is
what we're given.

Now, for certain attributes that are user-configurable -- NameAttr for
example -- this can cause trouble: the query returns the right results,
but the LDAP connector can't make sense of it as it's looking for an
attribute called `fOO`.

The real life case underlying this conundrum is Active Directory's
sAMAccountName.

Note that this is a bit of a workaround. Ideally, we'd use
`(*ldap.Entry).GetAttributes(string) []string` to access the attributes;
but until the case sensitivity issue is fixed there[1], changing the way
we read the results doesn't help us.

[1]: https://github.com/go-ldap/ldap/pull/111

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"strings"
 
 	"gopkg.in/ldap.v2"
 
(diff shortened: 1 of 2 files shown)


srenatus:
@ericchiang Another small LDAP improvement I'd like to get in. Any thoughts? 😃

sks:
Does it make sense to do case insensitive test from for DN as well ?

I had small case dn in my config and it took me 30 mins to figure out this issue.

srenatus:
@sks I frequently find https://ldapwiki.com/ quite useful, too. 😃

PR #1266, opened 2018-08-02
fix timeout bug for etcd3 client connect

byxorna:
Heya! Subtle bug here for `etcd3` storage backend. The timeout was being unnecessarily multiplied by `time.Second` causing a near-infinite timeout before returning on connection errors. This makes debugging Dex with etcd storage very annoying :)

```
fmt.Printf("time 1: %v\n", 2 * time.Second)
fmt.Printf("time 2: %v\n", 2 * time.Second * time.Second)

output:
time 1: 2s
time 2: 555555h33m20s
```


byxorna:
Bump! Who should i @mention to get traction on getting this merged? Is there a CLA to sign?

PR #1269, opened 2018-08-05
[version_update] Update the version to the latest

tw3rp:
Version mentioned in this example is very old and was causing issues

### DIFF
--- a/examples/k8s/dex.yaml
+++ b/examples/k8s/dex.yaml
@@ -12,7 +12,7 @@ spec:
         app: dex
     spec:
       containers:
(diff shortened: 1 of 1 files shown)


jimangel:
I made the same PR a little bit ago, it seems like a trivial / quick change: https://github.com/coreos/dex/pull/1242

I will close mine in favor of this PR.

It seems like this could be merged regardless of the outcome of Dex ownership: https://github.com/coreos/dex/issues/1273

srenatus:
@tw3rp @jimangel thanks 🎉

Issue #1020, opened 2017-08-09
grpc connector


Rio:
Have there been any plans/thoughts on implementing a grpc connector?
One that will just implement the `PasswordConnector` and `RefreshConnector` interfaces.
Just to make it easier for people to use Dex as a frontend for token generation and all that jazz more in line with what Hydra is doing and without needing to fork the project to have your custom connector implemented.

fforootd:
Hi @Rio 

As of today, we decided to do exactly that in an forked version of dex.
Our idea was to test it first internally and contributed it afterwards.

I will ask the guy working on the connector tomorrow about the status.

Rio:
That's great! I was just implementing it just to see how it would work. Hopefully we could pull you fork back in.

ericchiang:
I think a pluggable connector for out-of-tree would be great. Committing to an interface is probably the biggest design hurdle here.

ericchiang:
Out of curiosity, what provider are you trying to hook into? My general experience is that making things pluggable is good but also doesn't magically produce more development. It'd be great to understand what the current pain points are.

fforootd:
We internally use a grpc services as part of our identity solution, somewhat like an ldap but it does only store identity's (username, password and user related data).  
The reason to not use an LDAP based approach is that we use the same authentication interface for the auth process between our microservices (every microservice gets itself a token from this service) 

So we thought about building a more generic grpc connector where one can define his endpoint and the rpc request name from protobuf, as well as the message parameters to map his own values against the connector's.

Rio:
We are in the process to move stuff over to a new architecture but we of course need to be backwards compatible until it's time. One of the things we're running into is not being able to extend the openid tokens with groups and what not in Dex's local password database using the grpc api. 

At the same time we were thinking to not store the users in there at all and since we're going to use grpc more and more we thought to try to see if we could connect Dex to our own user storage until we flesh out if and how we want Dex to do password storage for us and if we can live without the groups extendability for example.

We also have a legacy where the combination of username and password will determine which customer this user will be logged into. So same username but different passwords will get you different customers. Transitioning to a single canonical user record will take time as we use those different user ids for permissions. Which is why having extra logic during the grpc login call to our backend would still be necessary.

ericchiang:
For some background, dex is committed to not becoming an user management or authorization solution. That's why so many of the "local" user objects are so underdeveloped. If we could remove them we would, but they're too important for bootstrapping.

[...]

PR #1271, opened 2018-08-06
Expose GetPassword via the GRPC API

tylercloke:
GetPassword was already wired up to storage and everything, just took a bit more work to wire it through the API. This will be super helpful when building things on top of Dex as you can retrieve a password without enumerating the entire store via ListPasswords.


examples/grpc-client/README.md
```diff
@@ -48,6 +48,7 @@ Finally run the Dex client providing the CA certificate, client certificate and
 Running the gRPC client will cause the following API calls to be made to the server
 1. CreatePassword
 2. ListPasswords
+3. GetPassword
 3. DeletePassword
```
srenatus: [nit] I know it doesn't matter for presentation, but it's nicer to read this in raw form if the numbers are actually consecutive -- can you make `DeletePassword` a `4.`?

server/api.go
```diff
@@ -223,6 +223,33 @@ func (d dexAPI) ListPasswords(ctx context.Context, req *api.ListPasswordReq) (*a
...
+			return &api.GetPasswordResp{
+				NotFound: true,
+			}, nil
+		}
+		d.logger.Errorf("api: there was an error retrieving the password: %v", err)
+		return nil, fmt.Errorf("get password: %v", err)
+	}
+
+	return &api.GetPasswordResp{
+		NotFound: false,
```
srenatus: [nit] this line is superfluous, but being explicit is OK 👍

tylercloke: Yep was trying to be explicit.

examples/grpc-client/README.md
tylercloke: I was trying to maintain the style, but I'm just gonna make them all `1.`

srenatus:
@ericchiang What do you think about this one?

While #1020 would resolve our need for using dex for user storage, as long as that's the case, this addition would let us simplify all the place where we're currently going with
```go
	resp, err := s.c.ListPasswords(ctx, &api.ListPasswordReq{})
	if err != nil {
		return nil, errors.Wrap(err, "get password")
	}

	// searches returned users for user
	for _, p := range resp.Passwords {
		if p.Email == email {
			usr := users.ShowUser{
				ID:    p.UserId,
				Name:  p.Username,
				Email: p.Email,
			}
			return &usr, nil
		}
	}

	return nil, &users.NotFoundError{}
```
...and, since it doesn't seem too big a deal for dex' GRPC API, I'd think it's a good addition 😉

ericchiang:
Can we do #1272 instead?

srenatus:
@ericchiang well, these two are rather complementary... 🤔 Or am I missing something? `VerifyPassword` wouldn't be of much use [for simplifying this](https://github.com/dexidp/dex/pull/1271#issuecomment-423531244), would it?

api/api.proto
```diff
@@ -128,12 +128,23 @@ message RevokeRefreshReq {
...
   // Set to true is refresh token was not found and token could not be revoked.
   bool not_found = 1;
 }
 
+// GetPasswordReq is a request get a specific password by email.
+message GetPasswordReq {
+  string email = 1;
+}
+
+// GetPasswordResp a specific password by email.
```
ericchiang: This is really oddly named. Can we either rename it "GetUserResp" and "GetUserReq"? Or can we strongly clarify that the password isn't returned?

cippaciong:
@sagikazarmark @nabokihms I'm writing an internal component which relies on the gRPC API and I stumbled upon this issue while looking for a way the check if a certain Password/User exists in dex.

Is there any plan to include such an endpoint in the API?

PR #1274, opened 2018-08-08
Added option to attach groups to static password users

ingopansa:
This adds support for the groups scope for users in the staticPasswords list.

I found it in particular useful for quickly getting up a test environment, in which teams want to have an OIDC integration, but also be able to manage test users and their respective groups locally.

The groups information is provided through the config file. A demonstration of how this could look like is added to the example config file.

Happy to discuss!


ingopansa:
understood!

srenatus:
I'm closing this, @devonbarrett is spot-on with that comment. 😃

nabokihms:
Was fixed in https://github.com/dexidp/dex/releases/tag/v2.45.0

PR #1275, opened 2018-08-08
Extend the API with a function which updates the client configuration

ccojocar:


ccojocar:
@srenatus Any chance to get this merged soon? Thanks

ccojocar:
@srenatus Do you see any issues which prevent merging this PR?

api/api.proto
```diff
@@ -138,6 +153,8 @@ message RevokeRefreshResp {
 service Dex {
   // CreateClient creates a client.
   rpc CreateClient(CreateClientReq) returns (CreateClientResp) {};
+  // UpdateClient updates an exisitng client
```
srenatus: Typo
```suggestion
  // UpdateClient updates an existing client
```

server/api.go
```diff
@@ -79,6 +79,38 @@ func (d dexAPI) CreateClient(ctx context.Context, req *api.CreateClientReq) (*ap
...
+	}
+
+	err := d.s.UpdateClient(req.Id, func(old storage.Client) (storage.Client, error) {
+		if req.RedirectUris != nil && len(req.RedirectUris) > 0 {
+			old.RedirectURIs = req.RedirectUris
+		}
+		if req.TrustedPeers != nil && len(req.TrustedPeers) > 0 {
+			old.TrustedPeers = req.TrustedPeers
+		}
+		old.Public = req.Public
```
srenatus: So, we're "patchy" for `Name` and `LogoUrl` -- empty-string means no update -- but not for `Public`? There is ways to do this in GRPC, so I wonder if we want to do that.

Concretely, if `Public` was of type [`BoolValue`](https://godoc.org/github.com/golang/protobuf/ptypes/wrappers#BoolValue), we could differentiate "not provided" from "false". I suppose if we do that, we should do the same with the other fields; as you currently cannot
- unset trusted peers
- unset redirect URIs
- set the name to empty-string
- set the logoURL to empty-string

💭 I guess one could argue whether this is needed. But as it stands, you cannot update the client without _setting its public value_; that at least should be fixed IMHO

ccojocar:
@srenatus I addressed the review issues and also add some tests. Do you have any other concerns? Thanks

srenatus:
@ccojocar thanks for adding those tests 👍 -- I'm not sure I understand how the other comment has been resolved, though; so I've unresolved it 😄

server/api.go
ccojocar: What would be your suggestion in this case? if it is a private client and the user tries to update to public we should return an error?

ccojocar: It seems that I need to use an external package in the proto file in order to get these types working. 
```
import "google/protobuf/wrappers.proto";
```
Some details here: https://github.com/golang/protobuf/issues/438

I can't get it working. If you have any suggestion how to make this working, I am up to it. 

An alternative solution is to not update the `Public` field in the update operation.

ccojocar: I tried to address your concerns in the last commit. Please let me know if this solution is fine for you.

srenatus: Yeah, I'm not sure how to best deal with this one. An alternative to introducing wrappers could be to use a [FieldMask](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#fieldmask) in UpdateClientReq. And of course, we could just leave it as-is. What do you think, @ericchiang?

Re 49fa5ee6e8271d433cd9a3b00c3533495e285c3e, I'm not sure I follow... From what I can tell, it's now an append-only operation? I.e. whatever is in the UpdateClientReq.RedirectUris will be _added_ to old.RedirectUris; but nothing could ever get removed? I'm not sure if this really is what we want... I think something like
```go
		if req.RedirectUris != nil {
			old.RedirectURIs = req.RedirectUris
		}
```
would do the trick -- it lets you determine what the RedirectUris are _after the update_, and it allows you to set them to `[]string{}`. Before, what last thing wasn't possible (unless I'm mistaken). Likewise for TrustedClients.

Thanks for bearing with me here 😃

ccojocar: 😄 This was what I understood from your first comment. I reverted the last commit, and also allow to update the redirect URIs and trusted peers to []string{}. 

The `Public` filed update was removed as well. I think is not required to be updated after the initial registration. 

Hope this is what you are expecting! Thanks

ccojocar:
@srenatus Are there any pending issue related to this PR?

srenatus:
@ccojocar thanks for the bump. I'll look into this again today. 😃

api/api.proto
```diff
@@ -36,6 +36,21 @@ message DeleteClientResp {
   bool not_found = 1;
 }
 
+// UpdateClientReq is a request to update an exisitng client.
+message UpdateClientReq {
+    string id = 1;
+    repeated string redirect_uris = 2;
+    repeated string trusted_peers = 3;
+    bool public = 4;
```
srenatus: I think we should either just overwrite the _public_ field with whatever was in the `UpdateClientReq` (what you had pushed before; sorry about that) -- and **ensure that documentation** makes it obvious that the behaviour isn't _patchy_ when it comes to `public`; **or** keep the code as-is and remove the field from this message.

server/api_test.go
```diff
@@ -290,3 +290,174 @@ func TestRefreshToken(t *testing.T) {
...
+			t.Fatalf("delete client delete client returned no response")
+		}
+	}
+
+	tests := map[string]struct {
+		setup   func(t *testing.T, clientId string)
+		cleanup func(t *testing.T, clientId string)
+		req     *api.UpdateClientReq
+		wantErr bool
+		want    *api.UpdateClientResp
```
srenatus: These tests only assert stuff on the `UpdateClientResp`, but don't cover what that function has done to the client storage. What would you think, is there in adding that?

api/api.proto
ccojocar: I would prefer to remove the `public` field since is not used. Please could you point me to the documentation, which should I update?

server/api_test.go
ccojocar: Actually it does, see `L420`. In addition of checking the `UpdateClientResp`, it also retrieves the updated client from the storage and compares its configuration with the values from `UpdateClientReq`

api/api.proto
ccojocar: The filed is removed.

srenatus:
Ooops, I've merged this without checking the commits. I guess this could have been a candidate for a squash before merging. My fault! 😅

Issue #1277, opened 2018-08-26
Support used self-signed certificates LDAP.


veily:
For LDAP server that used self-signed certificates and  demanded TLS client verification,dex should load self-signed  certificates for client.
This is my solution.
1. add clientCA & clientKey config.
```go
	// Path to a trusted root certificate file.
	RootCA string `json:"rootCA"`
	// Path to a client cert file generated by rootCA.
	ClientCert string `json:"clientCert"`
	// Path to a client private key file generated by rootCA.
	ClientKey string `json:"clientKey"`
	// Base64 encoded PEM data containing root CAs.
	RootCAData []byte `json:"rootCAData"`
```
2. Load self-signed certificates.
```go
        if c.ClientKey != "" && c.ClientCert != "" {
                cert, err := tls.LoadX509KeyPair(c.ClientCert, c.ClientKey)
                if err != nil {
                        return nil, fmt.Errorf("ldap: load self signed certs file failed.error:%s", err)
                }
                tlsConfig.Certificates = append(tlsConfig.Certificates, cert)
        }
```
3. Setting  self-signed  certificates in config file.
```yml
connectors:
- type: ldap
  name: OpenLDAP
  id: ldap
  config:
    host: localhost:636
    rootCA: /etc/dex/sslcerts/ca.crt
    clientCA: /etc/dex/sslcerts/client.crt
    clientKey: /etc/dex/sslcerts/client.key
```

ericchiang:
Which providers do this? Got some docs?

veily:
Used for LDAP server that used self-signed certificates (RootCA, server cert) and setting TLS verify client. In this case, dex should use the client certificates generated by the Ldap server RootCA.
Without the client certificates, dex all requests to ldap server will be rejected.


http://www.openldap.org/doc/admin24/tls.html
> 16.1.2. Client Certificates
The DN of a client certificate can be used directly as an authentication DN.
...

srenatus:
Fixed in v2.12.0 (#1314), thanks to your help in #1278. 🎉

PR #1278, opened 2018-08-26
Support used self-signed certificates LDAP.

veily:
support used self-signed certificates LDAP. #1277


connector/ldap/ldap.go
```diff
@@ -68,7 +68,10 @@ type Config struct {
 
 	// Path to a trusted root certificate file.
 	RootCA string `json:"rootCA"`
-
+	// Path to a self-signed certificates ca file.
+	ClientCA string `json:"clientCA"`
```
srenatus: ❓ Wouldn't this mostly match the `RootCA`? I'm not sure I understand the usecase where it differs. Can you please elaborate?

connector/ldap/ldap.go
```diff
@@ -68,7 +68,10 @@ type Config struct {
...
+	//Path to a self-signed certificates private key file.
```
srenatus: [nit] Can we make that `// Path` (with a space)?

srenatus:
@veily Thanks for your contribution 😃

connector/ldap/ldap.go
veily: If ldap server setting verify client cert is true, only setting RootCA cannot work.

connector/ldap/ldap.go
```diff
@@ -68,7 +68,10 @@ type Config struct {
 
 	// Path to a trusted root certificate file.
 	RootCA string `json:"rootCA"`
-
+	// Path to a self-signed certificates cert file.
```
srenatus: Can we remove the "self-signed" there? I don't think it is necessary to have self-signed cert here.

veily: Sorry, I am not clear enough. The self-signed certificate is not dex. The ldap server uses a self-signed certificate (RootCA, server cert) and opens the certificate for the client. In this case, if dex Ability to authenticate client certificates through the ldap server, using client certificates generated by the Ldap server RootCA

veily: http://www.openldap.org/doc/admin24/tls.html
> 16.1.2. Client Certificates
The DN of a client certificate can be used directly as an authentication DN.
...

veily: without the client certificates, dex all requests to ldap server will be rejected.

veily:
Used for LDAP server that used self-signed certificates (RootCA, server cert) and setting TLS verify client. In this case, dex should use the client certificates generated by the Ldap server RootCA.
Without the client certificates, dex all requests to ldap server will be rejected.


http://www.openldap.org/doc/admin24/tls.html
> 16.1.2. Client Certificates
The DN of a client certificate can be used directly as an authentication DN.
...

connector/ldap/ldap.go
```diff
@@ -225,6 +227,14 @@ func (c *Config) openConnector(logger logrus.FieldLogger) (*ldapConnector, error
 		}
 		tlsConfig.RootCAs = rootCAs
 	}
+
+	if c.ClientKey != "" && c.ClientCert != "" {
+		cert, err := tls.LoadX509KeyPair(c.ClientCert, c.ClientKey)
+		if err != nil {
+			return nil, fmt.Errorf("ldap: load self signed certs file failed.error:%s", err)
```
srenatus: [nit] Can we please make that 
```go
fmt.Errorf("ldap: load client cert failed: %v", err)
```
just to be in line with the updated comment (thanks, btw), and the other error printouts above?

ericchiang:
FYI there are commits that aren't attributed to the author of this PR. Please squash.

srenatus:
@veily If you need help squashing commits, let me know! Once squashed, and having only you as author, we'll get this in 🎉

veily:
@srenatus Okay, Please squash.  In addition, I used different git username  during  the completion of this PR, this is my mistake.

srenatus:
Ah, I see. Could you please squash and do a force push onto your branch
(the one from this pr)? We could then merge it using the normal "merge"
button on github. Just to be in line with how we normally merge things.

veily:
@srenatus Thanks, I has squashed commits under your guidance.

srenatus:
Thank you! :)

PR #1279, opened 2018-08-28
fix default baseURL for GitLab connector

AnianZ:
The URL `https://www.gitlab.com` which is the default baseURL for the GitLab connector does not work any more. According to the DNS entries it's still pointing to an Azure ip adress while the whole service has moved to google cloud. Since `https://gitlab.com` is main URL for GitLab anyway that's probably a better default anyway. 

Documentation also updated.

### DIFF
--- a/Documentation/connectors/gitlab.md
+++ b/Documentation/connectors/gitlab.md
@@ -20,8 +20,8 @@ connectors:
     # Required field for connector name.
     name: GitLab
     config:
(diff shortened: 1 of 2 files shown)


AnianZ:
I notified GitLab about the problem as well but they are not sure what to do about that subdomain: https://gitlab.com/gitlab-com/production/issues/441#note_97442708

Issue #1281, opened 2018-09-01
Determine criteria for repo write access


ericchiang:
We had quite a few people volunteer in https://github.com/dexidp/dex/issues/1273, but we ideally want to have some pre-conditions. Helping merge PRs, fixing issues, etc.

Open to ideas and extra points for something written down.

srenatus:
@jimangel This sounds like a solid choice -- do you know if that approach documented in some place?

devonbarrett:
+1, it's documented here: https://github.com/kubernetes/community/blob/master/community-membership.md

jimangel:
Thinking more about this, and trying not to turn a molehill into a mountain, there are 2 other things I would add:

1. Since the k8s flow is more for community org membership and less about repo governance, we could simplify the process by having the user open a PR (vs. emailing a group) against the OWNERS file with the exact permission they're requesting (to be reviewed and approved by 2 people). I'm not against dexidp becoming an org / having members, but I don't consider it a day 1 need.

2. Someone (*cough* @srenatus  *cough*) should have the keys to the castle (ability to revert merges, remove reviewers / approvers if malicious or inactive, and cut releases). This person should have a top priority of getting 1 or 2 more people to their status as fast as possible so the community isn't relying on one admin for everything.

This approach can be clearly documented in a CONTRIBUTING file which can be referenced in the README

Side note: what is the plan for cutting releases? Should the person in the second item create a quay account and maintain that with their elevated repo access?

Thoughts?

srenatus:
Yeah, I suppose keeping this simple is preferable.

Please correct me if I'm wrong, but I think the obvious next steps here are
1. finish the move
   - [ ] replace `github.com/coreos/` by `github.com/dexidp`
   - [ ] ensure travis picks up tests in PRs (see #1280)
2. cut a release 
3. cleanup issues and PRs
4. (maybe?) find a channel somewhere to have do real-time communication (IRC? Matrix? a channel in the gophers' slack?)

With this in place, we can move this forward, discuss new features, release cycles, etc.

While I'm confident merging the smaller PRs, I'd appreciate someone outside of my employer's org for the bigger ones, to not drive the project into a direction that is unwanted by the community -- So, I don't know about the planned involvement (and time available) of @rithujohn191 and @ericchiang, but, to _+1_ @jimangel, we should get a few more people (from different orgs) on board quickly.

[...]

PR #1280, opened 2018-08-31
*: update the maintainers list

rithujohn191:

PR #1283, opened 2018-09-03
Finish GitHub org move

srenatus:
As mentioned in #1281, this seems necessary to finish the move.

However, we probably shouldn't merge this until Travis has been updated -- I'm lacking authorization for that, but I've requested it (see #1280).

With this branch, `docker build -t test .` looks successful; and `make test` passes locally ✔️


ericchiang:
@rithujohn191 can you make me an owner of the dexidp org? I need that to re-enable travis

.travis.yml
```diff
@@ -7,7 +7,7 @@ go:
   - 1.9
   - '1.10'
 
-go_import_path: github.com/coreos/dex
```
ericchiang: In hindsight, this should let us forgo the package rename. We might be able to split out removing the bill of materials.

srenatus: Technically, yeah... but if someone wanted to depend on post-move dex, those names would have to match, wouldn't they? (Ignoring the new go modules stuff for a moment.)

ericchiang: I would feel really bad if someone was importing dex 😬 we don't maintain these packages for general use.

srenatus: 😄 It's quite convenient to have an all-in-memory-dex in tests. We do it in a few places... 🤔

srenatus: I suppose the same holds for using its grpc client definitions.

rithujohn191:
@ericchiang changed your role to owner

ericchiang:
(closing and re-opening to see if this kicks travis)

srenatus:
@ericchiang Ok, nuked the bill of materials, and updated the quay references anticipating that we  publish this into `quay.io/dexidp` (which I get a 403 on right now?), #1284.

I'd really rather have the package imports match the repository. Even if we want no one to import dex, and document how to generate the gRPC client package yourself, it feels like an unnecessary extra step for contributors to ensure that the checkout matches a different import path (like, "clone this into `${GOPATH}/coreos/dex`"). While doing a rename now might necessitate a few fixes in open PRs, on the long run, _renaming-all-the-things now_ seems like the simpler solution to me. What do you think?

Dockerfile
```diff
@@ -2,8 +2,8 @@ FROM golang:1.10.2-alpine
 
 RUN apk add --no-cache --update alpine-sdk
 
-COPY . /go/src/github.com/coreos/dex
-RUN cd /go/src/github.com/coreos/dex && make release-binary
+COPY . /go/src/github.com/dexidp/dex
```
ericchiang: Probably early to update the quay references. Let's do this once we get that account setup.

I'll need your quay username too.

ericchiang: added

srenatus:
@ericchiang ✔️ squashed

PR #1285, opened 2018-09-04
connectors/ldap: treat 'constraint violation' on bind as bad credentials

srenatus:
Some directory servers (I think it's Oracle) return

    Constraint Violation: Exceed password retry limit. Account locked.

when attempting to login too many times. While constraint violation can
mean many things, we're checking this as an error on BIND, so it's
more likely that something like this has happened than any other thing.

Hence, we should treat it as an "incorrect password" situation, not an
internal error.

It would of course be preferrable to surface more information about this
precise error (and similar ones), but I think this is beyond this small
change.

(This is some prior art: https://github.com/vesse/passport-ldapauth/issues/21 )


srenatus:
I feel like I'm skipping the queue, but since the change is quite minimal, I'll merge this. (A quick look around didn't reveal any low-hanging fruit in terms of improving test coverage for this, so, let's leave it at that.)

PR #1288, opened 2018-09-05
Allow embedding dex in other cobra projects with custom connectors

amdonov:
Moved most of the code from main package to a subfolder cmd. Expose both the serve and version commands. Included a sample cmd and connector for others to use as a reference for using their own connectors.

Closes #1282


ericchiang:
I don't think this solves https://github.com/dexidp/dex/pull/1282. You'd still need to modify the server, which we shouldn't expose.

ericchiang:
Also per https://github.com/dexidp/dex/issues/1020#issuecomment-321849533, we don't commit to the interfaces you're exposing. Anyone who does this is basically forking dex, so why not just fork it?

ericchiang:
Libraries require designing around APIs with compatibility in mind. We change stuff all the time, plumb new types through dex, and users shouldn't depend on these internal packages. One day we may split out the OAuth2 code into something like https://github.com/RangelReale/osin, but until then I don't want users importing the dex server or depending on the connector interfaces being stable. That's not what it's intended for and would place a huge burden on the dex maintainers.

Closing to save time going back and forth. I'm fine refactoring code, but not having other projects depend on dex's internal types.

Please fork dex if you need to support connectors we don't want to merge into main, or consider chiming in on some of our other efforts to support more generic user backends https://github.com/dexidp/dex/issues/1020

amdonov:
I understand, but I wish you felt differently. You're not signing a support contract with folks. 

Of course, you're free to change your API. That's what versions are for. :-)

Looking forward to seeing #1020. Not likely to commit anything. Merge approval is too opinionated for me.

Issue #1284, opened 2018-09-03
Release v2.11.0


srenatus:
[There have only been a few minor changes since v2.10.0](https://github.com/dexidp/dex/compare/v2.10.0...master), but the repo move makes for a good occasion to cut a release. Also, we should include #1253 (thanks @vasartori).

Open questions:
- Should this become `quay.io/dexidp/dex:v2.10.0`? I suppose so, keeping `quay.io/coreos/dex` feels odd.

From glancing at them, I think the [how to create a release](https://github.com/dexidp/dex/blob/master/Documentation/dev-releases.md) docs are up-to-date, but we'll see.

@rithujohn191 @ericchiang What do you think?

ericchiang:
I'm for moving this to a unique registry. This is in line with what etcd did after it moved out of the CoreOS org (https://github.com/etcd-io/etcd/releases). I've gone ahead and registered quay.io/dexidp.

I've created a google group to handle the quay.io admin emails. Let me add a maintainer doc with the groups new maintainers have to be added to. That doc can also include a few words covering the discussion in https://github.com/dexidp/dex/issues/1281

We probably need to update the scripts to.

Ideally, we could automate the container pushes. It always feels weird to have users pushing from their laptop and puts a lot of power in individual maintainers.

srenatus:
> Ideally, we could automate the container pushes. It always feels weird to have users pushing from their laptop and puts a lot of power in individual maintainers.

I agree completely and will set aside a little time today to look into the wiring between github and quay.

srenatus:
✔️ I've set up a build trigger, Quay is now building for every banch/tag pushed to `github.com/dexidp/dex`. I'll update the instructions and docs accordingly.

srenatus:
I've manually triggered builds for `master` and the latest release, `v2.10.0`. (I've included that release so that we can update `examples/k8s/dex.yml`... I don't suppose it makes much sense to re-build all the previous releases, as long as `quay.io/coreos/dex` doesn't go away.)

srenatus:
https://github.com/dexidp/dex/releases/tag/v2.11.0 ✔️

PR #1290, opened 2018-09-06
release process: update quay notes

srenatus:
This complements #1284. 

For every push to master (this should include merges, but I haven't merged anything since setting this up), and every pushed tag, quay will now build (and tag, respectively) docker images in the https://quay.io/dexidp/dex repository.

### DIFF
--- a/Documentation/dev-releases.md
+++ b/Documentation/dev-releases.md
@@ -3,16 +3,12 @@
 Making a dex release involves:
 
 * Tagging a git commit and pushing the tag to GitHub.
-* Building and pushing a Docker image.
 
-This requires the following tools.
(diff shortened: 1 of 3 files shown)


srenatus:
> Also, can you please destroy the private key of the SSH key you created? If it's in quay, I don't want that on a developer's local workstation :)

I don't have anything but the public key myself. The setup was done entirely by clicking through the wizard on Quay. 😄

ericchiang:
> @ericchiang I would merge this and push a tag (#1284) if you agreed with it. I don't know what to add to the notes, saying "set this up once by clicking through the quay wizard" seems a little lacking in content 😅)

Okay sounds good to me.

PR #1293, opened 2018-09-09
RPM Build

amdonov:
Relates to #811. Looked like progress on an RPM had stalled. I made one that works well on CentOS 7 with systemd. I made some assumptions that I'm happy to discuss/change. I figured you'd want it to run, so I included a sample configuration that's largely the example one except that it writes the database to /var/lib/dex/. It creates and runs as the dex user. I realize that all the cool kids are building RPMs with Bazel :-), but I didn't want to introduce that into your project at this time. I kept it simple and build with tito. If you're not running on a RHEL/CentOS system with tito installed, you can test the build by running docker build -f rpm/Dockerfile .

### DIFF
--- a/.tito/packages/.readme
+++ b/.tito/packages/.readme
@@ -1,0 +1,3 @@
+the .tito/packages directory contains metadata files
+named after their packages. Each file has the latest tagged
+version and the project's relative directory.
--- a/.tito/tito.props
+++ b/.tito/tito.props
@@ -1,0 +1,5 @@
+[buildconfig]
+builder = tito.builder.Builder
+tagger = tito.tagger.VersionTagger
+changelog_do_not_remove_cherrypick = 0
+changelog_format = %s (%ae)
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ check-go-version:
 clean:
 	@rm -rf bin/
 
-rpm:
+rpm: FORCE
 	tito build --rpm --test
 
 testall: testrace vet fmt lint
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ group=$(shell id -g -n)
 
 export GOBIN=$(PWD)/bin
 
-LD_FLAGS="-w -X $(REPO_PATH)/version.Version=$(VERSION)"
+LD_FLAGS="$(LDFLAGS) -w -X $(REPO_PATH)/version.Version=$(VERSION)"
 
 build: bin/dex bin/example-app bin/grpc-client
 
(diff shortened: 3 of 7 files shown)


srenatus:
Interesting. I'm afraid I don't have a good understanding of the big picture yet -- this PR will include the scaffolding to build an RPM, but would that be enough for users to be a useful enhancement?

I do think it's a great contribution -- all the more so as I have no idea about building (or using) RPMs. We're consuming Dex through [a habitat package](https://bldr.habitat.sh/#/pkgs/core/dex/latest) built from [a public `plan.sh` file](https://github.com/habitat-sh/core-plans/blob/e302a788bc7ca41b163b4eb4f332523d46fe8959/dex/plan.sh). Including a `Dockerfile` is a great idea for that. 👍 

The only thing I wonder is how to go forward with this:

- If every consumer is supposed to build their own RPMs, should this maybe go into a `contrib/` folder (or to-be-created repo)?
- Is there a Quay-like build service we could leverage? (I think I'd prefer not to have a maintainer build and upload an RPM for a release -- since this could be a slippery slope into `deb`, `apkg`, etc. It wouldn't be so bad if some service took care of this.)

amdonov:
Unfortunately, I had to fork dex for my use case, so I'd fall into the use case of building my own RPM. It could helpful to others in a similar situation.

I've never published to it, but it looks like https://packagecloud.io might be a place to publish RPMs. It supports Travis CI and seems to have a free option for open source projects.

srenatus:
I think, looking at how other open source projects do it, there could be a viable compromise in **adding the building and publishing of a linux binary (to github releases page) to the release process**. Then, we could have another repo for the RPM scaffolding? That way, there's a clean separation between where "the main project" ends (building a binary for linux) and distribution-related contrib starts (building an RPM from it).

What do you think, @ericchiang @rithujohn191 ?

ericchiang:
> Is there a Quay-like build service we could leverage? (I think I'd prefer not to have a maintainer build and upload an RPM for a release -- since this could be a slippery slope into deb, apkg, etc. It wouldn't be so bad if some service took care of this.)

We could probably just build an RPM and deb via docker in our travis build scripts.

I'm not terribly opposed to having more code in contrib, but it MUST have automated tests. Maybe a multi-stage docker file where the first stage builds the RPM and the second installs it.

ericchiang:
We're also going to have to decide what a the RPM should default install. I don't think it's a good thing to have the RPM stand up an insecure dex.

amdonov:
@ericchiang What changes would you like to see made before a merge? I'd like to finish this and move on to SAML. 

I agree on the Docker suggestion. I'll make the change to install the built RPM in a second stage.

What changes would you like to see made the default configuration file?

nabokihms:
I will close this one since providing package builds for different Linux distributions has yet to be planned. Feel free to open a new issue or PR if you still need one.

PR #1297, opened 2018-09-11
Use GitHub Team slug instead of name

tburko:
GitHub Team slug is arguably more script-friendly, and less surprising field to use for team names.

Compare `acme:Full-time Technical Staff` using name, and `acme:full-time-technical-staff` using slug.

The team slug is already being used by Terraform data module https://www.terraform.io/docs/providers/github/d/team.html for team lookups. Also https://github.com/google/go-github/ refers to teams by slug in a few places.

Our own tooling uses slug across internal systems. Not a strong argument in itself, just suspect that many other organizations are doing the same.

I have considered making the field choice configurable. Could easily add a connector Config property say `teamNameField` to accept `name|slug|both` and default to `name` to keep backwards compatibility. On the other hand, it feels like unnecessary code and flexibility. Hopefully project owners can help to make a call.


tburko:
Implemented config property

connector/github/github.go
```diff
@@ -586,7 +594,7 @@ func (c *githubConnector) teamsForOrg(ctx context.Context, client *http.Client,
 
 		for _, team := range teams {
 			if team.Org.Login == orgName {
-				groups = append(groups, team.Name)
+				groups = append(groups, c.teamNamesExtractor(team)...)
```
ericchiang: "teamNamesExtractor" is overkill. Just use a switch statement :) It's waaay more readable.

Documentation/connectors/github.md
```diff
@@ -51,6 +52,15 @@ connectors:
...
       - red-team
       - blue-team
+
+    # Optional choice between 'name' (default), 'slug' or 'both'.
+    #
+    # As an example, group claims for member of 'Site Reliability Engineers' in
+    # Acme organization would yield:
+    #   - ['acme:Site Reliability Engineers'] for 'name'
+    #   - ['acme:site-reliability-engineers'] for 'slug'
+    #   - ['acme:Site Reliability Engineers', 'acme:site-reliability-engineers'] for 'both'
```
ericchiang: Do you have a use case for 'both'? I'd rather we drop it if not. If we ever add another field (maybe ID?) 'both' becomes a weird setting.

Issue #66, opened 2015-08-19
cmd/dex: user filters


mjibson:
We'd like to be able to restrict login to a service to any valid google account, for example `@coreos.com`.

philips:
Has anyone thought about how this would be implemented? This is something I feel is necessary before we run a public dex.coreos.systems that is used to authenticate our users into our k8s clusters.

bobbyrullo:
I think it could be done relatively easily with a flag + some logic in the registration code.

bcwaldon:
Working on this.

bcwaldon:
My first attempt was to add a check to the OIDC connector's callback handler, which effectively prevents any user from logging in to a remote OIDC Identity Provider. I left the code for this here: https://github.com/coreos/dex/tree/oidc-domain-filter. The UX provided by this solution isn't quite right, so I'm holding off on this until we figure out a better solution. The specific issue I ran into looks like this: you have two connectors, local and oidc, and registration is enabled (globally). My first approach would prevent users from registering via oidc if they weren't in the correct domain, but they could still create a local identity. 

It feels like we need to be able to disable registration globally, and per-connector. The specific use-case we are trying to solve for here requires that existing users can still log in regardless of whatever domain filtering is set up on the registration flow.

ericchiang:
To expand on this, I think we'd like some form of user filters. Either globally to dex, per client, or per connector.

```
filter:
    whitelist:
        groups: ["foo"]
        emails: ["bar@coreos.com"]
```

Maybe expand into regexps or email domain.

ericchiang:
FYI, after a lot of sitting on this I think this should probably be done by client apps instead of dex.

ericchiang:
> FYI, after a lot of sitting on this I think this should probably be done by client apps instead of dex.

This approach actually doesn't work for users who want to use the public client or device flows (if we ever implement the latter). So there's still value in this proposal.

srenatus:
#1371 seems to be a front page issue right now (😉) with a rather big overlap. 🧹 GCing this one, please bear with me.

PR #1298, opened 2018-09-11
[WIP] PreferredEmailDomain GitHub connector Config option

tburko:
Considering implementing this. Will help identify users by verified company email without asking them to mark it as primary.

Sample config:

```
-connectors:
  - type: github
    id: github
    name: GitHub
    config:
      clientID: 'REDACTED'
      clientSecret: 'REDACTED'
      redirectURI: 'https://dex.acme.org/dex/callback'
      orgs:
        - name: acme
      preferredEmailDomain: 'acme.org'
```

### DIFF
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -40,13 +40,14 @@ var reLast = regexp.MustCompile("<([^>]+)>; rel=\"last\"")
 
 // Config holds configuration options for github logins.
 type Config struct {
-	ClientID     string `json:"clientID"`
-	ClientSecret string `json:"clientSecret"`
-	RedirectURI  string `json:"redirectURI"`
-	Org          string `json:"org"`
-	Orgs         []Org  `json:"orgs"`
-	HostName     string `json:"hostName"`
-	RootCA       string `json:"rootCA"`
+	ClientID             string `json:"clientID"`
+	ClientSecret         string `json:"clientSecret"`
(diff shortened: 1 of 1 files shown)


ericchiang:
Actually this should probably just be a global thing. https://github.com/dexidp/dex/issues/66

So the config would have something like:

```
userFilters:
  requiredEmailDomains:
  - "@foo.com"
  - "@example.com"
```

tburko:
> Actually this should probably just be a global thing. #66
> 
> So the config would have something like:
> 
> ```
> userFilters:
>   requiredEmailDomains:
>   - "@foo.com"
>   - "@example.com"
> ```

Hmm but that steps into authorization domain, I thought Dex was supposed to deal with identity. And a bigger change, if you want my help with this I would prefer to hear your thoughts on implementation first, we seem to have different approaches to structuring code :).

ericchiang:
> Hmm but that steps into authorization domain, I thought Dex was supposed to deal with identity. And a bigger change, if you want my help with this I would prefer to hear your thoughts on implementation first, we seem to have different approaches to structuring code :).

Sorry, I realized I'm getting these mixing up concepts. Yes, preferred email sounds good without my last suggestion :)

povils:
Any updates?

nabokihms:
Closed by https://github.com/dexidp/dex/pull/2740

PR #1305, opened 2018-09-26
connector/saml: make unparsable (trailing) data an error

### LINKED ISSUE #1304: SAML connector: if second cert is not formatted correctly, no error surfaces
[When parsing the PEM-formatted CA data](https://github.com/dexidp/dex/blob/ff70c0453f28c24eed959e6349b5dda79223b9d2/connector/saml/saml.go#L201-L204), we _stop reading in certs_ when `block == nil`. That could very well be the case when the configured certs are formatted badly, like
```
-----BEGIN CERTIFICATE-----
[...]
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
[...]
-----END CERTIFICATE----
```
(note that there's one `-` too little in the last `END CERTIFICATE` line). In that case, we'll read in one cert (the first one), and silently ignore the rest.

A low-hanging fruit fix here could be to **log the trailing bytes**, so that it at least becomes apparent that the second piece couldn't be made sense of.

srenatus:
Fixes #1304, if we want to be harsh.

However, I think if it was the user's intention to pass two certs, and
the second one couldn't be read, that shouldn't just disappear. After
all, when attempting to login later, that might fail because the
expected IdP cert data isn't there.


srenatus:
I've wanted to get a feel for the direction of the change first -- I'll happily add tests. Will do tomorrow.

srenatus:
Didn't get to it today. :/ 🤞 tomorrow.

srenatus:
✔️Added tests, updated code to cope with trailing spaces/newlines.

ericchiang:
thanks! just a couple nits

srenatus:
✔️ updated

PR #1320, opened 2018-10-18
Made work with Cisco Ping Federate Public OIDC Implementation

dcwangmit01:
* Smoothed over the quirks of Cisco Ping Federate
* The Cisco OIDC provider does not return email in claims.
  * Get email from userInfo if it doesn't exist.
* Get groups from Ping Federate userinfo "memberof" key.
  * Parse/filter groups to only match 'CN=(kube-.*?)'
    groupnames such as 'kube-admin'
* Implemented the passthrough of Refresh token

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 bin
 dist
 _output
+vendor
--- a/Dockerfile.localbuild
+++ b/Dockerfile.localbuild
@@ -1,0 +1,14 @@
+FROM debian:latest
+
+RUN echo "Install OS packages" && \
+    apt-get -y update && apt-get --no-install-recommends -y install \
+        ca-certificates && \
+    apt-get -y clean && apt-get -y autoremove && rm -rf /var/lib/apt/lists/*
+
+# Import frontend assets and set the correct CWD directory so the assets
+# are in the default path.
+COPY web /web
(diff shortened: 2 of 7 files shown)


dcwangmit01:
Oops.  Hit the wrong button.  Closing.

dcwangmit01:
Actually @ericchiang , if any parts of this are upstream-able, let me know.  There are two pieces that I think might qualify.

1) First, I've modified dex to pull the email claim from the userinfo endpoint if the upstream id_token doesn't have it.  Many downstream Dex clients such as gangway and oauth2_proxy expect "email" to be returned as a claim in the id_token.  Most identity providers do this automatically as a convenience, but some IDPs such will not return the email in the id_token, and thus expect the client to query the userinfo endpoint.  The latter behavior is still in-line with the OIDC spec.  See this [stackoverlow](https://stackoverflow.com/a/50928319) for explanation.

2) Second, I've cut and pasted code from the microsoft connector to allow the passthrough of the refresh token when downstream clients request scope "offline_access".

The rest of this accidental PR is specific to copying the group userinfo claim from the upstream provider "memberof" key to the downstream clients as "group".  It's pretty specific to this IDP provider, and if this functionality sounds useful for others I can try to generalize it.

Thanks!
-dave

PR #1324, opened 2018-10-22
Add meta content for google to not try to translate the page

NickLavrov:
This stops the prompt from Chrome to translate the page. To test:

Create index.html and copy-paste the html template that was changed. Comment out the notranslate tag. Make sure you put in a k8s token for {{ .IDToken }} (replace some characters with x's) to trigger the prompt.

Start a server to visit index.html (python -m SimpleHTTPServer is fast) in Google Chrome. Observe translation prompt.

Uncomment the notranslate tag and refresh. No more prompt.

### DIFF
--- a/cmd/example-app/templates.go
+++ b/cmd/example-app/templates.go
@@ -36,6 +36,7 @@ type tokenTmplData struct {
 
 var tokenTmpl = template.Must(template.New("token.html").Parse(`<html>
   <head>
+    <meta name="google" content="notranslate">
     <style>
 /* make pre wrap */
 pre {


ericchiang:
lgtm, but any motivation here? If you dislike this feature shouldn't you disable it in chrome?

NickLavrov:
It's a very minor change, I know :)

Chrome kept telling me the page was in Polish, probably due to the IDToken. Nothing in the example app looks like it would ever need to be translated, so this would just turn off any translation suggestions. That way, no matter what Chrome preferences the user has, it will not prompt translation.

ericchiang:
> Nothing in the example app looks like it would ever need to be translated, so this would just turn off any translation suggestions.

Well... if you don't speak English then it might need to be :)

Maybe adding "notranslate" to the JWT would be better? https://cloud.google.com/translate/faq#technical_questions

NickLavrov:
That `span` tag solution unfortunately seems to only apply to when you pass text into the translate API. Putting those tags around the tokens on the HTML still causes the translate prompt.

I was thinking that the only words on the page are "token," "claims," "refresh token," and "Redeem refresh token." If those words are indeed worth translating, then I suppose this PR can be closed.

ericchiang:
Was curious but this is fine as is. I imagine the travis builder is being impacted by the github outage https://status.github.com/messages so we may have to wait to merge.

NickLavrov:
Closing this PR since it is outdated.

PR #1325, opened 2018-10-23
*: update go version

ericchiang:
Using an old version of Go was breaking the build. Update travis and the Docker image while we're at it.

https://travis-ci.org/dexidp/dex/builds/444775302


ericchiang:
Using admin to fix the build.

PR #1328, opened 2018-10-23
Bump Go versions

HaraldNordgren:


srenatus:
Thank you for looking into this! Between this one and #1325, there's still some issues to resolve, it seems. If you'd like to push this further, let's drop 1.8 and 1.9, and see what happens next 😉

Issue #1332, opened 2018-11-01
have a namespace scoped role for dex, instead of cluster-wide


SongGithub:
our current situation is that we have already got a DEX 2.10.0 deployed in kube-system namespace which is using `cluster-role`. This DEX has created necessary CRDs for all namespaces. 

For some reason, one of our teams want to deploy a separate DEX in their own namespace. And for security reasons, we only defined a `Role` that is namespace scoped. This Role leads to a problem that DEX can't create CRDs at cluster level and giving massive amount of RBAC-DENY alerts that triggering off our security alerts. 

The config for storage is kubernetes-in cluster 

I am wondering, can DEX not trying to create CRDs especially at cluster scope, if the CRDs are already existing in the namespace?

SongGithub:
it appears to relate to this issue https://github.com/dexidp/dex/issues/1308 #1308

SongGithub:
In particular, the code that leads to the problem is here
https://github.com/dexidp/dex/blob/218d671a96865df2a4cf7f310efb99b8bfc5a5e2/storage/kubernetes/storage.go#L149

could it check if CRDs already exist in the current namespace before going out to create CRDs?

ericchiang:
I think this is a dup of #1308. We should probably just fix that.

SongGithub:
Can I submit a PR for the fix please?

SongGithub:
very happy to see the issue addressed by the above PR #1333, so I would like to close this issue.

Issue #1308, opened 2018-10-01
Disable CRD management


ernoaapa:
Hi,
we want to manage the CRDs by our self. What do you think of having `--disable-crd-creation` flag? Now if I create the CRDs and don't give RBAC to create customresourcedefinitions, I get a lot of spam to the logs. I can provide PR for this.

srenatus:
I have to admit that I know little about this. Do other CRD-using services provide similar knobs?

ericchiang:
I think others do. This should actually be a easy fix. Add an option here: https://github.com/dexidp/dex/blob/a1b6ba9bbad8bafbd680931a2bd5b6176df27dbb/storage/kubernetes/storage.go#L39

Then skip the create call later on.

SongGithub:
hi @ernoaapa, this issue was addressed in PR #1333 by me, would you like to close this one?

I happened to have similar issue to yours that our team doesn't like to give DEX a clusterRole to create CRDs, providing this DEX instance is running from a client team namespace, and we have a DEX setup already in kube-system namespace with ClusterRole who has created required CRD types in all namespaces.

ernoaapa:
Yeah, this can be closed. The #1333 solves my issue.

PR #1333, opened 2018-11-02
allow it to disable CRD creation

SongGithub:
responding to this issue #1332 #1308 , this PR will provide a simple fix to disable CRD creation step. 

In our case, this feature will greatly reduce the need for cluster scope permission to create CRDs, providing that we already have got a DEX running at cluster level which means all necessary CRDs have been created already, and we want to deploy a separate DEX in a team's namespace with minimal permissions(kubernetes Role only, no clusterRole allowed.) in order to de-couple this DEX with the existing cluster-wide DEX.

Thanks!

### DIFF
--- a/storage/kubernetes/storage.go
+++ b/storage/kubernetes/storage.go
@@ -40,7 +40,6 @@ type Config struct {
 	InCluster      bool   `json:"inCluster"`
 	KubeConfigFile string `json:"kubeConfigFile"`
 	UseTPR         bool   `json:"useTPR"` // Flag option to use TPRs instead of CRDs
-	NoCrdCreation  bool   `json:"noCrdCreation"` // Flag to disable creation of CRDs at cluster level
 }
 
 // Open returns a storage using Kubernetes third party resource.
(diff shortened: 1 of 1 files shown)


SongGithub:
on Ci build: there is a strange error that I can reproduce locally. When I ran `make verify-proto`, it passed.

ericchiang:
I think you need to run go fmt

ericchiang:
That avoids a flag and accomplishes the same thing. If the resource already exists dex just continues on its way.

SongGithub:
I think a perfect but not so simple solution would be: 

- if flag/namespaced == true: {
- check if CRDs in current namespace (LIST at namespace scope)
- if yes, continue
- if not create those CRD in current namespace. (CREATE at namespace scope)
}

that require a new CLI flag, then both a LIST method and a POST method could operate at namespace scope

SongGithub:
which path would you prefer to go? 

- stick with current one which is originated from your idea at #1308 for now
- have a namespace-scoped "perfect" solution

ericchiang:
> cli.list seems to need cluster scope LIST access by default.

List is namespaced

SongGithub:
Thanks for you hint, I have updated the code according to it. It does reduced need for a new flag :)

storage/kubernetes/storage.go
```diff
@@ -148,6 +148,12 @@ func (cli *client) registerCustomResources(useTPR bool) (ok bool) {
 			resourceName = r.ObjectMeta.Name
 		} else {
 			r := customResourceDefinitions[i]
+			var i interface{}
+			cli.logger.Infof("checking if resource %s has been created already...", r.ObjectMeta.Name)
+			if err := cli.list(r.Spec.Names.Plural, &i); err == nil {
+				cli.logger.Infof("The custom resource already created %s, hence continue", r.ObjectMeta.Name)
```
ericchiang: nit: "custom resource already available %s, skipping create"

storage/kubernetes/storage.go
```diff
@@ -148,6 +148,12 @@ func (cli *client) registerCustomResources(useTPR bool) (ok bool) {
 			resourceName = r.ObjectMeta.Name
 		} else {
 			r := customResourceDefinitions[i]
+			var i interface{}
+			cli.logger.Infof("checking if resource %s has been created already...", r.ObjectMeta.Name)
```
ericchiang: nit: if you're going to log this, I think it'd be good to log the error as well.

Right now these logs look like:

```
checking if resource %s has been created already...
create custom resource %s
```

Can you catch the error from the list and print that too?

```go
if err := cli.list(r.Spec.Names.Plural, &i); err == nil {
    // ...
} else {
    cli.logger.Infof("failed to list custom resource %s, attempting to create: %v", r.ObjectMeta,Name, err)
}
```

SongGithub:
It would be great if it could be released soon as patch version v2.12.1, as one of our projects would rather wait for an official release than a release of our own fork that makes solution looking hacky.

ericchiang:
We can just cut v2.13.x

SongGithub:
Thanks for merging @ericchiang! do you know when is it due to release v2.13.x?

Issue #1341, opened 2018-11-15
Postgres: forcing transactions to be serializable breaks concurrent login


vito:
Hi! Really glad to see Dex coming back to life lately - props to those who have volunteered!

We've got a long-lived fork with a bunch of changes we'd like to PR back: https://github.com/concourse/dex

One change in our fork was to remove the Postgres serialization level, because it was causing failures in our integration tests.

Prior context: https://github.com/dexidp/dex/issues/628, https://github.com/dexidp/dex/issues/600, https://github.com/concourse/concourse/issues/2531

I've noticed that when running our integration suites in parallel, if they all try to log in at the same time, a bunch of them fail with the following message:

```
Failed to set connector ID on auth request: update auth request: pq: could not serialize access due to read/write dependencies among transactions
```

Here are the logs from Postgres:

```
2018-11-15 16:55:47.558 UTC [797] ERROR:  could not serialize access due to read/write dependencies among transactions
2018-11-15 16:55:47.558 UTC [797] DETAIL:  Reason code: Canceled on identification as a pivot, during write.
2018-11-15 16:55:47.558 UTC [797] HINT:  The transaction might succeed if retried.
2018-11-15 16:55:47.558 UTC [797] STATEMENT:  
				update auth_request
				set
					client_id = $1, response_types = $2, scopes = $3, redirect_uri = $4,
					nonce = $5, state = $6, force_approval_prompt = $7, logged_in = $8,
					claims_user_id = $9, claims_username = $10, claims_email = $11,
					claims_email_verified = $12,
					claims_groups = $13,
					connector_id = $14, connector_data = $15,
					expiry = $16
				where id = $17;
```

I've confirmed that each login request has its own distinct request ID (the `where id = $17`). And these tests aren't sharing any cookies or state, so they shouldn't be clobbering each others' request IDs. So I don't really see how this would legitimately cause any problems without the strict isolation level. I'm not really sure what's causing this to conflict, either, since they should be updating distinct rows.

The Postgres docs say "applications using this level must be prepared to retry transactions due to serialization failures", which Dex doesn't seem to be doing, and it doesn't feel right to force the client to retry API/login requests due to this, and I don't think they'd even have enough information to recognize that it's a serialization failure to retry in the first place.

One thing we tried was simply ripping out the isolation level config: https://github.com/concourse/dex/commit/9b88d6750b32c3fb79c02961e13011d9f1286170 - however there are explicit tests checking for this exact sort of failure.

[...]

PR #1342, opened 2018-11-15
retry on serialization errors

vito:
Fixes https://github.com/dexidp/dex/issues/1341

Turns out it's easy enough to just capture the specific serialization error code and retry the transaction.

Tested this locally - the storage conformance tests pass (they failed when removing serialization altogether), and our integration tests no longer fail on concurrent logins.


storage/sql/sql.go
```diff
@@ -59,19 +59,34 @@ var (
 		// NOTE(ericchiang): For some reason using `SET SESSION CHARACTERISTICS AS TRANSACTION` at a
 		// session level didn't work for some edge cases. Might be something worth exploring.
 		executeTx: func(db *sql.DB, fn func(sqlTx *sql.Tx) error) error {
-			tx, err := db.Begin()
-			if err != nil {
-				return err
-			}
-			defer tx.Rollback()
+			for {
```
srenatus: ❓ This is addressing an edge case, but can we limit the number of retries? Also, do we care that it's hammering or does this need some wait times?

storage/sql/sql.go
```diff
@@ -59,19 +59,34 @@ var (
...
+				tx, err := db.Begin()
```
srenatus: I wonder if it would be worth switching to `BeginTx(context.TODO(), sql.LevelSerializable)`, which would, I think, mean we can ditch the explicit `SET TRANSACTION ISOLATION LEVEL SERIALIZABLE`.

srenatus: I suppose if we did that we could also go all-in on contexts and do something like
```go
ctx, cancel := context.WithCancel(context.TODO())
defer cancel() // will rollback transaction
tx, err := db.BeginTx(ctx, nil /* or some proper level? */)
if err != nil {
  return err
}
if err := fn(tx); err != nil {
  return err
}
err = tx.Commit()
// etc
```

storage/sql/sql.go
```diff
@@ -59,19 +59,34 @@ var (
...
+					return err
+				}
+
+				if err := fn(tx); err != nil {
+					return err
+				}
+
+				err = tx.Commit()
+				if err != nil {
+					if pqErr, ok := err.(*pq.Error); ok && pqErr.Code == "40001" {
```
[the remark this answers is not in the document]
vito: Ah yeah that's much better.

storage/sql/sql.go
vito: Hmmm, to be honest I'm not sure. The Postgres docs just say "retry" but don't say anything about backoff and attempts.

I'm a little nervous about limiting it as it could potentially mean bubbling up the error, given enough concurrency. I agree that it's scary to have it unbounded, but so long as we're getting serialization failures, I'm not sure there's any resolution other than to retry until it works (which should eventually happen as requests calm down). It helps that we're at least catching this very specific error code and only retrying on that one.

Not sure about the hammering/wait times. In my experience it's sufficient to just try again immediately to resolve concurrency "disputes", since they tend to resolve fairly quickly, and the net effect is usually just a few more queries (which the database should be able to handle). If it were a failure that could come from load or network connectivity, I'd totally agree, but this almost feels more like regular handling of database behavior. 🤷‍♂️

storage/sql/sql.go
vito: Neat, didn't know that was a thing! I'll try it out.

vito:
@srenatus Yeah, I'm a bit confused by it too. Maybe the failures have something to do with concurrent reads, not just writes? For example if there's any other query that runs against that table at the same time someone may be logging in - maybe auth request garbage-collection?

storage/sql/sql.go
```diff
@@ -59,19 +60,35 @@ var (
...
-				return err
+
+			for {
+				tx, err := db.BeginTx(ctx, opts)
+				if err != nil {
+					return err
+				}
+
+				if err := fn(tx); err != nil {
+					return err
```
vito: Ack, I think this error also needs to be checked. Our integration tests just failed, so maybe it was luck that it passed for me. I think the `fn` here returns an error because whatever query it's running sees the serialization failure. Unfortunately the `fn` also sometimes wraps the error, making it impossible to check the error code.

vito:
@srenatus Alright, I've pushed another commit that refactors all the error wrapping out of the `ExecTx` fn, and I left a comment warning future developers to not wrap errors in there. Without this last commit, my initial fix only applied some of the time, since the failure could either happen on the individual queries in `fn` or the `Commit` at the end, which was the only thing I added the check for. =

Along the way I tried to keep the error wrapping/handling consistent across all the `conn` methods. This change ended up being kinda big, but if it makes you feel any better there *were* tests that failed along the way, so the test coverage seems decent. I didn't end up having to change any tests, since no user-facing behavior changed once I fixed the things the tests told me to.

With this all cleaned up, I've got our integration tests successfully running in a loop with ~23 concurrent logins.

storage/sql/sql.go
```diff
@@ -51,27 +52,49 @@ var (
...
+
+			for {
+				tx, err := db.BeginTx(ctx, opts)
+				if err != nil {
+					return err
+				}
+
+				if err := fn(tx); err != nil {
+					if pqErr, ok := err.(*pq.Error); ok && pqErr.Code.Name() == "serialization_failure" {
+						// serialization error; retry
```
ericchiang: In the initial bug report you mentioned

```
The Postgres docs say "applications using this level must be prepared to retry transactions due to serialization failures", which Dex doesn't seem to be doing, and it doesn't feel right to force the client to retry API/login requests due to this, and I don't think they'd even have enough information to recognize that it's a serialization failure to retry in the first place.
```

Can you add a link to those docs here?

glide.yaml
```diff
@@ -114,7 +114,7 @@ import:
 - package: github.com/mattn/go-sqlite3
   version: 3fb7a0e792edd47bf0cf1e919dfc14e2be412e15
 - package: github.com/lib/pq
-  version: 50761b0867bd1d9d069276790bcd4a3bccf2324a
+  version: 9eb73efc1fcc404148b56765b0d3f61d9a5ef8ee
```
srenatus: ❓ There's a lot going on between [these two commits](https://github.com/lib/pq/compare/50761b0867bd1d9d069276790bcd4a3bccf2324a...9eb73efc1fcc404148b56765b0d3f61d9a5ef8ee) -- is there anything in particular that you're after, or is this a chore-type update?

vito: This was necessary for `BeginTx` with transaction level support: https://github.com/lib/pq/commit/ed1a26d9e1fd4eb1cc491e6d688467f4e81a413c

storage/sql/sql.go
vito: Sure!

storage/sql/sql.go
```diff
@@ -74,6 +75,11 @@ var (
 				}
 
 				if err := fn(tx); err != nil {
+					if pqErr, ok := err.(*pq.Error); ok && pqErr.Code.Name() == "serialization_failure" {
```
srenatus: 💭 So, the `serialization_failure` error _potentially_ comes back from the `fn(tx)`, and not only from the `Commit` below, as previously thought? Also, do we know if it _ever_ comes back from the `tx.Commit()`?

vito: Yes, I saw it come from both. I believe it just comes down to timing.

srenatus:
@vito thanks for hanging on to this 👍 looks like in the meantime, some other PRs have gotten into this PR's way (conflicts).

vito:
@srenatus Yep, just working on fixing the dependencies up now and adding a comment per @ericchiang's suggestion. Will have it ready soon.

vito:
OK, all rebased and good to go!

srenatus:
ℹ️ This is probably good to go, but I'd like to take a dive into PG doc land to wrap my head around this some more. However, I'll do that tomorrow morning, probably. In the mean time, there's not going to be a release out of thin air, so I'm fine either way -- shall we merge this now or later?

vito:
@srenatus Doesn't matter to me, feel free to take your time. We've still got a few things on our fork so we won't be blocked on a merge either way.

PR #1343, opened 2018-11-15
remove incomplete mysql and cockroachdb support

vito:
These storage drivers were never completed and can't be configured. The CockroachDB dependency is a bit out of date, so I figured it'd be easier to just remove these until someone has the bandwidth to implement them.


storage/sql/standup.sh
```diff
@@ -36,25 +36,21 @@ function main {
      case "$2" in
      "postgres")
         create_postgres;;
-     "mysql")
-        create_mysql;;
```
srenatus: ℹ️ Looks like `create_mysql` and `destroy_mysql`  have never been defined.

PR #1345, opened 2018-11-16
Add 'both' option to use team name AND slug in TeamNameField

vito:
This allows the connector to be configured with `both` so that both team names and slugs can be returned in the groups. This makes configuring teams a bit more foolproof in Concourse; we would often have confusion over whether the team name or slug should be given by the user, so it's easier to just allow both, since collisions shouldn't be possible anyway.

### DIFF
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -16,10 +16,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/sirupsen/logrus"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/github"
-
-	"github.com/sirupsen/logrus"
 
 	"github.com/dexidp/dex/connector"
 )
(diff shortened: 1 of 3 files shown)


srenatus:
@vito forgot to mention -- thank you for your contribution. 😃 👍

vito:
Updated! And no problem. There should be more to come over the coming weeks as we try and slim down our fork's deltas.

Thanks for the tip re: the header length, I'll keep an eye out for that. The reason we want both is that Concourse's dex integration allows whitelisting GitHub teams for Concourse teams, something like:

```
fly set-team -n team-name --github-team concourse:Pivotal
```

Initially we just used the team name, but users kept trying to specify the slug when configuring teams and getting confused as to why auth wouldn't work (just because of a lowercase `p`). So we figured it's easiest to just allow both.

connector/github/github.go
```diff
@@ -16,12 +16,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/dexidp/dex/connector"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/github"
-
```
srenatus: ❓ Would you mind much making this _three blocks_? Stdlib, third party, and anything starting with `github.com/dexidp/dex`? `goimports` has a setting for automatically doing this.

vito: sure!

Issue #1347, opened 2018-11-19
Gitlab connector should not require the api scope.


gypsydiver:
As of now, the Gitlab connector requires users to grant the `api` scope to configured applications in order to retrieve information that could be just as easily pulled using read only scopes (`read_user`, `openid`).

The `/oauth/userinfo` endpoint provides all group information needed.

Can provide PR for this.

Related: #1222 
Solves: #1246

srenatus:
Happy to take a PR there. ❤️  Could you please include some pointer to some current API docs?

PR #1351, opened 2018-11-20
Issue 1347 - Gitlab connector should not require the api scope.

gypsydiver:
#1347


srenatus:
Update -- I've read #1347 again. So, this is meant to use the simpler way, with a less permissive scope? If so, 👍

Still interested in the user-facing changes, if there are any.

gypsydiver:
Changes here would break applications that don't have the `openid` scope enabled and require to pull groups from Gitlab. Admins need to at least enable it in the Gitlab Application settings for groups to work; and ideally disable the `api` (though not required).

srenatus:
👍 sounds great. Do you happen to know when that feature as introduced into Gitlab?

gypsydiver:
[Gitlab CE](https://github.com/gitlabhq/gitlabhq/blob/master/CHANGELOG.md#900-2017-03-22) Version 9.0.0, 2017-03.

srenatus:
Thanks for looking that up. Judging from what, it's been out for more than 18 months, *and* reducing the application scope there is a bit of a significant security improvement, I suppose. I'll merge this.

⏩ Let's make sure to include this information in the next release notes, so nobody is in for a bad surprise.

Issue #1355, opened 2018-11-21
storage/sql/conformance: concurrent access tests fail for the wrong reason


srenatus:
Seems like a bug introduced in #1342 -- **not all serialization failures are expected to go away on retry**.

Running postgres in a docker container, and our conformance tests in a different terminal,
```
$ env DEX_POSTGRES_DATABASE=dex DEX_POSTGRES_USER=postgres DEX_POSTGRES_PASSWORD=postgres DEX
_POSTGRES_HOST=127.0.0.1:5432 go test -count=1 -p=1 -v github.com/dexidp/dex/storage/sql -run TestPostgres
=== RUN   TestPostgres
=== RUN   TestPostgres/AuthCodeCRUD
=== RUN   TestPostgres/AuthRequestCRUD
=== RUN   TestPostgres/ClientCRUD
=== RUN   TestPostgres/RefreshTokenCRUD
=== RUN   TestPostgres/PasswordCRUD
=== RUN   TestPostgres/KeysCRUD
=== RUN   TestPostgres/OfflineSessionCRUD
=== RUN   TestPostgres/ConnectorCRUD
=== RUN   TestPostgres/GarbageCollection
=== RUN   TestPostgres/TimezoneSupport
=== RUN   TestPostgres/AuthRequestConcurrentUpdate
=== RUN   TestPostgres/ClientConcurrentUpdate
=== RUN   TestPostgres/PasswordConcurrentUpdate
=== RUN   TestPostgres/KeysConcurrentUpdate
--- PASS: TestPostgres (6.57s)
    --- PASS: TestPostgres/AuthCodeCRUD (0.03s)
    --- PASS: TestPostgres/AuthRequestCRUD (0.02s)
    --- PASS: TestPostgres/ClientCRUD (0.03s)
    --- PASS: TestPostgres/RefreshTokenCRUD (0.03s)
    --- PASS: TestPostgres/PasswordCRUD (0.03s)
    --- PASS: TestPostgres/KeysCRUD (0.02s)
    --- PASS: TestPostgres/OfflineSessionCRUD (0.02s)
    --- PASS: TestPostgres/ConnectorCRUD (0.03s)
    --- PASS: TestPostgres/GarbageCollection (0.04s)
    --- PASS: TestPostgres/TimezoneSupport (0.01s)
    --- PASS: TestPostgres/AuthRequestConcurrentUpdate (1.17s)
    --- PASS: TestPostgres/ClientConcurrentUpdate (1.27s)
[...]

Issue #1292, opened 2018-09-07
Accumulating new AuthRequest objects can cause a denial of service on storage.


pjmichael:
We are running dex in a kubernetes cluster using kubernetes CRD's as our storage implementation. We recently had an incident in which a misconfiguration resulted in the storage of a large number of authorization requests. 

The implications of this scenario with our implementation was that:
- Each authorization request would result in the creation of a new AuthRequest object, hard coded with an expiry of 24 hours (see: https://github.com/dexidp/dex/blob/master/server/handlers.go#L163)
- Even though garbage collection would run every 5 minutes, AuthRequest objects continued to accumulate for 24 hours (ie. as they have not yet expired)
- Garbage collection lists all AuthRequest objects without pagination or filtering at the start of each execution
- Eventually, the list AuthRequests call began to timeout, causing Garbage collection to fail
- By the time AuthRequest objects began to expire they we not being cleaned up as Garbage collection could no longer list AuthRequests successfully
- The flow on effect was to cause a degradation of the kubernetes apiserver (ie. the storage implementation)

Do you have any thoughts as to how this issue could be avoided?

Some possible ideas I was thinking which may help were:
- making AuthRequest expiry configurable to reduce period in which they could accumulate prior to them being garbage collected
- perhaps adding rate limiting on the creation of AuthRequest objects

ericchiang:
I think the crux of the issue is "prevent unauthenticated users from creating too much backend state."

Rate limiting would be good, but overall we probably want to tie the auth request to an IP or something. The proposal in #646 is one way, but doesn't stop a motivated attacker from messing with your cluster.

Removing "help wanted" for now. I don't think there's a clear enough solution to warrant the label.

srenatus:
@mxey `help wanted` or not, exposing that as a configurable seems like a welcome PR to me. 😃

ericchiang:
I bet there's a way we can get away from storing state until we know a user's authenticated by using a cookie. The naive way would be to serialize the AuthRequest, sign it using dex's signing keys, and shoving it in the user's session:

https://godoc.org/github.com/coreos/dex/storage#AuthRequest

Then only persist the AuthRequest once we get a valid response from the backend (e.g. the LDAP server), once we'd fill in the Claims and ConnectorData.

[...]

Issue #646, opened 2016-10-28
server: refreshing initial login screen creates new auth requests


ericchiang:
Few things we could do:
1. Do an initial redirect so the user gets dropped at a screen with the `req=` parameter (fixes the refreshing).
2. Don't persist auth data until we redirect the user to a remote provider..

Also, we need to do rate limiting.

ericchiang:
> Don't persist auth data until we redirect the user to a remote provider..

FYI this also fixes issues when users suspend their screen and their auth request gets gc'd. I think the second option is a more correct one and would solve other problem.

nabokihms:
The issue was fixed in Dex v2.29.0
https://github.com/dexidp/dex/pull/1865

There is no need to keep track of this issue further. If you found this issue and still have a problem with auth requests, please open a new issue or discussion.

PR #1356, opened 2018-11-21
quickfix #1355: only try three times

srenatus:
cc @vito 😉

### DIFF
--- a/storage/sql/sql.go
+++ b/storage/sql/sql.go
@@ -4,6 +4,7 @@ package sql
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"regexp"
 	"time"
 
@@ -55,6 +56,8 @@ func isRetryableSerializationFailure(err error) bool {
 	return false
 }
 
+const maxRetries = 2
+
 var (
 	// The "github.com/lib/pq" driver is the default flavor. All others are
 	// translations of this.
@@ -75,14 +78,14 @@ var (
 				Isolation: sql.LevelSerializable,
 			}
 
-			for {
+			for i := 0; i < maxRetries; i++ {
 				tx, err := db.BeginTx(ctx, opts)
 				if err != nil {
 					return err
 				}
 
 				if err := fn(tx); err != nil {
-					if isRetryableSerializationFailure(err) {
+					if isRetryableSerializationFailure(err) && i != maxRetries {
 						continue
 					}
 
@@ -91,7 +94,7 @@ var (
 
 				err = tx.Commit()
 				if err != nil {
-					if isRetryableSerializationFailure(err) {
+					if isRetryableSerializationFailure(err) && i != maxRetries {
 						continue
 					}
 
@@ -100,6 +103,8 @@ var (
 
 				return nil
 			}
+
+			return errors.New("unable to resolve serialization failure")
 		},
 
 		supportsTimezones: true,


vito:
Fair enough! I can test this out Monday (on vacation atm) but I'm a little worried this limit will be too low. It was super easy to trigger the failure before the retry logic was added.

I can also try to work on a test to verify concurrent logins.

srenatus:
@vito Sure, enjoy your time off. We're just not able to cut a release right now, but that's OK.

I'm having a hard time coming up with a way to differentiate _legitimate serialization failures_ from _retryable ones that'll eventually succeed_... might try to find out what other folks do there.

vito:
@srenatus Ok, so here's my theory. I don't think there's really such a thing as legitimate vs. retryable serialization failures - I think that was a symptom of the tests, because they literally force an unsatisfiable concurrency situation which I would expect is unrealistic compared to actual API usage. All serialization failures should be retriable, otherwise it's a potential bug in the code, and part of the point of setting serializable transaction level would be to notice and fix those before they end up in the wild.

The tests passed for me because, as you found, my implementation leaked connections, eventually exhausting the server connection pool and resulting in an error. All the tests checked for was that an error was hit, so, well, they passed.

But I think the real fix here is to fix the connection leak by calling `tx.Rollback()` in the error case, and update the tests to *eventually* give up on their concurrent update attempts, and the test should see that this limit is hit. This would effectively test the retry logic, too, which is nice (since my PR initially had no test changes).

So far I've confirmed the connection leak theory by adding a `tx.Rollback()` in the error path. With this added, the tests went into an infinite loop (yay!). I'm thinking the next step is to have the tests eventually give up on the concurrent update, and assert that it reached the desired number of attempts before succeeding.

srenatus:
@vito thanks for looking into this. I'm wondering about this, too, 

> which I would expect is unrealistic compared to actual API usage. 

-- when would this happen? I think it could happen if you "choose" two different connectors at the same time. When you see the connector list, like
```
[ Log in with Username ]
[ Log in with SAML ]
```
the first link goes to `/dex/auth/local?req=authnreq-id`, and the second one to `/dex/auth/saml?req=authnreq-id`, and [the handler code behind these](https://github.com/dexidp/dex/blob/f3acec0b1b456aa7/server/handlers.go#L205-L237) will
- fetch the authn request with that id
- attempt to set its `ConnectorID` to `local` or `saml`

Now, if you'd do the two requests concurrently, this would lead to a serialization failure that isn't entirely wrong, and can't succeed on retry (unless I'm missing something). 👉 Pushed a commit here illustrating the situation a little, https://github.com/srenatus/dex/commit/ad2e5301754bb8bfa1aa315841d46e68004c4cf6. (Also, it would be a condition that can be triggered by an unauthenticated user 🤔)

What do you think? I guess there's a few options here -- for example, there has been evidence that restructuring the authn request code could be a good idea (#646, #1292).

srenatus:
> I haven't run your test change there yet, but I'm guessing you're seeing it fail (or hang)?

Yeah, depending on whether I'll `tx.Rollback()`, it's one or the other. Didn't keep close track, just tried to come up with a more realistic scenario. I'll dig in again tomorrow.

What's our desired behaviour, though? I'd think if the two-connectors-selected situation is happening, I would want the first one to success (whichever is first) and the second one to _fail_. In some way we need to ensure there's no cross-over, or mixup, of connectors. You cannot start an LDAP flow, do something to have the connector ID switch to SAML, and finish the LDAP flow. Or you can do that, but the result **must not** look like you had logged in using SAML 😉 

@ericchiang Could you chime in here? Do you know what edge cases this was supposed to catch?

vito:
Fair; I really don't know enough about the full Dex API and auth flow to know what kinds of guarantees are critical for consistency reasons, and if there are good reasons to fail user-initiated requests due to concurrency. So I'd appreciate @ericchiang's input here too. 🙂 

I typically expect last-write-wins; to me, two simultaneous requests with different intents have no absolutely "correct" resolution because it's subject to network jitter, but I could totally see why someone would want to start one login flow and then start a different one soon after, expecting the last one to work - they may have mistakenly initiated the first request via a simple misclick and just gone back to start over. This case is more trivially serializable, of course, but I'd guess supporting that would mean going with last-write-wins if it really comes down to the wire.

ericchiang:
Just to clarify about the tests. The tests just try to make sure that dex does the right thing in the case of concurrency. When they were written, no storage implementations retried, so the if there was a racy call, at least one of them should fail.

Since you've added some retries for optimistic concurrency, the test now passes. Which is the right behavior now, but the test is still locked into the old assumption. We should probably change the test.

I'm actually more concerned about the tx leak and HEAD having a failing test. To get HEAD back to a good state, can we:

1. revert #1342 
2. remerge the vendor changes in #1342
3. submit a PR that implements #1342 while taking into account #1355, #1363 and this PR?
 
I'm fine with retrying on any error. Maybe we should bake that into the storage layer so all implementations do that?

srenatus:
- [X] revert #1342
- [ ] remerge the vendor changes in #1342
- [ ] submit a PR that implements #1342 while taking into account #1355, #1363 and this PR?

srenatus:
The code of this PR is not useful anymore -- I'm keeping it open regardless, to track the https://github.com/dexidp/dex/pull/1356#issuecomment-442523779 progress above.

srenatus:
This is pretty much out of data right now, the discussion happens in #1341. Next-up relevant PR is #1370.

Issue #1262, opened 2018-07-17
The created ssl/ca.pem must then be mounted into your API server deployment


enenum:
The title is a line in the instructions for running dex with kubernetes.

I am new to all this, and I can't see how to get apiserver to trust dex. When I setup, I get apiserver complaining that the dex cert was signed by an unknown issuer.

So I see there is oidc-ca-file that I can set, but that needs to be a path in the apiserver container. This argument is for starting apiserver, so the I can't give a path as the apiserver container has not started, or secrets, for the same reason. Or so I think.

help anyone?

enenum:
I got around this in a  virtualbox VM environment. This can apply to a bare metal setup, but I don't see that it can work in a minikube environment.

I started the cluster (kubeadm init) and noticed that the apiserver mounts the host's /etc/ssl/certs to a path by the same name in its container. So I put the oidc ca file in that folder and gave the kubeadm config argument the path and cert name. 

I imagine there is a better way to do this out there. For something this unobvious, the terse instruction can keep folks away from a good-looking tool like dex.

srenatus:
@enenum Is this still something you're interested in? Looks like a welcome documentation improvement. 😃

ericchiang:
I think we probably want to update our Kubernetes examples.

OwenTuz:
I'm happy to pick up some Kubernetes documentation improvements, if nobody else has anything in the pipeline? I've been working through these today and noticed a few things missing so was thinking of submitting my working examples and clarifying a few points.

srenatus:
@OwenTuz That would be awesome! 🎈

PR #1358, opened 2018-11-22
Kubernetes docs: clarify steps around use/creation of TLS assets.

OwenTuz:
Related to issue #1262 .

While not strictly related to TLS, I'd like to make one or more follow-up PRs to fix issues around old references to TPRs, missing instructions on creating the serviceAccount and roles for RBAC etc. But this is a first pass, hopefully it makes reviewing easier.

### DIFF
--- a/Documentation/connectors/ldap.md
+++ b/Documentation/connectors/ldap.md
@@ -262,7 +262,7 @@ connectors:
     # freeIPA server's CA
     rootCA: ca.crt
     userSearch:
-      # Would translate to the query "(&(objectClass=person)(uid=<username>))".
+      # Would translate to the query "(&(objectClass=posixAccount)(uid=<username>))".
       baseDN: cn=users,dc=freeipa,dc=example,dc=com
       filter: "(objectClass=posixAccount)"
       username: uid
--- a/Documentation/kubernetes.md
+++ b/Documentation/kubernetes.md
@@ -3,6 +3,7 @@
 ## Overview
 
 This document covers setting up the [Kubernetes OpenID Connect token authenticator plugin][k8s-oidc] with dex.
+It also contains a worked example showing how the Dex server can be deployed within Kubernetes.
 
 Token responses from OpenID Connect providers include a signed JWT called an ID Token. ID Tokens contain names, emails, unique identifiers, and in dex's case, a set of groups that can be used to identify the user. OpenID Connect providers, like dex, publish public keys; the Kubernetes API server understands how to use these to verify ID Tokens.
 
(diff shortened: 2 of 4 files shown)


OwenTuz:
Got my issue numbers wrong, embarassing

OwenTuz:
Thanks for the fast review! I've updated (sorry) with some of the follow-ups I mentioned above

Documentation/kubernetes.md
```diff
@@ -84,14 +131,14 @@ $ kubectl create secret \
     --from-literal=client-secret=$GITHUB_CLIENT_SECRET
 ```
 
-Create the dex deployment, configmap, and node port service.
+### Deploy the Dex server
+
+Create the dex deployment, configmap, and node port service. This will also create RBAC bindings allowing the Dex pod access to manage [Custom Resource Definitions](https://github.com/dexidp/dex/blob/master/Documentation/storage.md#kubernetes-custom-resource-definitions-crds) within Kubernetes.
```
srenatus: Could you try making this a relative link instead? Maybe this could work:
```suggestion
Create the dex deployment, configmap, and node port service. This will also create RBAC bindings allowing the Dex pod access to manage [Custom Resource Definitions](storage.md#kubernetes-custom-resource-definitions-crds) within Kubernetes.
```

Documentation/connectors/ldap.md
```diff
@@ -262,7 +262,7 @@ connectors:
     # freeIPA server's CA
     rootCA: ca.crt
     userSearch:
-      # Would translate to the query "(&(objectClass=person)(uid=<username>))".
+      # Would translate to the query "(&(objectClass=posixAccount)(uid=<username>))".
```
srenatus: 👍 good catch! 😄

Documentation/kubernetes.md
OwenTuz: Thanks for catching that! I've updated it.

srenatus:
Looks good to me. However, I have no idea if it's correct. Please bear with me, let's get someone with sufficient knowledge of the domain to review this. (If you know someone, feel free to pull them in, it doesn't have to be someone with "commit bit". I'll merge when this has been reviewed, then.)

OwenTuz:
That sounds good to me, I'm not sure who to call on as this is all based on my own testing but I'd definitely appreciate a second set of eyes.

OwenTuz:
@whereisaaron @rendhalver would this be something one of you could weigh in on at all?

rendhalver:
Nice work!
Note to self: We should align the chart with these docs.

whereisaaron:
Good update

Issue #1190, opened 2018-02-14
[LDAP connector] Document that 'DN' must be in capitals


whereisaaron:
I was perplexed why my LDAP group searches were not working. But I noted on the otherwise excellent LDAP connector documentation page, that ['DN' was used in capitals](https://github.com/coreos/dex/blob/master/Documentation/connectors/ldap.md#example-mapping-a-schema-to-a-search-config), even though nothing else was.

I didn't think 'dn' was supposed to be case sensitive? Either way, I think this is worth specifically calling out in the documentation.

Fails with 'Login Error' screen:
```
      userSearch:
        baseDN: "ou=Users,dc=example,dc=com"
        filter: "(objectClass=inetorgperson)"
        username: cn
        idAttr: dn
        emailAttr: mail
        nameAttr: displayName
      groupSearch:
        baseDN: "ou=Groups,dc=example,dc=com"
        filter: "(objectClass=groupOfUniqueNames)"
        userAttr: dn
        groupAttr: uniqueMember
        nameAttr: cn
```

Works great, including groups:
```
      userSearch:
        baseDN: "ou=Users,dc=example,dc=com"
        filter: "(objectClass=inetorgperson)"
        username: cn
        idAttr: DN
        emailAttr: mail
        nameAttr: displayName
      groupSearch:
        baseDN: "ou=Groups,dc=example,dc=com"
        filter: "(objectClass=groupOfUniqueNames)"
        userAttr: DN
        groupAttr: uniqueMember
        nameAttr: cn
```

whereisaaron:
I found a [mention of the magic capitalization behavior](https://github.com/coreos/dex/blob/master/examples/config-ldap.yaml) here:
https://github.com/coreos/dex/blob/master/examples/config-ldap.yaml

Just need the same in the connector documentation page.

srenatus:
@whereisaaron I guess @OwenTuz's fine little change does the trick here, what do you think? (Optimistically, I'm closing this issue. Feel free to re-open.)

whereisaaron:
It all helps @srenatus! I don't think 'DN' ought to be case specific when LDAP normally isn't, but so long as you know this connector behavior you can work with it.

benjaminapetersen:
Does this mean ANY reference of DN? I am assuming yes.  A couple examples from the above docs page just for clarification:

```yaml
# Dex config example from dex docs
userSearch:
  idAttr: DN # this is caps, if it is DN 

# as well as this:
groupSearch:
[...]

PR #1359, opened 2018-11-26
LDAP connector - Document that 'DN' must be in capitals

OwenTuz:
As in #1190

### DIFF
--- a/Documentation/connectors/ldap.md
+++ b/Documentation/connectors/ldap.md
@@ -43,7 +43,7 @@ Dex currently allows insecure connections because the project is still verifying
 
 User entries are expected to have an email attribute (configurable through `emailAttr`), and a display name attribute (configurable through `nameAttr`). `*Attr` attributes could be set to "DN" in situations where it is needed but not available elsewhere, and if "DN" attribute does not exist in the record.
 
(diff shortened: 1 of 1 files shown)


Documentation/connectors/ldap.md
```diff
@@ -43,6 +43,8 @@ Dex currently allows insecure connections because the project is still verifying
 
 User entries are expected to have an email attribute (configurable through `emailAttr`), and a display name attribute (configurable through `nameAttr`). `*Attr` attributes could be set to "DN" in situations where it is needed but not available elsewhere, and if "DN" attribute does not exist in the record.
 
+For the purposes of configuring this connector, "DN" is case-sensitive and should always be capitalised. This is treated as a special case because, while all other configuration values refer to single attributes on an entity, an LDAP Distinguished Name consists of multiple attributes which uniquely identify a resource.
```
srenatus: 🤔 Does a DN really _consists of multiple attributes_? I've always though about these like file paths, referencing a node in directory tree... 
Looking at [`ldapwiki.com`](https://ldapwiki.com/wiki/Distinguished%20Names), I guess both is somewhat correct,
> A Distinguished Names is comprised of zero or more Relative Distinguished Name components that identify the location of the entry in the DIT. 

I guess this is OK, although I'd lean towards ditching the second sentence. What do you think? 😃

OwenTuz: I hesitated over the second sentence as well, let's just skip it

OwenTuz: (Updated)

srenatus: 👍 thank you! 🎉

PR #1364, opened 2018-11-27
Migrate to dep

josdotso:
Looking for ideas on how to improve this PR until it's acceptable for merge.  I would really like to see dex adopt the dep over glide.  

Some caveats:
- It seems dep doesn't prune/omit unused subpackages as aggressively as glide-vc. This is probably because I haven't found a way to tell dep in `Gopkg.toml` a list of subpackages we actually want, as we do with subpackage lists in `glide.yaml`. We could probably edit the `Gopkg.lock` file manually but that seems like a bad idea.
- The new `Gopkg.toml` was initially the result of running `dep init` and letting dep try to migrate from glide. I had to change `constraint` to `override` in a couple of places to keep `dep` from throwing errors about pinning those transitive dependencies.
- Some deps, transitive or not, didn't arrive in the new `Gopkg.toml`.  I think this is because `dep` didn't detect those packages were in use.  `glide` isn't explicit about transitive dependency pinning but `dep` is (see `[[override]]` stanzas). We can definitely force re-inclusion of those dropped packages (see `required` property in dep docs), but `make` works just fine, so I'm not sure we need them.

See: https://golang.github.io/dep/docs/Gopkg.toml.html

Note: There are 3 commits in this PR to separate config changes from delete/modified vendor files from new vendor files.


ericchiang:
Why not switch to go modules?

josdotso:
@ericchiang  I'll look into that.  Thanks for the tip.

ericchiang:
FYI, as part of our build we vendor "main" packages. I had to add the following file to trick `go mod vendor` to include them

```go
// +build vendor

package main

// This file exists to trick "go mod vendor" to include "main" packages.
// It is not expected to build, the build tag above is only to prevent this
// file from being included in builds.

import (
        _ "github.com/golang/protobuf/protoc-gen-go"
        _ "golang.org/x/lint/golint"
)

func main() {}
```

josdotso:
Thanks, all!

Closing this PR in favor of pursuing Go modules. See #1365.

PR #1365, opened 2018-11-28
Migrate to go modules

josdotso:
This is a follow up to #1364 where `go mod` was suggested as a better `glide` replacement than `dep`.  I'll close #1364 in a few minutes in favor of this PR.

Notes:
- I didn't make verbose comments in `go.mod` as were in `glide.yaml` because it's not yet clear to me if these will be automatically dropped by `go mod`.
- `go mod` doesn't seem to have an option for pruning non-go files.
- Updated `Makefile` to replace `glide` calls with `go dep` calls.

TODO:
- [x] Implement the ~~hack~~ workaround :) suggested by @ericchiang to enable vendoring main packages.
  - Question on that: Where should I write out the file you defined in #1364, @ericchiang? Would a top level `main.go` be the right place?
- [x] Fix the CI build when it probably breaks due to my first pass at this.


josdotso:
@ericchiang  Please see my question in the comment above. I'm eager to complete this PR.

srenatus:
> Implement the hack suggested by...

😄 I'd consider it a workaround. But this this fine.

🎉 Thanks for doing this, very nice to get up to speed with the ecosystem changes 👍

ericchiang:
> Question on that: Where should I write out the file you defined in #1364, @ericchiang? Would a top level main.go be the right place?

Can we call the file `vendor.go`?

josdotso:
Workaround fixed the build. Thanks!

ericchiang:
Thanks! @josdotso per our docs, can you squash this PR into two commits? One with the regular file changes, and one with the vendor changes?

```
*: remove glide and add Go module files
vendor: make revendor
```

https://github.com/dexidp/dex/blob/master/Documentation/dev-dependencies.md

ericchiang:
We can update that doc separately.

josdotso:
Thanks @ericchiang. Does it look better now?

PR #1370, opened 2018-12-10
Fix concurrent access for SQL storage.

kepkin:
SERIALIZABLE isolation level is not viable solution, as it will slow down the whole Postgres. And again there must be retries as you will always receive errors like "could not serialize access...".

Previous pull request for retries like https://github.com/dexidp/dex/pull/1356 were wrong as it resulted in deadlock in Database (you open transaction inside another transaction in conformance tests).

In SQL world the better solution is to lock the row explicitly with SELECT FOR UPDATE.

### DIFF
--- a/storage/sql/crud.go
+++ b/storage/sql/crud.go
@@ -386,13 +386,12 @@ func (c *conn) UpdateKeys(updater func(old storage.Keys) (storage.Keys, error)) 
 		firstUpdate := false
 		// TODO(ericchiang): errors may cause a transaction be rolled back by the SQL
 		// server. Test this, and consider adding a COUNT() command beforehand.
-		old := storage.Keys{}
 		err := c.flavor.lockForUpdate(tx, "keys", "id", keysRowID)
 		if err != nil {
 			return fmt.Errorf("get keys: %v", err)
 		}
 
-		old, err = getKeys(tx)
+		old, err := getKeys(tx)
 		if err != nil {
 			if err != storage.ErrNotFound {
 				return fmt.Errorf("get keys: %v", err)


storage/sql/crud.go
```diff
@@ -380,13 +386,23 @@ func (c *conn) UpdateKeys(updater func(old storage.Keys) (storage.Keys, error))
 		firstUpdate := false
 		// TODO(ericchiang): errors may cause a transaction be rolled back by the SQL
 		// server. Test this, and consider adding a COUNT() command beforehand.
```
srenatus: 🤔 Is that comment still accurate now?

storage/sql/sql.go
```diff
@@ -22,6 +22,10 @@ type flavor struct {
 	// Optional function to create and finish a transaction.
 	executeTx func(db *sql.DB, fn func(*sql.Tx) error) error
```
srenatus: This field could go away now, can't it?

storage/sql/crud.go
```diff
@@ -140,10 +140,16 @@ func (c *conn) UpdateAuthRequest(id string, updater func(a storage.AuthRequest)
 			return err
 		}
 
+		err = c.flavor.lockForUpdate(tx, "auth_request", "id", r.ID)
+		if err != nil {
+			return fmt.Errorf("update auth request: %v", err)
```
srenatus: 💭 All these `fmt.Errorf` make me nervous -- they're hard to deal with in call sites -- but this is the style this package is written in, so it's fine at add a few. 👍

ericchiang: We can switch to github.com/pkg/errors in another PR :)

storage/sql/sql.go
```diff
@@ -43,27 +47,9 @@ var (
...
 	// translations of this.
 	flavorPostgres = flavor{
-		// The default behavior for Postgres transactions is consistent reads, not consistent writes.
-		// For each transaction opened, ensure it has the correct isolation level.
-		//
-		// See: https://www.postgresql.org/docs/9.3/static/sql-set-transaction.html
-		//
-		// NOTE(ericchiang): For some reason using `SET SESSION CHARACTERISTICS AS TRANSACTION` at a
-		// session level didn't work for some edge cases. Might be something worth exploring.
-		executeTx: func(db *sql.DB, fn func(sqlTx *sql.Tx) error) error {
```
ericchiang: Does anything use this anymore? can we nuke it?

storage/sql/crud.go
ericchiang: I don't remember. Ha

storage/sql/crud.go
```diff
@@ -380,13 +386,23 @@ func (c *conn) UpdateKeys(updater func(old storage.Keys) (storage.Keys, error))
...
+		err := c.flavor.lockForUpdate(tx, "keys", "id", keysRowID)
 		if err != nil {
 			if err != storage.ErrNotFound {
 				return fmt.Errorf("get keys: %v", err)
 			}
+
 			firstUpdate = true
-			old = storage.Keys{}
+		} else {
+			old, err = getKeys(tx)
```
ericchiang: I'm confused why there's an `else` statement. Can you please add a comment?

storage/sql/sql.go
```diff
@@ -22,6 +22,10 @@ type flavor struct {
 	// Optional function to create and finish a transaction.
 	executeTx func(db *sql.DB, fn func(*sql.Tx) error) error
 
+	// Function for locking rows. It must return error if locking failed. If underlying
```
ericchiang: This comment needs to be expanded to explain what `value` is and why it's required.

storage/sql/sql.go
```diff
@@ -43,27 +47,9 @@ var (
...
-
-			if _, err := tx.Exec(`SET TRANSACTION ISOLATION LEVEL SERIALIZABLE;`); err != nil {
-				return err
-			}
-			if err := fn(tx); err != nil {
-				return err
-			}
-			return tx.Commit()
+		lockForUpdate: func(tx *trans, table, column, value string) error {
+			_, err := tx.Exec("SELECT 1 FROM "+table+" WHERE "+column+" = $1 FOR UPDATE NOWAIT;", value)
```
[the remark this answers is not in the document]
ericchiang: Probably better to add an inline comment :)

This works, but the lockForUpdate API seems a bit risky. Ideally it could be baked into ExecTx or something, but nothing's coming to mind at the moment.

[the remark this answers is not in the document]
kepkin: Yes NOWAIT here is only because of tests. And yes, I think tests need to be redesigned a little bit, as changing the same auth_request concurrently wouldn't actually harm a login flow, and deadlock which current tests produce is not possible in real life situations.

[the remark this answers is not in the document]
ericchiang: We should probably just update/remove the tests then, right?

storage/sql/crud.go
```diff
@@ -388,21 +388,17 @@ func (c *conn) UpdateKeys(updater func(old storage.Keys) (storage.Keys, error))
 		// server. Test this, and consider adding a COUNT() command beforehand.
 		old := storage.Keys{}
 		err := c.flavor.lockForUpdate(tx, "keys", "id", keysRowID)
+		if err != nil {
+			return fmt.Errorf("get keys: %v", err)
+		}
+
+		old, err = getKeys(tx)
```
srenatus: [nit] Let's ditch line 389 and go with `old, err :=` here, now that the code is rearranged.

krylovsk:
What about the issue with interleaving connectors in the concurrent login flow mentioned in https://github.com/dexidp/dex/pull/1356#issuecomment-442366861 ?

Failing one of the transactions seems to be the only way to handle it in current implementation of handler/storage.

storage/sql/sql.go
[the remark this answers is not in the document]
krylovsk: What about the issue with interleaving connectors in the concurrent login flow mentioned in https://github.com/dexidp/dex/pull/1356#issuecomment-442366861 ?

Failing one of the transactions seems to be the only way to handle it in current implementation of handler/storage.

bonifaido:
Can we revive this PR? I'm happy to help. We have an issue with the MySQL storage backend where according to the old conformance tests (the ones currently in master) SERIALIZABLE isolation level is needed, which has quite some issues on different MySQL implementations ( as in https://github.com/dexidp/dex/pull/1511). Percona XtraDB doesn't implement let you use it by default, AWS Aurora doesn't have it at all, on MariaDB transaction_isolation is called tx_isolation so it is a pain TBH.

FYI: I have rebased this branch on the current master and copied the `lockForUpdate` method from the Postgres implementation, removed the explicit SERIALIZABLE tx isolation level setting, all the MySQL conformance tests have passed on MariaDB, with MySQL the syntax is wrong, but as I understand correctly this is something we would like see removed.

srenatus:
💯 I'd be very interested in getting this resolved. I think the main blocker here is that we have no test suite that makes the error appear -- our contributors, however, seem to have that. This would help evaluating the different approaches that have been put forward when this issue had been on the table last winter....

PR #1374, opened 2018-12-14
Keystone connector

kbalka:
PR contains connector for openstack keystone.

Features:
- access tokens
- refresh tokens 
- groups

Requirements:
- access to openstack keystone instance
- keystone administrative account credentials 

Enabling keystone connector specific tests:
- make sure docker is running 
- export DEX_TEST_KEYSTONE=1
- make tests


connector/keystone/keystone.go
```diff
@@ -0,0 +1,158 @@
...
+func (p keystoneConnector) Close() error { return nil }
+
+func (p keystoneConnector) Login(ctx context.Context, s connector.Scopes, username, password string) (
+		identity connector.Identity, validPassword bool, err error) {
+	response, err := p.getTokenResponse(username, password)
+
+	// Providing wrong password or wrong keystone URI throws error
+	if err == nil && response.StatusCode == 201 {
+    	token := response.Header["X-Subject-Token"][0]
+		data, _ := ioutil.ReadAll(response.Body)
```
srenatus: Let's please check all errors (same applies to other calls in this PR) -- it's fine to skip an error check or two in the test files, but let's please not do that here.

connector/keystone/keystone.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (p keystoneConnector) Close() error { return nil }
+
+func (p keystoneConnector) Login(ctx context.Context, s connector.Scopes, username, password string) (
+		identity connector.Identity, validPassword bool, err error) {
+	response, err := p.getTokenResponse(username, password)
+
+	// Providing wrong password or wrong keystone URI throws error
+	if err == nil && response.StatusCode == 201 {
```
srenatus: Let's stick to idiomatic Go code, like
```go
resp, err := p.getTokenResponse(username, password)
if err != nil {
    // handle err
}
if resp.StatusCode == 201 {
    // ... read body
```

Also, `resp.Body` will need to be closed, otherwise we risk leaking memory.

connector/keystone/keystone.go
```diff
@@ -0,0 +1,158 @@
...
+func (c *Config) Open(id string, logger logrus.FieldLogger) (connector.Connector, error) {
+	return &keystoneConnector{c.Domain, c.KeystoneHost,
+	c.KeystoneUsername, c.KeystonePassword, logger}, nil
+}
+
+func (p keystoneConnector) Close() error { return nil }
+
+func (p keystoneConnector) Login(ctx context.Context, s connector.Scopes, username, password string) (
+		identity connector.Identity, validPassword bool, err error) {
+	response, err := p.getTokenResponse(username, password)
```
srenatus: It's better to thread the `ctx` parameter through here, and use it when making the request to Keystone -- see [the github connector for example](https://github.com/dexidp/dex/blob/4329f407dcfefad56aa81b4d3da1a8e487bd023e/connector/github/github.go#L483-L488).

connector/keystone/keystone.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// Providing wrong password or wrong keystone URI throws error
+	if err == nil && response.StatusCode == 201 {
+    	token := response.Header["X-Subject-Token"][0]
```
srenatus: [nit] We could use [`(*Header).Get()`](https://golang.org/pkg/net/http/#Header.Get) here to not depend on the case of the header key.

connector/keystone/types.go
```diff
@@ -0,0 +1,154 @@
...
+	ID     string         `json:"id"`
+	Name   string         `json:"name"`
+}
+
+type domainKeystone struct {
+	ID string   `json:"id"`
+	Name string `json:"name"`
+}
+
+// Config holds the configuration parameters for Keystone connector.
```
srenatus: Do you know which (API) version of keystone we're requiring here? There's a few https://developer.openstack.org /api-ref/identity/ 😉

connector/keystone/types.go
```diff
@@ -0,0 +1,154 @@
...
+	Methods   []string               `json:"methods"`
+	ExpiresAt string                 `json:"expires_at"`
+	User      userKeystone           `json:"user"`
+}
+
+type tokenResponse struct {
+	Token token `json:"token"`
+}
+
+type createUserRequest struct {
```
srenatus: Can we put the definitions which we only need for tests into the test file, please?

examples/config-keystone.yaml
```diff
@@ -0,0 +1,55 @@
+# The base path of dex and the external name of the OpenID Connect service.
```
srenatus: I don't think we need one of these example files for every connector. Let's remove it. 😃

connector/keystone/keystone_test.go
```diff
@@ -0,0 +1,326 @@
...
+const invalidPass = "WRONG_PASS"
+
+const testUser = "test_user"
+const testPass = "test_pass"
+const testEmail = "test@example.com"
+const testGroup = "test_group"
+
+const domain = "default"
+
+func startKeystoneContainer() string {
```
srenatus: This is a nice integration test suite.

But let's thread in a `t *testing.T` argument use `t.Helper()` and `t.Fatal` in the helper methods, instead of `fmt.Printf` calls, please 😉

connector/keystone/keystone_test.go
```diff
@@ -0,0 +1,326 @@
...
+   		}
+   		time.Sleep(10 * time.Second)
+  	}
+  	if !running {
+    	fmt.Printf("Failed to start keystone container")
+    	os.Exit(1)
+  	}
+  	defer cleanKeystoneContainer(dockerID)
+  	// run all tests
+	m.Run()
```
srenatus: ❓ Do we need `os.Exit(m.Run())` here to have the test run fail when one of the tests fails?

connector/keystone/types.go
kbalka: We are requiring API version 3. Could you have a second look at this PR ? It seems that CI builds are passing now ;-)

connector/keystone/keystone_test.go
```diff
@@ -0,0 +1,360 @@
...
+	"bytes"
+	"encoding/json"
+	"golang.org/x/net/context"
+	"io/ioutil"
+	"reflect"
+	"strings"
+)
+
+const (
+	adminUser   = "demo"
```
srenatus: ❓ Would it make sense to pass these in, too, at least in so far as they're specific to https://hub.docker.com/r/openio/openstack-keystone?

.travis.yml
```diff
@@ -13,13 +13,14 @@ services:
...
 env:
-  - DEX_POSTGRES_DATABASE=postgres DEX_POSTGRES_USER=postgres DEX_POSTGRES_HOST="localhost" DEX_ETCD_ENDPOINTS=http://localhost:2379 DEX_LDAP_TESTS=1 DEBIAN_FRONTEND=noninteractive
+  - DEX_POSTGRES_DATABASE=postgres DEX_POSTGRES_USER=postgres DEX_POSTGRES_HOST="localhost" DEX_ETCD_ENDPOINTS=http://localhost:2379 DEX_LDAP_TESTS=1 DEBIAN_FRONTEND=noninteractive DEX_KEYSTONE_URL=http://localhost:5000 DEX_KEYSTONE_ADMIN_URL=http://localhost:35357
 
 install:
   - sudo -E apt-get install -y --force-yes slapd time ldap-utils
   - sudo /etc/init.d/slapd stop
   - docker run -d --net=host gcr.io/etcd-development/etcd:v3.2.9
-
+  - docker run -d -p 0.0.0.0:5000:5000 -p 0.0.0.0:35357:35357 openio/openstack-keystone
```
srenatus: ❓ Is there any version pins we could add? Right now, this will take the "latest" tag, and that'll potentially break a month (or a year) from now, and will cause some PR to go ":x:" for the wrong reasons :wink:

.travis.yml
```diff
@@ -13,13 +13,14 @@ services:
...
+  - sleep 60s
```
srenatus: 🤔 Can we do something that is a bit more robust? 
Maybe (untested)
```yaml
  - |
    until curl --fail http://127.0.0.1:5000/v3; do
      echo 'Waiting for keystone...'
      sleep 1;
    done;
```
This would potentially be faster (wait for less than 60s) -- OTOH, it could take longer (if keystone does not come up), but I'd hope travis will fail the build with some timeout in that case.

Either way, this new additional dependency in our test suite is something to keep an eye on when this is merged. 🔍

connector/keystone/keystone.go
```diff
@@ -0,0 +1,189 @@
...
+	if resp.StatusCode == 201 {
+		token := resp.Header.Get("X-Subject-Token")
+		data, err := ioutil.ReadAll(resp.Body)
+		defer resp.Body.Close()
+
+		if err != nil {
+			return identity, false, err
+		}
+
+		var tokenResponse = new(tokenResponse)
```
srenatus: [nit] I'm not a fan of clashing names between types and variables; but if you are, that's tine 😉

.travis.yml
kbalka: This image has only "latest" tag, so the only thing we can do here is to fork it on dockerhub and add specific tag.

srenatus: 👍 Let's be optimistic. We can still fork it if the need arises down the road. (But really, I've rather avoid that 😅)

srenatus:
🎗 Note to self: *let's squash these before merging*.

srenatus:
Not right now, though! I appreciate the separate commits a lot for review 👍

joannanosek:
Hi @srenatus! Sorry for torpedoing with the commits 😅 And thank you for valuable feedback!

.travis.yml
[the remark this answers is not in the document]
kbalka: Good question :-) For me, the advantage here is only readability - you can see which ports are exposed and compare it with exported variables DEX_KEYSTONE*.

kbalka:
the application which can possibly leverage openstack cluster authentication is https://github.com/IntelAI/inference-model-manager :-)

connector/keystone/types.go
```diff
@@ -0,0 +1,102 @@
...
+	ID string `json:"id"`
+}
+
+type token struct {
+	IssuedAt  string                 `json:"issued_at"`
+	Extras    map[string]interface{} `json:"extras"`
+	Methods   []string               `json:"methods"`
+	ExpiresAt string                 `json:"expires_at"`
+	User      userKeystone           `json:"user"`
+}
```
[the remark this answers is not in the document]
kbalka: I think we don't need to care about ExpiresAt - because we are using token immediately after it's obtained. Keystone tokens are quite short living (I think by default 1h). Keeping it in identity.ConnectorData and using for Refresh would limit liveness of dex refresh token by expiration of keystone user token. That's why we require keystone admin account credentials. Admin should have access to other's users and their groups. When we check refresh token, the admin account gets fresh token from keystone and verifies user existence and groups. I hope that it makes some sense :-)

.travis.yml
ericchiang: Is there an official image for Keystone? I don't trust a "latest" tag from a organization that doesn't maintain the project.

connector/keystone/types.go
```diff
@@ -0,0 +1,87 @@
+package keystone
```
ericchiang: Please don't separate types and methods. These should all go in the same file

connector/keystone/keystone.go
```diff
@@ -0,0 +1,187 @@
...
+	resp, err := p.getTokenResponse(ctx, p.KeystoneUsername, p.KeystonePassword)
+	if err != nil {
+		return "", err
+	}
+	token := resp.Header.Get("X-Subject-Token")
+	return token, nil
+}
+
+func (p *keystoneConnector) checkIfUserExists(ctx context.Context, userID string, token string) (bool, error) {
+	userURL := p.KeystoneHost + "/v3/users/" + userID
```
ericchiang: please link to upstream docs describing this endpoint

connector/keystone/keystone.go
```diff
@@ -0,0 +1,187 @@
...
+
+	return identity, false, nil
+}
+
+func (p *keystoneConnector) Prompt() string { return "username" }
+
+func (p *keystoneConnector) Refresh(
+	ctx context.Context, s connector.Scopes, identity connector.Identity) (connector.Identity, error) {
+
+	token, err := p.getAdminToken(ctx)
```
ericchiang: Should this token be cached?

connector/keystone/keystone.go
```diff
@@ -0,0 +1,187 @@
...
+
+	ok, err := p.checkIfUserExists(ctx, identity.UserID, token)
+	if err != nil {
+		return identity, err
+	}
+	if !ok {
+		return identity, fmt.Errorf("keystone: user %q does not exist", identity.UserID)
+	}
+
+	groups, err := p.getUserGroups(ctx, identity.UserID, token)
```
ericchiang: groups should only be set if the scope is requested

connector/keystone/keystone.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (p *keystoneConnector) Login(ctx context.Context, s connector.Scopes, username, password string) (
+	identity connector.Identity, validPassword bool, err error) {
+	resp, err := p.getTokenResponse(ctx, username, password)
+	if err != nil {
+		return identity, false, fmt.Errorf("keystone: error %v", err)
+	}
+
+	// Providing wrong password or wrong keystone URI throws error
+	if resp.StatusCode == 201 {
```
ericchiang: Needs to check for non 2xx status codes. Also, just return early:

```
if resp.StatusCode / 100 != 2 {
    // return error describing status code
}
if resp.StatusCode != 201 {
    return identity, false, nil
}
// other logic
```

connector/keystone/keystone.go
```diff
@@ -0,0 +1,187 @@
...
+// Open returns an authentication strategy using Keystone.
+func (c *Config) Open(id string, logger logrus.FieldLogger) (connector.Connector, error) {
+	return &keystoneConnector{c.Domain, c.KeystoneHost,
+		c.KeystoneUsername, c.KeystonePassword, logger}, nil
+}
+
+func (p *keystoneConnector) Close() error { return nil }
+
+func (p *keystoneConnector) Login(ctx context.Context, s connector.Scopes, username, password string) (
+	identity connector.Identity, validPassword bool, err error) {
```
ericchiang: nit: keep these arguments on one line

connector/keystone/types.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"github.com/sirupsen/logrus"
+)
+
+type keystoneConnector struct {
```
ericchiang: already in the keystone package, so this can just be called "connector"

connector/keystone/keystone.go
```diff
@@ -0,0 +1,187 @@
...
+			return identity, false, err
+		}
+		defer resp.Body.Close()
+
+		var tokenResp = new(tokenResponse)
+		err = json.Unmarshal(data, &tokenResp)
+		if err != nil {
+			return identity, false, fmt.Errorf("keystone: invalid token response: %v", err)
+		}
+		groups, err := p.getUserGroups(ctx, tokenResp.Token.User.ID, token)
```
ericchiang: Groups should only be fetched if the scope is requested

connector/keystone/keystone.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+var (
+	_ connector.PasswordConnector = &keystoneConnector{}
+	_ connector.RefreshConnector  = &keystoneConnector{}
+)
+
+// Open returns an authentication strategy using Keystone.
+func (c *Config) Open(id string, logger logrus.FieldLogger) (connector.Connector, error) {
+	return &keystoneConnector{c.Domain, c.KeystoneHost,
```
ericchiang: nit: format this with fields on different lines

```
return &keystoneConnector{
    c.Doman,
    c.KeystoneHost,
    c.KeystoneUsername,
    c.KeystonePassword,
    logger,
}
```

connector/keystone/types.go
```diff
@@ -0,0 +1,87 @@
...
+	Domain           string
+	KeystoneHost     string
```
ericchiang: These should be `Host`, `AdminUsername`, `AdminPassword`, right?

connector/keystone/types.go
```diff
@@ -0,0 +1,87 @@
...
+//		id: keystone
+//		name: Keystone
+//		config:
+//			keystoneHost: http://example:5000
+//			domain: default
+//      keystoneUsername: demo
+//      keystonePassword: DEMO_PASS
+type Config struct {
+	Domain           string `json:"domain"`
+	KeystoneHost     string `json:"keystoneHost"`
```
ericchiang: Yeah, "host", "username", "password"

connector/keystone/keystone.go
kbalka: Good point, I've added a logic and test case

connector/keystone/keystone.go
kbalka: We can add caching token across sessions if needed, but my opinion is that at this moment we should keep the code simple. Is it ok for you ?

kbalka:
Hi @ericchiang, thanks for a thorough review. I'd addressed most of the issues. Please check if it looks better now. Regarding use of openio/openstack-keystone:latest image, I've asked the maintainer to add specific version tag, waiting for response. But, the most important, merry 🎄 and 🎅 to you together with @joannanosek @srenatus and @soggiest !

soggiest:
Happy Holidays and New Years!

.travis.yml
kbalka: I could not find an "official" image for Keystone. I've asked the maintainer of the image to add tag, so right now we are using "pike" tag. Is this good enough ?

connector/keystone/types.go
srenatus: How about `conn` or plain `c`? Currently, other connectors use a similar name to what's introduced here (e.g. `microsoftConnector`), but I do agree that shorter is nicer, the context is clear here.

.travis.yml
srenatus: I think it's a good compromise. Thanks for engaging with the maintainer there 👍

connector/keystone/types.go
kbalka: Sure, it makes sense. I've changed to `conn`.

connector/keystone/keystone.go
```diff
@@ -0,0 +1,189 @@
...
+		identity.UserID = tokenResponse.Token.User.ID
+		identity.Groups = groups
+		return identity, true, nil
+
+	}
+
+	return identity, false, nil
+}
+
+func (p keystoneConnector) Prompt() string { return "username" }
```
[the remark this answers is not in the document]
kbalka: I know it can be integrated with LDAP and ActiveDirectory.

kbalka:
Hi @srenatus. Thanks for the approval !!! I've squashed the commits into number of 4, added authors and force-pushed the branch on my fork. The code seems to be ready for merge. The only strange thing I can observe that there is a "tail" of 8 commits in the PR which are already on the master. They will probably disappear after merge, but I'm not sure. I can fix the PR if needed.

srenatus:
They should disappear if you fetch origin and rebase upon origin/master :)

I'd prefer getting this straightened out before hitting the button...

srenatus:
@kbalka Thank you so much! 🎉 

@ericchiang sorry I didn't wait for your "👍" after all. Feel free to assign me anything that's left 😉

PR #1380, opened 2019-01-07
LDAP connector - add emailSuffix config option

dkess:
In my organization, we don't have an LDAP mail attribute because email addresses are always based on username. This PR adds an `emailSuffix` option for this use case, where an email domain can be specified in place of a mail attribute.

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -107,6 +107,10 @@ type Config struct {
 		IDAttr    string `json:"idAttr"`    // Defaults to "uid"
 		EmailAttr string `json:"emailAttr"` // Defaults to "mail"
 		NameAttr  string `json:"nameAttr"`  // No default.
+
+		// If this is set, the email claim of the id token will be constructed from the idAttr and
+		// value of emailSuffix. This should not include the @ character.
+		EmailSuffix string `json:"emailSuffix"` // No default.
 	} `json:"userSearch"`
 
 	// Group search configuration.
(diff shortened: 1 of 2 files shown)


connector/ldap/ldap_test.go
```diff
@@ -123,6 +123,80 @@ userpassword: bar
...
+		{
+			name:      "invaliduser",
+			username:  "idontexist",
+			password:  "foo",
+			wantBadPW: true, // Want invalid password, not a query error.
+		},
+	}
+
+	runTests(t, schema, connectLDAP, c, tests)
+}
```
srenatus: [nit] missing newline

connector/ldap/ldap_test.go
```diff
@@ -123,6 +123,80 @@ userpassword: bar
 	runTests(t, schema, connectLDAP, c, tests)
 }
 
+func TestQuerySuffix(t *testing.T) {
```
srenatus: Thanks for adding this test -- would you mind cutting it down a little?

I'd think two cases would be enough: the happy path, and one with the `EmailAttr` missing 🤔 Invalid password, and not-found-user don't seem required there, it's not related to the to email suffix.

[nit] Let's add Email there: 
```suggestion
func TestQueryWithEmailSuffix(t *testing.T) {
```

connector/ldap/ldap.go
```diff
@@ -107,6 +107,9 @@ type Config struct {
 		IDAttr    string `json:"idAttr"`    // Defaults to "uid"
 		EmailAttr string `json:"emailAttr"` // Defaults to "mail"
 		NameAttr  string `json:"nameAttr"`  // No default.
+
+		// Use a username suffix instead of attribute for email.
```
srenatus: Could you elaborate what this means a little? Maybe something along the lines of _"if this is set, the email claim of the id token will be constructed from ... etc etc"_

dkess:
Thanks for the feedback! I believe I've addressed all the comments, let me know if anything else is necessary.

dkess:
Done!

PR #1387, opened 2019-01-22
Fix the error messages

hainesc:
I am reading the source code and I found maybe the error messages are reversed, or swapped.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -646,7 +646,7 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 
 	authCode, err := s.storage.GetAuthCode(code)
 	if err != nil || s.now().After(authCode.Expiry) || authCode.ClientID != client.ID {
-		if err != storage.ErrNotFound {
+		if err == storage.ErrNotFound {
 			s.logger.Errorf("failed to get auth code: %v", err)
 			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
 		} else {


server/handlers.go
```diff
@@ -646,7 +646,7 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 
 	authCode, err := s.storage.GetAuthCode(code)
 	if err != nil || s.now().After(authCode.Expiry) || authCode.ClientID != client.ID {
-		if err != storage.ErrNotFound {
+		if err == storage.ErrNotFound {
 			s.logger.Errorf("failed to get auth code: %v", err)
 			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
 		} else {
```
srenatus: 🤔 I think together with line 648 (`err != nil`), what was here before is correct:
- first branch: if it's NOT `storage.ErrNotFound`, something weird happened, and we'll inform the user about it -- it's a 500-ish thing (hence `errServerError`)
- second branch: if **is `storage.ErrNotFound`, the user has given us a bad code, and we tell them, using `errInvalidRequest` (probably a 400)

Does this make sense? 😃

srenatus: 🤔 how about having a bit `switch err { ... } ` instead? I don't know... maybe that could clarify stuff...

hainesc: I think a `switch` maybe better here. I will have a try tomorrow.

Thanks for your review.

hainesc: I survey some, I think it seems better that we wrap all internal errors(eg, timeout for etcd) as `github.com/pkg/errors.Wrap(err, "Server Error or some other words")`. It will import another [errors](https://github.com/pkg/errors) package which aims to [Go2 error proposals](https://go.googlesource.com/proposal/+/master/design/go2draft.md), but it is not final release. 

By using that, the piece of code will look like this:
```go
switch err := errors.Cause(err).(type) {
case *MyError:
        // handle specifically
default:
        // unknown error
}
```

It is trivial work to import a big `switch` without `Go2 error` package, the `errors` package offers little now. I think it's better that I close this PR and wait the `Go2 error` final release patiently. What do you think about it?

hainesc:
Close this PR.

PR #1396, opened 2019-02-01
Use github login as the id

jtnord:
Allow an option to use the github user handle rather than an id.

For downstream apps using a github handle is much simpler than working with numbers.

Whilst the number is stable and the handle is not - GitHub does give you a big scary warning if you try and change it that bad things may happen to you, and generally few users ever change it.

This can be enabled with a configuration option `useLoginAsId`

### DIFF
--- a/Documentation/connectors/github.md
+++ b/Documentation/connectors/github.md
@@ -69,7 +69,7 @@ connectors:
     teamNameField: slug
     # flag which will switch from using the internal GitHub id to the users handle (@mention) as the user id.
     # It is possible for a user to change their own user name but it is very rare for them to do so
-    useLoginAsId: false
+    useLoginAsID: false
 ```
 
 ## GitHub Enterprise
--- a/Documentation/connectors/github.md
+++ b/Documentation/connectors/github.md
@@ -67,6 +67,9 @@ connectors:
     #   - ['acme:site-reliability-engineers'] for 'slug'
     #   - ['acme:Site Reliability Engineers', 'acme:site-reliability-engineers'] for 'both'
     teamNameField: slug
+    # flag which will switch from using the internal GitHub id to the users handle (@mention) as the user id.
+    # It is possible for a user to change their own user name but it is very rare for them to do so
+    useLoginAsId: false
 ```
 
 ## GitHub Enterprise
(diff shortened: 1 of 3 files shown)


connector/github/github.go
```diff
@@ -260,12 +264,17 @@ func (c *githubConnector) HandleCallback(s connector.Scopes, r *http.Request) (i
...
 	identity = connector.Identity{
 		UserID:        strconv.Itoa(user.ID),
 		Username:      username,
 		Email:         user.Email,
 		EmailVerified: true,
 	}
+	if c.useLoginAsId {
+		identity.UserID = user.Login
+	}
+
```
srenatus: Travis fails because of this extra newline, I believe.

Documentation/connectors/github.md
```diff
@@ -67,6 +67,9 @@ connectors:
     #   - ['acme:site-reliability-engineers'] for 'slug'
     #   - ['acme:Site Reliability Engineers', 'acme:site-reliability-engineers'] for 'both'
     teamNameField: slug
+    # flag which will switch from using the internal GitHub id to the users handle (@mention) as the user id.
+    # It is possible for a user to change their own user name but it is very rare for them to do so
+    useLoginAsId: false
```
srenatus: [nit] let's make id `Id`? What do you think?
```suggestion
    useLoginAsID: false
```

jtnord: I normally prefer `Id` but will change as it is not my codebase (there are 135 occurences of `UserID` and 32 of `UserId`)  :)

connector/github/github.go
jtnord: removed the newline

jtnord:
> One question for clarification: there's no way to not have a login, right?

never say never - but I do not see how you can authenticate against github without an account which requires a login.

jtnord:
> Would you mind squashing those commits?

Yes I would mind; I consider it bad form (for some reasons see why [here](http://www.davidchudzicki.com/posts/first-parent/))

However it is not my repo and if you are saying this will not be merged until I squash it then I will squash it (but this does not seem o be a rule looking at the history)

srenatus:
We don't have any established, agreed-upon, strict rules; I guess it's a matter of taste mostly. (And a matter for forgetting about cleaning up the history most of the time -- at least when it comes to PRs I approve 😅)

So, fine. I'll just merge this. Thanks again 🎉

jtnord:
Thank you @srenatus

Issue #1114, opened 2017-11-06
PKCE support for public clients


srenatus:
It's intended to remedy possible attacks against public clients. See [RFC7636: Proof Key for Code Exchange by OAuth Public Clients](https://tools.ietf.org/html/rfc7636). It's a small twist to the authorization code flow:

```
                                                 +-------------------+
                                                 |   Authz Server    |
       +--------+                                | +---------------+ |
       |        |--(A)- Authorization Request ---->|               | |
       |        |       + t(code_verifier), t_m  | | Authorization | |
       |        |                                | |    Endpoint   | |
       |        |<-(B)---- Authorization Code -----|               | |
       |        |                                | +---------------+ |
       | Client |                                |                   |
       |        |                                | +---------------+ |
       |        |--(C)-- Access Token Request ---->|               | |
       |        |          + code_verifier       | |    Token      | |
       |        |                                | |   Endpoint    | |
       |        |<-(D)------ Access Token ---------|               | |
       +--------+                                | +---------------+ |
                                                 +-------------------+

                     Figure 2: Abstract Protocol Flow
```

Anyone interested in supporting this? 😃

RichardLindhout:
Recommended for public clients according to this article: https://www.ory.sh/oauth2-for-mobile-app-spa-browser/

Myself are a bit sceptic about that because most companies would not like to open a webpage via the app itself.

Teeed:
Adding link to pull request here: https://github.com/dexidp/dex/pull/1407

tacurran:
@RichardLindhout would you open an issue in ory/hydra with an alternative flow to "open a webpage via the app". We would consider some other approach compatible with your point above.

RichardLindhout:
Hey @tacurran, I'm not sure I think Hydra can handle this since you can build your own authorization logic e.g. username, password login for your own app while letting third parties do the login via the webpage.

However I think that's by definition something which could be handled by public clients since they are internal. I will make an issue because it can be an interesting discussion.

Would be nice if Hydra could handle these things more 'out of the box' but I'm not really sure.

(Hydra is really amazing I tried it 2 years ago I think but I think the documentation was not big enough because I could not get things to work)

By far the best approach best approach to oAuth I think!

RichardLindhout:
Issue -> https://github.com/ory/hydra/issues/1938

PR #1407, opened 2019-02-20
Implements PKCE (Proof Key for Code Exchange)

babariviere:
RFC 7636 (https://tools.ietf.org/html/rfc7636)


ericchiang:
Thanks! This is interesting. I'll need some time to review the RFC to review this PR.

stpabhi:
@babariviere Any plans on updating other storage options? I created a fork to implement k8s storage and others (wip) here https://github.com/stpabhi/dex/commit/6c184bbcb2ca5d0a037cb9ed592d641dbfa8311d

babariviere:
@stpabhi I plan to implement Redis storage, I may do it this week

stpabhi:
I meant etcd and k8s storage options for pkce. I don’t see the code for them in this PR.

babariviere:
Oh sorry, I have misunderstood. I will do it today

babariviere:
@stpabhi it's done if you want to use it

stpabhi:
@babariviere thanks, we have already forked dex and implemented pkce internally. I’m interested in getting it done upstream rather than branching out.

stpabhi:
I have few suggestions on the PR. 

- [ ] Add code challenge methods to discovery
- [ ] Add check for unsupported code challenge methods [here](https://tools.ietf.org/html/rfc7636#section-4.4.1)
- [ ] Add tests and update conformance test.

server/handlers.go
```diff
@@ -716,6 +721,27 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 		return
 	}
 
+	// Check for code challenge validity
+	if authCode.CodeChallenge != "" {
+		if verifier == "" {
+			s.tokenErrHelper(w, errAccessDenied, "Require code_verifier", http.StatusUnauthorized)
```
stpabhi: According to [spec](https://tools.ietf.org/html/rfc6749#section-5.2), if client doesn't pass code_verifier it would be invalid_request instead of access denied. Let me know if i'm mistaken.

server/handlers.go
```diff
@@ -716,6 +721,27 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
+			return
+		}
+		var challenge string
+		switch strings.ToLower(authCode.CodeChallengeMethod) {
```
stpabhi: client would lookup the discovery to know the supported challenge methods which are "plain" and "S256" in this case. case conversion not necessary.

server/handlers.go
```diff
@@ -716,6 +721,27 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
+		case "s256":
```
stpabhi: Could we have constants? something like [this](https://github.com/dexidp/dex/blob/master/server/oauth2.go#L124)

server/handlers.go
```diff
@@ -716,6 +721,27 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
+			sum := sha256.Sum256([]byte(verifier))
+			challenge = base64.RawURLEncoding.EncodeToString(sum[:])
+		// default to plain: insecure
+		default:
+			challenge = verifier
+		}
+		if challenge != authCode.CodeChallenge {
+			s.tokenErrHelper(w, errAccessDenied, "code_verifier doesn't have the same hash as code_challenge", http.StatusUnauthorized)
```
stpabhi: According to [spec](https://tools.ietf.org/html/rfc7636#section-4.6), it's invalid_grant instead of access_denied.

babariviere:
I will add test later

storage/sql/migrate.go
```diff
@@ -100,7 +100,10 @@ var migrations = []migration{
 				connector_id text not null,
 				connector_data bytea,
 		
-				expiry timestamptz not null
+				expiry timestamptz not null,
+
+				code_challenge text,
+				code_challenge_method text
```
srenatus: This will not work for existing databases, I think. Adding to the table needs to be another migration altogether, I suppose...

server/handlers.go
```diff
@@ -716,6 +721,30 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
+		if verifier == "" {
+			s.tokenErrHelper(w, errInvalidRequest, "Require code_verifier", http.StatusUnauthorized)
+			return
+		}
+		var challenge string
+		switch authCode.CodeChallengeMethod {
+		case codeChallengeSHA256:
+			sum := sha256.Sum256([]byte(verifier))
+			challenge = base64.RawURLEncoding.EncodeToString(sum[:])
+		// default to plain: insecure
```
srenatus: Defaulting to something insecure... Anything else we could do?

storage/sql/migrate.go
babariviere: yes you are right, I will add another statement

server/handlers.go
babariviere: It's the default in the RFC, if the client wants S256, it has to specify it. See section 4.3

Teeed:
Adding link to issue here: https://github.com/dexidp/dex/issues/1114

mackenzie-orange:
What's the status of this PR? Looks like the downstream branch got deleted :/.

babariviere:
Yes, sorry. I have to close the source as they were some code written for my company.
I will recreate it if necessary.

Teeed:
It is sad that this was not merged year (!) ago. As I can see code is still available in this PR. I have mine implementation of PKCE laying around if necessary (if this PR could not be merged for some reason).

PR #1408, opened 2019-02-22
Add logger interface and stop relying on Logrus directly

### LINKED ISSUE #868: Switch logrus uses in connectors to take an interface
Importing connectors shouldn't depend on logrus.

sagikazarmark:
Fixes #868

Before merging: I placed the interface in `pkg/log` package because that's the standard place for reusable stuff. If you want it somewhere else, let me know.

Further (possible) improvements:

- implement the whole set of log functions (currently only the used ones are implemented, adding new ones later would mean breaking the interface)
- remove `WithField` and replace it with a constructor (used in one place only). It would result in a cleaner interface which does not reference itself, thus external implementations would not have to import dex itself to provide integration


cmd/dex/config.go
```diff
@@ -6,9 +6,9 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/sirupsen/logrus"
 	"golang.org/x/crypto/bcrypt"
 
+	"github.com/dexidp/dex/pkg/log"
```
ericchiang: nit: can this be `internal/log` instead of `pkg/log`? :)

pkg/log/logrus.go
```diff
@@ -0,0 +1,45 @@
+package log
+
+import "github.com/sirupsen/logrus"
```
ericchiang: Instead of the log package importing logrus, this should just be done in dex's `main`, and passed as a logger.

pkg/log/logger.go
```diff
@@ -0,0 +1,17 @@
+// Package log provides a logger interface for logger libraries
+// so that dex does not depend on any of them directly.
+// It also includes a default implementation using Logrus (used by dex previously).
+package log
+
+// Logger serves as an adapter interface for logger libraries
+// so that dex does not depend on any of them directly.
+type Logger interface {
```
ericchiang: How many of these functions are actually used by dex? Can we limit this to only that subset?

storage/memory/static_test.go
```diff
@@ -99,11 +100,11 @@ func TestStaticClients(t *testing.T) {
 }
 
 func TestStaticPasswords(t *testing.T) {
-	logger := &logrus.Logger{
+	logger := log.NewLogrusLogger(&logrus.Logger{
```
ericchiang: I feel like `log.NewLogrusLogger` shouldn't be necessary. *logrus.Logger should just satisfy the interface. Is there something I'm missing here?

sagikazarmark:
@ericchiang let me answer here, as most of your comments are related to a single problem.

The interface I added includes the following function:

```go
WithField(key string, value interface{}) Logger
```

The same function in Logrus looks like this:

```go
WithField(key string, value interface{}) *Entry
```

The problem is the interface has a reference to itself in order to remove dependency on Logrus. Because of this:

- the interface must be importable by external packages in order to implement it, thus it cannot be an internal package
- Logrus itself cannot satisfy the interface, so we need a wrapper

That WithField function is used in a single place: https://github.com/dexidp/dex/blob/8935a1479c9b59b588d032b059709ae72dbfc44f/server/server.go#L489

It's inside the library part, that's why I added the function to the interface. But I didn't really have any heuristics about the implementation, I just created and implemented the interface.

So if dex code is not supposed to be used from external packages, the logger can be moved to an internal package (although then dex code should probably be moved to internal as well).

In case WithField method can somehow be replaced, then the interface wouldn't have to reference itself, it could be moved to an internal package and logrus would satisfy the interface without a wrapper.

pkg/log/logger.go
sagikazarmark: Currently it includes only those functions that are used by dex. However, as I mentioned in the PR description, I would consider implementing the full set. See my comment bellow.

pkg/log/logrus.go
sagikazarmark: See my comment bellow.

cmd/dex/config.go
sagikazarmark: See my comment bellow.

storage/memory/static_test.go
sagikazarmark: See my comment bellow.

ericchiang:
I vote we drop our uses of `WithField` to simplify the Logger interface.

sagikazarmark:
Well, I'm generally in favor of structured logging, but given there is a single place where it is used right now, I'm inclined to remove it as well, at least for now.

sagikazarmark:
@ericchiang I removed the wrapper type and the structural logging.

I still think if any dex code is supposed to be called externally, then the logger interface should be exported too, but if you want, I can move it under `internal/`.

storage/sql/sql.go
```diff
@@ -3,11 +3,10 @@ package sql
 
 import (
 	"database/sql"
+	"github.com/dexidp/dex/pkg/log"
```
ericchiang: nit: this should be in a separate import block from the stdlib

import (
    "database/sql"

    "github.com/dexidp/dex/pkg/log"

    // third party drivers
    // ...
)

sagikazarmark: Fixed

PR #1415, opened 2019-03-04
Add name claim to distinguish between name and username

pivotal-jwinters:
Right now there is no way of exposing a third party username through existing claims. Some providers hijack the `name` claim, while others use the `name` claim for a user's full name. By separating these into two different claims we can expose both.

For backwards compatibility all existing connectors map the name and username to the same value where applicable. Moving forward connectors can use the username claim to differentiate between the user's full name and username.

In the end we expose the `username` in the federated claims. Maybe we want to make the username a top level claim?

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -186,7 +186,8 @@ func (c *oidcConnector) HandleCallback(s connector.Scopes, r *http.Request) (ide
 	}
 
 	var claims struct {
-		Username      string `json:"name"`
+		Name          string `json:"name"`
+		Username      string `json:"username"`
 		Email         string `json:"email"`
 		EmailVerified bool   `json:"email_verified"`
 		HostedDomain  string `json:"hd"`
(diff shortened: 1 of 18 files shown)


ericchiang:
Instead of making any changes to the connectors, please just add a `name` claim and set it to the `username` value if the connector doesn't modify it. There shouldn't be any diff to the individual connectors for the first PR.

connector/oidc/oidc.go
```diff
@@ -186,7 +186,8 @@ func (c *oidcConnector) HandleCallback(s connector.Scopes, r *http.Request) (ide
 	}
 
 	var claims struct {
-		Username      string `json:"name"`
+		Name          string `json:"name"`
```
pivotal-jwinters: This internal variable name has been changed to be more consistent. This means we need to change all the connectors in order to remain backwards compatible. If we don't update the connectors they will be setting the `username` claim instead of the `name` claim.

annProg:
Any progress on this pr? I need this feature too

pivotal-jwinters:
@ericchiang I believe we need to modify all the connectors to remain backwards compatible. I left a comment inline, but that may have gotten lost.

pivotal-jwinters:
closing in favour of "preferred_username"

PR #1416, opened 2019-03-04
Use http.FileSystem for web assets

jwntrs:
This uses an `http.FileSystem` for web assets so that we can embed static assets into the dex binary. This makes it much easier to deploy. 

We use [packr](https://github.com/gobuffalo/packr), but there are many options available.


ericchiang:
Can we add a flag for os.Open or template.Load operations to use an in memory lookup instead to reduce the diff? :)

pivotal-jwinters:
Doesn't that mean we'd be introducing two separate code paths that need to be maintained? If it's necessary I don't mind doing it, just wondering what the reasoning is?

ericchiang:
> Doesn't that mean we'd be introducing two separate code paths that need to be maintained? If it's necessary I don't mind doing it, just wondering what the reasoning is?

Sorry, this diff is just really hard to read. I agree that faking the filesystem is probably the best way going forward, but there's a lot of extra changes that are confusing (like modifying the `join` method or change to `server/server.go`).

As a start, can you just replace any interaction with the file system in `loadWebConfig` to use an interface? There shouldn't be any changes to the method signatures yet. The smaller the diff is, the easier it is for me to review :)

> We use packr, but there are many options available.

I don't think we need an external package for this. It's reasonable for dex to implement itself.

xtremerui:
@bonifaido here is another legacy PR we have made. It has been updated recently to resolve conflicts and adapt changes.Is there a chance that you or one of the collaborator review this again so we could move forward?  

We know the [documentation of templates](https://github.com/dexidp/dex/blob/master/Documentation/templates.md) needs to be updated but just wanted to gather some early suggestions before that.

Thank you!

xtremerui:
@sagikazarmark Is it possible to give some directions about moving the PR forward? I have synced it again with upstream. 

Originally we have another PR for https://github.com/dexidp/dex/pull/1416/commits/1265e8660e9727faf301cce5ef2454ee9f01f986 but since it is built on top off this file system PR, I just cherry-picked the changes of `hostURL` into this PR.

Any input is appreciated. Thx!

sagikazarmark:
@xtremerui sorry for the delay. Just for you to know: this is on my radar, but I'm super busy lately, so I'm not sure when I get to reviewing this.

In general, I like the idea a lot, although using packr and embedded assets will essentially make it impossible to extend the docker image with a custom theme without recompiling the binary itself. I'd even say that it's a breaking change.

That being said, we could probably embed static resources from this repo and then look at the filesystem for other themes.

I realize this change does not actually include packr, just wanted to note the above for the future. Also, pkger seems to be the state of the art tool these days: https://github.com/markbates/pkger

I'll review this PR as soon as I can. Thanks for your patience!

tkleczek:
Current version panics when run `./bin/dex serve examples/config-dev.yaml`. I believe Server.Web.Dir is never initialized.

```
[tkleczek@thinkpad dex-http-filesystem]$ ./bin/dex serve examples/config-dev.yaml 
time="2020-09-29T19:05:10Z" level=info msg="config issuer: http://127.0.0.1:5556/dex"
time="2020-09-29T19:05:11Z" level=info msg="config storage: sqlite3"
time="2020-09-29T19:05:11Z" level=info msg="config static client: Example App"
time="2020-09-29T19:05:11Z" level=info msg="config connector: mock"
time="2020-09-29T19:05:11Z" level=info msg="config connector: local passwords enabled"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xa51f87]

goroutine 1 [running]:
github.com/dexidp/dex/server.loadTemplate(0x0, 0x0, 0x101f6e9, 0xa, 0xc00035e9d8, 0xc000490720, 0x0, 0x0, 0x0)
        /home/tkleczek/Projects/dex-http-filesystem/server/templates.go:193 +0xb7
github.com/dexidp/dex/server.loadTemplates(0x0, 0x0, 0x1022a78, 0xe, 0x1018bdb, 0x3, 0x101b1f7, 0x6, 0x0, 0x0, ...)
        /home/tkleczek/Projects/dex-http-filesystem/server/templates.go:72 +0x596
github.com/dexidp/dex/server.newServer(0x1193f40, 0xc000130040, 0xc00021f2c0, 0x19, 0x11aff40, 0xc0004904e0, 0xc00048c440, 0x1, 0x1, 0x0, ...)
        /home/tkleczek/Projects/dex-http-filesystem/server/server.go:201 +0x192
github.com/dexidp/dex/server.NewServer(0x1193f40, 0xc000130040, 0xc00021f2c0, 0x19, 0x11aff40, 0xc0004904e0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/tkleczek/Projects/dex-http-filesystem/server/server.go:172 +0xe4
main.serve(0xc0003c9180, 0xc0002d1eb0, 0x1, 0x1, 0x4, 0xcdbf70)
        /home/tkleczek/Projects/dex-http-filesystem/cmd/dex/serve.go:280 +0x1749
main.commandServe.func1(0xc0003c9180, 0xc0002d1eb0, 0x1, 0x1)
        /home/tkleczek/Projects/dex-http-filesystem/cmd/dex/serve.go:39 +0x4d
github.com/spf13/cobra.(*Command).execute(0xc0003c9180, 0xc0002d1e80, 0x1, 0x1, 0xc0003c9180, 0xc0002d1e80)
        /home/tkleczek/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:830 +0x29d
github.com/spf13/cobra.(*Command).ExecuteC(0xc0003c8f00, 0x5, 0xc00035ff78, 0x40d3df)
        /home/tkleczek/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x2fb
github.com/spf13/cobra.(*Command).Execute(...)
        /home/tkleczek/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
main.main()
        /home/tkleczek/Projects/dex-http-filesystem/cmd/dex/main.go:24 +0x2b
```

xtremerui:
I will fix the error and update the PR accordingly. Thx!

xtremerui:
@sagikazarmark the PR is updated to use pkger for embedding static contents. What do you think? Thx.

sagikazarmark:
I'm sorry this took so long. I suggest using Go 1.16 embedding for this. Let's keep this open for now, but chances are I'll rework it once Go 1.16 is out.

xtremerui:
@sagikazarmark we are also targeting the go embeded feature so we could use it in Concourse itself. Will update this PR after 1.16 is released.

sagikazarmark:
Awesome! Thanks a lot! Sorry for the long delay again.

xtremerui:
@sagikazarmark I have updated the PR to utilize go 1.16 new features. And also simplified the PR to keep only minimum changes that it needs. I have tested this PR in Concourse and the fs.FS works well. Please let me know. Thx!

server/server.go
```diff
@@ -110,6 +111,10 @@ type WebConfig struct {
 	//
 	Dir string
 
+	// A file system includes web static. Will be overwritten by Dir
```
WGH-, line 114: The current wording is somewhat confusing. How about something like this:

"Alternative way to configure web static filesystem. Dir overrides this. It's expected to contain the same files and directories as mentioned above in Dir doc."

server/templates.go
```diff
@@ -89,37 +77,32 @@ func loadWebConfig(c webConfig) (http.Handler, http.Handler, *templates, error)
 	if c.issuer == "" {
 		c.issuer = "dex"
 	}
-	if c.dir == "" {
-		c.dir = "./web"
```
WGH-, line 93: So it used to default to `"./web"`, and now it would crash if both `Dir == ""` and `WebFS == nil`?

xtremerui: Good call. It shouldn't assume one of `dir` or `webFS` is set. Will make it default to `./web` if both not set.

sagikazarmark: I think this should be resolved as high in the stack as possible and only an FS instance should be injected:

- if dir is set use it
- if not, fallback to ./web

Eventually, we should embed assets anyway and use that as a default instead of ./web

sagikazarmark (approved):
Thanks @xtremerui !

I think this is fine for now.

I still would like to see WebFS passed down as early as possible (getting rid of the directory as soon as possible), but we can do that once we embed the static assets.

PR #1423, opened 2019-04-06
Fix error output on storage list

simonswine:


server/handlers.go
```diff
@@ -224,8 +224,8 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 	}
 
 	connectors, e := s.storage.ListConnectors()
-	if e != nil {
-		s.logger.Errorf("Failed to get list of connectors: %v", err)
+	if err != nil {
```
srenatus: Should this not stay `if e != nil {`? Looks like you've fixed the bug twice 😉

simonswine: Well that's embarrassing. I have fixed it in the latest push

srenatus:
Looks like we have some issue with the Travis build that's unrelated to this change...

srenatus:
Huh! And I thought I had seen that before when merging #1443. Sorry, @simonswine, that your contribution fell off the map here. On the positive side, the correct error is printed now 🎉 😉

PR #1424, opened 2019-04-06
Fix name of json field

simonswine:
Not too sure of the impact of this, just a problem I saw navigating through the code


srenatus:
💭 If you've got an existing config stored under the (wrong) `email` key, it wouldn't be read back properly by a version of dex with this fix applied, would it? I suppose this case needs some extra handling. Not sure it's worth the hassle, though... 😉

ericchiang:
Yeah, it might be easier to just leave a comment that this is a bug, and
that it'd be a pain to fix it.

Issue #1425, opened 2019-04-16
Modify Docker image to run dex as non-root user


koushik82:
Is there a valid reason for Dex to run as a root?
```
$ kubectl -n auth exec dex-568bbc54fb-kfth2 -ti -- sh
/ # id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
```

ericchiang:
Nope :) I've gone ahead and updated the issue to reflect the needed change. Feel free to send a PR

justaugustus:
@ericchiang -- sent --> #1426 :)

PR #1426, opened 2019-04-16
Update Docker build/image

justaugustus:
- Update build container to golang:1.12.4-alpine
- Update dex image to alpine:3.9
- Run dex as non-root user

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

Fixes: #1425

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
-FROM golang:1.11.5-alpine
+FROM golang:1.12.4-alpine
 
 RUN apk add --no-cache --update alpine-sdk
 
 COPY . /go/src/github.com/dexidp/dex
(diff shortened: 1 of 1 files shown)


justaugustus:
cc: @koushik82

Makefile
```diff
@@ -55,7 +55,7 @@ lint: bin/golint
 
 .PHONY: docker-image
 docker-image:
-	@sudo docker build -t $(DOCKER_IMAGE) .
```
ericchiang: can you revert this change? docker requires (or should require) sudo on linux

justaugustus: @ericchiang -- Good point. Reverted!

PR #1431, opened 2019-04-24
storage: implements redis connector

babariviere:
This implements redis as a storage backend.
All commands are supported (tested with conformance tests + mock redis database).


ericchiang:
I'm really sorry, but dex doesn't have review bandwidth to support another storage backend.

PR #1433, opened 2019-04-24
Add option in oidc to hit the optional userinfo endpoint

jacksontj:
Some oauth providers return "thin tokens" which won't include all of the
claims requested. This simply adds an option which will make the oidc
connector use the userinfo endpoint to fetch all the claims.

as an example of an IDP that does this: https://support.okta.com/help/s/article/Okta-Groups-or-Attribute-Missing-from-Id-Token

### DIFF
--- a/Documentation/connectors/oidc.md
+++ b/Documentation/connectors/oidc.md
@@ -60,6 +60,12 @@ connectors:
     # or if they are acting as a proxy for another IDP etc AWS Cognito with an upstream SAML IDP
     # This can be overridden with the below option
     # insecureSkipEmailVerified: true 
+
+    # When enabled, the OpenID Connector will query the UserInfo endpoint for additional claims. UserInfo claims
+    # take priority over claims returned by the IDToken. This option should be used when the IDToken doesn't contain
+    # all the claims requested.
+    # https://openid.net/specs/openid-connect-core-1_0.html#UserInfo
+    # getUserInfo: true
 ```
 
 [oidc-doc]: openid-connect.md
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -39,6 +39,11 @@ type Config struct {
 
 	// Override the value of email_verifed to true in the returned claims
 	InsecureSkipEmailVerified bool `json:"insecureSkipEmailVerified"`
+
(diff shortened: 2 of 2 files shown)


jacksontj:
cc @ericchiang (as you seem to be the one working with PRs :) )

ericchiang:
Why is okta not returning username, email or email verified in their id token?

jacksontj:
I'm not sure I follow the question? This PR (and the linked docs) talk about some claims missing due to "think tokens". I have provided both a link to their documentation which explains it and code which supports this functionality. As to *why* this provider decided to do thin tokens I cannot say, but from what I understand this is allowed by spec -- so it seems reasonable to me that dex should optionally support it.

Documentation/connectors/oidc.md
```diff
@@ -60,6 +60,12 @@ connectors:
     # or if they are acting as a proxy for another IDP etc AWS Cognito with an upstream SAML IDP
     # This can be overridden with the below option
     # insecureSkipEmailVerified: true 
+
+    # Some providers return this id tokens, when the claim you requested is not in the initial token
```
ericchiang: nit: UserInfo is an OpenID Connect feature not OAuth2.
nit: no need to say this is disabled by default
nit: link to the UserInfo spec

Something like?

```
When enabled, the OpenID Connector will query the UserInfo endpoint for additional claims. UserInfo claims
take priority over claims returned by the IDToken. This option should be used when the IDToken doesn't contain
all the claims necessary to identify the user.
https://openid.net/specs/openid-connect-core-1_0.html#UserInfo
getUserInfo: true
```

connector/oidc/oidc.go
```diff
@@ -219,6 +228,18 @@ func (c *oidcConnector) HandleCallback(s connector.Scopes, r *http.Request) (ide
 
 	}
 
+	if c.getUserInfo {
+		userInfo, err := c.provider.UserInfo(r.Context(), oauth2.StaticTokenSource(token))
+		if err != nil {
+			return identity, fmt.Errorf("oidc: error loading userinfo: %v", err)
+		}
+		if err := userInfo.Claims(&claims); err != nil {
+			if err != nil {
```
ericchiang: redundant if statement

jacksontj:
@ericchiang Updated to address comments.

PR #1442, opened 2019-05-03
Added Email of Keystone to Identity

chrigl:
After the successful login to keystone, the Email of the logged in user
is fetch from keystone and provided to `identity.Email`.

This is useful for upstream software that uses the Email as the primary
identification.

### DIFF
--- a/connector/keystone/keystone.go
+++ b/connector/keystone/keystone.go
@@ -95,6 +95,14 @@ type groupsResponse struct {
 	Groups []group `json:"groups"`
 }
 
+type userResponse struct {
+	User struct {
+		Name  string `json:"name"`
+		Email string `json:"email"`
+		ID    string `json:"id"`
+	} `json:"user"`
+}
+
 var (
 	_ connector.PasswordConnector = &conn{}
 	_ connector.RefreshConnector  = &conn{}
@@ -143,6 +151,16 @@ func (p *conn) Login(ctx context.Context, scopes connector.Scopes, username, pas
 	}
 	identity.Username = username
 	identity.UserID = tokenResp.Token.User.ID
+
+	user, err := p.getUser(ctx, tokenResp.Token.User.ID, token)
+	if err != nil {
+		return identity, false, err
+	}
+	if user.User.Email != "" {
+		identity.Email = user.User.Email
+		identity.EmailVerified = true
+	}
+
 	return identity, true, nil
 }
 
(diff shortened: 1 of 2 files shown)


connector/keystone/keystone.go
```diff
@@ -215,25 +233,49 @@ func (p *conn) getAdminToken(ctx context.Context) (string, error) {
...
 func (p *conn) checkIfUserExists(ctx context.Context, userID string, token string) (bool, error) {
+	user, err := p.getUser(ctx, userID, token)
+	if err != nil {
+		return false, err
+	}
+	if user == nil {
+		return false, err
+	}
+	return true, nil
+}
```
srenatus: Nit: looks like this function is almost equivalent to
```go
func (p *conn) checkIfUserExists(ctx context.Context, userID string, token string) (bool, error) {
	user, err := p.getUser(ctx, userID, token)
		
	return user == nil, err
} 
```

connector/keystone/keystone_test.go
```diff
@@ -327,29 +387,29 @@ func TestUseRefreshTokenGroupsChanged(t *testing.T) {
...
 	addUserToGroup(t, token, groupID, userID)
+	defer deleteResource(t, token, groupID, groupsURL)
 
 	identityRefresh, err = c.Refresh(context.Background(), s, identityLogin)
 	if err != nil {
 		t.Fatal(err.Error())
 	}
 
-	delete(t, token, groupID, groupsURL)
-	delete(t, token, userID, usersURL)
```
srenatus: The user deletion doesn't seem to be put into a defer statement - oversight or intentional?

connector/keystone/keystone_test.go
```diff
@@ -363,8 +423,6 @@ func TestNoGroupsInScope(t *testing.T) {
 	}
 	expectEquals(t, 0, len(identityRefresh.Groups))
 
-	delete(t, token, groupID, groupsURL)
-	delete(t, token, userID, usersURL)
```
srenatus: Same here

connector/keystone/keystone.go
chrigl: Thanks for your Feedback!

Definitely... changing to `return user != nil, err` (`!=` instead of `==`).

connector/keystone/keystone_test.go
chrigl: Defer went to L370, right after creating the user.

connector/keystone/keystone_test.go
chrigl: lives on L404 now.

connector/keystone/keystone.go
```diff
@@ -215,25 +233,43 @@ func (p *conn) getAdminToken(ctx context.Context) (string, error) {
...
+	}
+
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
 	}
 
-	if resp.StatusCode == 200 {
-		return true, nil
+	user := new(userResponse)
```
srenatus: Nit: can we do `user := userResponse{}` and pass `&user` to unmarshal? It's trivial but a bit more consistent with the rest of the connector code, I think.

chrigl:
I also can't see a scenario where this must be configurable. If someone wants to ignore the email value, one can still ignore it. I think another roundtrip to keystone does not hurt at all. Fewer code to maintain and test, the better it is.

jrperritt:
@srenatus Is this still being considered for merge?

kenperkins:
We've (Rackspace) discovered a case where https://github.com/pusher/oauth2_proxy > 3.2.0 requires an email to be returned from the identity provider. We were forced to upgrade from 3.2.0 due to a [security vulnerability](https://github.com/pusher/oauth2_proxy/security/advisories/GHSA-qqxw-m5fj-f7gv) in https://github.com/pusher/oauth2_proxy. This required us to take this patch and fork for functionality with Keystone. 

Are we able to re-consider and merge/release with this capability?

bonifaido:
Yes, of course, please rebase on latest master @kenperkins

kenperkins:
@chrigl are you able to rebase this? if not, I will pick it up and address this.

kenperkins:
This has been rebased, and pushed as PR #1681.

srenatus:
✔️ #1681

PR #1444, opened 2019-05-06
Create a non-priviledged user in runtime container

deric:
@justaugustus introduced (56f02b95c6490981602dbf8726a5c4a0e3d858dc) running dex as a non-root user, which is cool but such user doesn't exist inside Alpine container.

This PR makes sure user `1001:1001` exists:
```
dex:x:1001:1001:Linux User,,,:/:/bin/ash
```
other option would be to leverage some existing account, like:
```
guest:x:405:100:guest:/dev/null:/sbin/nologin
nobody:x:65534:65534:nobody:/:/sbin/nologin
```
unlike `root`
```
root:x:0:0:root:/root:/bin/ash
```
these accounts doesn't have shell.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,9 @@ FROM alpine:3.9
 # experience when this doesn't work out of the box.
 #
 # OpenSSL is required so wget can query HTTPS endpoints for health checking.
-RUN apk add --update ca-certificates openssl
+RUN apk add --update ca-certificates openssl\
+  && addgroup --gid 1001 dex\
+  && adduser --disabled-password --no-create-home --home "$(pwd)" -u 1001 -G dex dex
 
 USER 1001:1001
 COPY --from=0 /go/bin/dex /usr/local/bin/dex


ericchiang:
Does it matter that the user doesn't exist in the container? What's the actual impact here?

deric:
When you use mounted volumes (e.g. with a certificate), it does matter. Linux will refuse access to files owned by an unknown UID/GID.

ericchiang:
Ah thanks for that! We don't have any test coverage for the image currently. What have you done to test this change? I can check it out later if you've not.

deric:
You're right, it doesn't have much impact. The only "problem" is that `whoami` fails:
```
$ docker run --entrypoint /usr/bin/whoami -it dex:latest
whoami: unknown uid 1001
```
When you're mounting files that aren't readable by the whole world you'd either have to:
* run e.g. `chown -R 1001:1001 certs` as UID, GID are mapped from host system
* run container with different `--user 1000:1000` that would override `USER` directive in current `Dockerfile`.

```
mkdir certs && touch certs/priv.key && chmod 0600 certs/priv.key
docker run --entrypoint /bin/ash -v $(pwd)/certs:/certs:ro --user $UID -it dex:latest
cat /certs/priv.key
```
So, this change won't improve much, except having readable username.

Dockerfile
```diff
@@ -11,7 +11,9 @@ FROM alpine:3.9
 # experience when this doesn't work out of the box.
 #
 # OpenSSL is required so wget can query HTTPS endpoints for health checking.
-RUN apk add --update ca-certificates openssl
+RUN apk add --update ca-certificates openssl\
+  && addgroup --gid 1001 dex\
+  && adduser --disabled-password --no-create-home --home "$(pwd)" -u 1001 -G dex dex
```
ericchiang: nit: dex doesn't need a home directory, why provide `--home "$(pwd)"`?

deric: It will set home dir to `/`. Even users with `nologin` have some home directories:
```
daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
bin:x:2:2:bin:/bin:/usr/sbin/nologin
sys:x:3:3:sys:/dev:/usr/sbin/nologin
```
AFAIK if the `--home` argument is not provided it will set home to `/home/dex` which will be non-existing directory. Setting home to `$(pwd)` ensures that the directory exists.

justaugustus:
What's the status on this?

PR #1446, opened 2019-05-08
microsoft: option for group UUIDs instead of name and group whitelist

maksd:
This PR add a `useGroupUUID` option to return groups UUIDs instead of names. And, it change the behavior of the `groups` option to not only allow user member of a particular group but also act as a whitelist for the groups return in the claim.

The main reason for using group UUIDs, is that group names are not unique in Microsoft Azure AD.

### DIFF
--- a/connector/microsoft/microsoft.go
+++ b/connector/microsoft/microsoft.go
@@ -19,40 +19,64 @@ import (
 	"github.com/dexidp/dex/pkg/log"
 )
 
+// GroupNameFormat represents the format of the group identifier
+// we use type of string instead of int because it's easier to
+// marshall/unmarshall
+type GroupNameFormat string
+
+// Possible values for GroupNameFormat
+const (
+	GroupID   GroupNameFormat = "id"
+	GroupName GroupNameFormat = "name"
+)
+
 const (
 	apiURL = "https://graph.microsoft.com"
 	// Microsoft requires this scope to access user's profile
 	scopeUser = "user.read"
 	// Microsoft requires this scope to list groups the user is a member of
-	// and resolve their UUIDs to groups names.
+	// and resolve their ids to groups names.
 	scopeGroups = "directory.read.all"
 )
 
 // Config holds configuration options for microsoft logins.
 type Config struct {
-	ClientID           string   `json:"clientID"`
-	ClientSecret       string   `json:"clientSecret"`
-	RedirectURI        string   `json:"redirectURI"`
-	Tenant             string   `json:"tenant"`
-	OnlySecurityGroups bool     `json:"onlySecurityGroups"`
-	Groups             []string `json:"groups"`
+	ClientID             string          `json:"clientID"`
+	ClientSecret         string          `json:"clientSecret"`
+	RedirectURI          string          `json:"redirectURI"`
+	Tenant               string          `json:"tenant"`
+	OnlySecurityGroups   bool            `json:"onlySecurityGroups"`
+	Groups               []string        `json:"groups"`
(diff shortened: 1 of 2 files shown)


ericchiang:
Can we make this an enum instead of a boolean? Something like:

```
groupNameFormat: id # defaults to name
```

Also are these actually UUIDs? (https://en.wikipedia.org/wiki/Universally_unique_identifier) That's a pretty specific format and I only see "id" referenced in the API 😃

maksd:
@ericchiang Yes I can change it for an enum. About the "uuids", I took it from the Microsoft Graph documentation for the [directoryObject resource type](https://docs.microsoft.com/en-us/graph/api/resources/directoryobject?view=graph-rest-1.0#properties) :

Property | Type | Description
-- | -- | --
id | String | A Guid that is the unique  identifier for the object; for example,  12345678-9abc-def0-1234-56789abcde. Key. Not nullable. Read-only.

connector/microsoft/microsoft.go
```diff
@@ -54,6 +67,13 @@ func (c *Config) Open(id string, logger log.Logger) (connector.Connector, error)
 		m.tenant = "common"
 	}
 
+	// By default, use group names
+	if m.groupNameFormat == "" {
```
ericchiang: nit: just use a switch

```go
switch m.groupNameFormat {
case "":
    m.groupNameFormat = GroupName
case GroupID, GroupName:
default:
    return nil, fmt.Errorf("invalid groupNameFormat")
}
```

connector/microsoft/microsoft.go
```diff
@@ -299,37 +320,40 @@ type group struct {
...
 			}
 		}
-	}
-	if len(c.groups) > 0 && !isInGroups {
-		return nil, fmt.Errorf("microsoft: user %v not in required groups", userID)
+
+		if len(filteredGroups) == 0 {
+			return nil, fmt.Errorf("microsoft: user %v not in required groups", userID)
+		}
+		return filteredGroups, nil
```
ericchiang: Returning a subset of groups is a breaking change. Please revert this logic or use a flag.

connector/microsoft/microsoft.go
maksd: Thanks for your feedback, fixed it

connector/microsoft/microsoft.go
maksd: 🤦‍♂ 
thanks, I added a flag `useGroupsAsWhitelist`

srenatus:
Could you rebase this once more, please? I had merged a change refactoring the group filtering into a helper method that affected this PR, sorry. 😃

maksd:
Hi @srenatus, rebase done.

connector/microsoft/microsoft.go
```diff
@@ -19,42 +19,66 @@ import (
...
 		m.tenant = "common"
 	}
 
+	// By default, use group names
+	switch m.groupNameFormat {
+	case "":
+		m.groupNameFormat = GroupName
+	case GroupID, GroupName:
+	default:
+		return nil, fmt.Errorf("invalid groupNameFormat")
```
srenatus: [nit] Let's either do 
```suggestion
		return nil, errors.New("invalid groupNameFormat")
```
or
```suggestion
		return nil, fmt.Errorf("invalid groupNameFormat: %s", m.groupNameFormat)
```

maksd:
@srenatus `nit` fixed, thanks for the catch. About the tests, I would be happy to do it, but it will have to wait a bit, maybe in another PR.

PR #1448, opened 2019-05-15
oidc: Make userID configurable

cappyzawa:
I make userID configurable.

The claims are changed `map[string]interface{}` from struct because these vary depending on the OIDC provider.

And userID is made configurable.
There are cases where users want to use other than sub-claim as userID.
(related https://github.com/concourse/dex/pull/8)

Thanks!

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -44,6 +44,9 @@ type Config struct {
 	// the token. This is especially useful where upstreams return "thin"
 	// id tokens
 	GetUserInfo bool `json:"getUserInfo"`
+
+	// Configurable key which contains the user id claim
+	UserIDKey string `json:"userIDKey"`
 }
 
 // Domains that don't support basic auth. golang.org/x/oauth2 has an internal
@@ -127,6 +130,7 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 		hostedDomains:             c.HostedDomains,
 		insecureSkipEmailVerified: c.InsecureSkipEmailVerified,
 		getUserInfo:               c.GetUserInfo,
+		userIDKey:                 c.UserIDKey,
 	}, nil
 }
 
@@ -146,6 +150,7 @@ type oidcConnector struct {
 	hostedDomains             []string
 	insecureSkipEmailVerified bool
 	getUserInfo               bool
+	userIDKey                 string
 }
 
 func (c *oidcConnector) Close() error {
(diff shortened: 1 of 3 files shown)


cappyzawa:
@srenatus Could you review, please 🙏 (if you have time)

connector/oidc/oidc.go
```diff
@@ -190,41 +195,49 @@ func (c *oidcConnector) HandleCallback(s connector.Scopes, r *http.Request) (ide
...
-		Email         string `json:"email"`
-		EmailVerified bool   `json:"email_verified"`
-		HostedDomain  string `json:"hd"`
-	}
+	var claims map[string]interface{}
 	if err := idToken.Claims(&claims); err != nil {
 		return identity, fmt.Errorf("oidc: failed to decode claims: %v", err)
 	}
 
+	name := claims["name"].(string)
```
srenatus: ❓ Does this have the same error semantics as the struct before? I.e. should the `idToken` _not_ have an `email` key, this would explode -- `claims["email"]` would be nil, `.(string)` would fail. What used to happen with `idToken.Claims(&claims)`? (Genuine question, I haven't checked...)

cappyzawa: Thanks for your feedback!
I fixed it:+1: I added error handlings for missing specified claims.
https://github.com/dexidp/dex/pull/1448/commits/601fa713aa8617640cea21ed92052c023d5ca72d

srenatus:
ℹ️ Please don't merge, unless we're OK with it as-is -- I'm not done reproducing the original issue, and verifying that this solves it. (In the narrowest scope, at least, this will make `openssl_1.1.1` an accepted version.)

cappyzawa:
@srenatus 
Sorry, I accidentally bumped `go-jose` :bow:
I deleted bumped commit

srenatus:
Thanks! I'll review this again tomorrow :)

connector/oidc/oidc.go
```diff
@@ -39,6 +39,9 @@ type Config struct {
 
 	// Override the value of email_verifed to true in the returned claims
 	InsecureSkipEmailVerified bool `json:"insecureSkipEmailVerified"`
+
+	// Configurable key which contains the user id claims
```
srenatus: [nit] It's just one claim, isn't it? 
```suggestion
	// Configurable key which contains the user id claim
```

connector/oidc/oidc.go
```diff
@@ -190,41 +195,58 @@ func (c *oidcConnector) HandleCallback(s connector.Scopes, r *http.Request) (ide
...
-			return identity, fmt.Errorf("oidc: unexpected hd claim %v", claims.HostedDomain)
+			return identity, fmt.Errorf("oidc: unexpected hd claim %v", hostedDomain)
 		}
 	}
 
 	if c.insecureSkipEmailVerified {
-		claims.EmailVerified = true
-
+		emailVerified = true
 	}
```
srenatus: [nit] That's the same as
```go
emailVerified = c.insecureSkipEmailVerified
```
but it really doesn't matter :smile:

connector/oidc/oidc.go
```diff
@@ -190,41 +195,58 @@ func (c *oidcConnector) HandleCallback(s connector.Scopes, r *http.Request) (ide
...
+	email, found := claims["email"].(string)
+	if !found {
+		return identity, errors.New("missing \"email\" claim")
+	}
+	emailVerified, found := claims["email_verified"].(bool)
+	if !found {
+		return identity, errors.New("missing \"email_verified\" claim")
+	}
+	hostedDomain, _ := claims["hd"].(string)
+
```
srenatus: 💭 So, if there's no `hd` claim and `c.hostedDomains` is not configured, it won't matter. If there's a no `hd` claim and `c.hostedDomains` _is configured_, the following check will fail. So, yeah, looks like we don't need to check `found` here. 👍

connector/oidc/oidc.go
cappyzawa: Thanks for your suggestion 👍 https://github.com/dexidp/dex/pull/1448/commits/5e68d43af2aeb4802105937f9102bfcd52ff8f24

cappyzawa:
@srenatus squashed 👍

srenatus:
https://github.com/dexidp/dex/pull/1448/commits/7f41bdaf89f96df3971b3488cb6536dc90b4e5fb What's up with this commit? 🤔

cappyzawa:
I try to resolve conflict.

srenatus:
Ah, I see. Typically you would rebase your PR branch, like this:

- `git checkout YOURBRANCH`
- `git fetch origin`
- `git rebase origin/master`
- `git push -f YOURORIGIN YOURBANCH` (⚠️ force push, take care)

cappyzawa:
@srenatus Thanks!!

srenatus:
@cappyzawa Would you mind adding a short note into the connector docs configuration section? https://github.com/dexidp/dex/blob/master/Documentation/connectors/oidc.md some example setting would be enough 😃 -- sorry for coming up with this in the last minute.

cappyzawa:
@srenatus Added 👍

srenatus:
Thank you very much! :tada: :ballon:

cappyzawa:
@srenatus Thank you! 👍

PR #1453, opened 2019-05-27
Add userinfo_endpoint

mdbraber:
Based on https://github.com/dexidp/dex/pull/1201 from @jackielii - redid some of the code to fit with the current codebase. Works for me now with traefik-forward-auth.

PR #1201, opened 2018-03-08
A simple implementation of userinfo endpoint

jackielii:
A stop gap to solve #376. Borrowed mostly from #1133.

The way it works is we create a signed jwt token as access_token (same claims as id token), on requesting /userinfo, it verifies the token and check it's not expired etc, and return the claims in json

IMHO, the full solution should involve the storage to implement similar logic for refresh_token. i.e. serialise access token and save in storage and retrieve upon requesting user info. 

Because access token is opaque to the user, we can later roll in the proper solution without breaking API.

PR #1454, opened 2019-05-27
Add userinfo_endpoint

mdbraber:
Supersedes https://github.com/dexidp/dex/pull/1453

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"encoding/json"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"net/http"
@@ -20,6 +21,10 @@ import (
 	"github.com/dexidp/dex/connector"
 	"github.com/dexidp/dex/server/internal"
 	"github.com/dexidp/dex/storage"
+)
+
+var (
+	errTokenExpired = errors.New("token has expired")
 )
 
 // newHealthChecker returns the healthz handler. The handler runs until the
@@ -151,6 +156,7 @@ type discovery struct {
 	Auth          string   `json:"authorization_endpoint"`
 	Token         string   `json:"token_endpoint"`
 	Keys          string   `json:"jwks_uri"`
+	UserInfo      string   `json:"userinfo_endpoint"`
 	ResponseTypes []string `json:"response_types_supported"`
 	Subjects      []string `json:"subject_types_supported"`
 	IDTokenAlgs   []string `json:"id_token_signing_alg_values_supported"`
@@ -165,6 +171,7 @@ func (s *Server) discoveryHandler() (http.HandlerFunc, error) {
 		Auth:        s.absURL("/auth"),
 		Token:       s.absURL("/token"),
 		Keys:        s.absURL("/keys"),
+		Keys:        s.absURL("/userinfo"),
 		Subjects:    []string{"public"},
 		IDTokenAlgs: []string{string(jose.RS256)},
 		Scopes:      []string{"openid", "email", "groups", "profile", "offline_access"},
@@ -559,7 +566,12 @@ func (s *Server) sendCodeResponse(w http.ResponseWriter, r *http.Request, authRe
 		idToken       string
 		idTokenExpiry time.Time
 
-		accessToken = storage.NewID()
+i		accessToken, err := s.newAccessToken(client.ID, authCode.Claims, authCode.Scopes, authCode.Nonce, authCode.ConnectorID)
+		if err != nil {
+			s.logger.Errorf("failed to create new access token: %v", err)
+			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
+			return
+		}
 	)
 
 	for _, responseType := range authReq.ResponseTypes {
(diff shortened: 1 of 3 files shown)


mdbraber:
@srenatus would be great if you could take a look at this (build checks out)!

srenatus:
I'll try to get this in this week 🤞

srenatus:
Sorry about that. _This week_ it is, then. 😅

server/oauth2.go
```diff
@@ -265,6 +265,11 @@ type federatedIDClaims struct {
 	UserID      string `json:"user_id,omitempty"`
 }
 
+func (s *Server) newAccessToken(clientID string, claims storage.Claims, scopes []string, nonce, connID string) (accessToken string, err error) {
+	idToken, _, err := s.newIDToken(clientID, claims, scopes, nonce, storage.NewID(), connID)
```
srenatus: Hmm! So we'll be using a JWT for `access_token`... but it's a different one; we don't hand out the same we give as `id_token`. 💭

server/handlers.go
```diff
@@ -1026,6 +1054,88 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
...
+	if keys.SigningKey == nil {
+		return nil, fmt.Errorf("no private keys found")
+	}
+
+	object, err := jose.ParseSigned(token)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse signed message")
+	}
+
+	// Parse the message to check expiry, as it jose doesn't distinguish expiry error from others
```
srenatus: I don't understand how the comment relates to the code below. 🤔

server/handlers.go
```diff
@@ -1026,6 +1054,88 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
...
+	parts := strings.Split(token, ".")
+	if len(parts) != 3 {
+		return nil, fmt.Errorf("square/go-jose: compact JWS format must have three parts")
```
srenatus: Also, we should probably not claim we're `square/go-jose`. We aren't 😄

mdbraber:
> This is looking good so far -- _thank you!_

Thanks for taking the time to review!
 
> * Would you mind throwing in a test or three for the new HTTP handler, please?

I'm not proficient in writing any sort of tests so that'll take me a lot of figuring out, I'm happy to try if someone might have some pointers how to do this.

> * @jackielii: Are you OK with @mdbraber adopting your PR? @mdbraber could you add a `Co-authored-by` line with @jackielii on the relevant commits, please? (Or massage the history to include the commit of #1201?)

Well I think @jackielii did most of the work as it's their patch. I only butchered the patch to make it fit the current code base :-) Happy to add a Co-authored line on the relevant commit. Massaging the history seems more work so I'll just do the first. 

> * Oh this has quite some history.... @fjbsantiago had a proposal up, too, in #1133. I suppose we'll better put a bunch of hat-tips into the merge commit 🤔

So it seems, so more hat tips for everyone :-) 🎩

server/handlers.go
mdbraber: Well that seems an easy one to fix

server/handlers.go
mdbraber: This might be me - doing some too agressive butchering of the code. Sorry.

server/oauth2.go
mdbraber: Maybe @jackielii can answer this one?

jackielii: Like in my original PR comment: "this is a stop gap solution". In my understanding, access_token is a "key" that should be saved away and later used to retrieve "user_info". Here I'm abusing the ID token signature verification: Use an ID token as the access token, and use signature verification to verify the JWT and just return the claims as the user_info response.

Because the access_token is opaque to the user, so this change is forward compatible with the "proper solution". I even created this function to later swap out for the proper "save to db" solution.

jackielii:
@mdbraber & @srenatus absolutely. Feel free to use my name or otherwise, no probs for me. I was lazy and looking for a solution hastily, and just figured this trick should be the easiest one without changing too much. It utilises existing id_token signature verification mechanism to implement the access_token + user_info verification. It has been in semi-production in our internal system for since I submitted the other PR over a year ago.

BTW, I realised on that my PR wasn't similar to the original #1133 . That one I still couldn't fully understand, and mine definitely took a different approach. But it helped to understand the code base a bit better.

Thanks & Regards,

server/handlers.go
```diff
@@ -1026,6 +1054,88 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
...
+	}
+
+	payload, err := base64.RawURLEncoding.DecodeString(parts[1])
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: check other claims
```
jackielii: This could be a "big" TODO. I realised later on that the jose or any other JWT token lib should do all the checks, but I'm not sure they do all the needed. Or maybe they do, and I'm just being over paranoid here to check again. Need a JWT expert or jose expert here to answer the question.

server/oauth2.go
jackielii: @srenatus sorry forgot to answer your original question. Within one auth request, the id_token won't be exactly the same as the access_token, albeit access_token looks and "can" behave like a jwt id_token. We're just abusing the fact that access_token is just a "random string".

mdbraber:
Added the right attribution: added Co-authored-by to @jackielii and @fjbsantiago acknowledging their work.

server/oauth2.go
srenatus: Quite clever! And thanks for considering _"forward compat with a proper solution"_ 😄

srenatus:
> I'm not proficient in writing any sort of tests so that'll take me a lot of figuring out, I'm happy to try if someone might have some pointers how to do this.

https://github.com/dexidp/dex/blob/master/server/server_test.go has some tests that could be added to or expanded, I suppose. LMK if you need help. 😃

alindeman:
How's this coming along? Is there anything I can do to help? I'd like to see this land in Dex so I can better integrate it with services like AWS ALBs that require userinfo endpoints.

srenatus:
@alindeman This needs tests. If you'd like to contribute, you could PR against `mdbraber:add-user-endpoint` or a open a new one against `dexidp/dex` with these commits included. (We can then merge this PR first and rebase the other one then.)

alindeman:
> @alindeman This needs tests. If you'd like to contribute, you could PR against `mdbraber:add-user-endpoint` or a open a new one against `dexidp/dex` with these commits included. (We can then merge this PR first and rebase the other one then.)

Sounds good. I will plan to tackle adding some tests tomorrow.

PR #1460, opened 2019-05-30
Update LinkedIn connector to use v2 APIs

### LINKED ISSUE #1452: LinkedIn connector fails with v1 APIs
The LinkedIn connector fails with an error like this:

```
time="...." level=error msg="Failed to authenticate: linkedin: get profile: 410 Gone: {
  \"errorCode\": 0,
  \"message\": \"This resource is no longer available under v1 APIs\",
  \"requestId\": \"....\",
  \"status\": 410,
  \"timestamp\": ....
}"
```

Looks like it needs to be updated to use v2 APIs, maybe these?
- https://docs.microsoft.com/en-us/linkedin/shared/integrations/people/profile-api?context=linkedin/compliance/context#retrieve-other-members-profile
- https://docs.microsoft.com/en-us/linkedin/shared/integrations/people/primary-contact-api?context=linkedin/compliance/context#retrieve-email-address
- https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/migration-faq#how-do-i-retrieve-the-members-email-address

tanmaykm:
This updates LinkedIn connector to use the more recent v2 APIs. Necessary because v1 APIs are not able to retrieve email ids any more with the default permissions.

The API URLs are now different. Fetching the email address is now a separate call, made after fetching the profile details. The `r_basicprofile` permission is not needed any more, and `r_liteprofile` (which seems to be the one assigned by default) is sufficient.

The relevant API specifications are at:
- https://docs.microsoft.com/en-us/linkedin/shared/integrations/people/profile-api
- https://docs.microsoft.com/en-us/linkedin/shared/integrations/people/primary-contact-api
- https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/migration-faq#how-do-i-retrieve-the-members-email-address

fixes #1452

### DIFF
--- a/connector/linkedin/linkedin.go
+++ b/connector/linkedin/linkedin.go
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	apiURL   = "https://api.linkedin.com/v1"
+	apiURL   = "https://api.linkedin.com/v2"
 	authURL  = "https://www.linkedin.com/oauth/v2/authorization"
 	tokenURL = "https://www.linkedin.com/oauth/v2/accessToken"
 )
@@ -38,7 +38,7 @@ func (c *Config) Open(id string, logger log.Logger) (connector.Connector, error)
 				AuthURL:  authURL,
 				TokenURL: tokenURL,
 			},
-			Scopes:      []string{"r_basicprofile", "r_emailaddress"},
+			Scopes:      []string{"r_liteprofile", "r_emailaddress"},
 			RedirectURL: c.RedirectURI,
 		},
 		logger: logger,
@@ -133,9 +133,17 @@ func (c *linkedInConnector) Refresh(ctx context.Context, s connector.Scopes, ide
 
 type profile struct {
 	ID        string `json:"id"`
-	FirstName string `json:"firstName"`
-	LastName  string `json:"lastName"`
+	FirstName string `json:"localizedFirstName"`
+	LastName  string `json:"localizedLastName"`
 	Email     string `json:"emailAddress"`
+}
+
+type emailresp struct {
+	Elements []struct {
+		Handle struct {
+			EmailAddress string `json:"emailAddress"`
+		} `json:"handle~"`
+	} `json:"elements"`
 }
 
 // fullname returns a full name of a person, or email if the resulting name is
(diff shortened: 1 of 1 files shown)


connector/linkedin/linkedin.go
```diff
@@ -149,15 +149,52 @@ func (p profile) fullname() string {
...
+	if err != nil {
+		return email, fmt.Errorf("read body: %v", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return email, fmt.Errorf("%s: %s", resp.Status, body)
+	}
+
+	var emailResp map[string]interface{}
+	json.Unmarshal(body, &emailResp)
```
srenatus: This needs error checking, I think.

connector/linkedin/linkedin.go
```diff
@@ -149,15 +149,52 @@ func (p profile) fullname() string {
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return email, fmt.Errorf("read body: %v", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return email, fmt.Errorf("%s: %s", resp.Status, body)
+	}
+
+	var emailResp map[string]interface{}
```
srenatus: I wonder if this was a bit simpler with a nested struct?

tanmaykm:
Thanks @srenatus for checking. I have added a commit to unmarshal into a nested struct. Hope that addresses both your comments?

connector/linkedin/linkedin.go
```diff
@@ -149,15 +157,46 @@ func (p profile) fullname() string {
 	return fname
 }
 
+func (c *linkedInConnector) primaryEmail(ctx context.Context, client *http.Client) (email string, err error) {
```
srenatus: I'd prefer non-named return values, but that's mostly a matter of taste and not consistent even within this code base. So, let's leave it as is, since we're not using _naked returns_, there's little harm done. 😉

connector/linkedin/linkedin.go
```diff
@@ -149,15 +157,46 @@ func (p profile) fullname() string {
...
+		return email, fmt.Errorf("%s: %s", resp.Status, body)
+	}
+
+	var parsedResp emailresp
+	err = json.Unmarshal(body, &parsedResp)
+	if err == nil {
+		for _, elem := range parsedResp.Elements {
+			email = elem.Handle.EmailAddress
+		}
+	}
```
srenatus: ❓ Is this validation good enough? From glancing at the API docs (thanks for including the links, much appreciated!), [`handle~` could be missing](https://docs.microsoft.com/en-us/linkedin/shared/integrations/people/primary-contact-api#sample-decoration-error) because something (no idea what!) went wrong...?

With the current code, that would mean `email` remains `""` (unassigned) and `err` would still be `nil` because of the way JSON unmarshaling works. (Its errors aren't for cases where expected fields aren't found, but rather more low-level stuff. If you try to unmarshal `{this:"garba"ge}}` for example.)

💡 looks like we'd catch that below, [in this check](https://github.com/dexidp/dex/pull/1460/files#diff-306cce14d2c51d1b284c11a81280c3acR225) -- I think I'd propose moving that into this helper method: "if `primaryEmail` returns a non-nil error, it has successfully retrieved the email address" sounds like a decent property for a function with that name. What do you think?

connector/linkedin/linkedin.go
tanmaykm: Okay.

connector/linkedin/linkedin.go
tanmaykm: Yes, agree. Will make that change in a bit.

srenatus:
@tanmaykm thank you! would you mind squashing the commits? 😃

tanmaykm:
Yup, done!

PR #1464, opened 2019-06-07
Fix typo in JSON tag name

alindeman:


srenatus:
This came up before, but I don't have the PR number handy. Changing this means we've got to deal with with _old key_ stored in existing configs. So, either we fix the typo and deal with this issue; or keep it as-is, and add a comment to keep future code readers from fixing it again 😅 

Either approach would be appreciated, but just fixing the tag is trouble, I'm afraid.

srenatus:
@alindeman do you feel like taking up one of the two options? 😉

alindeman:
Thanks for the clarification. I'm feeling like this is probably not worth the effort at the current moment. Closing for now.

PR #1467, opened 2019-06-12
Allowed connectors for clients

zetaab:
We have currently use-case that we have multiple connectors (and multiple clients) configured to dex. However, we want limit which connectors are active in each client. The default behaviour is not changed in this PR, if connectors are not defined all connectors are listed.


mmrath:
Client has allowed connectors and also auth request. So the allowed connecters are configured per auth request or per client? if per auth request then we don't need it in client?
If it is per client storage api should be implemented to store the connectors per client?

PR #1473, opened 2019-06-20
Add UserInfo endpoint

### LINKED ISSUE #376: UserInfo endpoint
I am currently trying to get dex working in a ruby environment. On the RP side I am currently working with [omniauth-openid-connect](https://github.com/jjbohn/omniauth-openid-connect), which has some issues of its own.

However: Working around some of those initial issues, I am now stuck with omniauth-openid-connect trying to get additional claims from the userinfo endpoint, which is [not implemented by dex](https://github.com/coreos/dex/blob/master/Documentation/oidc-notes.md).

In reading the [OIDC specification](http://openid.net/specs/openid-connect-core-1_0.html#UserInfo), I failed to recognize whether this is a mandatory endpoint or an optional endpoint, but since it appears right away in section 1.3 of the spec, I would _assume_ it is mandatory.

Are there any plans on implementing this endpoint?

If not, I _might_ start working on it, but that depends on a number of other factors and it would be my first larger golang contribution anywhere, not to mention that I have barely read enough of the specs...

alindeman:
This PR takes the great work in #1454, reworks it a bit to lean on `oidc.Verifier` for verification, and adds tests.

Fixes #376
Closes #1133
Closes #1201
Closes #1453
Closes #1454


srenatus:
I'll review this next week. Thanks for contributing! :smiley: :tada:

server/handlers.go
```diff
@@ -1055,84 +1051,30 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
...
 
-	if len(parts) != 2 || !strings.EqualFold(parts[0], "bearer") {
-		msg := "invalid authorization header"
-		w.Header().Set("WWW-Authenticate", fmt.Sprintf(`Bearer realm="dex", error="%s", error_description="%s"`, errInvalidRequest, msg))
-		s.tokenErrHelper(w, errInvalidRequest, msg, http.StatusBadRequest)
-		return
+	auth := r.Header.Get("authorization")
+	if len(auth) < len(prefix) || !strings.EqualFold(prefix, auth[:len(prefix)]) {
+		w.Header().Set("WWW-Authenticate", "Bearer")
+		s.tokenErrHelper(w, errAccessDenied, "Invalid bearer token.", http.StatusUnauthorized)
```
srenatus: ❓ This will need a `return`, or otherwise the processing is going to continue on.

alindeman: Thanks, fixed.

server/handlers.go
```diff
@@ -1026,6 +1050,34 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
...
+
+	auth := r.Header.Get("authorization")
+	if len(auth) < len(prefix) || !strings.EqualFold(prefix, auth[:len(prefix)]) {
+		w.Header().Set("WWW-Authenticate", "Bearer")
+		s.tokenErrHelper(w, errAccessDenied, "Invalid bearer token.", http.StatusUnauthorized)
+		return
+	}
+	rawIDToken := auth[len(prefix):]
+
+	verifier := oidc.NewVerifier(s.issuerURL.String(), &storageKeySet{s.storage}, &oidc.Config{SkipClientIDCheck: true})
```
srenatus: Is this a costly operation? (I.e., do we want to store and re-use `oidc.NewVerifier`? (I have no idea, genuine question.)

server/handlers.go
```diff
@@ -1026,6 +1050,34 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
...
+		s.tokenErrHelper(w, errServerError, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(claims)
+}
+
 func (s *Server) writeAccessToken(w http.ResponseWriter, idToken, accessToken, refreshToken string, expiry time.Time) {
 	// TODO(ericchiang): figure out an access token story and support the user info
```
srenatus: This comment could use some updating now, I suppose? 😉

server/oauth2.go
```diff
@@ -561,3 +567,41 @@ func validateRedirectURI(client storage.Client, redirectURI string) bool {
 	host, _, err := net.SplitHostPort(u.Host)
 	return err == nil && host == "localhost"
 }
+
+// storageKeySet implements the oidc.KeySet interface backed by Dex storage
+type storageKeySet struct {
+	storage.Storage
+}
+
+func (s *storageKeySet) VerifySignature(ctx context.Context, jwt string) (payload []byte, err error) {
```
srenatus: [nit] `ctx` isn't used:
```suggestion
func (s *storageKeySet) VerifySignature(_ context.Context, jwt string) (payload []byte, err error) {
```

server/server_test.go
```diff
@@ -201,6 +200,16 @@ func TestOAuth2CodeFlow(t *testing.T) {
 				return nil
 			},
 		},
+		{
+			name: "fetch userinfo",
+			handleToken: func(ctx context.Context, p *oidc.Provider, config *oauth2.Config, token *oauth2.Token) error {
+				_, err := p.UserInfo(ctx, config.TokenSource(ctx, token))
```
srenatus: ❓ Would it be worthwhile to assert something on the response?

server/handlers.go
alindeman: It's not, as far as I can tell. The implementation is currently:

```go
func NewVerifier(issuerURL string, keySet KeySet, config *Config) *IDTokenVerifier {
	return &IDTokenVerifier{keySet: keySet, config: config, issuer: issuerURL}
}
```

Somewhat notably, constructing an `oidc.Provider` is a relatively expensive operation, but we don't have a need to do that in non-test code.

server/server_test.go
alindeman: I went back and forth on it. Not many of the other tests do, but since it ended up being easy enough, I've added something in 840065f. Let me know what you think.

server/handlers.go
alindeman: Good catch, I've deleted it.

alindeman:
> 🎉 It's nice to see this come together, thanks for contributing. Some nitpicks/questions inside 😃

@srenatus Thanks for the review. Let me know if you'd like to see anything else addressed.

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/boltdb/bolt v1.3.1 // indirect
 	github.com/cockroachdb/cmux v0.0.0-20170110192607-30d10be49292 // indirect
 	github.com/coreos/etcd v3.2.9+incompatible
-	github.com/coreos/go-oidc v0.0.0-20170307191026-be73733bb8cc
+	github.com/coreos/go-oidc v2.0.0+incompatible
```
srenatus: ❓Is this a problem somehow?

server/handlers.go
```diff
@@ -151,6 +152,7 @@ type discovery struct {
 	Auth          string   `json:"authorization_endpoint"`
 	Token         string   `json:"token_endpoint"`
 	Keys          string   `json:"jwks_uri"`
+	UserInfo      string   `json:"userinfo_endpoint"`
```
srenatus: Could you expand this test case for the discovery json, please? https://github.com/dexidp/dex/blob/8959dc42750e1f574db74eec9f0fcc61b39b83ea/server/server_test.go#L150

go.mod
alindeman: No, it mostly means this repository uses a `v2` tag, but not in the native `go mod` way. In fact, it's not been converted to a go module yet at all (therefore `incompatible`).

I'm personally not concerned about it: it's pretty standard fare while the Go community converts things slowly to modules over time. If/when go-oidc becomes a module, we might have to update some imports, but otherwise shouldn't be a big deal.

server/handlers.go
```diff
@@ -716,7 +728,13 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 		return
 	}
 
-	accessToken := storage.NewID()
+	accessToken, err := s.newAccessToken(client.ID, authCode.Claims, authCode.Scopes, authCode.Nonce, authCode.ConnectorID)
```
joshua-rutherford: On line 556 you define the `accessToken` but here you use `:=` instead of `=`. Thus, when you set it in the callback query parameter on line 624 you are using the variable declared on line 556, not this one.  Thus, `access_token` is always empty in the callback.

alindeman: Thank you. This came in via one of the commits I built on top of, and I did not notice the error. Fixed in 5b66bf0.

alindeman:
Any reviewers have any additional feedback or requests to change things?

srenatus:
If no one beats me to it, I'll merge this tomorrow and push a release tag

srenatus:
I've considered another round-trip with the @alindeman for squashing commits, but... I don't think it matters that much. I'm going to push the big green button. 🚀

mdbraber:
@srenatus @alindeman @jackielii @JoelSpeed thanks for all the work getting this implemented!

PR #1477, opened 2019-06-28
Add support for auth endpoint on separate URL

mhrabovcin:
This PR introduces a new optional configuration parameter `publicURL`. If not provided, default is the existing `issuer` value. The new option allows announcing Dex `authentication_endpoint` value on different URL than `issuer` and `token_endpoint` in the `discovery` endpoint. This can be useful in a setup where Dex runs on an internal network that is not publicly accessible.

We would like to configure Dex as an OpenID Connect provider (OP) in an internal network. The OpenID Connect relying parties (RPs) are also in the internal network. However, the user agents (UAs) are exclusively in the external network. The DNS names used in the internal network do not work in the external network and vice versa.

In our example, when the UAs want to reach the OP then they use the base URL `https://my-aws-01-elb.aws.amazon.com/` whereas when the RPs want to reach the OP they use the base URL `https://dex-kubeaddons.kubeaddons.svc.cluster.local:8080/`.

The RPs redirect the UAs to the OP via the `authorization_endpoint` URL exposed in the OP's `.well-known/openid-configuration` JSON doc. Therefore, there should be a way in Dex to set a custom `authorization_endpoint` URL which is what this patch tries to achieve.

Using two different base URLs in `issuer` and `authorization_endpoint` does not seem to be common but we believe that it is a valid configuration and that it does not impose a security risk. What do you think?

Example of discovery endpoint output with the following configuration:

```yaml
issuer: https://dex-kubeaddons.kubeaddons.svc.cluster.local:8080/dex
publicURL: https://my-aws-01-elb.aws.amazon.com/dex
```

```json
{
  "issuer": "https://dex-kubeaddons.kubeaddons.svc.cluster.local:8080/dex",
  "authorization_endpoint": "https://my-aws-01-elb.aws.amazon.com/dex/auth",
  "token_endpoint": "https://dex-kubeaddons.kubeaddons.svc.cluster.local:8080/dex/token",
  "jwks_uri": "https://dex-kubeaddons.kubeaddons.svc.cluster.local:8080/dex/keys",
  ...
}
```

This can be useful in scenarios when `kube-apiserver` needs to be configured to use OIDC authentication against the Dex instance that will eventually run on same kubernetes cluster. In such setup, the `publicURL` will point to public ingress that exposes Dex service.

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -28,6 +28,10 @@ type Config struct {
 	Expiry    Expiry    `json:"expiry"`
 	Logger    Logger    `json:"logger"`
 
+	// Allow to announce and expose auth and callback endpoints on different URL
+	// than token and issuer URLs.
+	PublicURL string `json:"publicURL"`
+
 	Frontend server.WebConfig `json:"frontend"`
 
 	// StaticConnectors are user defined connectors specified in the ConfigMap
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -95,6 +95,10 @@ func serve(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	if len(c.PublicURL) == 0 {
+		c.PublicURL = c.Issuer
+	}
+
 	logger.Infof("config issuer: %s", c.Issuer)
 
 	prometheusRegistry := prometheus.NewRegistry()
@@ -221,6 +225,7 @@ func serve(cmd *cobra.Command, args []string) error {
 		SkipApprovalScreen:     c.OAuth2.SkipApprovalScreen,
 		AllowedOrigins:         c.Web.AllowedOrigins,
 		Issuer:                 c.Issuer,
+		PublicURL:              c.PublicURL,
 		Storage:                s,
 		Web:                    c.Frontend,
 		Logger:                 logger,
--- a/examples/config-dev.yaml
+++ b/examples/config-dev.yaml
@@ -3,6 +3,14 @@
 # path is provided, dex's HTTP service will listen at a non-root URL.
 issuer: http://127.0.0.1:5556/dex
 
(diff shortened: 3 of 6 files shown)


srenatus:
Thank you for contributing this! 😃 🎈 

First: I'm sorry, but I've just merged the user info endpoint PR, so there's a bit of conflict introduced here.

When we've had the same problem, we've approach it like this:
In the clients (on the internal network), we've messed with the `http.Transport` to _rewrite requests to `https://my-aws-01-elb.aws.amazon.com/dex` to `https://dex-kubeaddons.kubeaddons.svc.cluster.local:8080/dex`_ ([see here for example](https://github.com/chef/automate/blob/a6a06045120f3a83c72d6c722308e00ae3821b30/components/authn-service/authenticator/oidc/oidc.go#L140-L158)).

I'll try to review the code, and the oidc spec's related parts, this week or next 😅

mhrabovcin:
Hi @srenatus, I've rebased the patch on top of `v2.17.0` tag.

jgehrcke:
> I'll try to review the code, and the oidc spec's related parts, this week or next 

Thanks for the feedback @srenatus -- we would definitely appreciate another review!

mhrabovcin:
Google OpenID configuration uses completely different URLs for different endpoints:

```json
{
 "issuer": "https://accounts.google.com",
 "authorization_endpoint": "https://accounts.google.com/o/oauth2/v2/auth",
 "token_endpoint": "https://oauth2.googleapis.com/token",
 "userinfo_endpoint": "https://openidconnect.googleapis.com/v1/userinfo",
 "revocation_endpoint": "https://oauth2.googleapis.com/revoke",
 "jwks_uri": "https://www.googleapis.com/oauth2/v3/certs",
}
```
(https://accounts.google.com/.well-known/openid-configuration)

PR #1481, opened 2019-07-08
Added "connector_id" to skip straight to a connector (similar to when len(connector) is 1.

LanceH:
This would allow a request to bypass to a specific provider by providing connector_id.


server/handlers.go
```diff
@@ -233,6 +233,15 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if authReq.ConnectorID != "" {
+		for _, c := range connectors {
+			if c.ID == authReq.ConnectorID {
```
srenatus: ❓ Should this return an error if `connector_id` was specified but doesn't match any?

srenatus:
❔ Could you reset the commit's author to match your github user account? Just so the attribution works out well.

server/handlers.go
LanceH: I debated this myself.  What could it ever do with an error other than break flow completely or just return to the index?

srenatus: That's true, but it's still better than 🙈ignoring it, isn't it? I'd propose calling `s.tokenErrHelper`, maybe?

b2cbre: Perhaps if `connector_id` does not match any, then all connectors are shown as they were prior to this PR?

LanceH: Gotcha.

LanceH:
Trying to find a good spot to document this change.

server/handlers.go
LanceH: This would not change anything with respect to showBacklink (#1123), since the number of connectors doesn't actually change.  It would need an additional test such as authReq.ConnectorID != "" to suppress the link.

server/oauth2.go
```diff
@@ -405,6 +407,19 @@ func (s *Server) parseAuthorizationRequest(r *http.Request) (req storage.AuthReq
...
+	if connectorID != "" {
+		connectors, err := s.storage.ListConnectors()
+		if err != nil {
+			return req, &authErr{"", "", errServerError, "Unable to retrieve connectors"}
+		}
+		for _, c := range connectors {
+			if c.ID == connectorID {
+				break
+			}
+			return req, &authErr{"", "", errInvalidRequest, "Invalid ConnectorID"}
```
srenatus: With this _inside the `for` loop_, it'll fail if there are more than one connectors, and the second one is the requested one. It would be nice if we had another test case for this.

server/oauth2.go
```diff
@@ -100,6 +100,7 @@ const (
 	errUnsupportedGrantType    = "unsupported_grant_type"
 	errInvalidGrant            = "invalid_grant"
 	errInvalidClient           = "invalid_client"
+	errInvalidConnectorID      = "invalide_connector_id"
```
srenatus: typo
```suggestion
	errInvalidConnectorID      = "invalid_connector_id"
```
Also, the others aren't made up, but part of the OAuth2 spec, see for example this section [5.2 Error Responses](https://tools.ietf.org/html/rfc6749#section-5.2). But, custom ones are allowed ([8.5.  Defining Additional Error Codes](https://tools.ietf.org/html/rfc6749#section-8.5)), so this is fine.

server/handlers.go
srenatus: I'm not sure we'd actually _want_ to suppress the link, would we? Maybe you'd like to use a different method...? 🤔

server/oauth2.go
LanceH: Added a validate function.   Added a second connector to the test server.  Added test.

LanceH: All right.So I ended up duplicating newTestServer to have newTestServerMultipleConnectors.  It seems that the code workflow tests may depend on having a single connector, so adding in a second connector in order to test broke all of that.

LanceH: This now tests for selecting a connector other than the first.

srenatus:
> @srenatus I'm looking for a place to document this, but I can't find any mention of the index page of connectors in the workflow. Docs like "using-dex.md" just say, "Dex determines user's identity". Any ideas?

Working code over documentation 😉 We could add a form field to example-app. I think that would be pretty neat. Let me know if you find this request just too annoying, and I'll pick up where you've left off.

At any rate, let's prepare to merge this -- would you mind squashing your commits? 😃

LanceH:
Commits are squashed.

Documentation/connectors/authproxy.md
```diff
@@ -34,7 +34,7 @@ callback URL path:
 ( dex issuer URL )/callback/( connector id )?( url query )
 ```
 
-For example, if dex is running at `https://auth.example.com/dex` and the connector
+For example, if dex is running at `https:/0/auth.example.com/dex` and the connector
```
srenatus: ...hm?

LanceH: ugh, I'll fix and resquash

srenatus:
Travis looks stuck...? @LanceH could you rebase upon master and f-push? 🤞 maybe it'll be picked up then.

LanceH:
Looks like Github had some outages, including this repo.  I've kicked it off again.

PR #1485, opened 2019-07-13
MySQL storage - Take 2

bonifaido:
This PR tries to continue where https://github.com/dexidp/dex/pull/924 dropped off, namely on the work of @pborzenkov, I have changed the following stuff:

- rebased his branch on the current master
- compile-time issues and tests fixed
- Travis based MySQL conformance test enabled (I have also changed to Xenial, which is a good starting point IMHO, but if you are not ok with it I change to MySQL 5.7 the other way)
- added basic documentation for MySQL storage

Tested with MySQL 5.7 and 8.0.

I know that maintaining a storage backend takes a lot of time, I can offer my help (becoming a maintainer if needed/acceptable).


storage/sql/migrate.go
```diff
@@ -61,14 +63,14 @@ func (c *conn) migrate() (int, error) {
 }
 
 type migration struct {
-	stmt string
+	stmts []string
```
srenatus: Why do we need this? I haven't dug into this code in a while, but I'd think we could wrap stuff in `BEGIN; ... COMMIT;`, maybe? What's wrong with the `;` concatenation? 🤔I guess it's difficult to spot what's wrong; transactions-wise, I think it should not matter, should it?

storage/sql/migrate.go
```diff
@@ -77,8 +79,8 @@ var migrations = []migration{
 				public boolean not null,
 				name text not null,
 				logo_url text not null
-			);
-		
+			);`,
```
srenatus: [nit] If we do this as multiple calls to `tx.Exec`, we don't _need_ the ending `;`, I think.

storage/sql/migrate.go
```diff
@@ -38,8 +38,10 @@ func (c *conn) migrate() (int, error) {
```
srenatus: Correct me if I'm wrong, but if this change has nothing to do with the mysql backend, I'd rather split it off... 😃

.travis.yml
```diff
@@ -2,19 +2,39 @@ language: go
...
 
+dist: xenial
+
 matrix:
   include:
     - go: '1.12.x'
 
 env:
-  global: DEX_POSTGRES_DATABASE=postgres DEX_POSTGRES_USER=postgres DEX_POSTGRES_HOST="localhost" DEX_ETCD_ENDPOINTS=http://localhost:2379 DEX_LDAP_TESTS=1 DEBIAN_FRONTEND=noninteractive DEX_KEYSTONE_URL=http://localhost:5000 DEX_KEYSTONE_ADMIN_URL=http://localhost:35357 DEX_KEYSTONE_ADMIN_USER=demo DEX_KEYSTONE_ADMIN_PASS=DEMO_PASS
+  global:
```
srenatus: 👍 It's about time this happens.

storage/sql/migrate.go
bonifaido: The issue here is that MySQL driver by default doesn't support multiple statements in one execution, in my original implementation I have used https://github.com/go-sql-driver/mysql#multistatements and this change is not needed. If you think that is better I can use that version.

storage/sql/migrate.go
bonifaido: This is because of the multistatement stuff :)

storage/sql/config_test.go
```diff
@@ -49,23 +51,39 @@ var logger = &logrus.Logger{
 	Level:     logrus.DebugLevel,
 }
 
-func TestSQLite3(t *testing.T) {
+type opener interface {
+	open(logger log.Logger) (*conn, error)
+}
+
+func testDB(t *testing.T, o opener, withTransactions bool) {
+	// t.Fatal has a bad habbit of not actually printing the error
```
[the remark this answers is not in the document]
bonifaido: Should I just change it back? TBH I haven't had issues with `Fatal`.

bonifaido:
The Postgres backend uses strict SSL verification by default, should this backend have the same semantics?

srenatus:
> The Postgres backend uses strict SSL verification by default, should this backend have the same semantics?

I think that would be fair. 🤔

bonifaido:
I made MySQL `tls=true` default, which the strictest available option for the MySQL Go driver.

srenatus:
@bonifaido would you mind squashing these? maybe one commit per author?

bonifaido:
Yes, rebase done! 👍

storage/sql/config_test.go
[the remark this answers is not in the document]
srenatus: It's fine, we can revisit that some other time :+1:

forsberg:
Question: Is there anything that specifically requires MySQL 5.7, or would 5.6 work? 

I'd like to run it against an Amazon Aurora Serverless database, and they claim they are compatible with MySQL 5.6.

bonifaido:
Hi @forsberg, the only issue that comes into my mind is the `transaction_isolation` system variable 
renaming issue (I just recompiled the code with `tx_isolation` and it works fine on MySQL 5.6):

https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_transaction_isolation

> Note
> transaction_isolation was added in MySQL 5.7.20 as an alias for tx_isolation, which is now deprecated and is removed in MySQL 8.0. Applications should be adjusted to use transaction_isolation in preference to tx_isolation.

So we need to add some kind of logic to detect MySQL 5.6 and use `tx_isolation` instead of `transaction_isolation`, or add it explicitly in the config. I'm checking what is possible right now with the driver.

But, I just checked and AWS Aurora supports 5.7 today: https://aws.amazon.com/about-aws/whats-new/2018/02/amazon-aurora-is-compatible-with-mysql-5-7/

It would be nice to run some tests on Aurora first since they say that SERIALIZABLE isolation level is not supported.

bonifaido:
Okay, I just created an AWS Aurora instance for a quick check with MySQL 5.7 (Aurora version of the db was 2.04.6) and the transaction_isolation variable is still unavailable, probably this version is not MySQL 5.7.20.

bonifaido:
https://github.com/dexidp/dex/pull/1550

forsberg:
AWS Aurora does indeed support what they call 5.7, but AWS Aurora Serverless (where you don't have to run instances yourself) only support 5.6. This is pretty hard to find in the documentation, but is documented [in the FAQ](https://aws.amazon.com/rds/aurora/faqs/#Serverless).

PR #1486, opened 2019-07-13
Add VerifyPassword to API

AlbanSeurat:
It takes in an email and plain text password to verify. If it fails to find a password stored for email, it returns not_found. If it finds the password hash stored but that hash doesn't match the password passed via the API, it returns verified = false, else it returns verified = true.

Re-open with fix to make it possible to merge


AlbanSeurat:
I have corrected the PR from tylerclocke from the previous comment and with feedback from the team.
Any chance to merge that code in master ?

srenatus:
I'd be curious about your use case -- how do you intend to make use of this? 😃

AlbanSeurat:
We are using it to check our user’s password (from a back-end service) before calling updatePassword. 
We are using dex as central repository for our users.

server/api.go
```diff
@@ -254,6 +254,36 @@ func (d dexAPI) ListPasswords(ctx context.Context, req *api.ListPasswordReq) (*a
 
 }
 
+func (d dexAPI) VerifyPassword(ctx context.Context, req *api.VerifyPasswordReq) (*api.VerifyPasswordResp, error) {
+	if req.Email == "" {
+		return nil, errors.New("no email supplied")
```
srenatus: This being GRPC, can we return `status.Error(codes.InvalidArguments, "no email supplied")` or something like that? `errors.New()`, or, any error would translate to `codes.Internal` or `codes.Unknown`, can't remember.

server/api.go
```diff
@@ -254,6 +254,36 @@ func (d dexAPI) ListPasswords(ctx context.Context, req *api.ListPasswordReq) (*a
...
+		if err == storage.ErrNotFound {
+			return &api.VerifyPasswordResp{
+				NotFound: true,
+			}, nil
+		}
+		d.logger.Errorf("api: there was an error retrieving the password: %v", err)
+		return nil, fmt.Errorf("verify password: %v", err)
+	}
+
+	if err := bcrypt.CompareHashAndPassword(password.Hash, []byte(req.Password)); err != nil {
```
srenatus: Is the `err` interesting? If so, we might log it. If not, we could do
```go
if bcrypt.CompareHashAndPassword(password.Hash, []byte(req.Password)) != nil {
```
instead. Or even

```go
verified := bcrypt.CompareHashAndPassword(password.Hash, []byte(req.Password)) == nil
return &api.VerifyPasswordResp{Verified: verified}, nil
```

server/api.go
AlbanSeurat: I can change the error returned, but that will be not consistent with the rest of errors returned by other api. Do you want me to do it anyway ? and you will have some other tasks where you will change the other functions ?

server/api.go
AlbanSeurat: I will add some log

AlbanSeurat:
I have updated the branch in regards to some comments - adding some logs.

server/api.go
srenatus: I'll deal with the others. :smiley: You know what, you're correct, this is fine, let's keep them as-is and we can update all the errors in one go later.

AlbanSeurat:
Great ! Thanks

srenatus:
@AlbanSeurat One small thing that kept me from merging this: you haven't added your name as author anywhere. I suppose it would make sense to amend this PR's commit and add
```
Co-authored-by: YOUREMAIL
```
so that github will pick it up as your contribution, too, properly. 😃

AlbanSeurat:
> @AlbanSeurat One small thing that kept me from merging this: you haven't added your name as author anywhere. I suppose it would make sense to amend this PR's commit and add
> 
> ```
> Co-authored-by: YOUREMAIL
> ```
> 
> so that github will pick it up as your contribution, too, properly. 😃

Done

PR #1487, opened 2019-07-15
Add configurable callback

tanushree-sharma-98:
Allows the user to set a custom callback URL (different than that of the issuer) for an OIDC connector. An organization that implements OIDC may have a domain limit (eg. Google), which would limit the number of applications that can authenticate against it. This feature uses redirect URIs, rather than domains to redirect to the correct application after authentication. 

Setup: Add a "callback" field in the config.yaml file (eg. https://auth.company.com/app1). The redirect URI for the connectors would then be the callback URL + /callback (eg. https://auth.company.com/app1/callback). Ensure that your OIDC connector has the callback URL for each application configured. Create a proxy that directs traffic back to the application and does a URI match and re-write for callback and approval back to dex (eg. URI match on /app1/callback and re-write it to /callback and a URI match on /app1/approval and rewrite it to /approval). 

You can now have a number of applications that use the same OIDC connector for authentication.

### DIFF
--- a/Documentation/connectors/oidc.md
+++ b/Documentation/connectors/oidc.md
@@ -30,7 +30,8 @@ connectors:
     clientID: $GOOGLE_CLIENT_ID
     clientSecret: $GOOGLE_CLIENT_SECRET
 
-    # Dex's issuer URL + "/callback" or if a custom callback field is declared, Dex's callback URL + "/callback"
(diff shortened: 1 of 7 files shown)


tanushree-sharma-98:
Does anyone have any interest and/or feedback on this?

tanushree-sharma-98:
@klarose

PR #1491, opened 2019-07-22
Allow env expansion of issuer URL

sks:
- This allows us to have config file that can be reused across different environment `dev/qa/prod`

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -75,6 +75,7 @@ func serve(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


nabokihms:
Dex image starting from version v2.27 has an entry point, rendering the config with [gomplate](https://docs.gomplate.ca/).
With this feature, you can expand an issuer from an env variable like this: 
```
issuer: {{ getenv "DEX_ISSUER" "http://127.0.0.1:5556/dex" }}
```

I am going to close this PR. If you still have a problem, please reopen this issue or open a new one.

PR #1500, opened 2019-07-30
*: fix some lint issues

srenatus:
Mostly gathered these using golangci-lint's deadcode and ineffassign
linters.


srenatus:
I would have played with getting golangci.com hooked up in this repo, but I always hit a 500 when trying to  do so... :/

PR #1411, opened 2019-02-22
go.mod: update all vendored code with 'go get -u'

ericchiang:
Fixes https://github.com/dexidp/dex/issues/1406
Fixes https://github.com/dexidp/dex/issues/1141

Let's see if this builds

PR #1501, opened 2019-07-30
update all deps

srenatus:
#1411 with the protobuf code updated, too.


srenatus:
I'm afraid this is one of my first experiences with go modules 😅 Some thorough review of what I've done here would be appreciated.

bonifaido:
Do we have some tests for the Proto client API? :)

srenatus:
> Do we have some tests for the Proto client API? :)

There's an example client that does a few calls:
```
make bin/dex
make bin/grpc-client
env SAN=IP.1:127.0.0.1 examples/grpc-client/cert-gen
bin/dex serve examples/grpc-client/config.yaml &
bin/grpc-client -ca-crt ca.crt  -client-crt client.crt -client-key client.key
```

but I don't think that this is wired up with travis or anything.

Also, thanks for getting me to try this, we've got a small bug in error reporting for VerifyPassword:
```
time="2019-07-30T12:11:37Z" level=info msg="api: password check failed : %vcrypto/bcrypt: hashedPassword is not the hash of the given password"
```
☝️This is what the server logs. Looks like we're using `Info()` somewhere `Infof()` should be used.

srenatus:
⏩ https://github.com/dexidp/dex/pull/1502

srenatus:
I'm still hesitant to merge this. This PR is `+354,709 −83,649`, #1411 was `+294,875 −75,858`. Am I doing something wrong? 🤔

bonifaido:
Please try `go mod tidy` and then `go build` and you should have the current correct state in go.mod and go.sum (and of course `go mod vendor` again).

srenatus:
I'll look into this tomorrow.

srenatus:
So, I've run only `go get -u; make revendor` once again, and adapted to the prometheus API change, and regenerated the PB code. I've also reviewed the docs on modules, I think we're OK doing what we do. I'd rather get rid of the `vendor.go` "trick" eventually, but it's probably OK for now.

I'm ready to assume that the five months that have passed since #1411 are the reason for the difference... 😄

bonifaido:
> I'm ready to assume that the five months that have passed since #1411 are the reason for the difference... 😄

This is totally normal 😄 👍

PR #1505, opened 2019-08-02
Add option to always display connector selection even if there's only one

MarcDufresne:
We are currently building an auth proxy system that will be used to authenticate users before being redirected to their destination. The fact that Dex automatically forwards to the connector if there's only one has been confusing for our users. They type `my-service.com` in their browsers and end up on Google login, not a great UX.

This PR allows to specify an option to always show the connector selection screen even if there's only one. That way our users will land on a themed login page, and select "Login with Google" before ending up on Google login.

Let me know what you think!

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -127,6 +127,8 @@ type OAuth2 struct {
 	// If specified, do not prompt the user to approve client authorization. The
 	// act of logging in implies authorization.
 	SkipApprovalScreen bool `json:"skipApprovalScreen"`
+	// If specified, show the connector selection screen even if there's only one
+	AlwaysShowLoginScreen bool `json:"alwaysShowLoginScreen"`
 }
 
 // Web is the config format for the HTTP server.
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -249,7 +249,7 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if len(connectors) == 1 {
+	if len(connectors) == 1 && !s.alwaysShowLogin {
 		for _, c := range connectors {
 			// TODO(ericchiang): Make this pass on r.URL.RawQuery and let something latter
 			// on create the auth request.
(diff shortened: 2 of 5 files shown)


srenatus:
Thank you for the contribution :tada:
Haven't looked closely at the code yet, but I wonder, could the `prompt` query parameter be used here, too? I haven't looked in the oidc spec of that in a while, it could mean something else.

MarcDufresne:
https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

I'm not sure `prompt` fits this use case, maybe `select_account`? But this feels a bit like a stretch

srenatus:
@MarcDufresne Ah, I see. Yeah, indeed; but thanks a lot for checking. ✔️

server/handlers.go
```diff
@@ -249,7 +249,7 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if len(connectors) == 1 {
+	if len(connectors) == 1 && !s.alwaysShowLogin {
```
srenatus: When there's only local users, I believe `len(connectors) == 1` also drives whether or not a "back link" is rendered:

image

I think that code needs adjustment, too, now? What would be desired if you click on "login with username/password" and then change your mind? Hm. If there's nothing else they can do, it might be ok to keep the logic as-is -- i.e. to not display the backlink regardless of `s.alwaysShowLogin`...

cmd/dex/config.go
```diff
@@ -127,6 +127,8 @@ type OAuth2 struct {
 	// If specified, do not prompt the user to approve client authorization. The
 	// act of logging in implies authorization.
 	SkipApprovalScreen bool `json:"skipApprovalScreen"`
+	// If specified, show the connector selection screen even if there's only one
+	AlwaysShowLoginScreen bool `json:"alwaysShowLoginScreen"`
 }
```
srenatus: Can we please add that statement to the tests in `cmd/dex/config_test.go`?

MarcDufresne: Will do!

server/handlers.go
MarcDufresne: I'd be in favor of keeping the behaviour the same as now

MarcDufresne:
Test added!

PR #1508, opened 2019-08-02
handle id_token_hint for implicit login flow

scotthew1:
Handle `id_token_hint` as described in #990


server/oauth2.go
```diff
@@ -408,6 +409,24 @@ func (s *Server) parseAuthorizationRequest(r *http.Request) (*storage.AuthReques
 		return nil, &authErr{"", "", errServerError, ""}
 	}
 
+	// TODO: we don't currently handle prompt but should verify prompt=none
+	if tokenHint != "" {
+		keystore := &storageKeySet{s.storage}
+		payload, err := keystore.VerifySignature(context.TODO(), tokenHint)
+		if err != nil {
+			return nil, &authErr{"", "", errInvalidRequest, "Could not verify id_token_hint"}
```
scotthew1: not sure that we actually care if we can verify the signature here... if the token's in a client session from days ago, dex might not not have the keys to verify it anymore. since this is just a 'hint' rather than any proof of authentication, it's probably better to just get the sub claim without caring about the signature. thoughts?

srenatus:
Thanks for looking into this! 🎉 I'll review this soon, sorry for keeping you waiting.

Toxicable:
We're still interested in this feature, is there anything I can do to help?

srenatus:
@Toxicable Sorry, dropped the ball on this. Could you review this?

Toxicable:
The code  that's there looks like it does what's intended.
However I cant comment on the actual process, I don't know what are other valid inputs values and how they're meant to be handled

Issue #1503, opened 2019-08-01
storage mysql error with Percona XtraDB and pxc_strict_mode


mkontani:
I tried mysql storage with Percona XtraDB and pxc_strict_mode enabled, and had an error on startup migration step.
It looks occurring because of primary key restriction.

Percona XtraDB and pxc_strict_mode validates whether tables have primary keys.
https://www.percona.com/doc/percona-xtradb-cluster/LATEST/features/pxc-strict-mode.html#tables-without-primary-keys

And now dex's migrations table has no primary key.
https://github.com/dexidp/dex/blob/master/storage/sql/migrate.go#L10-L13

Is it difficult to have primary key on migrations table?
If so, I can only switch PXC Strict Mode DISABLED..

bonifaido:
Hi @mkontani, no I think it shouldn't be that hard to make this work `num` looks like a good candidate for a primary key.

Could you please change:
```sql
num integer not null,
```

to 

```sql
num integer not null primary key,
```

recompile and try again? TBH I don't have access to a properly working Percona DB right now :)

If that works we are more than happy to receive a PR for it! 👍

mkontani:
@bonifaido Thanks for your confirmation!
I tried and got another problem like below.
```
failed to initialize storage: failed to perform migrations: creating migration table: Error 1105: Percona-XtraDB-Cluster doesn't recommend using SERIALIZABLE isolation with pxc_strict_mode = ENFORCING
```

The detail is [here](https://www.percona.com/doc/percona-xtradb-cluster/LATEST/features/pxc-strict-mode.html#explicit-table-locking).
> EXPLICIT TABLE LOCKING
> Percona XtraDB Cluster has only experimental support for explicit table locking operations, The following undesirable operations lead to explicit table locking and are covered by this validation:
> ...
> - Setting the SERIALIZABLE transaction level

Current dex mysql transaction level is set to "SERIALIZABLE", and if change to others i.e. "REPEATABLE-READ", dex works well.
https://github.com/dexidp/dex/blob/master/storage/sql/config.go#L279

Any idea which is the best solution ? 🤔
- Switch PXC Strict Mode "DISABLED"
- Set primarykey and change transaction level from "SERIALIZABLE" to "REPEATABLE-READ"
- Set primarykey and Switch PXC Strict Mode "MASTER"
(anyway, setting primary key looks better.)

Thanks.

bonifaido:
[...]

PR #1511, opened 2019-08-03
Fix for mysql xtraDB support

mkontani:
Fixes for #1503 

* fix for mysql with Percona XtraDB and pxc_strict_mode
* storage.md smallfix

Thanks @bonifaido .

### DIFF
--- a/Documentation/storage.md
+++ b/Documentation/storage.md
@@ -248,7 +248,7 @@ The flow of the migration process is as follows:
 
 ## SQL
 
-Dex supports two flavors of SQL: SQLite3 and Postgres.
+Dex supports three flavors of SQL: SQLite3, Postgres, and MySQL.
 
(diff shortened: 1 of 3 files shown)


bonifaido:
Okay, I was wrong, unfortunately, SERIALIZABLE is required for MySQL, we need to think about this how we can proceed 🤔

bonifaido:
And further on I have tried Percona XtraDB with REPEATABLE-READ and the conformance tests failed with the same result as with MySQL in Travis...

bonifaido:
Blocked by: https://github.com/dexidp/dex/pull/1370 as you mentioned previously @mkontani 

Please see [my comment](https://github.com/dexidp/dex/pull/1370#issuecomment-517924434), if #1370 gets finished, we can remove the serialization level setting at all, which will fix our issues 🙂

Issue #1499, opened 2019-07-30
gRPC API: use status codes in error returns


srenatus:
In [our gRPC handlers in `api.go`](https://github.com/dexidp/dex/blob/72f55966713838932fa782ce48ce5c39660650d6/server/api.go), we return `errors.New("bad something")` in a few places. For the callers, it would be more informative if those errors were "proper" gRPC errors, i.e. contained status codes.

See [the docs of `google.golang.org/grpc/codes` for available codes and their meanings](https://godoc.org/google.golang.org/grpc/codes).

srenatus:
As mentioned in #1501's comments, it might also be nice to have some external API consumption tests here, or in a different PR. Currently, [the gRPC API is tested using it's own generated client code here](https://github.com/dexidp/dex/blob/master/server/api_test.go).

I could imagine using some reflection-based client on this, like https://github.com/fullstorydev/grpcurl... 🤔(We' have to enable gRPC reflection first, of course.)

venezia:
I personally use grpcurl and the sister project, [grpcui](https://github.com/fullstorydev/grpcui) - but I do have to provide the `api.proto` file ... adding reflection shouldn't impact the program that much (I typically add reflection myself, but we could also make it dependent on an environmental variable)

I'm willing to work on this as I was personally working on add a `ListClients` grpc call (I added it, but then realized that it wasn't implemented using k8s as a store, so I'm working on that now) and I personally believe that using the official grpc errors would be best (common errors across all grpc projects)

Furthermore, if we use official grpc errors we might want to re-evaluate the decision to not support grpc-gateway.  In conjunction with grpc error codes you will get a reasonably valid [HTTP status code](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/errors.go#L17) For more context see [this](https://github.com/dexidp/dex/blob/master/Documentation/api.md#why-not-rest-or-grpc-gateway)

There's also some errors in the documentation about the api (probably outdated) - for example dex will not [self generate api certificates if they're not defined](https://github.com/dexidp/dex/blob/master/Documentation/api.md#why-not-rest-or-grpc-gateway) - it will just work in insecure mode.  I can clean that up and some other things.

Also updating to a release version of golang/protobuf (and using a protoc that isn't 2.5 years old) might be nice changes.

Sorry this got long, but I'm willing to help.

srenatus:
@venezia Thanks for the message, and thanks for offering help! 🎉

[...]

PR #1512, opened 2019-08-06
Add reflection to grpc api

venezia:
This will add reflection as an option for the grpc api.
It also makes a small correction in the documentation - dex will not create self signed certificates if certs are not provided, rather it runs in plaintext mode.

Does not fix, but this work was discussed in #1499


venezia:
travis wasn't able to [connect to github during the test](https://travis-ci.org/dexidp/dex/builds/568556609?utm_source=github_status&utm_medium=notification) - unsure how to get travis to try again

cmd/dex/serve.go
```diff
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"google.golang.org/grpc/reflection"
```
srenatus: Please put that into a section with the other 3rd-party imports :)

venezia: Sorry, I IDE did that automatically for me and I wasn't paying attention - sure

venezia:
> LGTM, thanks for this! 😃 Would you mind squashing this into two commits -- one for the code, one for the revendor? That would also trigger travis again, I hope 🤞

Sure, but just know I was only following [dex's guidelines](https://github.com/dexidp/dex/blob/master/Documentation/dev-dependencies.md#composing-commits)

That being said do you want, in the future, for commits involving dependencies to be in the same commit as regular code changes?

srenatus:
@venezia thanks for pointing out those guidelines. I had forgotten about that. They're quite agreeable -- at any rate, could you amend your commits to satisfy them? 😉 
Like
1. `cmd/dex: make reflection configurable`
2. `vendor: revendor`

and the third commit can be squashed into the first one?

srenatus:
If you want to split commits more (while adhering to the guidelines), please do so -- I'm sorry for misleading you here 😉

venezia:
> If you want to split commits more (while adhering to the guidelines), please do so -- I'm sorry for misleading you here 😉

no, this is what happens when I misread a comment while still waking up.  I think here one commit is just fine, sometimes if it is bigger maybe more than one commit is better (if they're doing separate things, etc.) but that's for another day. 

Regardless, things should be squashed and commented better now.

Issue #1149, opened 2017-12-15
SAML: IdP-initated login


srenatus:
Hey there.

When setting up Dex as a SAML SP in Okta, when you don't specify that it should _not be there_<sup>(a)</sup>, you end up with an icon in the user's home screen:

<img width="839" alt="screen shot 2017-12-15 at 08 32 15" src="https://user-images.githubusercontent.com/870638/34031650-a3e2862e-e172-11e7-8d45-0dc1b40a34a5.png">

To me at least, it's always tempting to click that, and hope that it would just sign me in. 😄 

Technically, it works by just sending an assertion to the configured endpoint; and depending on how this endpoint reacts, it will work or not work. Specifically, Okta allows you to define the _Default RelayState_:

<img width="692" alt="screen shot 2017-12-15 at 08 35 40" src="https://user-images.githubusercontent.com/870638/34031767-2c350fc4-e173-11e7-9edc-e3ce7f7b81f8.png">

### Proposal: Let's make this work!

Currently, the `RelayState` is used in the same way `state` is used for OAuth2 callbacks: to carry around the `authID` ([see here](https://github.com/coreos/dex/blob/ec5e2cc3c6229193d421d247f9c039be82ccfbac/server/handlers.go#L328-L338)).
This ID references the [AuthRequest](https://github.com/coreos/dex/blob/ec5e2cc3c6229193d421d247f9c039be82ccfbac/storage/storage.go#L148-L185); whose details will be initialized when the authentication-through-dex process starts, typically by a call to `/auth`.

Now, in the IdP-initiated setting, we don't have any of these. However, if we're able to specify the default RelayState, we could use that to provide those settings, say, as a `url.Values`-encoded string (and properly documentated).

Of course, some care has to be taken about not compromising our security by this in any unacceptable way, but I don't think that this would block this completely.

**What do you think?**

##### Notes

<details><summary>(a) this would be done by checking the first box here</summary>

<img width="739" alt="screen shot 2017-12-15 at 08 35 29" src="https://user-images.githubusercontent.com/870638/34031743-163e3cae-e173-11e7-8fa6-04acd8a1d235.png">

</details>

ericchiang:
> What do you think?

This scares me. It'd require a pretty intrusive change to dex and prevents us from hardening features like using the `state` to prevent CCRF attacks.

https://security.stackexchange.com/questions/20187/oauth2-cross-site-request-forgery-and-state-parameter

srenatus:
[...]

Issue #1295, opened 2018-09-10
SAML test scaffolding


srenatus:
In various SAML-related PRs, it became apparent that out lack of automated tests for the SAML connector a hurdle to developing and maintaining that code (see for example #1045, #1175, #1282).

👉 [The LDAP connector spins up an LDAP server (openldap) in the travis tests.](https://github.com/dexidp/dex/blob/4a6da13097aeebe4e3c418ace4d20e4ad470b331/connector/ldap/ldap_test.go)

It would be great if we had a similar thing going for SAML. There seem to be a few applicable projects to plug into our tests, like

- https://github.com/amdonov/lite-idp (@amdonov 👋)
- https://github.com/OpenConext/Mujina
- and of course the feature-full, not-meant-for-dev-only projects, like https://simplesamlphp.org/

While the artifact binding seems to be controversial, I don't think the redirect binding is -- so any project used should perhaps support that, so we don't restrict ourselves for no good reason.

amdonov:
I'd be happy to help and willing to make any changes to lite-idp required to facilitate testing.

srenatus:
@amdonov would you mind if I assigned this to you then? 😉

amdonov:
@srenatus I looked at this briefly this morning. My initial inclination would be to beef up an external SAML service provider implementation like https://github.com/russellhaering/gosaml2 and integrate that with dex. It would probably get more use and testing in other scenarios and potentially benefit other users/projects.

Not committing to using that specific library, but would you be open to that type of approach?

srenatus:
So, rephrasing to ensure I've grokked it, your proposal would be to replace the current SAML connector with one based on a purpose-built SAML library (over the current code which [uses libraries](https://github.com/dexidp/dex/blob/666356d22d9c83f764045340d956e4f2213540ae/connector/saml/saml.go#L15-L17) for making sense of XML and the signature verification).

(Assuming that's it, ) I think that's a good idea. The only concern I'd have is backwards-compatibility: it seems like either, **we aim for backwards-compat** (for example by replacing the current saml connector), or **we introduce a new connector** (I've seen the idea of having a `saml2` floating around in some PR already 😉 (⏩ #1282)).

If we went with the first option, I'd suppose we might _first_ get tests in place to latter know that we've kept our backwards-compat goals when swapping the connector.

[...]

PR #1514, opened 2019-08-08
support for saml idp initiated flow

scotthew1:
hey @srenatus i know #1149 was closed due to security concerns, but this flow has become a hard requirement for our use case. that original issue discusses the ability to use the 'state' parameter to handle this, but i found that relying on 'RelayState' from the SAML provider was fine for our use case and don't think this approach ended up being too obtrusive. perhaps this is a bit naive and/or too specific to our use case, but i thought i'd put it out there for discussion.

a few notables:
1. this relies on the ssoIssuer config (which is otherwise optional) to match the unsolicited post from the SAML provider with the connector in the dex config.
2. to determine the client the authorization is destined for, I match the RelayState with any redirectURI a known client. i'm not sure if it's valid for two different clients to have the same redirectURI which would make it difficult to determine which client to create authorization for.
3. i need to fabricate an auth request, so i have to assume some set of claims. it's probably best to make this configureable, but i've just got them hardcode right now.
4. i have the 'code' response type hardcoded as well. i'm not sure whether or not it's possible to make the implicit flow work with this setup (client would expect to validate a nonce that we can't fabricate).


srenatus:
Thanks! Haven't had a chance to look at this yet, but I'm excited about this. 😃

server/handlers.go
```diff
@@ -475,6 +479,94 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
...
+	}
+	// build our auth request (fake it til ya make it)
+	authID := storage.NewID()
+	req := storage.AuthRequest{
+		ID:                  authID,
+		ClientID:            client.ID,
+		State:               "",
+		Nonce:               "",
+		ForceApprovalPrompt: false,
+		// TODO: make IdP initiated scopes configurable as part of the saml connector
```
srenatus: 👍

connector/saml/saml.go
```diff
@@ -411,6 +411,21 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse, inResponseTo str
...
+	if err != nil {
+		return "", err
+	}
+
+	var resp response
+	if err := xml.Unmarshal(rawResp, &resp); err != nil {
+		return "", err
+	}
+
+	return resp.Issuer.Issuer, nil
```
srenatus: ❓ Could `resp.Issuer` be `nil`? (I really don't know! 😅)

server/handlers.go
```diff
@@ -475,6 +479,94 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
...
 }
 
+// handleSamlCallback handles a saml callback response with support for the IdP initiated flow.
+// if the RelayState is not one of our auth requests, we check to see if it is a valid redirectURI
+// for one of our clients. then, if the SAMLResponse is from a registered connector, we create a
+// new auth request in the database so we can contuine with our regular callback flow.
+func (s *Server) handleSamlCallback(r *http.Request) (string, string, int, error) {
+	relayState := r.PostFormValue("RelayState")
+	if relayState == "" {
+		return "", "", http.StatusBadRequest, fmt.Errorf("user session error")
```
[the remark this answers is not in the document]
scotthew1: whoops yeah can definitely use `errors.New()`

re: the various error message comments, i was trying to keep the errors returned to the user intentionally vague for security reasons. since this is already a flow prone to man-in-the-middle attacks i was trying to reduce information gleamed from specific error messages and provide the more detailed error in the logs.

if this isn't something we find important, i can certainly make the returned errors more specific.

server/handlers.go
```diff
@@ -475,6 +479,94 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
...
+		for _, u := range c.RedirectURIs {
+			if relayState == u {
+				client = &c
+				break
+			}
+		}
+	}
+	if client == nil {
+		s.logger.Errorf("Cannot find client with redirectURI: %s", relayState)
+		return "", "", http.StatusInternalServerError, fmt.Errorf("bad saml response")
```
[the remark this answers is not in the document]
scotthew1: oh yeah, agreed. copy/paste error

server/handlers.go
```diff
@@ -475,6 +479,94 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
...
+	}
+	var connID string
+	for _, c := range connectors {
+		if c.Type != "saml" {
+			continue
+		}
+		var cfg saml.Config
+		if err := json.Unmarshal(c.Config, &cfg); err != nil {
+			s.logger.Errorf("Cannot parse saml config for connector %s: %s", c.ID, err)
+			return "", "", http.StatusPreconditionFailed, fmt.Errorf("config error")
```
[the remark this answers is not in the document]
scotthew1: yeeeah agreed that this shouldn't ever really happened, but wanted to handle the error regardless. since it's config related i figured it made sense as a precondition failure

connector/saml/saml.go
scotthew1: good call, probably doesn't hurt to check

server/handlers.go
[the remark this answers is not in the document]
srenatus: You're correct, better safe than sorry. We'll log the details.

server/handlers.go
```diff
@@ -475,6 +479,94 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
...
+	}
+	// if our relaySate is a valid authId, this is not IdP initiated
+	if _, err := s.storage.GetAuthRequest(relayState); err == nil {
+		return relayState, relayState, http.StatusOK, nil
+	} else if err != storage.ErrNotFound {
+		s.logger.Errorf("Failed to get auth request: %v", err)
+		return "", "", http.StatusInternalServerError, fmt.Errorf("database error")
+	}
+	// find the client the RelayState is trying to send us to, stop when we match a redirectURI
+	// TODO: is it valid for differrent clients to have the same redirectURI? ¯\_(ツ)_/¯
```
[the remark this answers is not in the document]
scotthew1: another idea i had is to put the clientId in the RelayState and provide something like a samlRedirectURI config on the client to decide which URI we should send the code to once we've verified the SAML request. this would be a more substantial change but perhaps worth it?

scotthew1:
hey @srenatus here's a thought on supporting custom scopes rather than the hardcoded ones i've got now. update the Client struct in storage to something like...
```golang
type Client struct {
	...
	SAMLInitiated *SAMLInitiatedConfig
}

type SAMLInitiatedConfig struct {
	RedirectURI string 
	Scopes      []string
}
```

this way each client could have different scopes for dealing with an IdP initiated scope. this change also includes the idea i had for handling RelaySate differently were we don't have to try and match to some existing redirectURI on any client, we'd just check if the RelayState matches some existing clientID and then see if that client has this SAMLInitiatedConfig to figure out how this request is handled. finally, this means any client would need to opt-in to working with IdP initiated flows by providing this configuration. perhaps there's even additional configuration here to decide which connectors the client is willing to work with as you described in your original comment.

just wanted to get some feedback on this design before implementing since it impacts storage and thus a bit less trivial to implement.

srenatus:
I haven't dug into how much of an effort this is going to be, but -- taken at face value, having *the clients* determine if they want to allow *being selected* for IdP-initiated Sign In seems like a good choice :+1:

My only hesitation is that we're adding something to _clients_ that is **very** specific to a single type of connector. (💭 I wonder if this could be abstracted?) I mean, on the other hand, the SAML connector(s) already necessitate special handling in the server's handlers, and that's a thing differentiating SAML from the rest. (What I mean is that we've already gone down that road.)

scotthew1:
@srenatus i just pushed some changes with an initial attempt at the approach i described. i've not updated the various storage connectors, but what i had was enough to work with the static config i've been testing with. further scope would involve updating the various storage connectors (some of which look to be easier than others to update).

i feel you on being adverse to having much connector-specific config on the client, but i'm not sure of a better way to handle this.. the whole IdP initiated flow is outside of anything oauth intends to do (with good reason), so it's hard to not have SAML things creep outside of the connector.

if you've got any good ideas for better abstraction let me know! and let me know what you think about the current approach with my latest commit!

storage/sql/migrate.go
```diff
@@ -190,4 +190,13 @@ var migrations = []migration{
 			);`,
 		},
 	},
+	{
+		stmts: []string{`
+			alter table client
+				add column saml_init_redirect_uri text not null default '';`,
+			`
+			alter table client
+				add column saml_init_scopes bytea; -- JSON array of strings`,
```
scotthew1: @srenatus i took a stab at updating the storage connectors with my changes. tests are passing now, but one questionable bit is this migration. seems like the convention is to make all columns `not null` across every table, but i ran into a bit of a conflict here.

postgresql won't let you add a `not null` column to a table without specifying a default, however mysql won't let you specify a default on a `BLOB` column (what `bytea` translates to in the mysql flavor). so.... i just let this column be nullable. not sure if there's any potential issues with that.

srenatus:
Sorry for radio silence -- I have not forgotten about this. (I still _want this!_ 😄)
Will review soonish.

server/handlers.go
```diff
@@ -475,6 +479,93 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
...
+		return "", "", http.StatusInternalServerError, errors.New("database error")
+	}
+	redirectURI := client.SAMLInitiated.RedirectURI
+	if redirectURI == "" {
+		// this client does not have support for IdP initiated login configured
+		s.logger.Warnf("SAML IdP initiated login attempt made to client %q, but support for this flow is not configured. Update the client config to support this login flow.", relayState)
+		return "", "", http.StatusBadRequest, errors.New("user session error")
+	}
+	// TODO: should we check that 'idtoken' is one of the scopes? or just put the
+	// burden on the user to confiugre these scopes correctly?
```
srenatus:
```suggestion
	// burden on the user to configure these scopes correctly?
```

storage/kubernetes/types.go
```diff
@@ -167,6 +167,14 @@ type Client struct {
 
 	Name    string `json:"name,omitempty"`
 	LogoURL string `json:"logoURL,omitempty"`
+
+	SAMLInitiated SAMLInitiatedConfig `json:"samlInitiated"`
```
srenatus: ❓ How about adding `,omitempty`? I suspect _most_ people won't use this right away... 🤔

storage/sql/migrate.go
srenatus: We'd expect most clients _not_ to have a SAML config, wouldn't we? So, null seems alright for that...? 🤔

storage/sql/migrate.go
```diff
@@ -190,4 +190,13 @@ var migrations = []migration{
 			);`,
 		},
 	},
+	{
+		stmts: []string{`
+			alter table client
+				add column saml_init_redirect_uri text not null default '';`,
+			`
+			alter table client
```
srenatus: I think these could be combined into a single statement, `ALTER TABLE client ADD COLUMN ..., ADD COLUMN`. Not sure if MySQL prohibits this?

scotthew1: i'm not sure.. tbh i just followed the structure of the `refresh_token` migration above it which does three `alter`s each in a separate statement.

scotthew1:
hey @srenatus i updated with docs and fixes to some minor issues.

i took a look at what might be needed to add some tests for `handleSAMLCallback()` and it seems kinda difficult to do much without addressing #1295. `TestHandleInvalidSAMLCallbacks` in `handlers_test.go` already covers *some* of the negative testing, but adding more would require a more elaborate saml connector setup with the test server.

pschaumburg:
@scotthew1 @srenatus 
Is there any update on this? :)

mhaddon:
Is there anything we can do to assist in the delivery of this feature?

Saqibm128:
same, i would be interested in this feature as well for our organization

bonifaido:
First of all this branch has to be rebased on current master since some things around storage and CI have changed. If someone could give a stab at trying to resurrect this branch that would be awesome!

scotthew1:
hey i just rebased the branch. in regards to what's need for a merge... i can throw some docs together in the near future, but i'm not sure about the testing question. never really got a response on that one and iirc there were some requirements on another issue.

pkarrot:
any plans to merge this?

srenatus:
Sorry for having dropped the ball on this. Anyone listening? I'll try to re-review when this is rebased! 😃

heidemn-faro:
@srenatus @scotthew1  I didn't rebase, but merged it with recent master. Still worked for me (tested with Okta).
Feel free to continue. Our company will most likely not need this, so I'm currently not motivated to continue myself.
https://github.com/faro-oss/dex/pull/22
https://github.com/faro-oss/dex/tree/saml-idp-initiated

karaimin:
What is the status for this PR ? Any plans for moving ahead ?

erhudy:
> What is the status for this PR ? Any plans for moving ahead ?

This PR is of interest to me (specifically doing provider-initiated flows from Okta to Argo CD), so I am working on resolving the merge conflicts and getting it to build and pass tests again. Since this PR is obviously from a fork that I do not own, I will open a separate PR once I have it in an appropriate state.

Issue #1518, opened 2019-08-14
Authentication through Gitlab 


owelinux:
when `groups` is provided, but  the all user can authenticated. ，my config is：

data:
  dex.config: |
    connectors:
      - type: gitlab
        id: gitlab
        name: GitLab
        config:
          baseURL: https://gitlal.com
          clientID: xxx
          clientSecret: xxx
          redirectURI: http://127.0.0.1:5556/dex/callback
          groups:
          - argo-

eohtake:
I'm seeing the same behavior here. In the connector's docs it states:
```
# If `groups` is omitted, all of the user's GitLab groups are returned when the groups scope is present.
```
Where the group scope should be present?
Omitting the groups directive doesn't return any group with the following results from Dex logs:
```
level=info msg="login successful: connector \"gitlab\", username=\"My User\", email=\"my.email@myemail.com\", groups=[]"
```

Adding or not the groups in config.yaml for Dex, the payload returned is the same.

```json
Claims:

{
  "iss": "https://dex.mydomain.com:32000",
  "sub": "BrDeMjESBmdpdPioE",
  "aud": "example-app",
  "exp": 1565834057,
  "iat": 1565747657,
  "at_hash": "4KksPLAGKNzV2ja8dJucgQ",
  "email": "my.email@myemail.com",
  "email_verified": true,
  "name": "My Name"
}
```

**GitLab 12.0.3-ee**
**dex:v2.10.0**

bonifaido:
The `groups` scope should be present in the AuthCodeURL of the Oauth request, so for example in the example-app like here on the screenshot:

<img width="306" alt="Screen Shot 2019-08-14 at 12 16 01 PM" src="https://user-images.githubusercontent.com/23779/63014115-0cbf2e00-be8e-11e9-92ad-1e550962a2c9.png">

But to be honest this should be the default if filtering `groups` are defined in the configuration YAML, don't you think so?

eohtake:
I missed that. Adding the group to that extra scopes field in the example-app solves the problem. Also using **mintel/dex-k8s-authenticator** it sends the groups automatically.

I agree with you. It would make things simpler, just having that as default if filtering is being used.

bonifaido:
I will create fix PR for this where it will make it default when specifying groups.

eohtake:
That’s awesome! Thanks mate!
I’m looking forward to see the code to have a better understanding on how it works.

PR #1520, opened 2019-08-14
gitlab: add groups scope by default when filtering is requested

bonifaido:
Fixes: https://github.com/dexidp/dex/issues/1518


bonifaido:
I think now it works like as it should by default, this is the way as the github and bitbucket connectors work. They all add the group scope by default if the whitelisting/filtering is requested, this was missing from gitlab.

Adding this to the server handler would have its benefits, but GitHub connector which supports teams as well is an outlier in this case and I haven't checked the others.

srenatus:
> They all add the group scope by default if the whitelisting/filtering is requested, this was missing from gitlab.

Ah. I wasn't aware of this. Thanks.

PR #1522, opened 2019-08-14
storage/kubernetes: adding `skipCRDValidation` option

### LINKED ISSUE #1513: kubernetes storage: allow option to disable CRD creation

venezia:
Dex shouldn't require the ability to create custom resources defintion records
in the kubernetes api.  This will make it so that it can be configured to not
presume it has the ability to create the necessary CRD records.

Updated the documentation to detail this option and to give an example of a rbac
profile where dex does not have cluster-wide privileges to its resources nor
the ability to create CRD records.

Fixes #1513

### DIFF
--- a/Documentation/storage.md
+++ b/Documentation/storage.md
@@ -85,19 +85,20 @@ status:
 
 Once the `CustomResourceDefinition` is created, custom resources can be created and stored at a namespace level. The CRD type and the custom resources can be queried, deleted, and edited like any other resource using `kubectl`.
 
-dex requires access to the non-namespaced `CustomResourceDefinition` type. For example, clusters using RBAC authorization would need to create the following roles and bindings:
+### Permissions
+
+dex may require access to the non-namespaced `CustomResourceDefinition` type if it configured to create the necessary Custom Resource Definitions for you.
+For example, clusters using RBAC authorization would need to create the following roles and bindings:
+
 ```
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRole
 metadata:
   name: dex
 rules:
-- apiGroups: ["dex.coreos.com"] # API group created by dex
-  resources: ["*"]
-  verbs: ["*"]
 - apiGroups: ["apiextensions.k8s.io"]
   resources: ["customresourcedefinitions"]
-  verbs: ["create"] # To manage its own resources identity must be able to create customresourcedefinitions.
+  verbs: ["*"] # To manage its own resources identity must be able to create customresourcedefinitions.
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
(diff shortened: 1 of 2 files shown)


kenperkins:
Is the expectation here that you would then be required to create the CRDs externally from dex, thus having different permissions for creation of CRDs versus consumption?

venezia:
> Is the expectation here that you would then be required to create the CRDs externally from dex, thus having different permissions for creation of CRDs versus consumption?

@kenperkins yes - the ability to create a custom resource definition is global in scope (and obviously it can create many different CRDs) so it would be reasonable to say "dex shouldn't be able to do this" - so dex wouldn't be able to create `customresourcedefinitions.apiextensions.k8s.io` while still saying that dex should be able to manage custom resources that have been declared already (presumably by you or a cluster administrator, etc.) - namely things like `*.dex.coreos.com`

Another way to think about it, in the vernacular of say mysql, is that dex should be able to manage a database called `dex`, but it shouldn't be able to _create_ a database named `dex` - this isn't a perfect analog, but its similar in some respects.

kenperkins:
I wonder if it would be cleaner to not require the config value, and rather just configure the rbac, and have dex gracefully fail if crd's don't exist and it can't create them. (Just thinking outloud, I'm not an expert here)

lion7:
I know this is an old issue, but I think this is a great addition to Dex. 

In my use case, I am deploying Dex alongside my application to offload some of the OIDC complexities to Dex. I don't want Dex to have cluster rights, because this is deployed on a cluster that is shared by many tenants. Each tenant only has full admin access to their own namespace. So while it's fine to give full access to CR's in the same namespace, I don't want to give cluster-wide access to CRDs.

Right now I am creating the CRDs as a cluster admin upfront, and want Dex to only manage the CRs in it's own namespace. But for Dex to properly start I have to at least give it access to list CRDs cluster-wide. This is fine for now, but not desired.

nabokihms:
Fixed by https://github.com/dexidp/dex/pull/4543

PR #1530, opened 2019-08-22
connector/ldap: display login error

bonifaido:
`CallbackConnectors` display the error message to the user, which is quite nice for debugging purposes however, `PasswordConnectors` haven't returned it, this fixes that.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -373,7 +373,7 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
 		identity, ok, err := passwordConnector.Login(r.Context(), scopes, username, password)
 		if err != nil {
 			s.logger.Errorf("Failed to login user: %v", err)
-			s.renderError(w, http.StatusInternalServerError, "Login error.")
+			s.renderError(w, http.StatusInternalServerError, fmt.Sprintf("Login error: %v", err))
 			return
 		}
 		if !ok {


server/handlers.go
```diff
@@ -373,7 +373,7 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
 		identity, ok, err := passwordConnector.Login(r.Context(), scopes, username, password)
 		if err != nil {
 			s.logger.Errorf("Failed to login user: %v", err)
-			s.renderError(w, http.StatusInternalServerError, "Login error.")
+			s.renderError(w, http.StatusInternalServerError, fmt.Sprintf("Login error: %v", err))
```
srenatus: Which errors could this be? Is there anything potentially sensible about them? I.e. we don't want let someone guess usernames this way... 💭

bonifaido: I'm not sure, after checking the two backends LDAP and Keystone from Dex code point of view (all the PasswordConnectors) there is no error message which would reveal something like that. Of course, if the upstream provider/lib would return something like that (we wrap those errors) that could be dangerous, but in that case, there is a security concern in that software already.

Currently, Dex returns `Internal Server Error - Login error.` even in the case, the user mistyped the password or the LDAP/Keystone server is inaccessible.

bonifaido:
Welcome :)

PR #1538, opened 2019-08-30
Convert k8s connector config to encode in a string 

erwinvaneyk:
Currently, in the Kubernetes storage, the connector config is stored as a JSON object in a byte array. This reduces readability, and prevents creating or editing the config in the Kubernetes resource.

This PR changes this by introducing the `configString` field to the Kubernetes storage struct of the connector. DEX stores the configuration as a JSON object (to avoid impacting the whole system) into this field, and is able to read both a YAML or JSON-encoded config from it (ff9fc64).

By keeping the original config (for now), this change is backward-compatible. DEX will prefer `configString`, but fallback to reading from `config` to be able to support configuration using the old structure.

Old config structure:
```yaml
apiVersion: dex.coreos.com/v1
kind: Connector
metadata:
  name: github
type: github
id: github
config: e30K
```

New config structure:
```yaml
apiVersion: dex.coreos.com/v1
kind: Connector
metadata:
  name: github
id: github
type: github
configString: |
  clientID: $GITHUB_CLIENT_ID
  loadAllGroups: true
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
@@ -31,6 +30,7 @@ import (
 	"github.com/dexidp/dex/pkg/log"
 	"github.com/dexidp/dex/storage"
 	"github.com/felixge/httpsnoop"
+	"github.com/ghodss/yaml"
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus"
@@ -475,7 +475,7 @@ func openConnector(logger log.Logger, conn storage.Connector) (connector.Connect
 	connConfig := f()
 	if len(conn.Config) != 0 {
 		data := []byte(string(conn.Config))
-		if err := json.Unmarshal(data, connConfig); err != nil {
+		if err := yaml.Unmarshal(data, connConfig); err != nil {
 			return c, fmt.Errorf("parse connector config: %v", err)
 		}
 	}
--- a/storage/kubernetes/types.go
+++ b/storage/kubernetes/types.go
@@ -667,8 +667,10 @@ type Connector struct {
 	Type            string `json:"type,omitempty"`
 	Name            string `json:"name,omitempty"`
 	ResourceVersion string `json:"resourceVersion,omitempty"`
-	// Config holds connector specific configuration information
-	Config []byte `json:"config,omitempty"`
+	// Config holds connector specific configuration information in a YAML or JSON struct
+	// The config here is a string rather than a byte-array to improve
+	// readability/editing in the k8s Connector CRD.
+	Config string `json:"config,omitempty"`
 }
 
 func (cli *client) fromStorageConnector(c storage.Connector) Connector {
@@ -685,7 +687,7 @@ func (cli *client) fromStorageConnector(c storage.Connector) Connector {
 		Type:            c.Type,
 		Name:            c.Name,
 		ResourceVersion: c.ResourceVersion,
-		Config:          c.Config,
+		Config:          string(c.Config),
 	}
 }
 
@@ -695,7 +697,7 @@ func toStorageConnector(c Connector) storage.Connector {
 		Type:            c.Type,
 		Name:            c.Name,
 		ResourceVersion: c.ResourceVersion,
-		Config:          c.Config,
+		Config:          []byte(c.Config),
 	}
 }
 
--- a/storage/kubernetes/types.go
+++ b/storage/kubernetes/types.go
@@ -4,7 +4,7 @@ import (
 	"strings"
 	"time"
 
-	jose "gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2"
 
 	"github.com/dexidp/dex/storage"
 	"github.com/dexidp/dex/storage/kubernetes/k8sapi"
@@ -667,13 +667,19 @@ type Connector struct {
 	Type            string `json:"type,omitempty"`
 	Name            string `json:"name,omitempty"`
 	ResourceVersion string `json:"resourceVersion,omitempty"`
-	// Config holds connector specific configuration information in a YAML or JSON struct
+
+	// Config holds connector specific configuration information in a JSON object.
+	// Deprecated: ConfigString is used instead.
+	Config []byte `json:"config,omitempty"`
+
+	// ConfigString is an alternative option to store the connector config.
 	// The config here is a string rather than a byte-array to improve
 	// readability/editing in the k8s Connector CRD.
-	Config string `json:"config,omitempty"`
+	ConfigString string `json:"configString,omitempty"`
 }
 
 func (cli *client) fromStorageConnector(c storage.Connector) Connector {
+
 	return Connector{
 		TypeMeta: k8sapi.TypeMeta{
 			Kind:       kindConnector,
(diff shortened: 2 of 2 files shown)


venezia:
FWIW, per [the docs](https://github.com/dexidp/dex/blob/master/Documentation/storage.md#kubernetes-custom-resource-definitions-crds)

> ... These resources are only designed to store state and aren't meant to be consumed by end users. For modifying dex's state dynamically see the API documentation.

While you can store state using kubernetes custom resources, its not intended for you to actually do `kubectl ...` on these resources.  The storage mechanism doesn't implement listwatch (nor use client-go) - it expects that the only changes done are being done by dex itself.

I'm not against the change necessarily, I just want to make sure its clear that you shouldn't be playing with the kubernetes resources directly - dex has a grpc api to do such

erwinvaneyk:
Hey @venezia - thanks for your review! I appreciate you taking the time to do so 😁  

> FWIW, per the docs
> 
> > ... These resources are only designed to store state and aren't meant to be consumed by end users. For modifying dex's state dynamically see the API documentation.

Thanks for pointing this out, I missed this when reading the docs! Personally, I don't think we should preemptively discard improvements or alternative approaches based on a 2 year old comment in the documentation. I hope we can weigh this change purely on its own merits and demerits.

> While you can store state using kubernetes custom resources, its not intended for you to actually do kubectl ... on these resources. The storage mechanism doesn't implement listwatch (nor use client-go) - it expects that the only changes done are being done by dex itself.

That is a good point, I noticed that (lack of client-go and listwatch) too. It seems to me, that the lack of client-go and listwatch is more of a legacy thing than a deliberate choice/constraint, but, then again, I may be missing the full story?

> I'm not against the change necessarily, I just want to make sure its clear that you shouldn't be playing with the kubernetes resources directly - dex has a grpc api to do such

Okay, let's leave the configuring of DEX through the CRDs directly aside for now. 🙂 I think this PR has merit enough because it improves the readability/debuggability of the Connector CRD.

venezia:
> Thanks for pointing this out, I missed this when reading the docs! Personally, I don't think we should preemptively discard improvements or alternative approaches based on a 2 year old comment in the documentation. I hope we can weigh this change purely on its own merits and demerits.

Oh I'm not trying to discard improvements, I was just trying to assert that, at least at one point, this was considered and somebody decided they did not think it was a great idea.  Times change and I think that we should always be open to changing things, just like you suggested that it should be evaluated on its own.  I was just trying to bring in context.

> That is a good point, I noticed that (lack of client-go and listwatch) too. It seems to me, that the lack of client-go and listwatch is more of a legacy thing than a deliberate choice/constraint, but, then again, I may be missing the full story?

The silly answer would be "_Do you want the application to be double in size?_" (gosh client-go is huge!) but in all seriousness, I do think a few things should be sorted out:

* If we're going to use client-go, we're probably going to want to use the code generators for the custom resources
  * This will pin us to a specific version of client-go, which may not be desired 
  * Remember, every 1.X release of Kubernetes results in a _major_ release of client-go, and with that the possibility of breaking changes (I always like when "_Most APIs will work_" is a [caveat](https://github.com/kubernetes/client-go#compatibility-matrix))
* Using client-go might help us when Kubernetes changes how service account tokens are handled in the future (the whole `BoundServiceAccountTokenVolume` thing and that service account tokens will no longer be forever tokens but rather temporal and thus the file will change over time) is logic that I'm guessing is not supported in the current implementation that dex uses
  * Do we use the [dynamic client](https://github.com/kubernetes/client-go/tree/master/dynamic) to try to side step some of this?

> Okay, let's leave the configuring of DEX through the CRDs directly aside for now. 🙂 I think this PR has merit enough because it improves the readability/debuggability of the Connector CRD.

Sure thing!

erwinvaneyk:
> Oh I'm not trying to discard improvements, I was just trying to assert that, at least at one point, this was considered and somebody decided they did not think it was a great idea. Times change and I think that we should always be open to changing things, just like you suggested that it should be evaluated on its own. I was just trying to bring in context.

Maybe my comment about this was a bit too harsh. I really appreciate the context, and it made me dig a bit more in the history to see what has been suggested.

> The silly answer would be "Do you want the application to be double in size?" (gosh client-go is huge!) but in all seriousness, I do think a few things should be sorted out: [...]

Haha, that is indeed a legit concern with the Kubernetes tooling! On the other hand, there also seem to be some legitimate concerns with sticking to a self-rolled client too (maintenance cost, performance, etc). In any case, that discussion seems out of the scope of this PR. Since you put a lot of thought on this already, should we open an issue to at least catalog/centralize that discussion there?

venezia:
> Haha, that is indeed a legit concern with the Kubernetes tooling! On the other hand, there also seem to be some legitimate concerns with sticking to a self-rolled client too (maintenance cost, performance, etc). In any case, that discussion seems out of the scope of this PR. Since you put a lot of thought on this already, should we open an issue to at least catalog/centralize that discussion there?

@erwinvaneyk I actually had the same initial thought when I first looked at the dex code - "no client-go? eww! but that explains why dex is so small!" - as I almost always just use client-go for anything k8s related.  After looking at it a bit more, possibly becoming more jaded with client-go, I'm unsure.  

This code _does_ work, its pretty lean, and while things come and go with any API, the overall limited engagement for how dex interacts with the API (with it truly just treating it as a storage vehicle, not treating it as a message bus or other patterns that you see with k8s api usage) - I don't think we will see the URL paths that k8s' apiserver uses for rest commands to change anytime soon.  Thus for a specific section of "maintenance" I'm unsure if it applies.

However, like I mentioned in a previous comment, the intent with service account tokens is to change the token on an interval, unlike the current forever-token situation that is done.  I would presume that client-go handles this well, while this code may not.  So our use of client-go would be to simply update how we parse kubeconfigs/in-cluster-environments - which is kinda silly on the surface of it, but you are right to suggest that it is typically better to use someone else's library than to roll your own.

I don't think the two of us should make such a decision though - it probably needs a consensus from the maintainers, etc.  For me, it falls under the "reasonable people could disagree" pool of decisions.

I would also like to note that in some situations, with some crazy configs and use cases, client-go _has_ failed me and I had to result in just bundling `kubectl` into my app and shelling out to it.  It made me sad and perhaps has made me feel that people who don't use client-go aren't necessarily wrong.  Like I said, I have mixed feelings.  Also at the time, kubectl had its own logic (note the [wg-apply group](https://github.com/kubernetes/community/tree/master/wg-apply) trying to make `apply` usable by people using the rest API vs being `kubectl` exclusive) - so things could have changed but I'm uncertain they have.

To the point of the direct change of this PR - I don't see an obvious problem with changing how the encoding happens for this nested section of the resource.

If you take the approach of "nobody should read/modify this resource directly" then you shouldn't really care how its encoded, thus if this makes other people happy, why not.  Unlike #1536 I don't think we have to worry about some custom resources being done _one way_ and other custom resources being done another.

Cheers!

erwinvaneyk:
> @erwinvaneyk I actually had the same initial thought when I first looked at the dex code - "no client-go? eww! but that explains why dex is so small!" - as I almost always just use client-go for anything k8s related. After looking at it a bit more, possibly becoming more jaded with client-go, I'm unsure.
> 
> This code does work, its pretty lean, and while things come and go with any API, the overall limited engagement for how dex interacts with the API (with it truly just treating it as a storage vehicle, not treating it as a message bus or other patterns that you see with k8s api usage) - I don't think we will see the URL paths that k8s' apiserver uses for rest commands to change anytime soon. Thus for a specific section of "maintenance" I'm unsure if it applies.
> 
> However, like I mentioned in a previous comment, the intent with service account tokens is to change the token on an interval, unlike the current forever-token situation that is done. I would presume that client-go handles this well, while this code may not. So our use of client-go would be to simply update how we parse kubeconfigs/in-cluster-environments - which is kinda silly on the surface of it, but you are right to suggest that it is typically better to use someone else's library than to roll your own.
> 
> I don't think the two of us should make such a decision though - it probably needs a consensus from the maintainers, etc. For me, it falls under the "reasonable people could disagree" pool of decisions.
> 
> I would also like to note that in some situations, with some crazy configs and use cases, client-go has failed me and I had to result in just bundling kubectl into my app and shelling out to it. It made me sad and perhaps has made me feel that people who don't use client-go aren't necessarily wrong. Like I said, I have mixed feelings. Also at the time, kubectl had its own logic (note the wg-apply group trying to make apply usable by people using the rest API vs being kubectl exclusive) - so things could have changed but I'm uncertain they have.

The client-go vs. self-rolled is indeed a whole different discussion, far too large for this minor PR. I don't have a strong opinion on it though. It is just something that you notice when diving into the dex codebase.  

> To the point of the direct change of this PR - I don't see an obvious problem with changing how the encoding happens for this nested section of the resource.
>
> If you take the approach of "nobody should read/modify this resource directly" then you shouldn't really care how its encoded, thus if this makes other people happy, why not?

I am taking the literary freedom to interpret this as a "👍" from you. Great! 😜

PR #1539, opened 2019-08-30
Replace x/net/context with stdlib context

erwinvaneyk:
I was not able to build the project due to missing the "golang.org/x/net/context" (with `go version go1.12.9 darwin/amd64`). It looks like there are two options here to fix this: (1) add a dependency to golang.org/x/net/context, or (2) use the context from the standard library.

Since the comment in the code does no longer apply (it is two years old) and the linked issue seems to be resolved, I think we can safely move to use "context".

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -1,14 +1,11 @@
 package server
 
 import (
+	"context"
 	"errors"
 	"fmt"
 
 	"golang.org/x/crypto/bcrypt"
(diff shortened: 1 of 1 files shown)


venezia:
dex's official docker image is built against 1.12.9 already - see #1529 

perhaps there is something else amiss?

Issue #1523, opened 2019-08-14
Graceful shutdown should be supported


liyimeng:
Today, dex dose not handle any signal from user or kubernetes. This might cause some resource open when dex process are killed, e.g. the storage.Close function never get called.

nabokihms:
I think this issue was resolved in #1942

sagikazarmark:
Hm, is storage.Close also called somewhere?

nabokihms:
I don't think so 😅 . Looks like we have to add it.

PR #1541, opened 2019-08-31
WIP: Gracefully shutdown server

liyimeng:
A simple fix to allow graceful shutdown on the server. 
Try to resolve issue https://github.com/dexidp/dex/issues/1523
but only storage close is implemented at the moment. More should be done on http servers and gRPC server.


sagikazarmark:
Fixed in #1991

PR #1550, opened 2019-09-23
storage/mysql: support pre-5.7.20 instances with tx_isolation only

bonifaido:
MySQL 5.7.20 introduced the `transaction_isolation` system variable, so pre 5.7.20 versions won't work with Dex MySQL storage (like AWS Aurora currently). This PR adds some basic detection logic for those use cases. See https://github.com/dexidp/dex/pull/1485#issuecomment-533977858 for more details. Once we can't get rid of the `SERIALIZABLE` requirement, this logic can be removed.

### DIFF
--- a/storage/sql/config.go
+++ b/storage/sql/config.go
@@ -29,6 +29,7 @@ const (
 	// MySQL error codes
 	mysqlErrDupEntry            = 1062
 	mysqlErrDupEntryWithKeyName = 1586
+	mysqlErrUnknownSysVar       = 1193
 )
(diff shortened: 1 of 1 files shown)


bonifaido:
We could do a `db.Query("SELECT VERSION()")` and compare the semantic versions with `github.com/coreos/go-semver` (that lib is an indirect dependency in the project already).

bonifaido:
Merging this, for now, later on we can refine this check if any problems occur.

PR #1558, opened 2019-09-28
Fix URLs in curl cmd as stated in the overview doc.

aijingyc:
As the title.


bonifaido:
LGTM thanks @aijingyc !

PR #1560, opened 2019-10-07
Use system certs by default for oauth2 gitlab calls

hightoxicity:
If no httpClient is defined for gitlab connector, create one with system trusted ca bundle.


bonifaido:
Hi @hightoxicity, why is this change needed? I'm concerned a bit, that client authentication by certificates is not used in this case.

hightoxicity:
Hi guy, our gitlab instance is presenting a self-signed certificate (our internal root CA), dex must trust the root CA used to issue the gitlab certificate, I have not found any way to provide a trusted cert bundle... Thx

jenting:
There exists a config `tlsClientCA` which used to specify the root CA.

hightoxicity:
Afaik tlsClientCA is only  to configure

jenting:
> Afaik tlsClientCA is only used to configure grpc client

Okay, I was thinking is it better to add `tlsClientCA` by configuration? Because in this way, the user provides the CA certificate in config like LDAP connector https://github.com/dexidp/dex/blob/a901e2f2048de70c30df49781a4fbb9aa3413f4f/connector/ldap/ldap.go#L70-L71 and https://github.com/dexidp/dex/blob/a901e2f2048de70c30df49781a4fbb9aa3413f4f/connector/ldap/ldap.go#L76-L77

PR #1563, opened 2019-10-08
AuthErr should return a 500 if the error type is server_error

chrismcbride:


nabokihms:
The code was rewritten and is compliant with the OIDC specification now. Closing.

If you still have an issue, please feel free to open a new PR.

PR #1566, opened 2019-10-10
add preffered_username to idToken

bonifaido:
Signed-off-by: Nandor Kracser <bonifaido@gmail.com>

Implemented preferred_username in ID Token for:

- GitHub
- GitLab
- LDAP

TODO:
- [ ] update docs after this gets finalized

Fixes: #1076

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -39,6 +39,7 @@ import (
 //         idAttr: uid
 //         emailAttr: mail
 //         nameAttr: name
+//         preferredUsernameAttr: uid
 //       groupSearch:
 //         # Would translate to the query "(&(objectClass=group)(member=<user uid>))"
 //         baseDN: cn=groups,dc=example,dc=com
@@ -103,9 +104,10 @@ type Config struct {
 		Scope string `json:"scope"`
 
 		// A mapping of attributes on the user entry to claims.
-		IDAttr    string `json:"idAttr"`    // Defaults to "uid"
-		EmailAttr string `json:"emailAttr"` // Defaults to "mail"
-		NameAttr  string `json:"nameAttr"`  // No default.
+		IDAttr                    string `json:"idAttr"`                // Defaults to "uid"
+		EmailAttr                 string `json:"emailAttr"`             // Defaults to "mail"
+		NameAttr                  string `json:"nameAttr"`              // No default.
+		PreferredUsernameAttrAttr string `json:"preferredUsernameAttr"` // No default.
 
 		// If this is set, the email claim of the id token will be constructed from the idAttr and
 		// value of emailSuffix. This should not include the @ character.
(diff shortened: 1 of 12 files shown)


connector/ldap/ldap.go
```diff
@@ -39,6 +39,7 @@ import (
 //         idAttr: uid
 //         emailAttr: mail
 //         nameAttr: name
+//		   preferredUsernameAttr: uid
```
srenatus: [nit] indentation seems a bit off here

server/handlers.go
```diff
@@ -175,6 +175,7 @@ func (s *Server) discoveryHandler() (http.HandlerFunc, error) {
 		Claims: []string{
 			"aud", "email", "email_verified", "exp",
 			"iat", "iss", "locale", "name", "sub",
+			"preferred_username",
```
srenatus: Is it proper to say that it's supported if only a subset does actually provide it...? 🤔

bonifaido: TBH, I have implemented it for these 3 providers since I have experience with that only ;) But I'm happy to extend the implementation. Should we remove `preferred_username` from the discovered claims until that?

sagikazarmark: Just a note: I would probably add a column to the connector list table in the readme.

bonifaido: Should I just remove `preferred_username` from the list in the meantime, until we get all connectors to implement this @srenatus ?

ffdybuster: Since `groups` (which also isn't supported by every provider) is also not in this list, I guess it makes more sense to remove it. But adding a column to the list in the README is still a good idea IMO.

bonifaido: all right! resolved :)

server/handlers.go
```diff
@@ -501,8 +502,8 @@ func (s *Server) finalizeLogin(identity connector.Identity, authReq storage.Auth
 		email = email + " (unverified)"
 	}
 
-	s.logger.Infof("login successful: connector %q, username=%q, email=%q, groups=%q",
-		authReq.ConnectorID, claims.Username, email, claims.Groups)
+	s.logger.Infof("login successful: connector %q, username=%q, preferred_username=%q, email=%q, groups=%q",
+		authReq.ConnectorID, claims.Username, claims.PreferredUsername, claims.Email, claims.Groups)
```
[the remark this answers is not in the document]
bonifaido: sorry, probably a mistake!

bonifaido:
Changes are done (as requested).

ffdybuster:
I built a docker container of this branch and tested it (sqlite3 storage). Works fine!

bonifaido:
Do you think we are okay to merge this @srenatus ?

server/handlers.go
ffdybuster:
```suggestion
		authReq.ConnectorID, claims.Username, claims.PreferredUsername, email, claims.Groups)
```
Your latest force-push reverted this.

bonifaido: Sorry, I haven't pulled before it! Thanks!

bonifaido:
Great to hear that, welcome! We also missed it, and will introduce it in prod in the coming days :)

PR #1568, opened 2019-10-13
Add note for redirect uri

life1347:


bonifaido:
Fixed by https://github.com/banzaicloud/bank-vaults/pull/698

PR #1570, opened 2019-10-16
Configurable cors support

kgoguevgoget:
Adds support for configuring all options present in Gorilla's CORS handler. All options configurable via an entry under the `Web` object in the configuration yml file.

// Access-Control-Allow-Origin
* allowedOrigins: ["*"]
// Access-Control-Allow-Headers
* allowedHeaders: ["custom-header-1", "custom-header-2"]
// Access-Control-Allow-Methods
* allowedMethods: ["PATCH", "DELETE"]
// Access-Control-Expose-Headers
* exposedHeaders: ["custom-header-1", "custom-header-2"] 
* ignoreOptions: true
// Access-Control-Allow-Credentials
* allowCredentials: true
// Access-Control-Max-Age
* maxAge: 10
* optionsStatuscode: 204

See Documentation/cors.md for more details

### DIFF
--- a/Documentation/cors.md
+++ b/Documentation/cors.md
@@ -1,0 +1,91 @@
+# CORS
+
+CORS (cross-origin-resource-sharing) allows servers to specify not only who can access protected resources but also how those resources are accessed. For a more indepth explanation for CORS please refer to https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+
+To successfully make a request to one of Dex's protected endpoints a CORS configuration will have to be set up.
+
+Dex is built on top of Gorilla's router and is able to access the standard CORS package provided by Gorilla. This is accessed via the `handlers` package.
+See `"github.com/gorilla/handlers"` for mor information on included functionality.
+
+## Protected endpoints
+
+Any requests to the following endpoints provided by Dex during runtime is subject to CORS policies
+
+* /token
+* /keys
+* /userinfo
+* /.well-known/openid-configuration
+
+
+## Dex defaults
+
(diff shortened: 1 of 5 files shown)


sagikazarmark:
Isn't this something that should rather be handled by a frontend proxy?

Also, allowing to configure all of these settings sounds like an overkill. I mean headers, allow credentials, methods all sound like application specific. Changing them from config doesn't really makes sense to me.

kgoguevgoget:
> Isn't this something that should rather be handled by a frontend proxy?
Technically yes, CORS setting can be handled by a proxy 
> 
> Also, allowing to configure all of these settings sounds like an overkill. I mean headers, allow credentials, methods all sound like application specific. Changing them from config doesn't really makes sense to me.

The initial option to enable CORS was already in Dex, These extra options are exposed for those who for whatever reason simply can not make changes to their current proxy. Users do not have to set these if they don't want to. There is documentation detailing all of these additionally there are defaults that are set by the underlying Gorilla library. However during my work I was required to set more than was currently available. 

I see no reason not allow others the same degree of freedom for configuration.  Either use it or don't it should be up to the person deploying this service.

sagikazarmark:
> Either use it or don't, but it should be up to the person deploying this service.

I guess I don't see why someone would do that. I mean, the settings are application specific, not client specific. If the application is not configured properly, it's a bug that should be fixed.

Obviously there are settings that you want to control (like origin), but stuff like headers are specific to the application, not a running instance.

One could even consider this a security feature that you can't configure stuff that makes things more open.

PR #1572, opened 2019-10-23
Allow to prefix retrieved by gitlab groups

hightoxicity:
We have dex configured to connect to both AD and gitlab and acting as an OIDC for kubernetes, retrieved groups are used in RBAC...
Since users in gitlab can create groups, they are able to create a group in gitlab with the same name as the admin group in AD and benefit of a privileged escalation that way.

Letting us configure a group prefix for gitlab prevent them to do that.

Thx.


sagikazarmark:
Shouldn't this be a provider independent configuration instead? I mean, adding features like this to a single provider feels a bit weird.

My 2 cents.

jfrabaute:
I'm working on a similar feature for google groups as well.
Do we want to provide a regexp for filtering, so filtering can be done in a more powerful way?

(I haven't checked yet how to make this feature generic across provider, or if it's even possible)

sagikazarmark:
> (I haven't checked yet how to make this feature generic across provider, or if it's even possible)

That's exactly what we plan to do, but still looking for the best options.

jfrabaute:
From the current code, ConnectorConfig is an interface, and the structs are per connector. It seems to make sense to have the "FilterGroup" field per connector and filter in each connector.
A helper function could help to share some code maybe, but as a first iteration, I'll probably go with implementing this only for the google connector.
In the long term, if this becomes common to a lot of connector, then something more fancy, like having a base Config object shared between connectors (using embedding) with a shared function could be implemented, or something similar, but it seems overkill for just one or two cases so far.

jfrabaute:
Corresponding commit for the filtering on the google connector: https://github.com/jfrabaute/dex/commit/498f19eebd2ce2fd25c71dc2e516fa7233cf0a6d

sagikazarmark:
I've just opened the issue for the common middleware layer: #1635 

Personally, I'm against merging any more PRs adding per connector features. Would be nice to make this a top priority though.

PR #1574, opened 2019-10-29
[WIP] Add basic session management for certain connectors

felixfontein:
After reading #32 (and wanting such a feature, but for LDAP) I decided to try to add some basic session management. I've created a skeleton -- the cookie handling and configuration is missing -- to get some feedback, whether this is the right direction and approach.

The idea is as follows: when you authenticate with a connector which supports the `SessionConnector` interface (and when this is configured), a cookie is set which contains information on the used connector, the username, and when it expires (should be a JWT or something like that, so that the user can't just modify or fake it). If this cookie is found when the user wants to autenticate and it is still valid, the connector will be asked whether this user is still valid, and if it is, the user will be authenticated directly.

This mostly makes sense for password-based connectors. I've implemented it for the mock connector and the LDAP connector.

### DIFF
--- a/connector/connector.go
+++ b/connector/connector.go
@@ -37,6 +37,12 @@ type Identity struct {
 	ConnectorData []byte
 }
 
+type Session struct {
+	ConnectorID string
+	Username    string
+	ValidUntil  string
+}
+
 // PasswordConnector is an interface implemented by connectors which take a
 // username and password.
 // Prompt() is used to inform the handler what to display in the password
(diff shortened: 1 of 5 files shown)


sherifkayad:
@felixfontein do you plan on pursuing this further?

felixfontein:
@sherifkayad I haven't heard any kind of feedback so far, so I haven't looked at it for a long time. IMO this is still something that's very much needed.

sherifkayad:
@felixfontein I also believe so. Maybe one of the maintainers can give feedback? @ericchiang @rithujohn191 @sagikazarmark

PR #1576, opened 2019-10-30
Pick icons on login screen by connector type instead of ID

nabokihms:
We are picking icons by provider ID. For example:
1. User can type anything into ID field.
2. ID field don't match any css class name with icon
3. User have no icon even if he picked provider with predefined icon (github, gitlab, etc.)

Using type field will make user always have icons for all providers on login screen.
Because we know all possible provider types and can prepare icons for them.

Only icons that don't fit into the logic now is google-icon.svg and coreos-icon.svg.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -259,16 +259,15 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 	}
 
 	connectorInfos := make([]connectorInfo, len(connectors))
-	i := 0
-	for _, conn := range connectors {
-		connectorInfos[i] = connectorInfo{
+	for index, conn := range connectors {
+		connectorInfos[index] = connectorInfo{
 			ID:   conn.ID,
 			Name: conn.Name,
+			Type: conn.Type,
 			// TODO(ericchiang): Make this pass on r.URL.RawQuery and let something latter
 			// on create the auth request.
 			URL: s.absPath("/auth", conn.ID) + "?req=" + authReq.ID,
 		}
-		i++
 	}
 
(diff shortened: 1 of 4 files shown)


jenting:
With this PR change, how to deals with multiple LDAP connectors?

sagikazarmark:
> Only icons that don't fit into the logic now is google-icon.svg and coreos-icon.svg.

That's not quite true. There is a google provider now.

> With this PR change, how to deals with multiple LDAP connectors?

This PR only changes the logo, but not the provider name, so I think it should continue working just fine.

@nabokihms can you please rebase your branch from master to make GitHub actions happy?

jenting:
> > With this PR change, how to deals with multiple LDAP connectors?
> 
> This PR only changes the logo, but not the provider name, so I think it should continue working just fine.

Thanks, I got the point ;)

nabokihms:
@sagikazarmark rebased :)

server/handlers.go
```diff
@@ -259,16 +259,14 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 	}
 
 	connectorInfos := make([]connectorInfo, len(connectors))
-	i := 0
-	for _, conn := range connectors {
-		connectorInfos[i] = connectorInfo{
-			ID:   conn.ID,
```
sagikazarmark: @nabokihms Further thinking about this: How about leaving the ID here? The templates can be replaced and someone might depend on the ID in a custom template. The default templates can continue using the type.

nabokihms: It seems ok for me. I will fix it.

PR #1579, opened 2019-11-06
connector/authproxy: supports groups claim

gergo-nyiro-oi:
The groups are collected from the X-Remote-Group headers as it is
proposed at https://kubernetes.io/docs/admin/authentication/#authenticating-proxy.

The X-Remote-User header is handled as email address only if it contains
"@" character, and the Username is derived before the "@" sign.

The UserID of the Identity is not set because we do not have any
knowledge of the proxied authentication how should that be handled.

### DIFF
--- a/connector/authproxy/authproxy.go
+++ b/connector/authproxy/authproxy.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/dexidp/dex/connector"
 	"github.com/dexidp/dex/pkg/log"
@@ -47,11 +48,21 @@ func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connecto
 	if remoteUser == "" {
 		return connector.Identity{}, fmt.Errorf("required HTTP header X-Remote-User is not set")
 	}
-	// TODO: add support for X-Remote-Group, see
-	// https://kubernetes.io/docs/admin/authentication/#authenticating-proxy
+	remoteGroups, ok := r.Header["X-Remote-Group"]
+	if !ok {
+		remoteGroups = []string{}
+	}
+	if strings.Contains(remoteUser, "@") {
+		remoteUserName := strings.SplitN(remoteUser, "@", 2)[0]
+		return connector.Identity{
+			Username:      remoteUserName,
+			Email:         remoteUser,
+			EmailVerified: true,
+			Groups:        remoteGroups,
+		}, nil
+	}
 	return connector.Identity{
-		UserID:        remoteUser, // TODO: figure out if this is a bad ID value.
-		Email:         remoteUser,
-		EmailVerified: true,
+		Username: remoteUser,
+		Groups:   remoteGroups,
 	}, nil
 }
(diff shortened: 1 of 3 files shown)


bonifaido:
Hi @gergo-nyiro-oi, I just updated this branch to get the CI build fixed. Could you please rebase on master to get a straight commit history?

gergo-nyiro-oi:
Thanks @bonifaido for the update. I rebased my branch.

connector/authproxy/authproxy.go
```diff
@@ -47,11 +48,21 @@ func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connecto
...
+			Email:         remoteUser,
+			EmailVerified: true,
+			Groups:        remoteGroups,
+		}, nil
+	}
 	return connector.Identity{
-		UserID:        remoteUser, // TODO: figure out if this is a bad ID value.
-		Email:         remoteUser,
-		EmailVerified: true,
+		Username: remoteUser,
```
bonifaido: So now we return fewer fields here if there are no groups?

gergo-nyiro-oi: Groups could be empty list.
The UserID, Email and EmailVerified fields won't be set, if there was no '@' in the remoteUser.

gergo-nyiro-oi: However the UserID is not set in any case.
Does UserID have higher precedence than Username, should we set it in all the cases?

bonifaido: I think it has, but not sure!

The only thing I'm sure that we shouldn't break compatibility 🙂

connector/authproxy/authproxy.go
```diff
@@ -47,11 +48,21 @@ func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connecto
 	if remoteUser == "" {
 		return connector.Identity{}, fmt.Errorf("required HTTP header X-Remote-User is not set")
 	}
-	// TODO: add support for X-Remote-Group, see
-	// https://kubernetes.io/docs/admin/authentication/#authenticating-proxy
+	remoteGroups, ok := r.Header["X-Remote-Group"]
+	if !ok {
+		remoteGroups = []string{}
```
ereOn, line 53: I don't think this is necessary: `nil` slices have a perfectly valid `empty slice` semantic in Go.

It probably doesn't hurt, but it doesn't help either I think.

connector/authproxy/authproxy.go
ereOn: From what it's worse, I think not being too smart about the `email` field, and leaving it empty is probably a better option.

We don't know that the `X-Remote-User` contains an email and we probably shouldn't assume that unless the HTTP header had some precise email semantic (`X-Remote-Email` or something).

connector/authproxy/authproxy.go
gergo-nyiro-oi: The json presentation of the nil will be null, which may not the for the client, therefore I enforced empty list.

connector/authproxy/authproxy.go
gergo-nyiro-oi: Ok, I remove the Email and EmailVerified fields.
Can I set Username and UserID to the same value (remoteUser)?

caiorcferreira:
Any updates on this feature or why this PR was closed? This would be really useful!

gergo-nyiro-oi:
We had an ongoing discussion which field should be filled with `remoteUser`:
> Can I set Username and UserID to the same value (remoteUser)?

PR #1583, opened 2019-11-15
Add allowedGroups option for clients config

nabokihms:
### Description
Sometimes it may be necessary to restrict access to applications for certain groups. 
It's easy to achieve by adding allowedGroups options to client configuration.

For example: 
We want to allow access to all django admin panels for users in `django-admins` group. 
Using oauth2 proxy and and specifying allowedGroups to admin panels client is easy way to do so.


nabokihms:
Now, this branch grows from actual master. This feature can still be convenient for someone. So, I will try to keep this branch up to date.

evrardjp:
Are You still maintaining this? What's the reason this feature never merged? (I don't see it)

nabokihms:
There is the patch for the latest version
https://github.com/deckhouse/deckhouse/blob/cd2115ecdfabe6b0fc7c56663438632da08997bb/modules/150-user-authn/images/dex/patches/client-groups.patch

I think it was not merged because the feature is too specific, but I still want to have something like this in Dex

fhopfensperger:
@nabokihms I still think this very useful, e.g. if you are part of a lot of groups you might want to filter them out, otherwise the header will explode.
One suggestion, would it not better to support a regex instead? So we could filter the groups with a prefix.

PR #1602, opened 2019-12-17
Added testing of ListRefreshTokens in conformance tests

Teeed:
I have noticed that ListRefreshTokens is not tested while running testRefreshTokenCRUD. Implemented it analogically to testPasswordCRUD.


Teeed:
Ah, so that is because of k8s... Maybe k8s storage plugin should just opt-out of testPasswordCRUD?

PR #1607, opened 2019-12-25
feat: update etcd to go.etcd.io/etcd

choujimmy:
In [kubernetes](https://github.com/kubernetes/kubernetes) 1.17.0 and above, the `git.etcd.io/etcd` package has been referenced by default to replace `github.com/coreos/etcd`. If `dex` and `kubernetes` are referenced in the same project, it will cause etcd's authpb to be registered repeatedly and panic will occur.


yadzhang:
@ericchiang  Can you help to merge this pr? We really need this commit to help us use dex in kubernetes 1.17.0.

bonifaido:
Eric is not a maintainer anymore, let me review this @yadzhang .

sagikazarmark:
@choujimmy you are supposed to download the protobuf yourself and generate it. That way you don't have to import the whole application.

Upgrading etcd will probably cause issues for a different set of users, so even if we decide to go forward, the solution to the problem is not to import dex itself.

Issue #1608, opened 2019-12-27
Mysql Storage Upstream Issue workaround


PeopleRange:
Hi all,

I have set up a copy of dex on my Kubernetes cluster on GKE.
My setup does not use any connectors at this time (Setting up bit by bit).
I am still using the original template until the backend part works.
For storage, I opted for an in cluster deployment of Percona XtraDB Cluster deployed with Helm. (I know its only starting to be supported)

There appears to be an issue with the chosen package for MySql compatibility as visible in the following logs:

`time="2019-12-27T11:02:48Z" level=info msg="config issuer: https://oauth.peoplerange.com"
time="2019-12-27T11:02:48Z" level=info msg="reconnecting with MySQL pre-5.7.20 compatibility mode"
time="2019-12-27T11:02:48Z" level=info msg="config storage: mysql"
time="2019-12-27T11:02:48Z" level=info msg="config connector: local passwords enabled"
time="2019-12-27T11:02:48Z" level=info msg="listening (http) on 0.0.0.0:5556"
[mysql] 2019/12/27 12:48:53 packets.go:36: unexpected EOF
time="2019-12-27T12:48:53Z" level=error msg="Storage health check failed: create auth request: insert auth request: invalid connection"
`

I didn't see anything on the current issue list so I looked it up and came across the following issue in the upstream package https://github.com/go-sql-driver/mysql/issues/674

bonifaido:
Fixed by https://github.com/dexidp/dex/pull/1609

PeopleRange:
Happy to help :)

PR #1609, opened 2019-12-27
Fixed mysql EOF issue

PeopleRange:
The upstream package that Dex uses for MySQL storage support has the following bug https://github.com/go-sql-driver/mysql/issues/674.

This addition to the config.go file removes the dex based default of 5 idle DB connections when using the go-SQL-driver for MySQL.

fixes issue https://github.com/dexidp/dex/issues/1608


PeopleRange:
Happy to Help

PR #1610, opened 2019-12-27
Adding oidc email scope check

### LINKED ISSUE #1598: Failed to authenticate: missing "email" claim
I am connecting to a corporate OIDC provider, and our "privileged" accounts do not have an email associated to them. I am not emailing the user anyhow. Can we make email claim optional? Or mayhaps make it so that I can map it to another attribute, like username.

nabokihms:
This helps to avoid "no email claim" error if the email scope was not specified.

Closes #1598

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -262,15 +262,25 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 	if !found {
 		return identity, fmt.Errorf("missing \"%s\" claim", userNameKey)
 	}
+
+	hasEmailScope := false
+	for _, s := range c.oauth2Config.Scopes {
+		if s == "email" {
+			hasEmailScope = true
+			break
+		}
+	}
+
 	email, found := claims["email"].(string)
-	if !found {
+	if !found && hasEmailScope {
 		return identity, errors.New("missing \"email\" claim")
 	}
+
 	emailVerified, found := claims["email_verified"].(bool)
 	if !found {
 		if c.insecureSkipEmailVerified {
 			emailVerified = true
-		} else {
+		} else if hasEmailScope {
 			return identity, errors.New("missing \"email_verified\" claim")
 		}
 	}
--- a/connector/oidc/oidc_test.go
+++ b/connector/oidc/oidc_test.go
@@ -50,16 +50,19 @@ func TestHandleCallback(t *testing.T) {
 		userIDKey                 string
 		userNameKey               string
 		insecureSkipEmailVerified bool
+		scopes                    []string
 		expectUserID              string
 		expectUserName            string
+		expectedEmailField        string
 		token                     map[string]interface{}
 	}{
 		{
-			name:           "simpleCase",
-			userIDKey:      "", // not configured
-			userNameKey:    "", // not configured
-			expectUserID:   "subvalue",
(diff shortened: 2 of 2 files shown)


connector/oidc/oidc.go
```diff
@@ -302,10 +315,12 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 	identity = connector.Identity{
 		UserID:        idToken.Subject,
 		Username:      name,
-		Email:         email,
 		EmailVerified: emailVerified,
 		ConnectorData: connData,
 	}
+	if email != "" {
```
sagikazarmark: This check is not necessary because of default values.

connector/oidc/oidc.go
```diff
@@ -262,16 +262,29 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
 	}
-	emailVerified, found := claims["email_verified"].(bool)
-	if !found {
-		if c.insecureSkipEmailVerified {
+
+	for _, s := range c.oauth2Config.Scopes {
+		if s != "email" {
+			continue
+		}
+		email, found = claims["email"].(string)
```
sagikazarmark:
```suggestion

		email, found = claims["email"].(string)
```

connector/oidc/oidc.go
```diff
@@ -302,10 +315,12 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 	identity = connector.Identity{
 		UserID:        idToken.Subject,
 		Username:      name,
-		Email:         email,
```
sagikazarmark: I would revert this, because it doesn't change behavior at all.

sagikazarmark:
Further thinking about this: `email` claim will only be included in the returned token if email scope is requested. Given this wasn't the behavior before, one could interpret this as a breaking change. I'm not sure what the standard says about returning non-requested fields, but if nothing, then we should probably continue returning those fields in case the underlying provider returns them.

nabokihms:
Yep, it has sense. I will change the behavior according to your thoughts.

connector/oidc/oidc.go
```diff
@@ -262,15 +262,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 	if !found {
 		return identity, fmt.Errorf("missing \"%s\" claim", userNameKey)
 	}
+
+	emailInScopes := func() bool {
```
sagikazarmark: How about:

```
	hasEmailScope := false

	for _, s := range c.oauth2Config.Scopes {
		if s == "email" {
			hasEmailScope = true

			break
		}
	}
```

Feels a bit cleaner and more readable

connector/oidc/oidc.go
```diff
@@ -262,15 +262,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+				return true
+			}
+		}
+		return false
+	}()
+
 	email, found := claims["email"].(string)
 	if !found {
-		return identity, errors.New("missing \"email\" claim")
+		if emailInScopes {
```
sagikazarmark: You could merge these two `if`s. The default value of email is an empty string because of the type.

nabokihms: Yep, you are right. Fixed.

PR #1612, opened 2020-01-03
connector/ldap: add multiple user to group mapping

vi7:
**This PR changes Dex LDAP connector config structure!**

Add an ability to fetch user's membership from groups of a different type by specifying multiple group attribute to user attribute value matchers in the Dex config:
```yaml
groupSearch:
  userMatchers:
  - userAttr: uid
    groupAttr: memberUid
  - userAttr: DN
    groupAttr: member
```

  In other words the user's groups can be fetched now from ldap structure similar to the following:

    dn: cn=john,ou=People,dc=example,dc=org
    objectClass: person
    objectClass: inetOrgPerson
    sn: doe
    cn: john
    uid: johndoe
    mail: johndoe@example.com
    userpassword: bar

    dn: cn=qa,ou=Groups,ou=Portland,dc=example,dc=org
    objectClass: groupOfNames
    cn: qa
    member: cn=john,ou=People,dc=example,dc=org

    dn: cn=logger,ou=UnixGroups,ou=Portland,dc=example,dc=org
    objectClass: posixGroup
    gidNumber: 1000
    cn: logger
    memberUid: johndoe

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -134,6 +134,12 @@ type Config struct {
 
 		Scope string `json:"scope"` // Defaults to "sub"
 
+		// DEPRECATED config options. Those are left for backward compatibility.
+		// See "UserMatchers" below for the current group to user matching implementation
+		// TODO: should be eventually removed from the code
+		UserAttr  string `json:"userAttr"`
+		GroupAttr string `json:"groupAttr"`
+
 		// Array of the field pairs used to match a user to a group.
 		// See the "UserMatcher" struct for the exact field names
 		//
@@ -173,6 +179,23 @@ func parseScope(s string) (int, bool) {
 		return ldap.ScopeSingleLevel, true
 	}
 	return 0, false
+}
+
+// Build a list of group attr name to user attr value matchers.
+// Function exists here to allow backward compatibility between old and new
+// group to user matching implementations.
+// See "Config.GroupSearch.UserMatchers" comments for the details
+func (c *ldapConnector) userMatchers() []UserMatcher {
+	if len(c.GroupSearch.UserMatchers) > 0 && c.GroupSearch.UserMatchers[0].UserAttr != "" {
+		return c.GroupSearch.UserMatchers[:]
+	} else {
+		return []UserMatcher{
+			{
+				UserAttr:  c.GroupSearch.UserAttr,
+				GroupAttr: c.GroupSearch.GroupAttr,
+			},
+		}
+	}
 }
 
 // Open returns an authentication strategy using LDAP.
@@ -392,7 +415,7 @@ func (c *ldapConnector) userEntry(conn *ldap.Conn, username string) (user ldap.E
 		},
 	}
 
-	for _, matcher := range c.GroupSearch.UserMatchers {
+	for _, matcher := range c.userMatchers() {
 		req.Attributes = append(req.Attributes, matcher.UserAttr)
 	}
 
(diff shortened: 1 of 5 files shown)


vi7:
@bonifaido @sagikazarmark  @JoelSpeed  @srenatus @rithujohn191 please review

sagikazarmark:
@vi7 please be patient.

vi7:
@sagikazarmark sure, sorry for any inconvenience. Didn't know what's the proper process of a reviewer assignment

srenatus:
I've just had a brief look -- **Thank you for contributing this! 🎉** -- what do you think, would it be too much to ask for backwards-compatibility here? I suppose we could keep the old config keys, and process both old and new...? If it doesn't seem like too much of a burden, I'd say we'd better go the compatible way. 🤔

vi7:
hey @srenatus , thanks for the comment! I was actually thinking about that either but was probably too lazy :) Let me check however, will try to implement a backward compatible code

vi7:
@srenatus I've added `buildUserMatchers()` func which generates a valid group to user matcher from the previous connector config version for backward compatibility.
Backward compatibility tests added either. Please review

connector/ldap/ldap.go
```diff
@@ -165,6 +181,25 @@ func parseScope(s string) (int, bool) {
 	return 0, false
 }
 
+var userMatchers []UserMatcher
```
srenatus: Instead of making sure that this global is set properly in all the right code paths (I think this is what the calls sprinkled on the code below are for?) -- Can we make this a function that reads both config locations and just *returns* `[]UserMatcher`? Like

```go
func (c *ldapConnector) userMatchers() []UserMatcher {
		return c.GroupSearch.UserMatchers[:]
	} else {
		return []UserMatcher{
			{
				UserAttr:  c.GroupSearch.UserAttr,
				GroupAttr: c.GroupSearch.GroupAttr,
			},
		}
	}
}
```
and use *that function* where we use `userMatchers` currently?

(I don't believe performance is an issue here.)

vi7:
hey @srenatus ! thanks for the comment, that makes a perfect sense. Fix has been delivered, please review.

vi7:
@srenatus thanks, hope that'll happen soon :)

vi7:
hi guys! any updates on that?

vi7:
Hi @rvirdiz ,

are you talking about authorization or about authentication?

`groupSearch` parameters does not prevent users from logging in but rather help Dex to fetch user's groups from LDAP properly and further on provide a groups list to the K8S RBAC mechanism which will in turn decide whether to let user in to the cluster based on the Roles and Bindings

rvirdiz:
Hi @vi7 

Yes, I am looking for users returned from dex userSearch parameter, should not be allowed to login/authenticate if groupSearch parameter returns empty ldap group list.

Following are my dex logs, when user tries to login user dex ldap connector:
```
time="2020-06-15T19:40:55Z" level=info msg="performing ldap search ou=users,dc=example,dc=com sub (&(objectClass=posixAccount)(uid=myldapuser))"
time="2020-06-15T19:40:55Z" level=info msg="username \"myldapuser\" mapped to entry uid=myldapuser,ou=users,dc=example,dc=com"
time="2020-06-15T19:40:55Z" level=info msg="performing ldap search 
cn=kubeflow,ou=groups,dc=example,dc=com sub (&(objectClass=posixGroup)(memberUid=myldapuser))"
time="2020-06-15T19:40:55Z" level=info msg="login successful: connector \"ldap\", username=\"LDAP_user\", preferred_username=\"\", email=\"myldapuser@example.com\", groups=[\"kubeflow\"]"
```

In above logs, is there any way LDAP user "myldapuser" should not be allowed to login if not part of any groupSearch filter, groups=[].
Just want to avoid LDAP userSearch users to authenticate which are not part of whitelisted LDAP groups in groupSearch filter.

vi7:
@rvirdiz ,

I see what you mean and believe that it's not a part of the current LDAP connector functionality. You can either add that functionality urself and create a PR or shoot an issue and hope that somebody will implement

vudex:
I encourage you to not deprecate old config settings, because with them I can solve problem of indirect group membership in FreeIPA.
[See](https://github.com/dexidp/dex/discussions/2809)

I guess you are planning to change app architecture in the future, to remove connectors from the repo and just expose API, so there is also a no point of radically change connector and deprecating settings.

@nabokihms tagging you just in case

Issue #1619, opened 2020-01-10
config: No way to escape $ in connector-config


msiebuhr:
I've been playing with adding a SQL (Posgres) connector, as we have internal users there. To make that re-usable, I figured the query should be configurable, say:

```yaml
connectors:
- type: sql
  id: authdb
  name: Internal Auth DB 
  config:
    connection: host=localhost password=test ...
    query: "SELECT userId, ... WHERE username = $1 AND pwhash = crypt($2, pwhash)"
    groupQuery: ...
```

But the SQL placeholders (`$1` and `$2`) vanish because of the environment-expansion done in `cmd/dex/config.go`:

```go
connConfig := f()
if len(conn.Config) != 0 {
	data := []byte(os.ExpandEnv(string(conn.Config))) // - Here
	if err := json.Unmarshal(data, connConfig); err != nil {
		return fmt.Errorf("parse connector config: %v", err)
	}
}
```
I can see there's some vaguely related work going on in https://github.com/dexidp/dex/issues/1099

This could be fixed by forking `os.ExpandEnv` to leave unknown env's alone (which would break if anyone set `1=foo`) OR break out the reflection-package, walk the config-tree and check every string-key for starting with a `$` and matching some env name.

dfresh613:
I'm also affected by this problem. Any feedback on when this fix will be accepted and merged?

nabokihms:
With the latest Dex version you can use the `DEX_EXPAND_ENV` env variable to disable expansion.
See https://github.com/dexidp/dex/commit/4cb5577e1133132d42c702f9fb70d6742a343f9f

Issue #1099, opened 2017-10-18
config: all sensitive fields should be configurable through a file or env var


ericchiang:
Spun out of https://github.com/coreos/dex/issues/968 and to address https://github.com/coreos/dex/pull/943

Currently we let parts of dex config reference environment variables:

```yaml
storage:
- type: postgres
  config:
    password: $PASSWORD
```

We restrict this to storage and connectors since '$' can show up in things like bcrypt hashes. There have been additional request to add this to other parts of the config https://github.com/coreos/dex/pull/943. I'd like to propose we stop doing this.

Overall this is really bad since it lets the environment inject arbitrary bytes directly into the config. You could put an entire YAML object in that environment variable and it'd replace the environment variable reference. This is both hacky and prevents other tools from sanely parsing dex's config (https://github.com/coreos/dex/issues/968).

Add additional typed fields to sensitive config fields which let them be filled in by a file or env var explicitly. I should be able to configure a password like:

```yaml
staticPasswords:
- email: john.doe@example.com
  username: john
  hashFromFile: /var/run/dex/john-bcrypt-hash
- email: jane.doe@coreos.com
  username: jane
  hashFromEnv: JANE_BCRYPT_HASH
```

jungziege:
Agreed this is a cleaner approach. How do you plan to handle backwards compatibility? Or do you plan to continue to allow expansion where it exists now.

ericchiang:
> Agreed this is a cleaner approach. How do you plan to handle backwards compatibility? Or do you plan to continue to allow expansion where it exists now.

* Don't add new env var expansion logic.
* Continue to accept the literal values e.g. `hash`
* Add new optional fields like `hashFromFile` and `hashFromEnv`

Or we could do something like

```yaml
- email: john.doe@exapmle.com
  username: john
  password:
    file: /var/run/dex/john-bcrypt-hash
- email: jane.doe@example.com
  username: jane
  password:
    env: JANE_PASSWORD
```

Either way, we'll continue to be backward compatible. New fields will be optional.

dnascimento:
Any updates?

dafelipe130:
+1'ing here as well. I'm coming from #1158 and I could not put the identity configmap to source control with this still lingering around

AlexMorreale:
would love this feature! :+1:

SongGithub:
please, please, push this forward. I would love to see this feature got released soon. 

[...]

PR #1620, opened 2020-01-10
Only replace whole config-keys with environments

msiebuhr:
Reflect over and walk the loaded configuration and replace any $ENV keys
with their corresponding environment variable, but only of the whole key
matches.

This makes it possible to include '$' in configuration keys (e.g.
configured passwords; see #1619). It also removes the possibility for
injecting further JSON/yaml in the middle of a document through
env-expansion (see #1099).

On the flip-side, this makes it impossible to do env-expansion on
anything non-string (i.e. numbers) through environment variables, as
`"$name"` is an invalid number.

Previously, this was done by running the JSON through os.ExpandEnv(),
which does not do escaping, making it impossible to embed a '$' anywhere
in the storage or connector configurations.


msiebuhr:
I had a number of CI timeouts, so upped the linting-timeout to five minutes.

fmotrifork:
I believe that this PR would resolve this one https://github.com/dexidp/dex/pull/1428

msiebuhr:
Sort-of: This PR doesn't run environment-expansion on the overall configuration (only the parts it used to before). But fixing that should be quite trivial.

fmotrifork:
That would be a great solution in my opinion.
Which way do you guys think we should go? @bonifaido @sagikazarmark 

My goal is to keep all my secrets in Hashicorps Vault and use Banzaiclouds bank-vaults to inject them into Dex.

msiebuhr:
An alternative implementation could be to make a version of `os.ExpandEnv()` that respect escaping of dollar-signs, say `\$` or `$$` and continue to expand the raw input string.

msiebuhr:
Updated to replace keys in all of config. One gotcha is that it skips 'Hash' fields, as those tend to have non-environment names such as `$2a$...`

I did also consider skipping keys that aren't in env (i.e. `os.Getenv(name) == ""`). While that would leave such keys unchanged, it would likely expose accidentally unset environment names in the final config. I can't quite decide if that's a security issue or a help in debugging if `$DEX_SITE_NAME` or some such suddenly shows up in the UI.

nabokihms:
Starting from version 2.27 Dex's official container image has an entrypoint that runs a config file through [gomplate](https://docs.gomplate.ca/) templating engine. It is possible to use environment variables in config anywhere like this
```
telemetry:
  http: {{- getenv "DEX_TELEMETRY_HTTP" | quote }}
```
Previously added support for some field expansion with env variables is still in the code to keep backward compatibility.

I am going to close this PR. If you still have a question, please open a new issue or discussion. Thank you for your work!

PR #1623, opened 2020-01-13
microsoft: add useOnPremSamAccountName to get preferred_username

### LINKED ISSUE #1618: Microsoft Connector: allow access to on-prem username (onPremisesSamAccountName)
When using the Microsoft Connector, `sub` is a generated user id, and `username` is set to Microsoft Graph `displayName`. To integrate with existing systems, the on-premises username ("SAM Account Name") is often needed.

This can be fetched from Microsoft Graph by requesting the `onPremisesSamAccountName`:

[Microsoft Graph REST API v1.0 - User properties](https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0#properties)

## Proposal

Add a new boolean flag `useOnPremSamAccountName` to the Microsoft Connector configuration. The default value will be false and will not change the JWT-tokens/claims for existing installations. When set to `true`, the `preferred_username`-field will be populated with the value from `onPremisesSamAccountName`.

chlunde:
When Azure Active Directory is synchronized with an on-premises directory, the
boolean configuration parameter `useOnPremSamAccountName` can be set to get the
on-premises samAccountName as the `preferred_username` claim.

Fixes #1618

### DIFF
--- a/Documentation/connectors/microsoft.md
+++ b/Documentation/connectors/microsoft.md
@@ -80,6 +80,10 @@ connectors:
       tenant: organizations
 ```
 
+When Azure Active Directory is synchronized with an on-premises directory, the
+boolean configuration parameter `useOnPremSamAccountName` can be set to get the
+on-premises samAccountName as the `preferred_username` claim.
+
 ### Groups
 
 When the `groups` claim is present in a request to dex __and__ `tenant` is
--- a/connector/microsoft/microsoft.go
+++ b/connector/microsoft/microsoft.go
@@ -40,30 +40,32 @@ const (
 
 // Config holds configuration options for microsoft logins.
 type Config struct {
-	ClientID             string          `json:"clientID"`
-	ClientSecret         string          `json:"clientSecret"`
-	RedirectURI          string          `json:"redirectURI"`
-	Tenant               string          `json:"tenant"`
-	OnlySecurityGroups   bool            `json:"onlySecurityGroups"`
-	Groups               []string        `json:"groups"`
-	GroupNameFormat      GroupNameFormat `json:"groupNameFormat"`
-	UseGroupsAsWhitelist bool            `json:"useGroupsAsWhitelist"`
+	ClientID                string          `json:"clientID"`
+	ClientSecret            string          `json:"clientSecret"`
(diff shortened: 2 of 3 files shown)


chlunde:
@sagikazarmark Is there anything we can help with to move forward here?

chlunde:
@sagikazarmark @bonifaido hi, I noticed this comment

> Personally, I'm against merging any more PRs adding per connector features
https://github.com/dexidp/dex/pull/1633#issuecomment-576462786

is this why this PR is blocked?

I absolutely agree with this I think more per-connector features could be general and some could be per client (like required groups).

For this feature, do you have something like this in mind? A more dynamic remapping for all kinds of data to claims?

I think possibly
* CEL: https://github.com/google/cel-go
* JMESPath: https://grafana.com/docs/grafana/latest/auth/generic-oauth/#jmespath-examples
* OPA/Rego: https://www.openpolicyagent.org/docs/latest/integration/#integrating-with-the-go-api

could be used, but I have not looked into this at all.

On the downside, this would also increase the attack surface and make it easier to make mistakes...

sagikazarmark:
@chlunde that comment was rather meant for changes that duplicates logic between connectors.

I guess this one will always will be Microsoft specific.

TBH I'm not that familiar with it: does Microsoft ever return a `preferred_username` claim? If not, would it make sense to make this the default behavior?

chlunde:
@sagikazarmark I guess it depends on the user requirements - what field is the preferred_username? E-mail? "AD user"/on prem sam account name? Just a unique ID? Which is why I wonder if it would be better if connector implementations expose all fields/claims they have, without much configuration, and then you can have a generic implementation of mapping claims/fields (like email -> preferred_username, etc.), and filtering based on groups, not bound to each connector.

TristonianJones:
Hi, CEL maintainer here, if you're looking for a very limited DOS surface area, you can subset CEL to disable support for nearly every operator in the spec. I'd be happy to help explain how this can be accomplished if this is something you're interested in learning more about. 

Cheers!

chlunde:
@sagikazarmark should I just abandon this PR? I've moved to keycloak for now.

PR #1624, opened 2020-01-13
CloundFoundry connector

xtremerui:
Support for authenticating against [CloudFoundry](https://www.cloudfoundry.org/)

Documentation of [CloudFoundry UAA server](https://docs.cloudfoundry.org/api/uaa/version/74.4.0/index.html#overview)

cc @pivotal-jwinters


xtremerui:
Hi, just checking back. Since there is no feedback here yet so I am wondering are there certain requirements to meet for Dex to add a new connector? Is there a way to move forward for this PR or its just a simple `Nah not this connector`? 

Thank you for your time.

Albertoimpl:
Our team is also interested in this connector to adopt dex.
I saw it used to be supported but seems to be gone now: https://github.com/dexidp/dex/pull/542/files
Is there anything we can do to move this forward? 
Thanks!

mvdkleijn:
@xtremerui This also has merge conflicts... :smile:

xtremerui:
@sagikazarmark I kown there must be lots of  pending PRs in your list. Just wanted to get some feedback here so we could move forward. Thx!

sagikazarmark:
Generally, we are against adding new connectors to the codebase (at least until #1635 is figured out), but I'd hate to see all this work going to waste.

Another reason that keeps us from adding all potential connectors is the lack of resources to support/test them.

If we were to add CloudFoundry as a connector, it would be labeled as a "community maintained" connector, until we can figure out a way to move it out of the core as a plugin of some sort. We would refer to you whenever someone reports an issue.

Would that be acceptable for you @xtremerui ?

xtremerui:
@sagikazarmark that's fair enough. Concourse team would have a better set of tooling to maintain this connector. Once connectors are out of the core, Concourse org would take ownership of cloudfoundry connector as well.

connector/cf/cf.go
```diff
@@ -0,0 +1,394 @@
+package cf
```
sagikazarmark, line 1: I'd prefer if this would be something like `cloudfoundry`. Cf is too short and not descriptive.

connector/cf/cf.go
```diff
@@ -0,0 +1,394 @@
...
+	Name    string
+	GUID    string
+	OrgGUID string
+	Role    string
+}
+
+type Org struct {
+	Name string
+	GUID string
+}
```
sagikazarmark, line 78: Are these types exported for a reason?

sagikazarmark (changes requested):
A few nits

sagikazarmark:
@xtremerui let me know if you think this is ready.

And you are right: some sort of documentation would be nice. At least adding it to the list of connectors.

connector/cf/cf.go
xtremerui: Not sure the reason behind it. I will update them.

xtremerui:
@sagikazarmark the PR is updated. Thx for the reviewing.

We will submit a PR to dex docs too. BTW, in the readme `https://dexidp.io/docs/connectors/oauth/` is somehow broken  as in the connector listing, though the PR for oauth doc is already merged.

nabokihms:
I assume it fails because we changed the interface.

xtremerui:
@taylorsilva I assume we don't need Concourse to support Cloudfoundry anymore? Do we still looking to add this connector to Dex?

taylorsilva:
We do still need this. Last big update was a year ago here: https://github.com/concourse/dex/pull/148

I get that Dex may not want this though.

nabokihms:
@taylorsilva, this is not the means of wanting something. We want all the connectors! We just don't have the capacity to support them all. I personally have no CloudFoundry in my space, and I do not know any real clients (which doesn't mean there are not any, I just don't know them). Which means neither I nor any other maintainers can effectively answer community questions regarding CloudFoundry. 

My desire is that instead of accepting all the connectors, give users the way to easily add their own connectors.

The initial issue: https://github.com/dexidp/dex/issues/1907
The latest proposal: https://github.com/dexidp/dex/pull/4578

PR #1625, opened 2020-01-13
Optional Prometheus Registry

xtremerui:
Before the PR, Dex will break if `PrometheusRegistry` is not provided. IMHO that field doesn't need to be required unless there is a specfic reason (?).

With this PR, if user is not interesting at collecting Dex metric through Prometheus, one could config a Dex server without `PrometheusRegistry`.


sagikazarmark:
I would love to see OpenCensus or OpenTelemetry implemented in Dex. Until then, I think making Prometheus is okay.

PR #1626, opened 2020-01-14
SQL Connecor

msiebuhr:
Rudimentary support for using an SQL database for authentication (i.e. having a table of users, groups and passwords).

Requires #1620 (or something to the same effect) to work with Postgres (and others that use `$` for placeholders).


msiebuhr:
Re-reading things again, the I understand the `refresh()`-endpoint should be able to find out that users has been removed.

As we don't have the password around in `refresh()` (and I don't want to send the hash as an ID), it probably needs another SQL query to check if a user is still valid. I'm not very familiar with OIDC and friends, so I would appreciate if someone could point me in the right direction.

msiebuhr:
The code has been updated so `Refresh()` will return any new user data and reject users that no longer exist. It saves a copy of the username in `ConnectorData` if `OfflineAccess` is set.

msiebuhr:
Also, it now uses named placeholders, which aren't prefixed with dollar-signs (in any SQL-dialect that I can remember), so fixing the env-injection (#1620)) shouldn't be required for now.

PR #1627, opened 2020-01-14
google: Retrieve all the groups for a user

jfrabaute:
The list of groups is paginated (default page is 200), so when a user
has more than 200 groups, only the first 200 are retrieve.
This change is retrieving all the groups for a user by querying all the
pages.


jfrabaute:
For info, if this is merged, I also have a review to do filtering:

https://github.com/jfrabaute/dex/commit/498f19eebd2ce2fd25c71dc2e516fa7233cf0a6d

(Which seems to be related to @JoelSpeed comment: `// TODO (joelspeed): Make desried group key configurable`)

PR #1629, opened 2020-01-16
Client Credentials Grant

xtremerui:
so client can get access to Dex resources e.g. forwarding auth request and performing token exchange while handling callback from Dex.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -889,6 +889,8 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
 		s.withClientFromStorage(w, r, s.handlePasswordGrant)
 	case grantTypeTokenExchange:
 		s.withClientFromStorage(w, r, s.handleTokenExchange)
+	case grantTypeClientCredentials:
+		s.withClientFromStorage(w, r, s.handleClientCredentialsGrant)
 	default:
 		s.tokenErrHelper(w, errUnsupportedGrantType, "", http.StatusBadRequest)
 	}
@@ -1141,6 +1143,29 @@ func (s *Server) handleUserInfo(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", "application/json")
 	w.Write(claims)
+}
+
+func (s *Server) handleClientCredentialsGrant(w http.ResponseWriter, r *http.Request, client storage.Client) {
+	if err := r.ParseForm(); err != nil {
+		s.tokenErrHelper(w, errInvalidRequest, "Couldn't parse data", http.StatusBadRequest)
+		return
+	}
+	q := r.Form
+
+	nonce := q.Get("nonce")
+	scopes := strings.Fields(q.Get("scope"))
+
+	claims := storage.Claims{UserID: client.ID}
+
+	accessToken := storage.NewID()
+	idToken, expiry, err := s.newIDToken(r.Context(), client.ID, claims, scopes, nonce, accessToken, "", "client")
+	if err != nil {
+		s.tokenErrHelper(w, errServerError, fmt.Sprintf("failed to create ID token: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	resp := s.toAccessTokenResponse(idToken, accessToken, "", expiry)
+	s.writeAccessToken(w, resp)
 }
 
 func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, client storage.Client) {
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -1157,7 +1157,13 @@ func (s *Server) handleClientCredentialsGrant(w http.ResponseWriter, r *http.Req
 
 	claims := storage.Claims{UserID: client.ID}
 
-	accessToken := storage.NewID()
+	accessToken, _, err := s.newAccessToken(r.Context(), client.ID, claims, scopes, nonce, "client")
+	if err != nil {
+		s.logger.ErrorContext(r.Context(), "failed to create new access token", "err", err)
+		s.tokenErrHelper(w, errServerError, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
 	idToken, expiry, err := s.newIDToken(r.Context(), client.ID, claims, scopes, nonce, accessToken, "", "client")
 	if err != nil {
 		s.tokenErrHelper(w, errServerError, fmt.Sprintf("failed to create ID token: %v", err), http.StatusInternalServerError)
--- a/connector/authproxy/authproxy.go
+++ b/connector/authproxy/authproxy.go
@@ -83,20 +83,20 @@ type callback struct {
 }
 
 // LoginURL returns the URL to redirect the user to login with.
-func (m *callback) LoginURL(s connector.Scopes, callbackURL, state string) (string, error) {
+func (m *callback) LoginURL(s connector.Scopes, callbackURL, state string) (string, []byte, error) {
 	u, err := url.Parse(callbackURL)
 	if err != nil {
-		return "", fmt.Errorf("failed to parse callbackURL %q: %v", callbackURL, err)
+		return "", nil, fmt.Errorf("failed to parse callbackURL %q: %v", callbackURL, err)
 	}
 	u.Path += m.pathSuffix
 	v := u.Query()
 	v.Set("state", state)
 	u.RawQuery = v.Encode()
-	return u.String(), nil
+	return u.String(), nil, nil
 }
 
 // HandleCallback parses the request and returns the user's identity
-func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connector.Identity, error) {
+func (m *callback) HandleCallback(s connector.Scopes, _ []byte, r *http.Request) (connector.Identity, error) {
 	remoteUser := r.Header.Get(m.userHeader)
 	if remoteUser == "" {
 		return connector.Identity{}, fmt.Errorf("required HTTP header %s is not set", m.userHeader)
--- a/connector/authproxy/authproxy_test.go
+++ b/connector/authproxy/authproxy_test.go
@@ -36,7 +36,7 @@ func TestUser(t *testing.T) {
 		"X-Remote-User": {testUsername},
 	}
 
-	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, req)
+	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, nil, req)
 	expectNil(t, err)
 
 	// If not specified, the userID and email should fall back to the remote user
@@ -62,7 +62,7 @@ func TestExtraHeaders(t *testing.T) {
 		"X-Remote-User-Email": {testEmail},
 	}
 
-	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, req)
+	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, nil, req)
 	expectNil(t, err)
 
 	expectEquals(t, ident.UserID, testUserID)
@@ -85,7 +85,7 @@ func TestSingleGroup(t *testing.T) {
 		"X-Remote-Group": {testGroup1},
 	}
 
-	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, req)
+	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, nil, req)
 	expectNil(t, err)
 
 	expectEquals(t, ident.UserID, testEmail)
@@ -106,7 +106,7 @@ func TestMultipleGroup(t *testing.T) {
 		"X-Remote-Group": {testGroup1 + ", " + testGroup2 + ", " + testGroup3 + ", " + testGroup4},
 	}
 
-	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, req)
+	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, nil, req)
 	expectNil(t, err)
 
 	expectEquals(t, ident.UserID, testEmail)
@@ -132,7 +132,7 @@ func TestMultipleGroupWithCustomSeparator(t *testing.T) {
 		"X-Remote-Group": {testGroup1 + ";" + testGroup2 + ";" + testGroup3 + ";" + testGroup4},
 	}
 
-	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, req)
+	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, nil, req)
 	expectNil(t, err)
 
 	expectEquals(t, ident.UserID, testEmail)
@@ -158,7 +158,7 @@ func TestStaticGroup(t *testing.T) {
 		"X-Remote-Group": {testGroup1 + ", " + testGroup2 + ", " + testGroup3 + ", " + testGroup4},
 	}
 
-	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, req)
+	ident, err := callback.HandleCallback(connector.Scopes{OfflineAccess: true, Groups: true}, nil, req)
 	expectNil(t, err)
 
 	expectEquals(t, ident.UserID, testEmail)
--- a/connector/oauth/oauth.go
+++ b/connector/oauth/oauth.go
@@ -116,9 +116,9 @@ func (c *Config) Open(id string, logger *slog.Logger) (connector.Connector, erro
 	return oauthConn, err
 }
 
-func (c *oauthConnector) LoginURL(scopes connector.Scopes, callbackURL, state string) (string, error) {
+func (c *oauthConnector) LoginURL(scopes connector.Scopes, callbackURL, state string) (string, []byte, error) {
 	if c.redirectURI != callbackURL {
-		return "", fmt.Errorf("expected callback URL %q did not match the URL in the config %q", callbackURL, c.redirectURI)
+		return "", nil, fmt.Errorf("expected callback URL %q did not match the URL in the config %q", callbackURL, c.redirectURI)
 	}
 
 	oauth2Config := &oauth2.Config{
@@ -129,10 +129,10 @@ func (c *oauthConnector) LoginURL(scopes connector.Scopes, callbackURL, state st
 		Scopes:       c.scopes,
 	}
 
-	return oauth2Config.AuthCodeURL(state), nil
+	return oauth2Config.AuthCodeURL(state), nil, nil
 }
 
-func (c *oauthConnector) HandleCallback(s connector.Scopes, r *http.Request) (identity connector.Identity, err error) {
+func (c *oauthConnector) HandleCallback(s connector.Scopes, _ []byte, r *http.Request) (identity connector.Identity, err error) {
 	q := r.URL.Query()
 	if errType := q.Get("error"); errType != "" {
 		return identity, errors.New(q.Get("error_description"))
--- a/connector/oauth/oauth_test.go
+++ b/connector/oauth/oauth_test.go
@@ -50,7 +50,7 @@ func TestLoginURL(t *testing.T) {
 
 	conn := newConnector(t, testServer.URL)
 
-	loginURL, err := conn.LoginURL(connector.Scopes{}, conn.redirectURI, "some-state")
+	loginURL, _, err := conn.LoginURL(connector.Scopes{}, conn.redirectURI, "some-state")
 	assert.Equal(t, err, nil)
 
 	expectedURL, err := url.Parse(testServer.URL + "/authorize")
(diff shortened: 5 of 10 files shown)


karunchennuri:
@xtremerui I tested this PR on the master branch of dex. Worked perfectly fine. 
Somehow I had difficulty running local tests on the https://github.com/concourse/dex. That's the reason just cherry picked your PR. Thanks a lot for putting this together, I was reading some older posts on the reluctance to support client_credentials grant type, there must be a reason, I would be interested in knowing more on that from others on this thread.

xtremerui:
@karunchennuri glad it works. The test could be ran using the `Makefile` https://github.com/dexidp/dex/blob/2ca992e9b359eff69f9886d10620f255c4cff2f2/Makefile#L44-L45
```bash
make test
```

mvdkleijn:
@xtremerui Would you be open to update this PR since it has conflicts?

xtremerui:
@mvdkleijn absolutely. Most of my PRs got conflicts I will update all of them. Thank you!

xtremerui:
@sagikazarmark could I get some feedback on this PR please?

kellyma2:
@xtremerui I've added some changes to your original PR that fix a few problems.  See: https://github.com/kellyma2/dex/commit/a4414c271d5510d76ebff696da93f889ca1e6d7a

Can you add them to this PR?

xtremerui:
> @xtremerui I've added some changes to your original PR that fix a few problems. See: [kellyma2@a4414c2](https://github.com/kellyma2/dex/commit/a4414c271d5510d76ebff696da93f889ca1e6d7a)
> 
> Can you add them to this PR?

Sure thing. Thx for the fix!

Though, you will have to sign off your commit first so after I cherry-pick into this PR it won't fail github DCO checks.

kellyma2:
Done. https://github.com/kellyma2/dex/commit/19f212fb2846d900d77ce20fa49ea2bc14579afb

sagikazarmark:
Sorry for the delay reviewing this PR.

@xtremerui can you please clarify why this is useful?

From an OIDC perspective the `client_credentials` grant makes no sense because you don't receive an `id_token`.

From an OAuth2 perspective it might, but who do you want to talk to with the access token issued by Dex?

kellyma2:
We use it for non human, ie: robot, interactions.

sagikazarmark:
> We use it for non human, ie: robot, interactions.

Well, that much I've guessed.

> Not supporting it would force folks to use another provider such as Auth0.

I don't think that's necessarily true.

As far as I can tell, this is basically a convenience feature for issuing JWT tokens as access tokens (but that itself is a problem IMO, more on that later) and the fact that Dex is issuing these tokens is irrelevant.

I see a couple problems with this approach:

First of all, OpenID doesn't care about client credentials, because there is no identity involved. Dex is primarily an OpenID Connect provider. Although technically that also means it's an OAuth2 authorization server, it simply lacks the characteristics and integration with resource servers.

Moving onto the more pressing issues:

The current implementation accepts a list of scopes and issues an access token with that same list of scopes. Normally, during an authorization grant the user would have to authorize those scopes....or not. Without a user present, the server is supposed to make sure that a client cannot authorize itself for any scope (eg. by assigning a default list of approved scopes to the client or using any other solution the authorization server sees fit). None of that is happening in this case, which means the client can just authorize itself to do anything. Since there is no identity involved, the resource server can't even properly authorize these requests either other than trusting that the issued token can do all those things.

Another serious problem with using these access tokens is the targeted audience. If you take a closer look at the implementation access tokens are basically....ID tokens. The audience for ID tokens is the _client_ whereas the audience for access tokens is the _resource server_. One could say that this is a Dex implementation flaw, but the thing is OIDC (and Dex) is not traditional OAuth2, there is no target audience, other than the clients. If you decide to accept these tokens somewhere, it basically means you have to turn off audience validation (because the audience will always be the client requesting the token) which means all you have to do is get a valid token from somewhere (eg. an ID token from a public client auth flow) and you will be able to talk to the resource server, no problem.

Finally, there is the problem of the access token verification: the fact that access tokens issued by Dex today are JWTs is coincidental (basically they are JWTs so it would be easier for _Dex_ to validate them when the client calls the `/userinfo` endpoint). They could just be opaque tokens in which case the resource server would have to verify the token by talking to Dex....but Dex has no token verification endpoint. This is an internal detail of how Dex issues access tokens.


To sum it up:
- Using `client_credentials` grant in the context of OIDC is abusing the fact that OIDC builds on OAuth
- There is no scope authorization whatsoever which opens a window for privilege escalation
- You need to disable audience verification which means a user could easily gain access to an API by simply getting an ID token
- JWT as access token is not mandated by OIDC or OAuth, it's an implementation detail for Dex


For these reasons, I find the current implementation and the idea to use `client_credentials` grant with OIDC quite problematic.

The reason why I asked for a use case, because my guess you only need something that issues JWT tokens that you can verify in a backend API for example. I don't think you necessarily need Dex for that, you can simply integrate a basic OAuth2 server into your application.

Am I wrong in any of my assumptions above? If so, please prove me wrong and I'm happy to reconsider my position.

sagikazarmark:
I stand corrected on scopes: unknown scopes are in fact ignored, but that's also coincidental.

ericchiang:
Since client_credentials can't be used with 2FA (U2F, WebAuthN, even OTP), I think its value is limited in real world scenarios. Modern orgs are at the point where they're literally hunting down and disabling APIs that only use username/password auth without a second factor.

Dex's Access Tokens were never intended to have meaning outside of Dex itself. They're intended to be opaque to other services.

The concerns about scopes and audience could be resolved by configuration. E.g. you could limit what scopes or audiences a specific client_id could be issued.

If you need a flexible CA to issue credentials automation, you may want to look at HashiCorp Vault. Service-to-service credentials have different lifecycle and issuance requirements than human ones (e.g. you don't want a CI job to break when a user leaves the company). And you generally want to issue automation credentials based on the identity of the machine the automation is running on (e.g. a cloud VM identity[0][1] or kubernetes container identity[2]).

That being said, I don't work on Dex anymore.

The OAuth2 and OpenID Connect specs are routinely stretched in all sorts of interesting ways :) If you understand the draw backs but still feel that this is within Dex's scope, then please feel free to add support.

[0] https://cloud.google.com/compute/docs/instances/verifying-instance-identity
[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
[2] https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection

kellyma2:
> As far as I can tell, this is basically a convenience feature for issuing JWT tokens as access tokens (but that itself is a problem IMO, more on that later) and the fact that Dex is issuing these tokens is irrelevant.

Why does dex even issue access tokens at all, if not to use them?

> The current implementation accepts a list of scopes and issues an access token with that same list of scopes. Normally, during an authorization grant the user would have to authorize those scopes....or not. Without a user present, the server is supposed to make sure that a client cannot authorize itself for any scope (eg. by assigning a default list of approved scopes to the client or using any other solution the authorization server sees fit). None of that is happening in this case, which means the client can just authorize itself to do anything. Since there is no identity involved, the resource server can't even properly authorize these requests either other than trusting that the issued token can do all those things.

In practice, scopes are not super important in the client credentials sense.  Effectively, the robot in question is demonstrating that it is a valid client with respect to dex by providing a valid client id and client secret.

> 
> Another serious problem with using these access tokens is the targeted audience. If you take a closer look at the implementation access tokens are basically....ID tokens. The audience for ID tokens is the _client_ whereas the audience for access tokens is the _resource server_. One could say that this is a Dex implementation flaw, but the thing is OIDC (and Dex) is not traditional OAuth2, there is no target audience, other than the clients. If you decide to accept these tokens somewhere, it basically means you have to turn off audience validation (because the audience will always be the client requesting the token) which means all you have to do is get a valid token from somewhere (eg. an ID token from a public client auth flow) and you will be able to talk to the resource server, no problem.

I'm aware of the implementation - part of this PR is my work. :)  Regarding being able to use *any* access token, this assumes that the token is still valid (not expired and not revoked).  That's an interesting point though - probably want to restrict client credentials to require a client secret....

Regarding audience validation:

https://benohead.com/blog/2018/07/05/oauth-2-0-openid-connect-explained/#:~:text=of%20access%20permissions.-,Audience,only%20valid%20for%20certain%20purposes.

"It is the responsibility is the resource server to determine whether the token should be used or not. A resource server may choose to ignore the audience claim and accept any valid token."

For my purposes, that's a-ok.


> 
> Finally, there is the problem of the access token verification: the fact that access tokens issued by Dex today are JWTs is coincidental (basically they are JWTs so it would be easier for _Dex_ to validate them when the client calls the `/userinfo` endpoint). They could just be opaque tokens in which case the resource server would have to verify the token by talking to Dex....but Dex has no token verification endpoint. This is an internal detail of how Dex issues access tokens.

I'm happy to work up a token verification endpoint for dex. :)  I can even look at it next week.

> 
> To sum it up:
> 
> * Using `client_credentials` grant in the context of OIDC is abusing the fact that OIDC builds on OAuth
> * There is no scope authorization whatsoever which opens a window for privilege escalation
> * You need to disable audience verification which means a user could easily gain access to an API by simply getting an ID token
> * JWT as access token is not mandated by OIDC or OAuth, it's an implementation detail for Dex
> 
> For these reasons, I find the current implementation and the idea to use `client_credentials` grant with OIDC quite problematic.
> 
> The reason why I asked for a use case, because my guess you only need something that issues JWT tokens that you can verify in a backend API for example. I don't think you necessarily need Dex for that, you can simply integrate a basic OAuth2 server into your application.

Bigger picture is that we're integrating one or more external IdPs using dex on a per-cluster basis and wiring it all up to have Envoy talk to it.  It'd be a bit of a pain to do what you're desribing.

kellyma2:
> If you need a flexible CA to issue credentials automation, you may want to look at HashiCorp Vault. Service-to-service credentials have different lifecycle and issuance requirements than human ones (e.g. you don't want a CI job to break when a user leaves the company). And you generally want to issue automation credentials based on the identity of the machine the automation is running on (e.g. a cloud VM identity[0][1] or kubernetes container identity[2]).

Yes, that would be a sane approach generally speaking.  Unfortunately, in our case that'd be a bit messy.. we'd effectively be creating identies based on entire *classes* of containers.  This seems like the appropriate use case for client credentials, no?  As in, how is issuing a client id + client secret pair significantly different from issuing a username+password pair?

kellyma2:
fyi,I should point out that I'm not the original PR author.  I'm just using the PR in my setup

sagikazarmark:
> Why does dex even issue access tokens at all, if not to use them?

It should be used for accessing the `/userinfo` endpoint as per the OIDC standard.

> In practice, scopes are not super important in the client credentials sense. Effectively, the robot in question is demonstrating that it is a valid client with respect to dex by providing a valid client id and client secret.

It might not be important in _your_ use case, but APIs that use scopes for authorization would potentially be in a big trouble. Obviously, documentation can help, but it would be a funny thing to say "Hey, if you use scopes, client_credentials might make your API vulnerable if you issue tokens through Dex". Admittedly, configuration could be used to limit scopes.

> "It is the responsibility is the resource server to determine whether the token should be used or not. A resource server may choose to ignore the audience claim and accept any valid token."

Here is the deal though: if you accept any valid token, you accept ID tokens as well (because the way it's implemented in Dex). It's not that difficult to get an ID token for a user. Again, it might not be a problem for your use case, but it could mean a potential vulnerability in someone else's system.

> The concerns about scopes and audience could be resolved by configuration. E.g. you could limit what scopes or audiences a specific client_id could be issued.

I'm not 100% sure that's true for audiences. I mean, the audience is just hard-coded to the client ID.

> I'm happy to work up a token verification endpoint for dex.

Well, that's not my point at all. :)

> Bigger picture is that we're integrating one or more external IdPs using dex on a per-cluster basis and wiring it all up to have Envoy talk to it. It'd be a bit of a pain to do what you're desribing.

I'm not entirely sure why: can Envoy only talk to one OAuth2 server at a time. Anyway, it's still all too coincidental to me: The fact that Dex uses OAuth2 under the hood for OIDC doesn't mean it's a fully fledged OAuth2 server.

I'm afraid this is an incredibly deep rabbit hole: next thing we realize we are reviewing PRs for token introspection and revocation which are absolutely not in scope for Dex. And this PR basically makes an implementation detail (JWT access token) part of the "spec" without token introspection.

I still haven't heard an argument why introducing a separate OAuth2 server  for this purpose is impossible or a bad thing.

kellyma2:
> Here is the deal though: if you accept any valid token, you accept ID tokens as well (because the way it's implemented in Dex). It's not that difficult to get an ID token for a user. Again, it might not be a problem for your use case, but it could mean a potential vulnerability in someone else's system.

If you've gotten the id token or access token, you're vulnerable.  This is a given regardless, scopes or not. That's why access tokens are supposed to have a limited lifetime, no?

> 
> > The concerns about scopes and audience could be resolved by configuration. E.g. you could limit what scopes or audiences a specific client_id could be issued.
> 
> I'm not 100% sure that's true for audiences. I mean, the audience is just hard-coded to the client ID.

Would this concern not just be resolved by dealing with the audience correctly?  Why is this not a problem for existing user credentials?

> I'm not entirely sure why: can Envoy only talk to one OAuth2 server at a time. 

More specifically, in that this is Kubernetes we're talking about here, we're using Emissary-ingress, and no, it only allows a single AuthService to be configured. :)

> I'm afraid this is an incredibly deep rabbit hole: next thing we realize we are reviewing PRs for token introspection and revocation which are absolutely not in scope for Dex. And this PR basically makes an implementation detail (JWT access token) part of the "spec" without token introspection.

I'm not clear why you wouldn't want to support token revocation?  :)
As to the implementation detail being part of the spec, I hate to break it to you, but that horse has likely already left the barn long ago.  Implicit or not, it is effectively part of the spec.

sagikazarmark:
> If you've gotten the id token or access token, you're vulnerable. This is a given regardless, scopes or not. That's why access tokens are supposed to have a limited lifetime, no?

I'm not talking about _stolen_ tokens. You can legitimately get your own ID token (eg. when using any kind of public client auth). And without audience check, a user could access any of the services that accepts a Dex issued JWT token (id or access).

> Would this concern not just be resolved by dealing with the audience correctly? Why is this not a problem for existing user credentials?

What existing user credentials? :) Currently, the access token issued by Dex should only be consumed by Dex. So I guess technically you are right: Dex is vulnerable, although when it's the only audience apart from the client, this is probably less interesting.

> More specifically, in that this is Kubernetes we're talking about here, we're using Emissary-ingress, and no, it only allows a single AuthService to be configured. :)

Can you point me to some sort of documentation? I'd like to understand how they are consuming both OIDC and OAuth2 client credentials at the same time.

> As to the implementation detail being part of the spec, I hate to break it to you, but that horse has likely already left the barn long ago. Implicit or not, it is effectively part of the spec.

Technically, it's not. There is no token introspection implemented, so everything you rely on is implementation detail. That being said, I guess people would be mad if we changed it...

Let's pretend for a moment that we want to expand Dex and make it a fully fledged OAuth2 server. I'm afraid there is still a whole lot of work we would have to do to make it secure enough and we have a whole lot of other things on our plates right now. I guess we could hide it behind feature flags, exempt it from the backwards compatibility promise....

I need to think about it a bit more. I wouldn't mind to here from @xtremerui as well: how they use it, in what environment, etc.

kellyma2:
fwiw, the use case that @xtremerui described is very congruent with my own internal use case as well.

RealHarshThakur:
Hey folks, is there anything I can do to help this PR move forward?

cloudmarius:
Hi there, any intention still to merge this feature?

jarrettprosser:
I've tested the fork this change is based on and found that in the case of a public client, Dex will issue client credentials for the public client if I provide the client ID and leave out the secret value (or set it to null). This is an issue for our use; simply knowing the client ID (which is sent in plaintext by the UI when getting a token) is enough to get an access token. Not sure if that has been resolved in this PR though.

nabokihms:
With the [allowedGrants](https://github.com/dexidp/dex/blob/9a67dbd434da595e3a92d68454d24efa567ad886/server/server.go#L69C2-L69C19), I think now we can add the support for client credentials grant and make it not enabled by default.

@sagikazarmark what do you think?

smehboub:
Hello,

Thank you very much for your work.
Any news on progress ?

Thanks in advance :-)
Rgs

cardoe:
@sagikazarmark and @nabokihms wanted to re-raise what you mentioned about including this and leaving it disabled by default. What would you think?

Reason I ask is that I know folks are using this for robot / machine accounts. It might also be acceptable to add the ability to have the hidden connectors that robot / machine accounts utilize without having that prompt users.

testinfected:
@sagikazarmark We have exactly the same issue as @xtremerui and @kellyma2 with [Flyte](https://flyte.org). Flyte deploys in a Kubernetes env and has a concept of a worker which needs to authenticate to the Flyte control plane. We can only setup a single OIDC/OAuth2 server, and we use Dex for it (Okta connector). Flyte uses client credentials grant for the worker.

As per @nabokihms @cardoe suggestions, would it be fine to disable it by default. ~~Also, would need to add verification of the client secret as well. Currently only the ID is checked.~~

Thanks

sagikazarmark:
I'm inclined to accept this given the popular demand.

Unfortunately I have limited time, so if someone could run some tests, it would be immensely helpful.

The code change itself is relatively small, so once we have confirmation it works as expected, we should be able to get this merged relatively quickly.

(I'm constantly flooded by GitHub emails, so keep pinging me if I don't respond. Thanks! 🙏 )

cardoe:

testinfected:
I pulled this PR as well on my side and have tested it against Flyte (our use case). Works without issues. I have reviewed the code which looks alright (although I'm new to the codebase).

My 2 cents
Thanks guys

server/handlers.go
```diff
@@ -1116,6 +1118,35 @@ func (s *Server) handleUserInfo(w http.ResponseWriter, r *http.Request) {
 	w.Write(claims)
 }
 
+func (s *Server) handleClientCredentialsGrant(w http.ResponseWriter, r *http.Request, client storage.Client) {
```
sagikazarmark, line 1148: Correct me if I'm wrong, but client credentials grant isn't supposed to work with public clients, is it?

Shouldn't there be some sort of check here?

Technically, `withClientFromStorage` checks the secret, but public clients have empty secret, so it will always match.

If I'm right, this is a rather serious vulnerability.

@cardoe any chance you could give this a try in your build?

server/handlers.go
```diff
@@ -1116,6 +1118,35 @@ func (s *Server) handleUserInfo(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	q := r.Form
+
+	nonce := q.Get("nonce")
+	scopes := strings.Fields(q.Get("scope"))
+
+	claims := storage.Claims{UserID: client.ID}
+
+	accessToken, _, err := s.newAccessToken(r.Context(), client.ID, claims, scopes, nonce, "client")
```
sagikazarmark, line 1160: Is generating an access token always necessary? Shouldn't it be returned based on the requested scopes?

server/handlers.go
```diff
@@ -1116,6 +1118,35 @@ func (s *Server) handleUserInfo(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		s.logger.ErrorContext(r.Context(), "failed to create new access token", "err", err)
+		s.tokenErrHelper(w, errServerError, err.Error(), http.StatusInternalServerError)
```
sagikazarmark, line 1163: We've just merged #4457 removing internal error messages from responses. I think a generic error message should be returned here.

See the linked PR.

server/handlers.go
```diff
@@ -1116,6 +1118,35 @@ func (s *Server) handleUserInfo(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	idToken, expiry, err := s.newIDToken(r.Context(), client.ID, claims, scopes, nonce, accessToken, "", "client")
+	if err != nil {
+		s.tokenErrHelper(w, errServerError, fmt.Sprintf("failed to create ID token: %v", err), http.StatusInternalServerError)
```
sagikazarmark, line 1169: Same as above.

server/handlers.go
sagikazarmark: Also, it would probably be a good idea to cover at least the happy path with tests.

sagikazarmark (changes requested):
I really hate to be the gatekeeper (especially since @xtremerui has been rebasing this PR for years now), but I had some comments, one of them is potentially a blocker.

xtremerui:
> I really hate to be the gatekeeper (especially since @xtremerui has been rebasing this PR for years now), but I had some comments, one of them is potentially a blocker.

@taylorsilva are you interested to keep this going? I am not sure if Concourse still wants/needs this PR to be merged in Dex now.

@sagikazarmark thx for reviewing it again. ATM I don't have resource to move this PR forward but am happy to pass it to whomever.

taylorsilva:
Thanks for the ping Rui!

I'm happy to pick this up and incorporate the PR feedback. I have access to the branch so I can push changes to it. I don't work in the dex codebase often or look at this PR, so I'll need to carve out some time to pick up context. I'll throw this in my backlog. Happy to see others are using it though!

sagikazarmark:
Awesome, thanks @taylorsilva!

Feel free to ping me if you need me. I'm happy to fasttrack this, given the popular demand and the fact that's been open (and kept up-to-date) for a while now.

taylorsilva:
I was about to start working on this and was rebasing on `master` when I noticed https://github.com/dexidp/dex/pull/4583 landed a few weeks ago.

@sagikazarmark @nabokihms Correct me if I'm wrong, but I think this PR is moot now since #4583 implements the same thing, no?

taylorsilva:
Pinging again @sagikazarmark @nabokihms see previous comment.
TY 🙏

nabokihms:
It seems so, but I didn't have time to take a look on this PR.

cardoe:
Yes I think this is moot given #4583

taylorsilva:
@xtremerui or any Dex maintainers, feel free to close this PR. I can re-submit if I need to for some reason.

PR #1630, opened 2020-01-16
OAuth connector

xtremerui:
Hi, this implements an oauth connector, which could be used for generic authentication following OAuth2.0 framework.


mvdkleijn:
Nice! This would be most welcome in my opinion.

@xtremerui Would you mind adding a commit or two to

a) list the generic oauth connector in the table of the main readme; and
b) add documentation entry for those of use who are too lazy (or unable) to read the code (in Documentation/connectors)

I believe that would be helpful to get the PR to be accepted.

@bonifaido @sagikazarmark This PR has stalled basically on a single person's review, I believe it'd be nice to get this in (with above mentions additions) so I'm hoping one of you could take a look since @srenatus doesn't appear to have time?

**Disclaimer:** I'm in no way involved with either the Concourse or Dex team. I'm just a simple user trying to help the process as it were.

xtremerui:
I have updated the documents for oauth connector. Please let me know if there is anything I can do. Thx!

connector/oauth/oauth.go
```diff
@@ -0,0 +1,255 @@
...
+		c.userNameKey = "user_name"
+	}
+
+	if c.groupsKey == "" {
+		c.groupsKey = "groups"
+	}
+
+	identity.UserID, _ = userInfoResult[c.userIDKey].(string)
+	identity.Username, _ = userInfoResult[c.userNameKey].(string)
+	identity.PreferredUsername, _ = userInfoResult["name"].(string)
```
mvdkleijn, line 187: Unless I am mistaken, you guys wanted to make sure the `preferred_username` field is available for users.

However, hardcoding it to be the same as the `name` field does not seem desirable to me. Apart from that it also doesn't give the backend a chance to pass `preferred_username` if it tries to and effectively renders `preferred_username` fairly useless.

My suggestion would be to:
- default it to unset if its unset at the backend
- pass through the value unchanged if the backend **does** provide the field
- and allow a user to override the above two through a config entry which tells the connector to copy the value of another field to `preferred_username`

Please see the Atlassian Crowd connector for an example.

Again disclaimer: I'm not a member of Dex, just an occasional contributor trying to help both of you. :smile:

server/server.go
```diff
@@ -474,6 +475,7 @@ var ConnectorsConfig = map[string]func() ConnectorConfig{
 	"gitlab":          func() ConnectorConfig { return new(gitlab.Config) },
 	"google":          func() ConnectorConfig { return new(google.Config) },
 	"oidc":            func() ConnectorConfig { return new(oidc.Config) },
+	"oauth":           func() ConnectorConfig { return new(oauth.Config) },
```
mvdkleijn, line 478: some whitespace issues

connector/oauth/oauth.go
xtremerui: I think that makes sense. If there is connector that handles preferred_username already. I'd like to keep the behavior the same as much as possible.

server/server.go
xtremerui: hmm can't tell where the whitespace is. I took an existing line and just changed the name. It shown no difference.

mvdkleijn:
Hmm... I'm wondering if I made a booboo here... I believe that `preferred_username` is part of the OIDC spec and not the OAUTH spec. (whoops) If so, my apologies in advance.

Can we verify this? @sagikazarmark & @xtremerui ? I did not find any reference to preferred_username in the OAUTH 2 specs, only in the OIDC specs.

xtremerui:
@mvdkleijn I think you are right. Though I can't see how it affects the current implementation in this PR.

Upstream might or might not include `preferred_username` claim. 
If it does, users of oauth connection could leave the key not configured so it will pick up the claim. 
If it does not, the preferred_username field in `identity` could be used for any claim, which is more flexible since oauth connector meant to be a backup plan for auth provider that support Oauth but not OIDC.

mvdkleijn:
Sounds sensible enough, but I'll leave this decision to one of the Dex mainteiners like @sagikazarmark :smile: I'm juat happy to see there's some movement in these PRs again since they're so close to being merged.

Documentation/connectors/oauth.md
```diff
@@ -0,0 +1,49 @@
...
+    # Default: groups
+    # groupsKey:
+
+    # Optional: Configurable keys for user ID claim look up
+    # Default: user_id
+    # userIDKey:
+
+    # Optional: Configurable keys for preferred username claim look up
+    # Default: preferred_username
+    # preferredUsernameKey:
```
sagikazarmark, line 48: Can we add a claim mapping for these (except userIDKey) similar to the one introduced in #1634?

(`userIDKey` is not actually an OIDC claim, so it should remain separate)

connector/oauth/oauth.go
```diff
@@ -0,0 +1,261 @@
...
+		c.userNameKey = "user_name"
+	}
+
+	if c.groupsKey == "" {
+		c.groupsKey = "groups"
+	}
+
+	if c.preferredUsernameKey == "" {
+		c.preferredUsernameKey = "preferred_username"
+	}
```
sagikazarmark, line 189: I think these default should probably be assigned in the "constructor", no?

connector/oauth/oauth.go
```diff
@@ -0,0 +1,261 @@
...
+		identity.ConnectorData = connData
+	}
+
+	return identity, nil
+}
+
+func (c *oauthConnector) addGroupsFromMap(groups map[string]bool, result map[string]interface{}) error {
+	groupsClaim, ok := result[c.groupsKey].([]interface{})
+	if !ok {
+		return errors.New("cant convert to array")
```
sagikazarmark, line 223:
```suggestion
		return errors.New("cannot convert to slice")
```

connector/oauth/oauth_test.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	values := req.URL.Query()
+	values.Add("code", code)
+	req.URL.RawQuery = values.Encode()
+
+	return req
+}
+
+func expectEqual(t *testing.T, a interface{}, b interface{}) {
```
sagikazarmark, line 234: Maybe use testify instead?

sagikazarmark (changes requested):
Thanks @xtremerui 

I think it's a great feature.

Found a few nits, can you please take a look at those? Thanks!

Documentation/connectors/oauth.md
xtremerui: I wonder if `userIDKey` is not a standard claim (actullay i believe for oauth framework there is no standard claim, compared to OIDC), should we also include it in the claim mapping so user can configure it to match whatever they want from the auth provider?

I think the same case also apply to `email` and `email_verified`. Please correct me if there is a doc about standard claims of oauth.

sagikazarmark: I think in our case claims are interpreted from the Dex providing OIDC point of view. No, they are not oauth claims, but we map some information from oauth to the OIDC token.

The userIDKey is only used by Dex to determine which field to use as the user identifier for the specific connector. The rest of the keys are "mapped" into an OIDC token as standard claims.

That's how I see it at least, correct me if my thought process is wrong.

connector/oauth/oauth.go
xtremerui: I put them here to follow what OIDC does in the `handlerCallback()`

https://github.com/dexidp/dex/blob/2a282860fa8570433aa09827db6c5721b0ff4149/connector/oidc/oidc.go#L282-L310

Documentation/connectors/oauth.md
xtremerui: I have update the PR to include all identities mapping in a `ClaimMappings`. That is for all different kinds of non-standard user identity profile return by Oauth2 provider. I will list the user profile response from Reddit, AWS and LinkedIn for example.

[Reddit](https://github.com/reddit-archive/reddit/wiki/OAuth2-Quick-Start-Example#curl-example)
```
https://oauth.reddit.com/api/v1/me
{
    "comment_karma": 0, 
    "created": 1389649907.0, 
    "created_utc": 1389649907.0, 
    "has_mail": false,    
    "has_mod_mail": false, 
    "has_verified_email": null,   <--- emailVerifiedKey
    "id": "1",                               <--- userIDKey
    "is_gold": false, 
    "is_mod": true, 
    "link_karma": 1, 
    "name": "reddit_bot",           <--- preferredUsernameKey
    "over_18": true
}
```

[AWS](https://docs.aws.amazon.com/cognito/latest/developerguide/userinfo-endpoint.html)
```
/oauth2/userInfo
{
   "sub": "248289761001",      <--- userIDKey
   "name": "Jane Doe",           <--- userNameKey
   "given_name": "Jane",
   "family_name": "Doe",
   "preferred_username": "j.doe", <--- preferredUsernameKey
   "email": "janedoe@example.com"  <--- emailKey
}
```

[LinkedIn](https://docs.microsoft.com/en-us/linkedin/shared/integrations/people/profile-api)
```
   ...
   "localizedFirstName": "Bob",      <--- userNameKey
   "localizedHeadline": "API Enthusiast at LinkedIn",
   "vanityName": "bsmith",             <--- preferredUsernameKey
   "id":"yrZCpj2Z12",                      <--- userIDKey
   "localizedLastName": "Smith",
  ...
```
As you can see, the field for unique user identity could be `sub` or `id`, and the username field could has different names `vanityName` or `name`. So does the email verified key.

Id prefer this Oauth connector to be as general as possible so it could retrieve user profile from all kinds of responses from different provider and map them to dex identity.

xtremerui: I have put eveything but `userIDKey` under `ClaimMappings` and added an error if user ID is not found since it is critical. `EmailKey` and `EmailVerifiedKey` are added so it can handle a response like reddit:

[Reddit](https://github.com/reddit-archive/reddit/wiki/OAuth2-Quick-Start-Example#curl-example)
```
https://oauth.reddit.com/api/v1/me
{
    "comment_karma": 0, 
    "created": 1389649907.0, 
    "created_utc": 1389649907.0, 
    "has_mail": false,    
    "has_mod_mail": false, 
    "has_verified_email": null,   <--- emailVerifiedKey
    "id": "1",                               <--- userIDKey
    "is_gold": false, 
    "is_mod": true, 
    "link_karma": 1, 
    "name": "reddit_bot",           <--- preferredUsernameKey
    "over_18": true
}
```

b-jsshapiro:
@sagikazarmark This feature, which is absolutely essential to anyone developing for AWS, has been pending for more than a year. You requested changes in October that appear (?) to have been made. Merge is blocked, apparently because your requested changes are not addressed - at least so far as GitHub can see.

What needs to happen to move this forward? Can you say what remains to be addressed? If not, can you please accept the changes that have been made so that this can be released?

sagikazarmark:
@b-jsshapiro I don't really appreciate your tone.

As far as I can tell, this PR was scheduled for the last release and was potentially delayed because there were unresolved conflicts. It's scheduled for the next release.

Documentation/connectors/oauth.md
```diff
@@ -0,0 +1,64 @@
+# Authentication using Generic OAuth 2.0 provider
```
sagikazarmark, line 1: We've relocated the documentation to a new repository: https://github.com/dexidp/website

README.md
```diff
@@ -72,6 +72,7 @@ Dex implements the following connectors:
 | [SAML 2.0](https://dexidp.io/docs/connectors/saml/) | no | yes | no | stable | WARNING: Unmaintained and likely vulnerable to auth bypasses ([#1884](https://github.com/dexidp/dex/discussions/1884)) |
 | [GitLab](https://dexidp.io/docs/connectors/gitlab/) | yes | yes | yes | beta | |
 | [OpenID Connect](https://dexidp.io/docs/connectors/oidc/) | yes | yes | yes | beta | Includes Salesforce, Azure, etc. |
+| [Generic OAuth 2.0](https://dexidp.io/docs/connectors/oauth.md) | no | yes | yes | beta | |
```
sagikazarmark, line 75: Please add this to the documentation site. Also, it should be an alpha stability at first I think.

connector/oauth/oauth.go
```diff
@@ -0,0 +1,283 @@
...
+		c.groupsKey = "groups"
+	}
+
+	if c.emailKey == "" {
+		c.emailKey = "email"
+	}
+
+	if c.emailVerifiedKey == "" {
+		c.emailVerifiedKey = "email_verified"
+	}
```
sagikazarmark, line 212: Is it a good idea to set these in HandleCallback? I think these should be set when the connector is created.

connector/oauth/oauth.go
```diff
@@ -0,0 +1,283 @@
...
+	var userInfoResult map[string]interface{}
+	err = json.NewDecoder(userInfoResp.Body).Decode(&userInfoResult)
+
+	if err != nil {
+		return identity, fmt.Errorf("OAuth Connector: failed to parse userinfo: %v", err)
+	}
+
+	if c.userIDKey == "" {
+		c.userIDKey = "id"
+	}
```
sagikazarmark, line 186: Same as below

sagikazarmark (changes requested):

b-jsshapiro:
Mark:

No tone was intended, and I apologize.

Obviously, we’re frustrated. The delay on supporting LWA is forcing us to
implement our own solution, which seems really unfortunate.

Is there some way I can be concretely helpful moving this forward?

sagikazarmark:
No worries. The mean time between releases is definitely not ideal right now, we'll have to work on that.

I went through the PR again and there are no obvious red flags for me. There are a few nits to be fixed before being merged.

We'll try to get the next release out sooner.

connector/oauth/oauth.go
xtremerui: sure I have moved them into the construct fuction. thx!

README.md
xtremerui: The docs PR is submitted https://github.com/dexidp/website/pull/91. Please let me know. Thx.

connector/oauth/oauth.go
```diff
@@ -0,0 +1,283 @@
...
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+				DualStack: true,
+			}).DialContext,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+		},
```
JoelSpeed, line 147: I'm guessing this is mostly copied from the default transport? Perhaps instead of copying the values here we can use the `Clone` method on the default config and just override the fields we need? Otherwise perhaps we could add a comment to mention where it came from?

connector/oauth/oauth.go
```diff
@@ -0,0 +1,283 @@
...
+	userInfoResp, err := client.Get(c.userInfoURL)
+	if err != nil {
+		return identity, fmt.Errorf("OAuth Connector: failed to execute request to userinfo: %v", err)
+	}
+
+	if userInfoResp.StatusCode != http.StatusOK {
+		return identity, fmt.Errorf("OAuth Connector: failed to execute request to userinfo: status %d", userInfoResp.StatusCode)
+	}
+
+	defer userInfoResp.Body.Close()
```
JoelSpeed, line 199: This body close needs to come earlier else we might leak go routines if we exit earlier than line 199
```suggestion
	userInfoResp, err := client.Get(c.userInfoURL)
	if userInfoResp != nil {
		defer userInfoResp.Body.Close()
	}
	if err != nil {
		return identity, fmt.Errorf("OAuth Connector: failed to execute request to userinfo: %v", err)
	}

	if userInfoResp.StatusCode != http.StatusOK {
		return identity, fmt.Errorf("OAuth Connector: failed to execute request to userinfo: status %d", userInfoResp.StatusCode)
	}
```

We do this before the error check because there are occasions within the Go http client when it can initialise the response body and still return an error

connector/oauth/oauth.go
```diff
@@ -0,0 +1,283 @@
...
+
+	var userInfoResult map[string]interface{}
+	err = json.NewDecoder(userInfoResp.Body).Decode(&userInfoResult)
+
+	if err != nil {
```
JoelSpeed, line 204: Nit: please pair errors with their initialisation
```suggestion
	err = json.NewDecoder(userInfoResp.Body).Decode(&userInfoResult)
	if err != nil {
```

connector/oauth/oauth.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	identity.UserID = userID
+	identity.Username, _ = userInfoResult[c.userNameKey].(string)
+	identity.PreferredUsername, _ = userInfoResult[c.preferredUsernameKey].(string)
+	identity.Email, _ = userInfoResult[c.emailKey].(string)
+	identity.EmailVerified, _ = userInfoResult[c.emailVerifiedKey].(bool)
+
+	if s.Groups {
+		groups := map[string]bool{}
```
JoelSpeed, line 220: If this is being used as a set, it's technically more efficient to use `map[string]structu{}`, though this very much a nit

xtremerui: based on https://gist.github.com/davecheney/3be245c92b61e5045f75, it seems like struct{} is a bit memory efficient so i will update it. Thx!

connector/oauth/oauth.go
xtremerui: Thx for catching. Though while err is non-nil, based on the go lang documentation the userInfoResp.Body is already closed.

https://github.com/golang/go/blob/dc50683bf7ebdfde726d710131ba05fe97e10a07/src/net/http/client.go#L571-L573

so we could check if the err is nil first then do `defer userInfoResp.Body.Close()`

panxunying:
I have a similar scene,  i want to know  why this pr isn't merged now?

connector/oauth/oauth.go
sagikazarmark: I guess it's fine for now. We can decouple some code later.

connector/oauth/oauth.go
sagikazarmark: Looks good to me

connector/oauth/oauth.go
```diff
@@ -0,0 +1,286 @@
...
+		c.ClaimMapping.GroupsKey = "groups"
+	}
+
+	if c.ClaimMapping.EmailKey == "" {
+		c.ClaimMapping.EmailKey = "email"
+	}
+
+	if c.ClaimMapping.EmailVerifiedKey == "" {
+		c.ClaimMapping.EmailVerifiedKey = "email_verified"
+	}
```
sagikazarmark, line 90: Changing the struct itself is not very elegant. Using local variables would be nicer, but this is a nit.

README.md
```diff
@@ -80,6 +81,7 @@ Dex implements the following connectors:
 | [Atlassian Crowd](https://dexidp.io/docs/connectors/atlassiancrowd/) | yes | yes | yes * | beta | preferred_username claim must be configured through config |
 | [Gitea](https://dexidp.io/docs/connectors/gitea/) | yes | no | yes | alpha | |
 | [OpenStack Keystone](https://dexidp.io/docs/connectors/keystone/) | yes | yes | no | alpha | |
+| [Generic OAuth 2.0](https://dexidp.io/docs/connectors/oauth/) | no | yes | yes | alpha | |
```
sagikazarmark, line 84: Why is it added twice?

sagikazarmark (commented):
@xtremerui Thanks for the efforts you've put into this. I finally got the time around to give it a proper review. It looks good to me.

I had two small comments. Would you mind taking a look at them? We can finally get this merged after that.

xtremerui:
Thx for reviewing. I have updated it accordingly. @sagikazarmark

sagikazarmark (approved):

sagikazarmark:
Thanks @xtremerui

PR #1633, opened 2020-01-20
google: Option to filter groups by regexp

jfrabaute:
When fetching groups, this field allows to add a regexp to filter the groups
that will be returned and used.
This can be useful when the user is registered on a lot of groups because
he is part of multiple clusters. In this case, usually, the groups are named
with a pattern and a regexp can be used to filter the groups per cluster,
allowing to have a smaller group list retrieved for each cluster.
Default to empty string, and no filtering will be done in this case.


sagikazarmark:
We are planning to implement something like this for all providers for quite some time. I've just opened the issue to track its progress: #1635 

Personally, I'm against merging any more PRs adding per connector features. Would be nice to make this a top priority though.

tsuna:
Ping. Can we get this upstream pretty please?

alexisdondon:
Hi, is there some perspective on any kind of feature like that?

I also need similar feature to allow filter of groups or event better filter or regexp on any claim return by the upstream oidc.

nabokihms:
Hell @alexisdondon. We are a bit out of track, but we look forward to implementing a generic option to mutate groups soon this year.

PR #1634, opened 2020-01-20
Support username, email and groups claim in OIDC connector

xtremerui:
This PR leverage the `PreferredUsername` in `Identity` for OIDC connector
https://github.com/dexidp/dex/blob/aca67b083984982c11942ac505b883c1d75bf6f9/connector/connector.go#L28

User can config `PreferredUsernameKey` to lookup a custom claim that has user handle. Noted that it is not the same as `name` or `preferred_username` standard claims.

Updated: implementes #1777


sagikazarmark:
Thanks for the PR @xtremerui !

What's the reason behind letting the user change the claim for `PreferredUsername`?

connector/oidc/oidc.go
```diff
@@ -286,6 +291,11 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 	}
 	hostedDomain, _ := claims["hd"].(string)
 
+	if c.preferredUsernameKey == "" {
+		c.preferredUsernameKey = "username"
```
sagikazarmark: Shouldn't this be `preferred_username`?

xtremerui:
Thx for reviewing @sagikazarmark !

> What's the reason behind letting the user change the claim for `PreferredUsername`?

Since in OIDC spec `username` is not a standard claim (so the claim key is customizable i.e. not always  be `username`), making it configurable in OIDC config allows more flexibility.

connector/oidc/oidc.go
xtremerui: `preferred_username` is meant for `Shorthand name` according to OIDC spec. The PR here is aiming for `username` claim, which is usualy unique. Other connectors like github and gitlab has similar implementations

https://github.com/dexidp/dex/blob/aca67b083984982c11942ac505b883c1d75bf6f9/connector/gitlab/gitlab.go#L152

https://github.com/dexidp/dex/blob/aca67b083984982c11942ac505b883c1d75bf6f9/connector/github/github.go#L269

sagikazarmark: But `preferred_username` is actually a standard claim, so I would assume the OIDC coonnector would return that (given the underlying connector supports it).

Having an arbitrary, non-standard claim as the default value is confusing at best.

sagikazarmark:
Yeah, but `preferred_username` *is* a standard claim.

xtremerui:
Sure, I think that is reasonable. The PR is updated accordingly. 

From our end as long as there is a way for OIDC connector to return both `name` and `username`(or whatever it is) claim we are totally fine. Thx!

sagikazarmark:
I think there is a misunderstanding: `preferred_username` is a standard claim. Why do you want to change it? It's standard.

pivotal-jwinters:
@sagikazarmark the upstream provider might not support the "preferred_username" claim (dex itself only added support recently). So while I agree that the default should be "preferred_username", having it configureable is still valuable.

mvdkleijn:
@sagikazarmark @jwntrs @xtremerui - Can I try to help resolve this issue here? :smile: 

The problem is not that the name of the `preferred_username` field should be configurable, that's just a possible solution. The problem is that not all backends implement / provide the `preferred_username` field. (feel free to correct me at any time if needed!)

I can understand fully why the Dex maintainers would have an issue with changing the name of the `preferred_username` field as a potential solution to the above mentioned problem.

However, I propose that there is an alternative solution that keeps the name of the `preferred_username` field intact but still allows the end user to configure the connector so that the `preferred_username` field is set in Dex's response.

In fact, this is basically what I did in my (already merged) PR at https://github.com/dexidp/dex/pull/1684 for the Atlassian Crowd connector.

By default it would not supply a `preferred_username` response if none was set by the backend, however, the user could configure the connector to fill the `preferred_username` field with the value of another field that **is** supplied by the backend.

It'd be nice if we could drag this PR out of the current stalemate. :smile: :+1:

jwntrs:
@mvdkleijn yup I totally agree with your assessment. I had to re-check the code, because I thought this was how it was implemented, but it looks like the original behaviour changed in that second commit. I've stepped away from the project using dex, but maybe @xtremerui could update this PR.

xtremerui:
I have added tests for this @mvdkleijn . Thank you for the suggestion.

> By default it would not supply a preferred_username response if none was set by the backend, however, the user could configure the connector to fill the preferred_username field with the value of another field that is supplied by the backend.

sagikazarmark:
@mvdkleijn thanks for the suggestion. I think it makes sense.

So in short: if there is a `preferred_username` coming from the upstream use that value, otherwise let the user make it configurable and use the value of another claim. Is that right?

I think that sounds acceptable, so if @xtremerui updates the PR accordingly, I'd be happy to review it again.

mvdkleijn:
@sagikazarmark That is 100% correct.

I see that @xtremerui made most of the changes... the claim key's to use instead of `preferred_username` can now be configured.

However, the connector is not checking if `preferred_username` is set by upstream or not.

In pseudocode, I'd expect something like

```
if upstream['preferred_username'] is nil then
  preferred_username = upstream[config.preferred_username_key]
else
  preferred_username = upstream['preferred_username']
```

if @xtremerui wouldn't mind making that change, I think we can basically wrap this up.

(apologies for any typos, I'm writing from my phone)

mvdkleijn:
By the way @xtremerui ... all my comments here with regards to your use of preferred_username also apply to your oauth PR #1630 I believe

mvdkleijn:
Thanks @xtremerui !

LGTM

@sagikazarmark Your turn :smile:

sagikazarmark:
Please see #1777

xtremerui:
@sagikazarmark the PR is updated following the discussion in #1777 .

mvdkleijn:
@sagikazarmark Is anything still required for this apart from a review and merge?

sagikazarmark:
@mvdkleijn  See #1777

Documentation/connectors/oidc.md
```diff
@@ -74,23 +72,37 @@ connectors:
...
-
     # For offline_access, the prompt parameter is set by default to "prompt=consent". 
     # However this is not supported by all OIDC providers, some of them support different
     # value for prompt, like "prompt=login" or "prompt=none"
     # promptType: consent
+
+
+    # Some providers return no standard claim that is different to 
+    # claims list at https://openid.net/specs/openid-connect-core-1_0.html#Claims
+    # Use claimMapping to specify custom claim names
```
sagikazarmark, line 83:
```suggestion
    # Some providers return non-standard claims (eg. mail).
    # Use claimMapping to map those claims to standard claims:
    # https://openid.net/specs/openid-connect-core-1_0.html#Claims
    # claimMapping can only map a non-standard claim to a standard one if it's not returned in the id_token.
```

connector/oidc/oidc.go
```diff
@@ -281,9 +321,16 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+	emailKey := "email"
+	email, found = claims[emailKey].(string)
+	if !found && c.emailKey != "" {
+		emailKey = c.emailKey
+		email, found = claims[emailKey].(string)
+	}
+
 	if !found && hasEmailScope {
-		return identity, errors.New("missing \"email\" claim")
+		return identity, fmt.Errorf("missing \"%s\" claim", emailKey)
```
sagikazarmark, line 333: This error message won't tell the user that it's related to the email scope being set. Also, the original issue is still true: the email claim is missing as well.

I would probably leave email here and maybe mention the alias/mapping key as well.

Documentation/connectors/oidc.md
```diff
@@ -74,23 +72,37 @@ connectors:
...
+    claimMapping:
+      # The set claim is used as user id.
+      # Default: sub
+      # user_id: nickname
+
+      # The set claim is used as user name.
+      # Default: name
+      # user_name: nickname
```
sagikazarmark, line 91: UserID and Username are not actually claims, so I'm not sure we should include them here.

sagikazarmark (commented):
Couple nits, but I think it's great!

bonifaido (approved):

sagikazarmark:
@xtremerui can you take a look at the requested changes? I think it's time that we get this merged, it only needs a few final touches.

xtremerui:
@sagikazarmark done! Thx for waiting, it was long weekend last week and I took days off as well.

connector/oidc/oidc.go
```diff
@@ -49,14 +49,31 @@ type Config struct {
...
 
 	// PromptType will be used fot the prompt parameter (when offline_access, by default prompt=consent)
 	PromptType string `json:"promptType"`
+
+	ClaimMapping struct {
+		// Configurable key which contains the user id claim
+		UserIDKey string `json:"user_id"` // defaults to "sub"
+
+		// Configurable key which contains the username claim
+		UserNameKey string `json:"user_name"` // defaults to "name"
```
sagikazarmark, line 66: I guess these shouldn't be here either.

Documentation/connectors/oidc.md
```diff
@@ -74,23 +72,30 @@ connectors:
...
-    # The set claim is used as user id.
-    # Default: sub
-    # Claims list at https://openid.net/specs/openid-connect-core-1_0.html#Claims
-    #
-    # userIDKey: nickname
-    
-    # The set claim is used as user name.
-    # Default: name
-    # userNameKey: nickname
-
```
sagikazarmark, line 86: These should be added back as these are valid options, but not claims.

connector/oidc/oidc.go
```diff
@@ -49,14 +49,31 @@ type Config struct {
 	// id tokens
 	GetUserInfo bool `json:"getUserInfo"`
 
-	// Configurable key which contains the user id claim
+	// Deprecated: use UserIDKey in claimMapping instead
 	UserIDKey string `json:"userIDKey"`
 
-	// Configurable key which contains the user name claim
+	// Deprecated: use UserNameKey in claimMapping instead
```
sagikazarmark, line 55: These shouldn't be deprecated.

sagikazarmark (commented):
@xtremerui thanks a lot. I think you missed a few things though (or I wasn't entirely clear): I would keep the user id and username keys as they are, since they are not claims at all. We can always move them under claim mapping later if we think it's the right place, but for now it just seems logically wrong.

Thanks!

xtremerui:
That makes sense. It will make this PR more clean. I have reverted the changes for user id and user name.

sagikazarmark (approved):
Thanks a lot @xtremerui

mvdkleijn:
Yay! Done! :smile:

PR #1657, opened 2020-02-20
Setting email for OpenShift connector

sabre1041:
Automatically sets the email field to be based on the username for the OpenShift connector

### DIFF
--- a/connector/openshift/openshift.go
+++ b/connector/openshift/openshift.go
@@ -177,6 +177,7 @@ func (c *openshiftConnector) HandleCallback(s connector.Scopes, r *http.Request)
 		UserID:            user.UID,
 		Username:          user.Name,
 		PreferredUsername: user.Name,
+		Email:             user.Name,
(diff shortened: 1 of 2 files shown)


sabre1041:
Any chance the CI can be rerun? Appears to have failed due to a timeout

bonifaido:
Yes, just kicked it 👞

bonifaido:
Somehow the build can't pick up this commit from your remote, have you force pushed it (possibly)? Even that shouldn't cause an issue like this...

sabre1041:
Yes I did force push after squashing commits. but as you mentioned,  this should not cause an issue.

I can start over and re point this pr at a new branch or new pr

bonifaido:
Can you please rebase on master now? I think the issue was because GitHub actions recently updated the checkout action, and we used the previous version.

sabre1041:
@bonifaido I opened up #1661 which used latest master as I didnt get the your message notification in time :(

PR #1661, opened 2020-02-21
Setting email for OpenShift connector

sabre1041:
Automatically sets the email field to be based on the username for the OpenShift connector


sabre1041:
@bonifaido many thanks! cheers!

PR #1664, opened 2020-03-03
storage/static.go: expand environment variables in client ID and secret

lhotrifork:
Having ID and Secret in clear inside configuration files for static
clients is not ideal. This commit allows setting these from environment
variables.

Signed-off-by: Yann Soubeyrand <yann.soubeyrand@gmx.fr>

Rebased Yann's changes from: https://github.com/dexidp/dex/pull/1428

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -153,7 +153,28 @@ func serve(cmd *cobra.Command, args []string) error {
 	logger.Infof("config storage: %s", c.Storage.Type)
 
 	if len(c.StaticClients) > 0 {
-		for _, client := range c.StaticClients {
+		for i, client := range c.StaticClients {
+			if client.Name == "" {
+				return fmt.Errorf("invalid config: Name field is required for a client")
+			}
+			if client.ID == "" && client.IDEnv == "" {
+				return fmt.Errorf("invalid config: ID or IDEnv field is required for a client")
+			}
+			if client.IDEnv != "" {
+				if client.ID != "" {
+					return fmt.Errorf("invalid config: ID and IDEnv fields are exclusive for client %q", client.ID)
+				}
+				c.StaticClients[i].ID = os.Getenv(client.IDEnv)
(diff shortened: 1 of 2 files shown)


paul-theorem:
https://github.com/dexidp/dex/issues/1099

paul-theorem:
Apologies for after-the-fact commentary - but I have tested this and it works. For the benefit of others who might stumble across this (and for the authors and maintainers) - the use of the env var in static client config is _different_ that the use of env var in other areas of the same config file.  for example, in configuring the google connector - you'd do this:

      config:
        clientID: $GOOGLE_CLIENT_ID
        clientSecret: $GOOGLE_CLIENT_SECRET**

Later - in the same configmap, to ref "secretEnv" - you'd do this:
    staticClients:
    - id: kubelogin
      redirectURIs:
      - http://localhost:28000
      name: 'kubelogin'
      secretEnv: KUBELOGIN_SECRET

Note - no $ before the env var ref, when (by convention) the variable name ends in Env.  

I've built from source, and can confirm this does work. Ultimately i'd love to see more consistency in the use of env vars across the configmap yaml, but I do appreciate everyone's work.  This was a blocker for me - as of yesterday, and found your PR today. thanks !

yann-soubeyrand:
@paul-theorem you can see https://github.com/dexidp/dex/pull/1428#issuecomment-484634920 for an explanation of the inconsistency.

thesuperzapper:
@lhotrifork This has changed the behaviour of the config validation slightly, you could previously omit the `secret` field, but this will now throw an error.

For example, this is no longer a valid config: (from argo-cd)
```
staticClients:
- id: argo-cd-cli
  name: Argo CD CLI
  public: true
  redirectURIs:
  - http://localhost
```

srenatus:
@thesuperzapper an ugly but working workaround is to add `secretEnv: THISDOESNOTEXIST`, I think.

PR #1670, opened 2020-03-13
handlers: do not fail login if refresh token gone

### LINKED ISSUE #1669: Offline workflows can get into a bad state: failed to delete refresh token
We've been running into an issue occasionally with offline workflows. We aren't 100% sure what is causing it, but I can conceive of at least one case where it is possible.

The symptom is that when a user tries to log in, the attempt fails with the following error:

> {"level":"error","msg":"failed to get refresh token: not found","time":"2020-03-12T20:52:12Z"}

This persists until the user's `OfflineSession` is deleted from the backend storage. This is a fairly serious bug because it requires manual intervention the correct.

The issue seems to occur when the `OfflineSession` falls out of sync with the `RefreshToken` table. In particular, if the refresh token corresponding to the ID stored for the client in the offline storage is not present, then attempts to log in will fail in this block of code:

```
			if oldTokenRef, ok := session.Refresh[tokenRef.ClientID]; ok {
				// Delete old refresh token from storage.
				if err := s.storage.DeleteRefresh(oldTokenRef.ID); err != nil {
					s.logger.Errorf("failed to delete refresh token: %v", err)
					deleteToken = true
					return
				}
			}
```
One way this could happen is if dex crashes or is killed between when the above code executes and the code which updates the `OfflineSession` executes. In that case, the OfflineSession will still point to the refresh token we just deleted, so future calls to the code block will fail, leading to the auth request to fail.

klarose:
There is a chance that offline storage could fall out of sync with the
refresh token tables. One example is if dex crashes/is stopped in the
middle of handling a login request. If the old refresh token associated
with the offline session is deleted, and then the process stops, the
offline session will still refer to the old token.

Unfortunately, if this case occurs, there is no way to recover from it,
since further logins will be halted due to dex being unable to clean up
the old tokens till referenced in the offline session: the database is
essentially corrupted.

There doesn't seem to be a good reason to fail the auth request if the
old refresh token is gone. This changes the logic in `handleAuthCode` to
not fail the entire transaction if the old refresh token could not be
deleted because it was not present. This has the effect of installing
the new refresh token, and updating the offline storage, thereby fixing
the issue, however it occurred.

Fixes #1669

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -905,7 +905,7 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 		} else {
 			if oldTokenRef, ok := session.Refresh[tokenRef.ClientID]; ok {
 				// Delete old refresh token from storage.
-				if err := s.storage.DeleteRefresh(oldTokenRef.ID); err != nil {
+				if err := s.storage.DeleteRefresh(oldTokenRef.ID); err != nil && err != storage.ErrNotFound {
 					s.logger.Errorf("failed to delete refresh token: %v", err)
 					s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
 					deleteToken = true


klarose:
It looks like the CI may have failed for an unrelated reason. Is there an easy way to retry it?

bonifaido:
Please rebase on master, so the CI issue can get resolved.

swade1987:
It would be great to get this merged as its impacting us as well cc @willvrny

klarose:
@bonifaido Rebased, but it looks like the lint timed out this time. :(

bonifaido:
Build is fixed now! :)

swade1987:
@bonifaido any idea when a release with this will be cut?

bonifaido:
No planned released for now, but there are some fixes hanging around here, we will review a few and I will cut a patch release. Until that: https://quay.io/repository/dexidp/dex/manifest/sha256:2f5af70a79eae505d122ac40f684cff29fabda434f5fefeb49c0d888323a545d

PR #1673, opened 2020-03-17
storage/kubernetes: remove shadowed ResourceVersion from Connector

### LINKED ISSUE #1672: storage/kubernetes: Connectors are cached indefinitely
The type used to store connectors in kubernetes has a `ResourceVersion` field

https://github.com/dexidp/dex/blob/d820fd45d80cef74d4c65f5fcc5766ddb1fa514e/storage/kubernetes/types.go#L601

Which shadows the "actual" `ResourceVersion` specified by `ObjectMeta`. This causes the comparison here

https://github.com/dexidp/dex/blob/d820fd45d80cef74d4c65f5fcc5766ddb1fa514e/server/server.go#L550

To always report the same (empty) version, meaning cached connectors are never updated if they change in kubernetes.

ktravis:
Fix #1672

### DIFF
--- a/storage/conformance/conformance.go
+++ b/storage/conformance/conformance.go
@@ -591,11 +591,10 @@ func testConnectorCRUD(t *testing.T, s storage.Storage) {
 	id1 := storage.NewID()
 	config1 := []byte(`{"issuer": "https://accounts.google.com"}`)
 	c1 := storage.Connector{
-		ID:              id1,
-		Type:            "Default",
-		Name:            "Default",
-		ResourceVersion: "1",
-		Config:          config1,
+		ID:     id1,
+		Type:   "Default",
+		Name:   "Default",
+		Config: config1,
 	}
 
 	if err := s.CreateConnector(c1); err != nil {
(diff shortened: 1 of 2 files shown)


bonifaido:
Once https://github.com/dexidp/dex/pull/1674 gets merged, please rebase on master to have the CI fixed.

bonifaido:
I just rebased on master to overcome the MySQL CI issue which is now fixed on master. But the storage tests are failing now for k8s.

ktravis:
I'm not certain the best way to update these tests, it seems like they rely on a particular `ResourceVersion` being set, which will now conflict with the one set by the k8s API.

I can remove setting `ResourceVersion` at all in `fromStorageConnector` rather than setting it in `ObjectMeta` (which is probably the right thing to do, given that the k8s API will complain if it is set during a create) - but the tests are comparing the version to an expected value which will not be consistent after the API sets it. If I was writing the test from scratch I would probably set it up to skip the `ResourceVersion` comparison, if that is acceptable is there a standard way this is being done currently?

Any advice would be appreciated.

ktravis:
In order to make the connector storage tests pass I have ignored the `ResourceVersion` field from comparison. Testing that this individual field specifically is set does not add any particular extra value to the tests cases, from my perspective. If other behavior is to be tested, it seems reasonable that new/different tests should be written (i.e. `old.ResourceVersion != new.ResourceVersion` after `UpdateConnector` calls, etc). Please let me know what else should be done to ensure that this change is ready to merge.

PR #1678, opened 2020-03-25
Automatic consistency fixing in case of missing refresh token in db

Teeed:
In case of missing refresh token in db application will return 500 when user tries to issue new refresh token.

This PR adds automatic consistency fixing in case of missing refresh token in db by ignoring missing one. This could happen in case of interrupted write.


Teeed:
Could anybody have look at this? @bonifaido

PR #1681, opened 2020-03-31
Added Email of Keystone to Identity

kenperkins:
_Note: This is a replacement of #1442 as the original author hasn't been responsive, but is still a valuable contribution. I've rebased from master._

After the successful login to keystone, the Email of the logged in user
is fetch from keystone and provided to `identity.Email`.

This is useful for upstream software that uses the Email as the primary
identification.

_Note: Email is now required when using Dex with a Keystone Backend and pusher/oauth2_proxy@v5.0.0_

### DIFF
--- a/connector/keystone/keystone.go
+++ b/connector/keystone/keystone.go
@@ -95,6 +95,14 @@ type groupsResponse struct {
 	Groups []group `json:"groups"`
 }
 
+type userResponse struct {
+	User struct {
+		Name  string `json:"name"`
+		Email string `json:"email"`
+		ID    string `json:"id"`
+	} `json:"user"`
+}
+
 var (
 	_ connector.PasswordConnector = &conn{}
 	_ connector.RefreshConnector  = &conn{}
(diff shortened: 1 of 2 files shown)


kenperkins:
@srenatus this should be ready for review.

kenperkins:
@bonifaido also; as an aside what's the best channel for communication re: Dex development?

bonifaido:
Unfortunately, there is no centralized place currently to talk about Dex (at least no one that works, we have a Google group which is basically dead and of course it is not instant messaging). I think it would be nice to have one on the Kubernetes Slack, since most people are coming from that world.

kenperkins:
@bonifaido opened #1682

kenperkins:
@bonifaido this is ready for review after re-pushing based on changes to master. I don't know whom else might be available for a review.

bonifaido:
One last nit, can we squash this to one commit?

kenperkins:
I'm ok with squashing; on many of my projects I squash through GitHub. Would you prefer I squash directly through Git?

bonifaido:
Sorry, squashing wasn't enabled by this project, but now I enabled it, I think it is a good idea, since GitHub makes it easy to follow the commit back to the PR, and doesn't need to ping developers here.

kenperkins:
@bonifaido I agree; plus that keeps the original author's commits more visible in the PR, as I just re-pushed his changes. Thanks! What's the release schedule look like?

bonifaido:
We can cut a release in the coming days I guess, this week maybe?

PR #1683, opened 2020-04-03
Added X-Remote-Group header support in authproxy

ereOn:
Hi,

I might be missing something, but I felt like implementing this was pretty straightforward and I figured I would give it a shot.

Let me know if I'm completely mistaken ^^

P.S: Thanks for your amazing work on Dex! It's a wonderful thing we needed.


mvdkleijn:
Just passing by, but I noticed two things...

1) It should ben `r.Header["X-Remote-Group"]`

2) PR #1579 already exists for this and feels more complete than this to me. If you agree I'd suggest closing this PR in favor of #1579

ereOn:
@mvdkleijn You are absolutely right. I actually overestimated the time I had to work on this when I first submitted my vastly incomplete PR. Sorry for the bother, let's favor the other PR which seems much more advanced indeed.

Thank you!

PR #1684, opened 2020-04-04
Allow preferred_username claim to be set for Crowd connector

mvdkleijn:
**What this is**
In the current implementation, the `preferred_username` claim remains empty. Some clients might require that claim and Dex supports it, if set from the connector's side.

This PR allows the connector to **optionally** support the `preferred_username` claim. It achieves this by allowing a user to choose to fill the `preferred_username` claim by one of the available Crowd fields. Specifically: `key`, `name` or `email`. Invalid options cause the claim to remain empty.

**Breaking change:** No. Existing configuration should not notice this change.
**Optional:** Yes.

**Notes:**
- Added test for `identityFromCrowdUser` function which was not extensively tested before
- Added entry for the Crowd connector in the readme
- Updated documentation for the connector
- I'm fairly new to Go, so I welcome constructive feedback

### DIFF
--- a/connector/atlassiancrowd/atlassiancrowd.go
+++ b/connector/atlassiancrowd/atlassiancrowd.go
@@ -35,12 +35,18 @@ import (
 //       - admin
 //       # Prompt for username field
 //       usernamePrompt: Login
+//		 preferredUsernameField: name
 //
 type Config struct {
 	BaseURL      string   `json:"baseURL"`
 	ClientID     string   `json:"clientID"`
 	ClientSecret string   `json:"clientSecret"`
 	Groups       []string `json:"groups"`
+
+	// PreferredUsernameField allows users to set the field to any of the
+	// following values: "key", "name" or "email".
+	// If unset, the preferred_username field will remain empty.
+	PreferredUsernameField string `json:"preferredUsernameField"`
 
 	// UsernamePrompt allows users to override the username attribute (displayed
 	// in the username/password prompt). If unset, the handler will use.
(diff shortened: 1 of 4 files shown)


mvdkleijn:
Hi @srenatus @bonifaido @sagikazarmark - I was wondering if I could get some feedback on what this still needs or possibly just have this approved... It's for the recently added Crowd connector. :smiley:

Passes the checks and latest changes from master were merged in.

connector/atlassiancrowd/atlassiancrowd.go
```diff
@@ -368,6 +374,15 @@ func (c *crowdConnector) identityFromCrowdUser(user crowdUser) (connector.Identi
...
 	}
 
+	switch c.PreferredUsernameField {
+	case "key":
+		identity.PreferredUsername = user.Key
+	case "name":
+		identity.PreferredUsername = user.Name
+	case "email":
+		identity.PreferredUsername = user.Email
+	}
```
nabokihms, line 384: In my opinion, we need to add a warning (or return an error) if the preferredUserName field is set to the wrong value.

Users can start wondering why the claim is empty after setting the option value to 'email or another invalid value.

nabokihms:
@mvdkleijn thanks a lot for your work. This is wonderful! It would be great to get it merged.

mvdkleijn:
The build / checks failed due to stability issues with Github. Can they be re-triggered manually or do I need to push a dummy commit?

mvdkleijn:
Thanks for the retrigger @bonifaido Sadly Github is still having issues. I'll check back later. This PR should be good to merge anyway. I don't believe this needs anything else and the change is backwards compatible.

Issue #1635, opened 2020-01-20
Connector middleware


sagikazarmark:
Right now connectors implement certain types of behavior that could be reused. The most common example is filtering groups.

Consider creating a common middleware layer for connectors instead of implementing these features (and configuration for them) separately for each connector.

srenatus:
> The most common example is filtering groups.

💯 It would be great to refactor that at least. If it fits a more generic pattern (as you described) all the better for it. 👍

blaggacao:
related &rarr; #1790

merlindorin:
+1 for this kind of pattern.

Maybe there is something  to do/learn from ory oathkeeper: https://www.ory.sh/oathkeeper/docs/pipeline/mutator

(by the way... this kind of feature is awesome and we dream about it @renault :) : https://www.ory.sh/oathkeeper/docs/pipeline/mutator#hydrator)

al45tair:
We're going to want something like this too. If anyone has any suggestions about what designs they might prefer, I might have some time to work on this.

al45tair:
OK, so my assumption here is that the middleware layer will be passed the identity returned by a successful invocation of a Connector, and will return an (Identity, error) pair. Additionally, multiple middleware components can be stacked, in which case the identity output of one will be passed into the identity input of the next in the stack.

In addition, it looks like the Identity struct will need to be extended to enable custom claims, since middleware that adds custom claims is something people are likely to be interested in (in addition to being able to mutate group names and things). I know we are.

merlindorin:
yes, I'm fully agree with that.

As a use case : we operate an homemade extensions on Keycloak 😔 in order to retrieve custom claims in order to enrich the user identity. This extension is a simple http call. I easily imagine a Dex "connector" that can do the same.

This kind of feature is for us very precious.

janwillies:
It would be very interesting to issue an HTTP request from dex and let OPA do the authorization for which claims to include, maybe even handover additional infos like HTTP request headers

srenatus:
💭 Just sharing a thought here: It could end up nicely if Dex embedded OPA, exposing an interface similar to [what gatekeeper does](https://github.com/open-policy-agent/gatekeeper#how-to-use-gatekeeper). 

A sketchy adaptation of that model to Dex would look like this:

1. There's _contraints_, like

```yaml
constraints:
  - kind: allowedGroups
    match:
    - connector: ldap # some connector ID
    parameters:
      groups:
       - ops
       - admin
```

[...]

PR #1688, opened 2020-04-16
feat: Add team groups support to bitbucket connector

nabokihms:
## Description:
This PR provides support for team groups.  The only option added is `includeTeamGroups`. When the option value is `true`,  we make an additional request to Bitbucket API to get all team group names and combine them with the team name using `/`.

The groups claim in dex id_token looks like:
* Without option: `groups=["my_team"]`
* With option: `groups=["my_team", "my_team/administrators", "my_team/members"]`

## Motivation:
I started working on it because there was no chance to do authn/authz management in the frames of one team. This behavior is something that other connectors already has (github, gitlab).

## Concerns: 
My only concern is sending a request to deprecated API to get team groups. However, Atlassian said, that `/groups` endpoint would be supported until the equal method would appear in 2.0 API.

### DIFF
--- a/Documentation/connectors/bitbucketcloud.md
+++ b/Documentation/connectors/bitbucketcloud.md
@@ -31,4 +31,8 @@ connectors:
     # If `teams` is provided, this acts as a whitelist - only the user's Bitbucket teams that are in the configured `teams` below will go into the groups claim.  Conversely, if the user is not in any of the configured `teams`, the user will not be authenticated.
     teams:
     - my-team
+    # Optional parameter to include team groups.
+    # If enabled, the groups claim of dex id_token will looks like this:
+    # ["my_team", "my_team/administrators", "my_team/members"]
+    includeTeamGroups: true
 ```
--- a/connector/bitbucketcloud/bitbucketcloud.go
+++ b/connector/bitbucketcloud/bitbucketcloud.go
@@ -21,7 +21,8 @@ import (
 
 const (
 	apiURL = "https://api.bitbucket.org/2.0"
-
+	// Switch to API v2.0 when the Atlassian platform services are fully available in Bitbucket
+	legacyAPIURL = "https://api.bitbucket.org/1.0"
 	// Bitbucket requires this scope to access '/user' API endpoints.
 	scopeAccount = "account"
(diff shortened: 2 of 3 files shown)


crytome1995:
I really like this, and need this, but I can't seem to get the groups to be added to the claim. I believe this is due to the comparison of the original whitelisted teams in config compared to what is provided after group concatenation?

nabokihms:
@crytome1995 to make `includeTeamGroups: true` works with teams filter correctly, you should add additional entries to your filter like:
```
includeTeamGroups: true
teams: ["my_team", "my_team/everyone", "my_team/administrators"]
```
This behavior helps to filter unnecessary groups from the result ID token.

nabokihms:
I just rebased the branch to activate new CI jobs.

sagikazarmark:
@nabokihms Thanks! The PR looks OK.

Regarding the filter: I believe other connectors (GitHub, GitLab) behave the same way, but what pains me is that config looks differently (again). 😢

nabokihms:
In this particular case, we can avoid adding an option. We just need to inform users that default behavior will change (I'm not a big fan of breaking changes, but it has its pros and cons). By the way, I am looking forward to your decision about this PR :)

P.S.:
About difference in configs... It's hard to keep connectors flexible enough without kind specific parameters.

In my opinion, it will be good to move some options from providers config to a higher level, like baseURL, clientID, clientSecret, etc, and use the config section only for unique connector options.

sagikazarmark:
I'm actually a huge fan of how it is done in GitHub: https://github.com/banzaicloud/pipeline/blob/cb95d36/etc/config/dex.yml.dist#L104-L113

It's well structured, easy to understand.

In most of the cases just a concept or some rules to follow would be enough. But since there is no such thing, everyone just invents their own solutions.

sagikazarmark:
Also, there is this: #1635

sagikazarmark:
@nabokihms if you rebase this PR I'll merge it as well. Thanks!

nabokihms:
I rebased the branch and resolved conflicts.

sagikazarmark (approved):
Thanks @nabokihms

PR #1690, opened 2020-04-16
Fix templates which asset path points to external URL

nabokihms:
## Description:
#1554 breaks the ability to specify an icon for header using external URL like [this one](https://kubernetes.io/images/favicon.png).

I added check, that if assetURL points to external URL, no transformations required.

### DIFF
--- a/server/templates.go
+++ b/server/templates.go
@@ -176,6 +176,11 @@ func loadTemplates(c webConfig, templatesDir string) (*templates, error) {
 //assetPath is static/main.css
 //relativeURL("/dex", "/dex/auth", "static/main.css") = "../static/main.css"
 func relativeURL(serverPath, reqPath, assetPath string) string {
+	if u, err := url.ParseRequestURI(assetPath); err == nil && u.Scheme != "" {
+		// assetPath points to the external URL, no changes needed
+		return assetPath
+	}
+
 	splitPath := func(p string) []string {
 		res := []string{}
 		parts := strings.Split(path.Clean(p), "/")
--- a/server/templates_test.go
+++ b/server/templates_test.go
@@ -31,6 +31,13 @@ func TestRelativeURL(t *testing.T) {
 			assetPath:  "assets/css/main.css",
 			expected:   "../assets/css/main.css",
 		},
+		{
+			name:       "external-url",
+			serverPath: "/dex",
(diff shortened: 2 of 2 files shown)


sagikazarmark:
@nabokihms can you please rebase your branch?

server/templates.go
```diff
@@ -176,6 +176,11 @@ func loadTemplates(c webConfig, templatesDir string) (*templates, error) {
 //assetPath is static/main.css
 //relativeURL("/dex", "/dex/auth", "static/main.css") = "../static/main.css"
 func relativeURL(serverPath, reqPath, assetPath string) string {
+	if u, err := url.ParseRequestURI(assetPath); err == nil && u.Scheme != "" {
```
sagikazarmark, line 179: Not sure if this is actually correct. `//kubernetes.io/images/favicon.png` would be perfectly valid HTML as well.

nabokihms: `//kubernetes.io/images/favicon.png` will be treated as a relative path because of the URL scheme is empty. Do you think we should limit the scheme to http/https instead?

sagikazarmark: `//kubernetes.io/images/favicon.png` **is** a valid URI and and I'm not sure it should be rewritten as a relative path.

Furthermore, Dex should usually be served over HTTPS, which means having an HTTP URL on the site will probably trigger browser security behavior.

TBH, that makes me think if supporting external URLs is a good idea in the first place.

nabokihms: I made this PR because:
* It can be handy to use with prebuilt docker images from the official repo. You only need to specify one URL in the config to add an icon instead of adding it to the docker image. 
* It fixes the broken behavior. Someone might suffer because of such breaking changes.

But I agree with the concerns that you raised. Frankly, we need to make a decision here.

sagikazarmark: So after thinking about this for a looooooong time, I think we can assume that people should always use HTTPS URLs these days, so not supporting the schemaless URL should be fine. I'm also not concerned about using http urls on an https site: if someone does that, it's their problem.

Limiting the scheme, however, is an interesting question. Does that pose a security risk if we just allow using any schemes there? not sure.

sagikazarmark (approved):
Eh, let's just merge it. Thanks @nabokihms !

PR #1691, opened 2020-04-17
OIDC: Add parameter configuration to override email claim key

cyrilix:
Some OIDC providers don't return standard email claim key (ex: 'mail'), add configuration to override email claim name.


sagikazarmark:
Please see #1777

sagikazarmark:
Fixed in #1634

Issue #1695, opened 2020-04-21
Secret field is unnecessary required for a static client config. 


alexmt:
The commit https://github.com/dexidp/dex/commit/99c3ec68203c0b2764a49724ecb9ecc26b520482 has introduced the ability to use secret references in static client configuration and at the same time made this field required. 

Some authentication types require only client id and don't require secret. For example following configuration used to work :

```
- id: argo-cd-cli
  name: Argo CD CLI
  public: true
  redirectURIs:
  - http://localhost
```

Now it failed with the following error:

```
 invalid config: Secret or SecretEnv field is required for client "argo-cd-cli"
```

PR #1701, opened 2020-05-05
allow no secret for static public clients

tkleczek:
For statically-configured public clients it should be allowed for both
Secret and SecretEnv fields to be empty.


tkleczek:
This is fix for dexidp/dex#1695

PR #1704, opened 2020-05-11
connector/saml: add 'FilterGroups' setting

### LINKED ISSUE #1476: SAML: whitelist groups?
So, for LDAP through the group filter, and for some other connectors (like [github](https://github.com/dexidp/dex/blob/master/connector/github/github.go#L60-L64), [gitlab](https://github.com/dexidp/dex/blob/master/connector/gitlab/gitlab.go#L33), [bitbucket](https://github.com/dexidp/dex/blob/master/connector/bitbucketcloud/bitbucketcloud.go#L38)), we support whitelisting a bunch of groups -- so beyond authentication, only the groups listed will be considered...

❓ What do you think, should the SAML connector have a similar feature?

srenatus:
This should make AllowedGroups equivalent to an LDAP group filter:

When set to true, only the groups from AllowedGroups will be included in the
user's identity.

This is not the great refactoring where all the connectors use some common logic, but just another band-aid to provide some useful functionality to SAML users. 

------
Fixes #1476


srenatus:
ℹ️ rebased

PR #1711, opened 2020-05-14
Fix the go modules v2 path

### LINKED ISSUE #1710: Issue with using the client package v2 version in go modules project
I am trying to use client package latest v2 version in a go modules project but there is an error:

```
go get -u github.com/dexidp/dex@v2.24.0
go get github.com/dexidp/dex@v2.24.0: github.com/dexidp/dex@v2.24.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

I think the v2 version is not properly handled following these recommendations:
https://blog.golang.org/v2-go-modules

ccojocar:
The latest version of the api client cannot be used in a go modules project as a dependency. When running:

```
go get -u github.com/dexidp/dex@v2.24.0
```
returns:
```
go get github.com/dexidp/dex@v2.24.0: github.com/dexidp/dex@v2.24.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

The go modules v2 should follow the recommendations from https://blog.golang.org/v2-go-modules.

fixes #1710


sagikazarmark:
I don't think this is something we want to do right now. Personally I'm not a huge fan of importing this project anywhere. A better solution would be making the api package a module IMHO.

If you absolutely need to import Dex, you can do so by importing it with a specific commit hash.

PR #1715, opened 2020-05-19
Add Gitea connector

techknowlogick:
This PR adds Gitea as a connector for Dex.

Note: As documented, this new connector doesn't yet support Groups claim, however as Gitea offers a similar API to Github, it could be added in as a later PR.

### DIFF
--- a/Documentation/connectors/gitea.md
+++ b/Documentation/connectors/gitea.md
@@ -1,0 +1,29 @@
+# Authentication through Gitea
+
+## Overview
+
+One of the login options for dex uses the Gitea OAuth2 flow to identify the end user through their Gitea account.
+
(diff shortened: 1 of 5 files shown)


nabokihms:
Many thanks for this PR! Nowadays Gitea is really popular.

I just wonder if Gitea provides the same API as GitHub maybe we can just reuse the GitHub connector for Gitea.

HostName setting already presents for GitHub connector, the only thing we need to do is to change the API version.

Please feel free to correct me or disagree with me.

techknowlogick:
Thanks for kind words :)

I was unclear, in terms of API being similar I should've specified specified that I was talking about only the teams endpoint, so that if/when a future PR comes along to implement it, then the code would be similar. However even with teams response being similar, it does pagination differently than Github, as well as has a different URL (/api/v1/teams for Gitea vs /api/v3/teams for github). The oauth2 flow is different than Github, in one big area is that Gitea offers refresh tokens.

Happy to answer any other questions.

PR #1734, opened 2020-06-16
fix documentation ldap

bloudman:
Refer to this issue https://github.com/argoproj/argo-cd/issues/3112
I fix the documentation

### DIFF
--- a/Documentation/connectors/ldap.md
+++ b/Documentation/connectors/ldap.md
@@ -126,9 +126,8 @@ connectors:
       # Following list contains field pairs that are used to match a user to a group. It adds an additional
       # requirement to the filter that an attribute in the group must match the user's
       # attribute value.
-      userMatchers:
-      - userAttr: uid
-        groupAttr: member
+      userAttr: uid
+      groupAttr: member
 
       # Represents group name.
       nameAttr: name


Documentation/connectors/ldap.md
```diff
@@ -126,9 +126,8 @@ connectors:
       # Following list contains field pairs that are used to match a user to a group. It adds an additional
       # requirement to the filter that an attribute in the group must match the user's
       # attribute value.
-      userMatchers:
-      - userAttr: uid
-        groupAttr: member
+      userAttr: uid
```
bonifaido, line 129: I haven't checked the upstream issue here, but `userMatchers` are the way to go, please check: https://github.com/dexidp/dex/blob/30cd592801d493b9f253d7bf1c12b072b917d6a6/connector/ldap/ldap.go#L137

papanito:
My ldap config was not working until somebody pointed me to this. Please merge asap ;-)

bloudman:
@papanito Yes and it was the same for me ... But someone commented out that it should not be the case. it is a little bit weird I thought the documentation was not good but it seems that it is the helm chart that have a problem.

Documentation/connectors/ldap.md
bloudman: I checked and it is not working with userMatchers ... So maybe in fact the issue is in the chart

aliusmiles: well, in dex's master those are deprecated but look at dex v2.22.0, the one argocd uses: https://github.com/dexidp/dex/blob/26061f9558f3011bea53ef5a5bb02ac669d612e0/connector/ldap/ldap.go#L135

so it's either 'downgrade' docs to match dex config or upgrade dex i guess :)

sagikazarmark:
@bloudman can you please consider resubmitting this PR to our website repo if it is still valid: https://github.com/dexidp/website

PR #1743, opened 2020-06-30
add docker-compose.yaml for local testing

bonifaido:
This PR tries to resolve the pain of testing different storage options quickly. I have added different almost all storage options/versions (except Kubernetes CRDs) and included the corresponding Dex config counterparts as well.


sagikazarmark:
Maybe sync versions used in GitHub Actions and the compose file?

bonifaido:
Good idea, synced!

PR #1745, opened 2020-06-30
Remove vendor

### LINKED ISSUE #1597: Revisit vendoring

sagikazarmark:
Fixes #1597


sagikazarmark:
> Just manually verified that everything still works (as far as I can tell), noticed while I was doing that that our make target for the docker image requires sudo, I think we should probably drop that as we shouldn't strictly need sudo for it

I'm gonna address that in a different PR (as well as add a Docker job to actions pushing to dockerhub)

PR #1746, opened 2020-06-30
Update some dependencies

sagikazarmark:


sagikazarmark:
It looks like upgrading protobuf will require a bunch of other changes. I will do those first and then continue upgrading the rest of the dependencies.

PR #1762, opened 2020-07-13
[doc/example] Make LDAP example functional again by running OpenLDAP with docker-compose

### LINKED ISSUE #1654: missing ./scripts/slapd.sh
I would follow the [LDAP tutorial](https://github.com/dexidp/dex/blob/master/Documentation/connectors/ldap.md) but the script `./scripts/slapd.sh` is missing.

heidemn-faro:
Fixes #1654


examples/config-ldap.yaml
```diff
@@ -11,7 +11,7 @@ connectors:
   name: OpenLDAP
   id: ldap
   config:
-    host: localhost:10389
+    host: localhost:389
```
heidemn-faro, line 14: If there's a strong opinion to use the nonstandard port, I can revert it.
But then the port would be different here (outside the container) and when using "docker exec" (inside the container).

sagikazarmark:
Thanks @heidemn-faro . Personally, I'd much rather see a docker-compose based solution, without adding back that script. I believe the OpenLDAP image is able to add definitions automatically on startup.

heidemn-faro:
@sagikazarmark okay understand.
But since the issue is open since February, I would suggest that you accept this as a temporary fix until you have time to adapt it according to your preferences.

At least it works now if you follow the docs :-)

heidemn-faro:
@sagikazarmark as requested, I've reworked this to use docker-compose.
Any further changes required?

examples/config-ldap.yaml
sagikazarmark: I think the rationale behind using a non-standard port is avoiding potential collision with an existing service. But all the other examples (mysql, etc) use standard ports, so it should be okay.

heidemn-faro:
@sagikazarmark  @bonifaido thanks for reviewing.
The PR is ready from my side. I've verified that the example works.

I don't have permissions to merge, so you would have to do that.

PR #1772, opened 2020-07-31
Add c_hash to id_token

HEllRZA:
This PR adds support for c_hash on the id_token.
The c_hash is required in the hybrid flow, and optional otherwise.

According to https://openid.net/specs/openid-connect-core-1_0.html#CodeValidation
> If the ID Token is issued from the Authorization Endpoint with a code, which is the case for the response_type values code id_token and code id_token token, this is REQUIRED; otherwise, its inclusion is OPTIONAL.

**OpenID Connect Certification**
The test (OP-IDToken-c_hash) now succeeds

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -661,7 +661,7 @@ func (s *Server) sendCodeResponse(w http.ResponseWriter, r *http.Request, authRe
 				return
 			}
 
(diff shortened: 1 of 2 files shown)


HEllRZA:
see #1773

PR #1774, opened 2020-07-31
Add c_hash to id_token, issued on /auth endpoint, when in hybrid flow

HEllRZA:
* fixed name collision (renamed hash->hashFunc)


HEllRZA:
sorry ... this is not what i supposed to do. too many tabs open ...

PR #1631, opened 2020-01-20
Support customize oidc groups key

xtremerui:
Since groups claim is not a [standard claim in OIDC](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims), the groups claim name could be something other than "groups". This PR allows dex OIDC connector to take a custom key to look up groups claims.

cc @pivotal-jwinters

PR #1776, opened 2020-08-07
Add groupsClaimMapping to the OIDC connector

Lemmons:
The `groupsClaimMapping` setting allows one to specify which claim to pull group information from the OIDC provider.  Previously it assumed group information was always in the "groups" claim, but that isn't the case for many OIDC providers (such as AWS Cognito using the "cognito:groups" claim instead)

### DIFF
--- a/Documentation/connectors/oidc.md
+++ b/Documentation/connectors/oidc.md
@@ -68,6 +68,10 @@ connectors:
     # This can be overridden with the below option
     # insecureEnableGroups: true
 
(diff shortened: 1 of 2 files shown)


Lemmons:
This is related to #1065

Lemmons:
Looks like #1631 implemented this same logic back in Jan, but is still waiting for a review.

sagikazarmark:
I'm fine with resubmitting it.

Please see #1777 before reviewing and merging this.

sagikazarmark:
Fixed in #1634

PR #1782, opened 2020-08-18
Enable local build of example app

heidemn-faro:
Without these changes, it doesn't seem possible to build the example app from the **local** checkout.
(I'm not a Go expert, so there might be other ways to achieve this.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,10 +20,13 @@ LD_FLAGS="-w -X $(REPO_PATH)/version.Version=$(VERSION)"
 # Dependency versions
 GOLANGCI_VERSION = 1.21.0
 
-build: bin/dex bin/grpc-client
+build: bin/dex bin/example-app bin/grpc-client
 
 bin/dex:
 	@go install -v -ldflags $(LD_FLAGS) $(REPO_PATH)/cmd/dex
+
+bin/example-app:
+	@go install -v -ldflags $(LD_FLAGS) $(REPO_PATH)/examples/example-app
 
 bin/grpc-client:
 	@go install -v -ldflags $(LD_FLAGS) $(REPO_PATH)/examples/grpc-client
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -20,11 +20,14 @@ LD_FLAGS="-w -X $(REPO_PATH)/version.Version=$(VERSION)"
 # Dependency versions
 GOLANGCI_VERSION = 1.21.0
 
-build: bin/dex bin/grpc-client
+build: bin/dex bin/example-app bin/grpc-client
```
xUnholy, line 23: I'd recommend leaving it as its own target and not making it required whenever a regular `build` target is being called.

heidemn-faro: As you wish.
But I will then instead add a step `make bin/example-app` to `ci.yml`, to ensure that the example app builds correctly.

sagikazarmark:
Firstly, I agree with @xUnholy.

Secondly, this is a documentation issue. Adding a replace will basically make the example a dependency if the main repo which is a big no.

Instead, you need to change dir into the example directory and run go build there.

Makefile
sagikazarmark: That's actually a good idea. You could create an 'examples' target that builds all example binaries (including grpc-client)

heidemn-faro:
> Firstly, I agree with @xUnholy.
> 
> Secondly, this is a documentation issue. Adding a replace will basically make the example a dependency of the main repo which is a big no.
> 
> Instead, you need to change dir into the example directory and run go build there.

Thanks for your comments.
I think the main issue why I couldn't build the stuff was that I had set `GO111MODULE=on`. 
This seems to result in `go get` downloading the repo to a different location, and complaining "...no non-test Go files..."
I'll review my Go settings, and give it another try then.

sagikazarmark:
No, this is unrelated to your go mod settings. Since the example package is a submodule (in the sense of go modules), you simply cannot build it without changing directories, otherwise go thinks that you want to download it as a dependency.

PR #1220, opened 2018-04-16
Make static clients' RedirectURIs to be regular expressions

redbaron:

Issue #448, opened 2016-05-24
Allow wildcards in for valid RedirectURIs


hanikesn:
We want to use dex to together with a [Nginx](https://github.com/pingidentity/lua-resty-openidc) reverse proxy to secure a large number of internal services. That would mean creating a large number of dex clients or doing a bit of session trickery to share a single authentication via dex between mutiple subdomains.

Allowing wildcards for valid RedirectURIs would make this a lot easier to manage.

bobbyrullo:
I understand why that would be helpful @hanikesn, but I think it goes against the OIDC/OAuth2 spec.

You could exploit the `state` parameter for this purpose though...

hanikesn:
Sorry, to reopen this. I don't understand how this would violate the OIDC spec, e.g. Google allows multiple domains to be used for a single client:
![oidc](https://cloud.githubusercontent.com/assets/630711/15570343/de51d988-2335-11e6-9373-13f54ba43504.png)

What I basically want to be able to do is:
```
eval "$(./bin/dexctl --db-url=$DEX_DB_URL new-client http://*.example.com/callback)"
```

ericchiang:
> REQUIRED. Redirection URI to which the response will be sent. This __URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in Section 6.2.1 of [RFC3986] (Simple String Comparison)__. When using this flow, the Redirection URI SHOULD use the https scheme; however, it MAY use the http scheme, provided that the Client Type is confidential, as defined in Section 2.1 of OAuth 2.0, and provided the OP allows the use of http Redirection URIs in this case. The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application.

https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest (Emphasis my own)

Seems pretty clearly against the spec.

redbaron:
There is a spec and there are practical questions. Google took pragmatic approach and allowed wildcards, Dex took pragmatic approach and didn't ban HTTP scheme from redirects as specs says.

For dynamic development environments, such as created with help of Gitlab, where apps are deployed to unique URLs per branch, wildcard handling is essential. 

For those who wants it, here is PR where RedirectURIs can be a regex: https://github.com/coreos/dex/pull/1220

hanikesn:
I can confirm that Azure Active Directory also allows wildcards for redirect URIs now.

nabadger:
[...]

Issue #1300, opened 2018-09-19
RedirectURL's for public clients


RichardLindhout:
I'm currently adding a feature to our webapp where we want to login with the microsoft connector.
```
    - id: webapp
      secret: ********************************
      public: true
      name: 'Webapp'
      redirectURIs:
      - 'https://dexserver.com/accounts/callback' 
      - 'http://localhost:3000/connected-login'
      - 'https://webapp.com/connected-login'
```

```
    connectors:
    - type: microsoft
      id: microsoft
      name: Microsoft
      config:
        clientID: ***********************
        clientSecret: ***********************
        redirectURI: https://dexserver.com/accounts/callback
        tenant: ********.onmicrosoft.com

```

The windows connect redirects from the microsoft login to our dex server and the dex server redirects it to the public client with a code as parameter.

In our webapp we get our access_token with the code which is send to us via dex when authenticated with microsoft.

It works perfect when the redirectURI=http://localhost:3000/connected-login
But not when the redirectURI=https://webapp.com/connected-login

According to this code redirects are only allowed for  non-public
```
	if !client.Public {
		for _, uri := range client.RedirectURIs {
			if redirectURI == uri {
				return true
			}
		}
		return false
	}

```

I don't get that exactly since the redirectURI is checked when obtaining the access_tokens:

handleAuthCode (server/handler.go:643)

`authCode.ClientID != client.ID` > error
`authCode.RedirectURI != redirectURI` > error

Since the redirect and client id are checked why not allow redirectURI for public-clients.

![flow_dex_swimminglane](https://user-images.githubusercontent.com/6492229/45746983-ddb02600-bc04-11e8-81b1-0a47167f493b.png)
![flow-dex](https://user-images.githubusercontent.com/6492229/45746984-ddb02600-bc04-11e8-9b13-bd57aca84be4.png)

RichardLindhout:
I also can't obtain access_tokens with the microsoft connector login with a non-public client since it would expose the secret when obtaining the access_token via the authorization_code grant.

srenatus:
Have you seen the [docs on "public clients"](https://github.com/dexidp/dex/blob/316acbee03e1ae904170adfcb8452452ca6cea1b/Documentation/custom-scopes-claims-clients.md#public-clients)? They're trying to provide some colour re: how it's intended to be used.

> I also can't obtain access_tokens with the microsoft connector login with a non-public client since it would expose the secret when obtaining the access_token via the authorization_code grant.

[...]

PR #1783, opened 2020-08-19
Rule-restricted wildcard match of single subdomain #488

tom-haines:
Add support for rule-restricted wildcard match of lowest-level subdomain via LRU cache.

This PR:
- builds on [PR 1220](https://github.com/dexidp/dex/pull/1220), applying to any client created via API for example - rather than static clients only, see discussion at [issue 448](https://github.com/dexidp/dex/issues/448)
- far tigher security as per comments by fbsb on PR #1220
- in-memory LRU cache of max 500 redirect URI values -> empty string (no valid wildcard) or to regex string to match if all conditions met

### DIFF
--- a/server/oauth2.go
+++ b/server/oauth2.go
@@ -601,7 +601,8 @@ func validateRedirectURI(client storage.Client, wildcardCache *lru.ARCCache, red
 			if urlRequested, err := url.Parse(redirectURI); err == nil && urlRequested != nil {
 				// validly-formed url
 				clobberMatcher := getClobberHostMatcher(uri, wildcardCache)
-				if clobberMatcher.HostMatcher != nil && matchesClobber(clobberMatcher, urlRequested) {
+				validClobber := clobberMatcher.HostMatcher != nil && clobberMatcher.ClientRedirectURL != nil
+				if validClobber && matchesClobber(clobberMatcher, urlRequested) {
 					// clobber match success
 					return true
 				}
@@ -631,16 +632,16 @@ func validateRedirectURI(client storage.Client, wildcardCache *lru.ARCCache, red
 }
 
 func matchesClobber(clobberMatcher ClientRedirectClobberMatcher, urlRequested *url.URL) bool {
-	if !clobberMatcher.HostMatcher.MatchString(urlRequested.Host) {
-		// failed host match
+	if clobberMatcher.ClientRedirectURL.Path != urlRequested.Path {
+		// failed path match
 		return false
 	}
-	if clobberMatcher.ClientRedirectURLSections.Scheme != urlRequested.Scheme {
+	if clobberMatcher.ClientRedirectURL.Scheme != urlRequested.Scheme {
 		// failed scheme match
 		return false
 	}
-	if clobberMatcher.ClientRedirectURLSections.RawPath != urlRequested.Path {
-		// failed path match
+	if !clobberMatcher.HostMatcher.MatchString(urlRequested.Host) {
+		// failed host match
 		return false
 	}
 	return true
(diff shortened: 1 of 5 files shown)


server/oauth2_test.go
```diff
@@ -263,11 +264,70 @@ func TestValidRedirectURI(t *testing.T) {
...
+				RedirectURIs: []string{"http://pre-*-post.foo.com/bar"},
+			},
+			redirectURI: "http://pre-dinner-post.foo.com/bar",
+			wantValid:   true,
+		},
+		{
+			// valid wildcard will not match a URL more than one subdomain level in place of the wildcard
+			//  https://*.foo.com will not work with https://123.abc.foo.com.
+			client: storage.Client{
+				RedirectURIs: []string{"https://*.foo.com/"},
```
postmaxin, line 319: In our environment, we are currently authenticating second level hosts via OIDC using wildcards, eg, "https://hostname.namespace.k8s-cluster-name/".

Could we have an option to allow this by making it explicit that we want to match dots as well? maybe something like "https://**.foo.com/"?

tom-haines: Yes, this seems reasonable to me.  I'll update test cases & docs and last modify core flow.

tom-haines: I think explicit wildcards would be better, i.e.
``
https://*.*.foo.com/
```

Then the prefix and suffix function will still work / be available.

Comment (tom-haines) [reply server/oauth2_test.go:333]: To limit PR scope, I think it might be better to file new issue to consider multiple subdomains and modify as separate ticket as being an additional feature?

The change will be an enhancement and backwards comparible, so two steps seems logical?

Comment (postmaxin) [reply server/oauth2_test.go:333]: I support the idea of `*.*.foo.com`, however, I can see that causing a set of rules like *.foo.com, *.*.foo.com, *.*.*.foo.com, *.*.*.*.foo.com, depending on how people are using their domain names.

The **.foo.com syntax was loosely based off of shell globbing syntax (**/*.py).

I'm on vacation right now but when I get home I'll write more (maybe even some code! lol).

Currently, we've forked Dex so that we can support allowing it to authenticate entire subdomain trees.

Comment (tom-haines) [reply server/oauth2_test.go:333]: I've had a look further into this, and I have some concerns about the performance of the shell globbing syntax using the current approach, because of the runtime overhead associated with it.  Effectively a fork on the standard Go URL.parse function needs to be written in terms of carving out the scheme (as `*` is not permitted in hostname and returns nil).

https://tools.ietf.org/html/rfc3986#section-3.3

I'll have a quick look at whether it can be done efficiently today.

**Comment by tom-haines** [comment]
@postmaxin 
I've added globber support and wildcard (within single subdomain portion only).
I also took opportunity to clean up performance and testing aspects, and to use the matcher only on the hostname component to avoid injection type issues as we can enforce end of host stream matches e.g. "example.com" which could otherwise be added to the path component of the URL.
I'll update docs next.

**Comment by tom-haines** [comment]
Docs updated.  Ready for next review.

File: server/oauth2_test.go (hunk 1 above)
Comment (tom-haines) [reply server/oauth2_test.go:418]: Both are implemented; better caching; better comparison (use host portion only of URI) - conservatively implemented.
It would be good to have access to the logger somehow for debug statements, but could not see how this should be done cleanly.  Any input welcome!

File: server/server.go [hunk 2]
```diff
@@ -524,15 +536,15 @@ func (s *Server) OpenConnector(conn storage.Connector) (Connector, error) {
 		}
 	}
 
-	connector := Connector{
+	connectorInstance := Connector{
```
tom-haines, line 539: GoLand was complaining variable name was bad practice as overlaps with namespace - so I fixed while groking / unrelated to issue.

server/oauth2.go
```diff
@@ -575,7 +588,8 @@ func validateRedirectURI(client storage.Client, redirectURI string) bool {
 	if err != nil {
 		return false
 	}
-	if u.Scheme != "http" {
+	// public clients should use http or https (#1300)
+	if u.Scheme != "http" && u.Scheme != "https" {
```
tom-haines, line 592: This was discussed in open issue #1300 – it was a quick/logical quick so included here, but not relevant to main issue of wildcard support.

carpenike:
Would also like to see this as well.

wildman12 (approved):

jvanzyl:
This would be awesome to get into the next release! I know everyone is busy but it would be great if this can be reviewed!

kstreee-furiosa:
Any progress here? I'm looking for being merged this pr.

Smana:
Looking forward testing this, could we help making this feature its way to the next release.

PR #1652, opened 2020-02-14
Basic implementation of PKCE

Teeed:
This is my own implementation of PKCE, it is similar to #1407 but this version could be directly merged.

Issue #1114

Issue #1532, opened 2019-08-24
OPTIONS /userinfo denied on Access-Control-Request-Headers: Authorization


rikardev:
Hi,

I have a client that sends an OPTIONS request to the /userinfo endpoint with an `Access-Control-Request-Header: authorization`. This seems correct to me but dex replies with 403 since no Allowed CORS Headers are configured.

Implementing the handleWithCors func in server.go like this would be an easy fix:
```golang
handleWithCORS := func(p string, h http.HandlerFunc) {
	var handler http.Handler = h
	if len(c.AllowedOrigins) > 0 {
		allowedOrigins := handlers.AllowedOrigins(c.AllowedOrigins)
		allowedHeaders := handlers.AllowedHeaders([]string{"authorization"})
		handler = handlers.CORS(allowedOrigins, allowedHeaders)(handler)
	}
	r.Handle(path.Join(issuerURL.Path, p), instrumentHandlerCounter(p, handler))
}
```

Is that a viable solution? Am I missing something?

rikardev:
Clients running in Firefox seems also to complain about missing Access-Control-Allow-Methods header. That seems dirtier to solve however since gorilla isn't setting any unless the Access-Control-Request-Methods header contains a value separate from the Gorilla defaults (GET, HEAD, POST).

So just using `handlers.AllowedMethods([]string{"POST", "GET", "HEAD", "OPTIONS"})` for example with a request containing the header `Access-Control-Request-Method: GET` will not yield a 'Access-Control-Allow-Methods: GET' header in the response.

HEllRZA:
I had the same issue with CORS on the /userinfo endpoint. As i understand CORS, the preflight on OPTIONS should not return 4xx at all, but a 200 with the correct CORS headers. Then the browser should check the CORS Headers.

Is this behavior considered more secure?

al45tair:
This still needs fixing, and it's now October 2020. @HEllRZA has a fix in his PKCE implementation, but has said he'll remove it as it isn't related to PKCE. So, @HEllRZA, are you going to submit it as a separate PR? If not, please let me know and I'll raise one myself. It needs fixing for what I'm working on, and I don't want to have to maintain a Dex branch forever just for this.

sagikazarmark:
I'm wondering if it would make sense to add this header to the userinfo request only. But I guess it's fine to add to the "global" middleware and iterate if necessary.

[...]

PR #1784, opened 2020-08-26
PKCE implementation

### LINKED ISSUE #1114: PKCE support for public clients

HEllRZA:
This pull request adds the PKCE (Proof Key for Code Exchange) implementation. ([RFC 7636](https://tools.ietf.org/html/rfc7636)), Closes #1114.

Based on the implementation of @Teeed (PR #1652)
- add the remarks of @deric (return invalid_grand when wrong code_verifier)
- fix the issue, @mfmarche found.
- Enforce PKCE flow on /token when PKCE flow was started on /auth
- made sure, the correct errors are returned
- add Tests for the error cases
- make sure, the client_secret is only omitted, when code_verifier is received on /token, and grant_type is correct.
- add "Authorization" to allowed CORS headers

I am lookin forward to your review.

### Test PKCE
You should use a public client (without secret), so PKCE is not stopped by a missing client_secret
```yaml
staticClients:
- id: example-app
  name: 'Example App'
  public: true
```
### Note
* Configuring redirectURIs does currently not work with public clients (see #1822) but public clients allow all localhost callback URIs (e.g. http://localhost:5556/dex/callback)
* When testing with oidc-client.js if you use `loadUserInfo: true`, it will not work until #1819 is merged.

Thanks to @Teeed and i hope this PR is of interest to @mfmarche, @deric, @justin-slowik


Teeed:
Hi,

good job! :+1:

HEllRZA:
@sagikazarmark could you take a look at it? That would be great!

asoorm:
Is it necessary to support the plain code challenge method?

https://tools.ietf.org/html/rfc7636#section-7.2

>    Because of this, "plain" SHOULD NOT be used and exists only for
>    compatibility with deployed implementations where the request path is
>    already protected.  The "plain" method SHOULD NOT be used in new
>    implementations, unless they cannot support "S256" for some technical
>    reason.
> 
>    The "S256" code challenge method or other cryptographically secure
>    code challenge method extension SHOULD be used.  The "plain" code
>    challenge method relies on the operating system and transport
>    security not to disclose the request to an attacker.

https://tools.ietf.org/html/draft-ietf-oauth-v2-1-00#section-4.1.1.2

> The plain transformation is for compatibility with existing
   deployments and for constrained environments that can't use the
   "S256" transformation.

HEllRZA:
@asoorm 
Thanks for your input! I also dislike `"plain"`.

Reading the RFC, it sounds to me that the server has to support the `code_challenge_method "plain"`. It is even the default, when no `code_challenge_method` is specified. The passages you quote discourage the use of `"plain"` in the client that starts the PKCE request.

> Client Sends the Code Challenge with the Authorization Request
>
>   The client sends the code challenge as part of the OAuth 2.0
>   Authorization Request (Section 4.1.1 of [RFC6749]) using the
>   following additional parameters:
>
>   code_challenge
>      REQUIRED.  Code challenge.
>
>   code_challenge_method
>      OPTIONAL, defaults to "plain" if not present in the request.  Code
>      verifier transformation method is "S256" or "plain".

However the client should not use it. It should use `"S256"` and never downgrade to `"plain"`. `"S256"` is required on the server.
>Clients MUST NOT downgrade to "plain" after trying the "S256" method.
>   Servers that support PKCE are required to support "S256", and servers
>   that do not support PKCE will simply ignore the unknown
>   "code_verifier".

Other Auth Providers also support both (e.g. auth0).
Attacks should be prevented by disallowing to use `"plain"` on the /token endpoint, when `"S256"` was used as `code_challenge_method` on the /auth endpoint to start the flow. **I will make sure that i have a test for that!**

server/handlers.go
```diff
@@ -783,6 +811,32 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 		return
 	}
 
+	// RFC 7636 (PKCE)
+	codeChallengeFromStorage := authCode.CodeChallenge.CodeChallenge
+	providedCodeVerifier := r.PostFormValue("code_verifier")
+
+	if providedCodeVerifier != "" && codeChallengeFromStorage != "" {
+		providedCodeVerifier := r.PostFormValue("code_verifier")
```
tkleczek, line 819: already declared before 'if'
```suggestion
```

server/handlers.go
```diff
@@ -783,6 +811,32 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
+		calculatedCodeChallenge, err := s.calculateCodeChallenge(providedCodeVerifier, authCode.CodeChallenge.CodeChallengeMethod)
+		if err != nil {
+			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
```
tkleczek, line 822: I believe according to https://tools.ietf.org/html/rfc7636#section-4.4.1, errInvalidRequest and http.StatusBadRequest should be returned here (in case of unknown code challenge method)

```
If the server supporting PKCE does not support the requested
   transformation, the authorization endpoint MUST return the
   authorization error response with "error" value set to
   "invalid_request".  The "error_description" or the response of
   "error_uri" SHOULD explain the nature of error, e.g., transform
   algorithm not supported
```

server/oauth2.go
```diff
@@ -413,6 +413,18 @@ func (s *Server) parseAuthorizationRequest(r *http.Request) (*storage.AuthReques
...
+	codeChallenge := q.Get("code_challenge")
+	codeChallengeMethod := q.Get("code_challenge_method")
+
+	if codeChallengeMethod == "" {
+		codeChallengeMethod = CodeChallengeMethodPlain
+	}
+
+	if codeChallengeMethod != CodeChallengeMethodS256 && codeChallengeMethod != CodeChallengeMethodPlain {
+		description := fmt.Sprintf("Unsupported PKCE challenge method (%q).", codeChallengeMethod)
+		return nil, &authErr{"", "", errInvalidRequest, description}
```
tkleczek, line 425: I believe the error of unsupported PKCE challenge method should be returned as a query param in the redirect to a client callback URI.
So this check should be performed after redirectURI gets validated below and the newErr func defined there should be used.

server/handlers.go
```diff
@@ -749,12 +759,17 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
 		}
 		return
 	}
-	if client.Secret != clientSecret {
+
+	grantType := r.PostFormValue("grant_type")
+	codeVerifier := r.PostFormValue("code_verifier")
+
+	if grantType == grantTypeAuthorizationCode && codeVerifier != "" {
```
tkleczek, line 766: Is this really necessary? What's wrong with checking clientSecret even for PKCE?

server/oauth2.go
HEllRZA: This is the passage from the RFC. I think that it is implemented according to it. What do you think?
```
RFC 7636                       OAUTH PKCE                 September 2015
4.4.1.  Error Response
   [...]
   If the server supporting PKCE does not support the requested
   transformation, the authorization endpoint MUST return the
   authorization error response with "error" value set to
   "invalid_request".  The "error_description" or the response of
   "error_uri" SHOULD explain the nature of error, e.g., transform
   algorithm not supported.
```

server/server_test.go
```diff
@@ -1170,6 +1312,30 @@ func TestKeyCacher(t *testing.T) {
...
+		if err := json.Unmarshal(rErr.Body, details); err != nil {
+			t.Errorf("%s: could not parse return json: %s", tc.name, err)
+			return
+		}
+		if tc.tokenError.Error != "" && details.Error != tc.tokenError.Error {
+			t.Errorf("%s: got wrong Error in response: %s (%s). expected %s",
+				tc.name, details.Error, details.ErrorDescription, tc.tokenError.Error)
+		}
+	} else {
+		t.Errorf("%s: unexpedted error type: %s. expected *oauth2.RetrieveError", tc.name, reflect.TypeOf(err))
```
tkleczek, line 1335:
```suggestion
		t.Errorf("%s: unexpected error type: %s. expected *oauth2.RetrieveError", tc.name, reflect.TypeOf(err))
```

server/handlers.go
HEllRZA: If we would check the clientSecret for PKCE, the client would have to possess the client secret. PKCE is created for public clients that **cannot safely** store a clientSecret; e.g. single-page applications.

[Very good explanation of Auth Code flow with PKCE](https://auth0.com/docs/flows/authorization-code-flow-with-proof-key-for-code-exchange-pkce)

However; if someone does not want to have PKCE enabled, because he **always** wants to use the client secret, we should think about a configuration to enable/disable PKCE.

tkleczek (commented):
Would be nice to also add some tests to oauth2_test.go

server/oauth2.go
tkleczek: The thing is, oauth2 rfc https://tools.ietf.org/html/rfc6749#section-4.2.2.1 says
```
If the request fails due to a missing, invalid, or mismatching
   redirection URI, or if the client identifier is missing or invalid,
   the authorization server SHOULD inform the resource owner of the
   error and MUST NOT automatically redirect the user-agent to the
   invalid redirection URI.

   **If the resource owner denies the access request or if the request
   fails for reasons other than a missing or invalid redirection URI,
   the authorization server informs the client by adding the following
   parameters to the fragment component of the redirection URI** using the
   "application/x-www-form-urlencoded" format, per Appendix B:
```

Using authErr{"", "", errInvalidRequest, description}  (with first two params empty) would result in dex just displaying page with error instead of sending HTTP redirect back to client

server/handlers.go
tkleczek: I agree that PKCE was introduced to mitigate security concerns for public clients, but my reasoning is: 
If client is configured as public, then the check should pass as both clientSecret and client.Secret should be empty.
But if for whatever reason sb configures confidential client with PKCE, why not validate client secret as well.

storage/storage.go
```diff
@@ -169,6 +169,12 @@ type Claims struct {
 	Groups []string
 }
 
+// Data needed for PKCE (RFC 7636)
+type CodeChallenge struct {
```
tkleczek, line 173: I am wondering if the better names would be 
CodeChallenge -> Challenge
CodeChallengeMethod -> Method
so that we don't have excessive verbosity when used?: 
CodeChallenge.CodeChallengeMethod

alternatively the type could be renamed to PKCE

server/oauth2.go
HEllRZA: Ok, now I understand. I will take a look at it.

storage/storage.go
HEllRZA: `CodeChallenge.Challenge` also sounds verbose, so
I prefer `PKCE.CodeChallenge` and `PKCE.CodeChallengeMethod`
It makes it clear, for what it is used.

server/handlers.go
HEllRZA: Dex currently disallows an empty client secret with the error:
`invalid config: Secret or SecretEnv field is required for client "example-app"`
and I don't want to mess with that decision. The secret protects e.g. the refresh_token flow and the password_grant flow.

I guess it is best to have a config parameter for the client like
`allowPKCE: true` which is false by default. This way, old configurations still use client_secret only and are not affected by the PKCE implementation.

Sounds good?

HEllRZA: With "public" you mean setting the client.public parameter and thus enabling https://github.com/dexidp/dex/pull/471?

### Suggestion one
```suggestion
        if client.Public && grantType == grantTypeAuthorizationCode && codeVerifier != "" {
               // let it pass
        } else // check secret
```
This will accomplish the following:

if client is confidential, the clientSecret will always be checked (
if client is public, but the client has a client secret, PKCE will still work *without* a client secret.

### Suggestion two
introduce a new boolean to client configuration, which allows PKCE without checking the client secret. This way the client does not have to be set to public.

HEllRZA:
@tkleczek 
Thanks for the code review!
from the [RFC 7636](https://tools.ietf.org/html/rfc7636#section-4.4.1)
```
4.4.1.  Error Response
   If the server requires Proof Key for Code Exchange (PKCE) by OAuth
   public clients and the client does not send the "code_challenge" in
   the request, the authorization endpoint MUST return the authorization
   error response with the "error" value set to "invalid_request".  The
   "error_description" or the response of "error_uri" SHOULD explain the
   nature of error, e.g., code challenge required.
```

Sounds to me as if it would be a good idea to have a switch in the dex client configuration,
that sets PKCE as "required" for the code auth flow of public clients, what do you think?

server/handlers.go
tkleczek: `With "public" you mean setting the client.public parameter and thus enabling #471?`
Yes.

I am trying to understand the scenario you have in mind behind the this if-check. Are you concerned about allowing access in scenario of having a single client configuration dex-side (as confidential client) and then have 2 clients that use it - one is confidential (knows the secret and can protect it) and other is public (does not know the secret and uses PKCE)?

tkleczek:
@HEllRZA 
 
> Sounds to me as if it would be a good idea to have a switch in the dex client configuration,
> that sets PKCE as "required" for the code auth flow of public clients, what do you think?

Yeah, for me definitely nice-to-have - but not required. I was trying to think whether to have a server-wide or per-client option. I guess server-wide default could save your time if you would like every client to be forced to use PKCE. On the other hand, some clients could not be ready for it - so having a per-client config also seems to have benefits.

server/handlers.go
HEllRZA: > Are you concerned about allowing access in scenario of having a single client configuration dex-side (as confidential client) and then have 2 clients that use it - one is confidential (knows the secret and can protect it) and other is public (does not know the secret and uses PKCE)?

Yes.

tkleczek: I think this use-case might violate the OIDC spec:
https://openid.net/specs/openid-connect-core-1_0.html
```
9.  Client Authentication
This section defines a set of Client Authentication methods that are used by Clients to authenticate to the Authorization Server when using the Token Endpoint. During Client Registration, the RP (Client) MAY register a Client Authentication method. If no method is registered, the default method is client_secret_basic.

These Client Authentication methods are:

client_secret_basic
Clients that have received a client_secret value from the Authorization Server authenticate with the Authorization Server in accordance with Section 2.3.1 of OAuth 2.0 [RFC6749] using the HTTP Basic authentication scheme.
[...]
none
The Client does not authenticate itself at the Token Endpoint, either because it uses only the Implicit Flow (and so does not use the Token Endpoint) or because it is a Public Client with no Client Secret or other authentication mechanism.
```

This seems to imply that there should be a single authentication method used by the client and your scenario assumes two (some client instances would use `client_secret_basic`, others `none`)

HEllRZA: Also reading through [this](https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/issues/1456) and [this](https://stackoverflow.com/questions/56793968/should-authorization-flow-use-both-secret-and-pkce-for-a-web-app), I agree upon that. Will change it.

HEllRZA: Four pur application, the restriction of the callback uris implied by the setting "public" does not work out.

HEllRZA: Actually, i don't quite understand, why "public" clients don't respect the configured redirectURIs.

HEllRZA: The document even states that wildcard domains should be prevented. The current "private" implementation violates this by basically implementing redirect URI of type `http[s]://localhost/*`.
But implementing it according to that spec, would mean to possibly break existing dex deployments :/

HEllRZA: I think i misread it, so i mimiized above comment. Should I implement the usage of RedirectURI in "private" in this pull request, because PKCE does not work correctly for SPAs without it, or should that be discussed in #1300?

tkleczek: when you say "private" you mean "public"? I think this should be discussed in #1300. I wonder if there are any downsides to just using confidential client with dummy password as a workaround for your case, don't see any at first glance.
Also, I guess this would be a good time to get some actual maintainer to interested in this PR

HEllRZA: > when you say "private" you mean "public"?

yes. (D'oh!)
> I think this should be discussed in #1300. I wonder if there are any downsides to just using confidential client with dummy password as a workaround for your case, don't see any at first glance.

The SPA draft states `NOT RECOMMENDED` to use a client_secret in that case. I am worried that we would carry that useless client_secret around "forever".
> Also, I guess this would be a good time to get some actual maintainer to interested in this PR

definitively!

tkleczek (approved):
LGTM

tkleczek:
It would be great if this could be pulled-in, especially that it's straightforward RFC implementation of the feature that is considered as recommended for public clients protection (by OAuth 2.0 Security Best Current Practice document (https://tools.ietf.org/html/draft-ietf-oauth-security-topics-15)  Could anybody from maintainers team review? @sagikazarmark maybe?

sagikazarmark:
The first four open tabs in my browser are Dex PRs to review. I'll get there eventually.

server/handlers.go
```diff
@@ -767,6 +777,19 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
...
 
+func (s *Server) calculateCodeChallenge(codeVerifier, codeChallengeMethod string) (string, error) {
+	switch codeChallengeMethod {
+	case CodeChallengeMethodPlain:
+		return codeVerifier, nil
+	case CodeChallengeMethodS256:
+		shaSum := sha256.Sum256([]byte(codeVerifier))
+		return base64.RawURLEncoding.EncodeToString(shaSum[:]), nil
+	default:
+		s.logger.Errorf("unknown challenge method (%v)", codeChallengeMethod)
```
sagikazarmark, line 788:
```suggestion
```

server/handlers.go
```diff
@@ -783,6 +806,30 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
...
 	}
 
+	// RFC 7636 (PKCE)
+	codeChallengeFromStorage := authCode.PKCE.CodeChallenge
+	providedCodeVerifier := r.PostFormValue("code_verifier")
+
+	if providedCodeVerifier != "" && codeChallengeFromStorage != "" {
+		calculatedCodeChallenge, err := s.calculateCodeChallenge(providedCodeVerifier, authCode.PKCE.CodeChallengeMethod)
+		if err != nil {
+			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
```
sagikazarmark, line 816:
```suggestion
			s.logger.Error(err)
			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
```

server/server.go
```diff
@@ -287,7 +287,8 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
 		var handler http.Handler = h
 		if len(c.AllowedOrigins) > 0 {
 			corsOption := handlers.AllowedOrigins(c.AllowedOrigins)
-			handler = handlers.CORS(corsOption)(handler)
+			corsHeaders := handlers.AllowedHeaders([]string{"Authorization"})
```
sagikazarmark, line 290: Is this change related to PKCE? If not, what's the reason for adding it?

Also, I don't think it's enough if you need to send credentials in a cors request. I think you also need `AllowCredentials` as well.

But if it's not related to the PR and it's there for testing reasons, it should probably be removed.

sagikazarmark (changes requested):
LGTM, thanks a lot for working on it.

Found a few nits, nothing special.

server/server.go
HEllRZA: I added `"Authorization"` to the allowed headers that, from a browser, after a successful login with PKCE, the `/userinfo` endpoint can be called with the bearer token from the browser.
I tested it with oidc-client, which has a parameter `loadUserInfo`, which does just that successfully.

According to [this](https://www.sjoerdlangkemper.nl/2018/09/12/authorization-header-and-cors/), whitelisting `"Authorization"` in the `Access-Control-Allow-Headers` is sufficient, if we just want to send a bearer token (and do not want the browser to ask the user for a password).

But it is not essential for PKCE, i would consider it a bug. There is already an issue [1532](https://github.com/dexidp/dex/issues/1532) for it.

I will remove it the.

server/oauth2.go
```diff
@@ -572,12 +588,15 @@ func (s *Server) validateCrossClientTrust(clientID, peerID string) (trusted bool
 }
 
 func validateRedirectURI(client storage.Client, redirectURI string) bool {
-	if !client.Public {
-		for _, uri := range client.RedirectURIs {
-			if redirectURI == uri {
-				return true
-			}
+	// Allow named RedirectURIs for both public and non-public clients.
```
heidemn-faro, line 591: @sagikazarmark our use case is to use PKCE for a web app that doesn't run on "localhost".
Therefore, we would like that `RedirectURIs` are also considered for public clients.

The previous code to allow any localhost URI, `redirectURIOOB` or `deviceCallbackURI` for public clients is still active.
An alternative could be to no longer allow these special URLs when `RedirectURIs` is configured.
I don't have a strong opinion which option is better.

server/server.go
HEllRZA: A similar patch is handled in PR #1819

server/oauth2.go
tkleczek: I think it might be better to add this change in a separate PR (as it is not PKCE-specific) just to get the PKCE feature merged faster

heidemn-faro: @tkleczek OK, I just created a separate PR here: https://github.com/dexidp/dex/pull/1822

server/handlers.go
heidemn-faro: Done.

server/handlers.go
heidemn-faro: Done.

heidemn-faro:
@bonifaido  @sagikazarmark  it seems that all requested changes have been implemented.
Anything else that needs to be addressed before this can be merged?

server/server.go
heidemn-faro: Done (reverted).

trentis:
I could be doing something wrong but I just tried using this PR with a oidc-client-js implementation. The handleToken method would still always respond with unauthorized at the checking of client secret step. Any ideas why this would be happening?

HEllRZA:
@trentis It was decided to still use the client_secret with PKCE. 

So there are two ways to solve this issue for you:
1. Use a client secret with oidc-client-js (that's probably not what you want to do)
2. Configure your dex-client as **public client** with **no client secret**.
```yaml
staticClients:
- id: example-app
  name: 'Example App'
  public: true
```
Note that configuring `redirectURIs` does currently not work with public clients (see  #1822) but public clients allow all localhost callback URIs (e.g. http://localhost:5556/dex/callback) (IIRC 127.0.0.1 does not work!)

HEllRZA:
@trentis Another pitfall:
If you use ` loadUserInfo: true` in oidc-client-js, it will not work until #1819 is merged.

trentis:
Cheers for the assistance @HEllRZA . Yeah an error might be good took me a while to realize it wasn't oidc-client-js  as they do the code challenge automatically. So it particularly could help others. I'll probably end up using a public client and merge in the other PRs to get what I need. Hopefully it gets merged soon

server/handlers.go
```diff
@@ -749,6 +759,10 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
 		}
 		return
 	}
+	if clientSecret == "" && client.Secret != "" && r.PostFormValue("code_verifier") != "" {
```
tkleczek, line 762: Seems inconsistent that we want to specifically call out "missing credentials" in PKCE case and not in general case. 
Also I'm wondering whether not specifying client authentication error details but just general "Invalid client credentials." is for security reasons. E.g. the error you introduced would allow to determine whether the client having a specific client_id is registered on the server.

HEllRZA: I understand. Should we instead just log it?

tkleczek: I would say it's a  good idea, many such entries could indicate that some attack is in place. We could also log when client_id is invalid. Not sure if we would like to specifically call-out "missing credentials" case, but I'm not against it. And let's do it regardless of PKCE.

HEllRZA: Ok, Now, there is a log output instead.

tkleczek: I was thinking more in lines of:

```
           if client.Secret != clientSecret {
	        if clientSecret != "" {
	            s.logger.Info("missing client secret")
	        } else {
	            s.logger.Info("invalid client secret")
	        }
	        
		s.tokenErrHelper(w, errInvalidClient, "Invalid client credentials.", http.StatusUnauthorized)					 
                return
```

HEllRZA: Written text has so much potential for misunderstanding ;o)

sagikazarmark:
@tkleczek @HEllRZA is this PR ready for another look?

tkleczek:
@sagikazarmark  Yes, I believe so.

HEllRZA:
@tkleczek @sagikazarmark 
Yes, it is ripe.

server/handlers.go
```diff
@@ -750,6 +760,11 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if client.Secret != clientSecret {
+		if clientSecret == "" {
+			s.logger.Infof("missing client_secret on token request for client: %s", client.ID)
```
johanbrandhorst, line 764: Should this case be OK for PKCE? The whole idea is that the client doesn't have to provide the client secret. The role of the IdP in this case is just to verify Code verifier and Code challenge, which happens in `handleAuthCode`.

Ref: https://auth0.com/docs/flows/authorization-code-flow-with-proof-key-for-code-exchange-pkce

johanbrandhorst: I tried rewriting this locally to move the clientSecret validation into `handleAuthCode`, `handleRefreshToken` and `handlePasswordGrant` respectively. I think we have to remove it from this function for PKCE to work.

johanbrandhorst: Here's a rought sketch of the changes I made, though I'd suggest extracting the client secret validation into its own function instead:
```diff
diff --git a/server/handlers.go b/server/handlers.go
index 342849ee..1141dffc 100644
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -765,24 +765,15 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
 		}
 		return
 	}
-	if client.Secret != clientSecret {
-		if clientSecret == "" {
-			s.logger.Infof("missing client_secret on token request for client: %s", client.ID)
-		} else {
-			s.logger.Infof("invalid client_secret on token request for client: %s", client.ID)
-		}
-		s.tokenErrHelper(w, errInvalidClient, "Invalid client credentials.", http.StatusUnauthorized)
-		return
-	}
 
 	grantType := r.PostFormValue("grant_type")
 	switch grantType {
 	case grantTypeAuthorizationCode:
-		s.handleAuthCode(w, r, client)
+		s.handleAuthCode(w, r, client, clientSecret)
 	case grantTypeRefreshToken:
-		s.handleRefreshToken(w, r, client)
+		s.handleRefreshToken(w, r, client, clientSecret)
 	case grantTypePassword:
-		s.handlePasswordGrant(w, r, client)
+		s.handlePasswordGrant(w, r, client, clientSecret)
 	default:
 		s.tokenErrHelper(w, errInvalidGrant, "", http.StatusBadRequest)
 	}
@@ -801,7 +792,7 @@ func (s *Server) calculateCodeChallenge(codeVerifier, codeChallengeMethod string
 }
 
 // handle an access token request https://tools.ietf.org/html/rfc6749#section-4.1.3
-func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client storage.Client) {
+func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client storage.Client, clientSecret string) {
 	code := r.PostFormValue("code")
 	redirectURI := r.PostFormValue("redirect_uri")
 
@@ -839,6 +830,16 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 		// Received PKCE request on /auth, but no code_verifier on /token
 		s.tokenErrHelper(w, errInvalidGrant, "Expecting parameter code_verifier in PKCE flow.", http.StatusBadRequest)
 		return
+	} else {
+		if client.Secret != clientSecret {
+			if clientSecret == "" {
+				s.logger.Infof("missing client_secret on token request for client: %s", client.ID)
+			} else {
+				s.logger.Infof("invalid client_secret on token request for client: %s", client.ID)
+			}
+			s.tokenErrHelper(w, errInvalidClient, "Invalid client credentials.", http.StatusUnauthorized)
+			return
+		}
 	}
 
 	if authCode.RedirectURI != redirectURI {
@@ -1000,7 +1001,17 @@ func (s *Server) exchangeAuthCode(w http.ResponseWriter, authCode storage.AuthCo
 }
 
 // handle a refresh token request https://tools.ietf.org/html/rfc6749#section-6
-func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, client storage.Client) {
+func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, client storage.Client, clientSecret string) {
+	if client.Secret != clientSecret {
+		if clientSecret == "" {
+			s.logger.Infof("missing client_secret on token request for client: %s", client.ID)
+		} else {
+			s.logger.Infof("invalid client_secret on token request for client: %s", client.ID)
+		}
+		s.tokenErrHelper(w, errInvalidClient, "Invalid client credentials.", http.StatusUnauthorized)
+		return
+	}
+
 	code := r.PostFormValue("refresh_token")
 	scope := r.PostFormValue("scope")
 	if code == "" {
@@ -1227,7 +1238,17 @@ func (s *Server) handleUserInfo(w http.ResponseWriter, r *http.Request) {
 	w.Write(claims)
 }
 
-func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, client storage.Client) {
+func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, client storage.Client, clientSecret string) {
+	if client.Secret != clientSecret {
+		if clientSecret == "" {
+			s.logger.Infof("missing client_secret on token request for client: %s", client.ID)
+		} else {
+			s.logger.Infof("invalid client_secret on token request for client: %s", client.ID)
+		}
+		s.tokenErrHelper(w, errInvalidClient, "Invalid client credentials.", http.StatusUnauthorized)
+		return
+	}
+
 	// Parse the fields
 	if err := r.ParseForm(); err != nil {
 		s.tokenErrHelper(w, errInvalidRequest, "Couldn't parse data", http.StatusBadRequest)
```

HEllRZA: @johanbrandhorst

I initially implemented it in a similar way by skipping the client_secret validation, if the request was PKCE.

We decided that, for that case of using PKCE, the client should be configured as **public** and have no client_secret. And we should not omit the client_secret if the client is **confidential**.

Quote from @tkleczek:
> I agree that PKCE was introduced to mitigate security concerns for public clients, but my reasoning is:
> If client is configured as public, then the check should pass as both clientSecret and client.Secret should be empty.
> But if for whatever reason sb configures confidential client with PKCE, why not validate client secret as well.

johanbrandhorst: That makes sense, thanks!

sagikazarmark (approved):
LGTM, thanks @HEllRZA and @tkleczek for your work! ❤️

bonifaido (approved):
LGTM, thank you very much!

PR #1793, opened 2020-09-09
Fix example app build

lzeng27:
- Since the example app was moved the examples folder, the make command doesn't generate the example-app bin. Added a new make command to fix this issue.
- Github documentation updated


sagikazarmark:
Thanks for the contribution @lzeng27 !

As I pointed out in #1782, importing the examples module in the main module is not a good idea.

Anyway, I fixed this in #1794, but then I realized you had some doc fixes as well. Can you please open a separate PR for those? Thanks!

Issue #1158, opened 2017-12-21
LDAP password is stored in clear text ConfigMap


danilina-wsib:
Currently, tectonic-identity stores a clear text LDAP password in a ConfigMap. This is not secure, a secret is probably a better way to store it.

ericchiang:
Yep, we plan to address this with https://github.com/coreos/dex/issues/1099

ericchiang:
Note that you can already store it as part of the environment variable. Fields in the `storage.config` blocks are expanded from the environment.

```
storage:
- type: ldap
  config:
    password: $PASSWORD
```

PR #1797, opened 2020-09-13
Add the option to configure the bind password using an environment variable for the LDAP connector.

onkarbhat:
In a k8s deployment that uses Dex, this will allow us to define an environment variable using a k8s secret that contains the bind password instead of showing it in plaintext in a configmap. 

This partially meets requirements in #1099.

Related  issue - https://github.com/dexidp/dex/issues/1158 

cc: @sagikazarmark @bonifaido @danilina-wsib

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -191,7 +191,7 @@ func (c *ldapConnector) userMatchers() []UserMatcher {
 	if len(c.GroupSearch.UserMatchers) > 0 && c.GroupSearch.UserMatchers[0].UserAttr != "" {
 		return c.GroupSearch.UserMatchers[:]
 	}
-	
+
 	return []UserMatcher{
 		{
 			UserAttr:  c.GroupSearch.UserAttr,
@@ -247,7 +247,11 @@ func (c *Config) openConnector(logger log.Logger) (*ldapConnector, error) {
 		if c.BindPW != "" {
 			return nil, fmt.Errorf("ldap: bindPW and bindPWFromEnv are exclusive for the LDAP connector")
 		}
-		bindPW = os.Getenv(c.BindPWFromEnv)
+		var ok bool
+		bindPW, ok = os.LookupEnv(c.BindPWFromEnv)
+		if !ok || bindPW == "" {
+			return nil, fmt.Errorf("ldap: environment variable for the bind password was not set to a valid")
+		}
 	}
 	c.BindPW = bindPW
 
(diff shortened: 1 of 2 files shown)


connector/ldap/ldap.go
```diff
@@ -237,6 +239,18 @@ func (c *Config) openConnector(logger log.Logger) (*ldapConnector, error) {
...
 
+	if c.BindPW == "" && c.BindPWFromEnv == "" {
+		return nil, fmt.Errorf("ldap: bindPW or bindPWFromEnv are required for the LDAP connector")
+	}
+	bindPW := c.BindPW
+	if c.BindPWFromEnv != "" {
+		if c.BindPW != "" {
+			return nil, fmt.Errorf("ldap: bindPW and bindPWFromEnv are exclusive for the LDAP connector")
+		}
+		bindPW = os.Getenv(c.BindPWFromEnv)
```
vkamra, line 250: Use `LookupEnv` instead and fail if the environment variable is not present.

onkarbhat: Thanks @vkamra . I've addressed the comment and added a unit test to check for error cases related to LookupEnv.

heidemn:
@onkarbhat have you seen the new config templating feature, coming in the next release?
This should solve your problem already.
[examples/config-tmpl.yaml](https://github.com/dexidp/dex/pull/1965/files#diff-c7f8bd4d261664ee2b089de54d7a767b3a03409af065e9eb51fc95592cc384c0)

Note that the config file is just an example. You can define your own, using all the templating logic you need.

sagikazarmark:
Yeah, I don't think we'll be accepting this one. @onkarbhat please try the new templating feature.

onkarbhat:
Thank you @heidemn @sagikazarmark 👍🏼 .   I believe these are the steps that a user would have to execute in order to configure a password in the LDAP connector config in Dex in a K8s cluster. 

- Create a secret that contains the password.
- Mount the secret's entry as an environment variable in the Dex Pod. Lets call this env as LDAP_BIND_PW_ENV.
- Prepare a config.docker.yaml with the bindPW  set like this: `{{ getenv "LDAP_BIND_PW_ENV" "" }}`
- Store config.docker.yaml in a config map, and mount it at /etc/dex/config.docker.yaml .

sagikazarmark:
That sounds about right. I'd probably consider using [bank-vaults](https://github.com/banzaicloud/bank-vaults) or [kube-secrets-init](https://github.com/doitintl/kube-secrets-init) instead of Kubernetes secrets though.

PR #1802, opened 2020-09-17
Add constructor for static key strategy

xtremerui:
Allow user to provide a static key for signing without rotation.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"crypto/rsa"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


sagikazarmark (changes requested):
@xtremerui can you please rebase your PR?

Also, it'd be nice to hear more about your use case for Dex. I recall you maintaining a fork, but PRs like this rather tells that you use Dex as a "library" (or rather configure it in a different way in your own app).

TBH I'm usually not a huge fan of (re)using application code in other applications, but given you seem to be an active user, you might have a story to tell. Also, given that there is no way to add custom connectors at the moment, this could actually be the only way to provide better support for that use case as well.

xtremerui:
Thx for your time @sagikazarmark reviewing this.

The reason that we want to have a static key could be found here https://github.com/concourse/concourse/commit/1199aaebaa47bb0da74256d0f7416723872bb9ff

Concoures can run of multiple web nodes and each of that has its own dex server. If the key is by default rotating after 6/24 hours, the rotated key will be different in each of them. So we want to make sure there is a default way to deploy Concourse with multiple web nodes that operator don't need to do anything special to sync the signing keys between dex servers.

sagikazarmark:
Gotcha! I think the docker failure is normal at this point, so I think this is good to go.

sagikazarmark (approved):

PR #1804, opened 2020-09-21
Let Github actions always pull the base images

heidemn-faro:
Let Github actions always pull the base images, to avoid CVEs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,6 +103,7 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+          always_pull: true
           repository: dexidp/dex
           tags: ${{ steps.imagetag.outputs.value }}
           add_git_labels: true
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ fix: bin/golangci-lint ## Fix lint violations
 
 .PHONY: docker-image
 docker-image:
-	@sudo docker build -t $(DOCKER_IMAGE) .
+	@docker build -t $(DOCKER_IMAGE) .
 
 .PHONY: proto
 proto: bin/protoc bin/protoc-gen-go
--- a/Makefile
+++ b/Makefile
@@ -74,11 +74,7 @@ fix: bin/golangci-lint ## Fix lint violations
 	bin/golangci-lint run --fix
 
 .PHONY: docker-image
-docker-image: ## Pull base images, then build the image.
-	@sudo docker build --pull -t $(DOCKER_IMAGE) .
-
-.PHONY: docker-image-no-pull
-docker-image-no-pull: ## Skip pulling the base images for development.
+docker-image:
 	@sudo docker build -t $(DOCKER_IMAGE) .
 
 .PHONY: proto
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,11 @@ fix: bin/golangci-lint ## Fix lint violations
 	bin/golangci-lint run --fix
 
 .PHONY: docker-image
-docker-image:
+docker-image: ## Pull base images, then build the image.
+	@sudo docker build --pull -t $(DOCKER_IMAGE) .
+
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -74,7 +74,11 @@ fix: bin/golangci-lint ## Fix lint violations
 	bin/golangci-lint run --fix
 
 .PHONY: docker-image
-docker-image:
+docker-image: ## Pull base images, then build the image.
+	@sudo docker build --pull -t $(DOCKER_IMAGE) .
+
+.PHONY: docker-image-no-pull
+docker-image-no-pull: ## Skip pulling the base images for development.
```
heidemn-faro, line 81: An alternative would be:

```
# Makefile:
DOCKER_OPTIONS ?= --pull
docker-image:
	@sudo docker build $(DOCKER_OPTIONS) -t $(DOCKER_IMAGE) .

# Usage:
make docker-image
DOCKER_OPTIONS="" make docker-image
DOCKER_OPTIONS="--pull --no-cache" make docker-image
```

sagikazarmark:
To be completely honest, I don't really see why you need to pull anything in order to build an image. Docker build should just pull everything needed. Furthermore, it's a multi-stage build, so all you would pull is the final stage which basically nothing more than the built binaries.

So why exactly do you need this?

To be even more honest: I don't think we actually need this target. `docker build .` is not much longer than `make docker-image`, so I don't really see the point in having this target in the first place.

Makefile
```diff
@@ -74,7 +74,11 @@ fix: bin/golangci-lint ## Fix lint violations
 	bin/golangci-lint run --fix
 
 .PHONY: docker-image
-docker-image:
+docker-image: ## Pull base images, then build the image.
+	@sudo docker build --pull -t $(DOCKER_IMAGE) .
```
sagikazarmark, line 78: `sudo` is not gonna fly here for sure 🙂

heidemn-faro:
> To be completely honest, I don't really see why you need to pull anything in order to build an image. Docker build should just pull everything needed. Furthermore, it's a multi-stage build, so all you would pull is the final stage which basically nothing more than the built binaries.

My intention was to avoid any CVEs from outdated images, since I assume that `alpine:3.12` is updated regularly.
Even if these dependencies from the base image are not used, they will trigger alerts if you have a vulnerability scanner.

> To be even more honest: I don't think we actually need this target. `docker build .` is not much longer than `make docker-image`, so I don't really see the point in having this target in the first place.

Well I saw the build target, assuming that people use it, so I thought I could make it more useful.
And `make docker<TAB_FOR_AUTOCOMPLETE>` is more convienient to type compared to the Docker commandline.

So if you don't like this PR, we can handle this in our build system, no problem.

Makefile
heidemn-faro: Not sure what you mean by that?

sagikazarmark:
Yeah, I feel like the build system should make sure to always pull a fresh image. I don't see this target having a very long life in this repo to be honest.

Makefile
techknowlogick: This line hasn't been changed from existing makefile, it is only moved. So `sudo` isn't an addition.

With that being said, perhaps `sudo` should be removed, even if it wasn't added by this PR.

sagikazarmark: Sorry, missed that it was already there. :) Yeah, but I don't think this PR will happen anyway.

.github/workflows/ci.yml
```diff
@@ -103,6 +103,7 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+          always_pull: true
```
heidemn-faro, line 106: If we're afraid of Docker's new pull limits, we could only pull if branch = master.
Any opinions?

sagikazarmark: I don't think this option has any effect. AFAIK GitHub doesn't cache images.

heidemn-faro: If your assumption is true, it shouldn't harm to add it.

PR #1812, opened 2020-10-03
fix: Replace deprecated teams endpoint in bitbucket connector

### LINKED ISSUE #1811: Depreciation of BitbucketCloud teams and users endpoints
## Problem Summary
The `/2.0/users` and `/2.0/teams` endpoints are being deprecated but not the `/2.0/user`, so there is a probability that the current BitbucketCloud connector won't be working anymore from October 14th. Note that I don't know enough of the internal logic of the connector and dex, so maybe with only `/2.0/user`, the login might still work, but not sure.

https://developer.atlassian.com/cloud/bitbucket/bitbucket-api-teams-deprecation/
> On October 14, 2020 all of the /2.0/teams/ and most of the /2.0/users/ API endpoints will be removed. You need to update your Bitbucket Cloud integrations to use the new /2.0/workspaces/ endpoints. These endpoints are available today, so you can start updating your code immediately.

## Resolution
The api call [`/2.0/user/emails`](https://developer.atlassian.com/bitbucket/api/2/reference/resource/user/emails) can be kept.

The api call [`/2.0/teams?role=member`](https://developer.atlassian.com/bitbucket/api/2/reference/resource/teams) shall be transitioned to [`/2.0/user/permissions/teams`](https://developer.atlassian.com/bitbucket/api/2/reference/resource/user/permissions/teams).

I do not know if the scope should be changed or kept as-is, to access the `/2.0/user/permissions/teams` endpoint , as there is no mention of any new workspace scope in the [docs](https://developer.atlassian.com/bitbucket/api/2/reference/meta/authentication#scopes-bbc). So maybe it does not need to be changed.

nabokihms:
Replace deprecated `/teams` API endpoint with the actual one `/user/permissions/teams`.
Support of `/teams` endpoint ends 14 October 2020. 
* With this PR, we do not need `teams` scope to access user teams anymore
* Output of endpoints differs a little

Closes #1811

P.S. I also tested these changes with changes from #1688. Everything works fine and looks compatible despite there are some conflicts in tests between these PRs.

### DIFF
--- a/Documentation/connectors/bitbucketcloud.md
+++ b/Documentation/connectors/bitbucketcloud.md
@@ -10,7 +10,7 @@ When a client redeems a refresh token through dex, dex will re-query Bitbucket t
 
 Register a new OAuth consumer with [Bitbucket](https://confluence.atlassian.com/bitbucket/oauth-on-bitbucket-cloud-238027431.html) ensuring the callback URL is `(dex issuer)/callback`. For example if dex is listening at the non-root path `https://auth.example.com/dex` the callback would be `https://auth.example.com/dex/callback`.
 
-The application requires the user to grant the `Read Account` and `Read Team membership` permissions. The latter is required only if group membership is a desired claim.
+The application requires the user to grant only the `Read Account` permission.
 
 The following is an example of a configuration for `examples/config-dev.yaml`:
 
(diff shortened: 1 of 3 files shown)


sagikazarmark (approved):
Thanks @nabokihms for quickly fixing this!

Issue #1813, opened 2020-10-04
Password grant flow doesn't log errors from Login method


tomqwpl:
Standard flow:

		identity, ok, err := passwordConnector.Login(r.Context(), scopes, username, password)
		if err != nil {
			s.logger.Errorf("Failed to login user: %v", err)

Password flow:

	identity, ok, err := passwordConnector.Login(r.Context(), parseScopes(scopes), username, password)
	if err != nil {
		s.tokenErrHelper(w, errInvalidRequest, "Could not login user", http.StatusBadRequest)

PR #1814, opened 2020-10-04
fix: log errors from login during password grant

tomqwpl:
This simply logs the message in the same way it is for the normal code flow

Issues: #1813


al45tair:
This PR needs to be closed and resubmitted from a different branch. Sorry.

PR #1819, opened 2020-10-05
fix: allow Authorization header when doing CORS

### LINKED ISSUE #1532: OPTIONS /userinfo denied on Access-Control-Request-Headers: Authorization

al45tair:
The Authorization header needs to be allowed when doing CORS because
otherwise /userinfo can't work.  It isn't one of the headers
explicitly allowed by default by Gorilla, so we have to call
handlers.AllowedHeaders() to specify it.

Fixes #1532

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -294,8 +294,14 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
 	handleWithCORS := func(p string, h http.HandlerFunc) {
 		var handler http.Handler = h
 		if len(c.AllowedOrigins) > 0 {
-			corsOption := handlers.AllowedOrigins(c.AllowedOrigins)
-			handler = handlers.CORS(corsOption)(handler)
+			allowedHeaders := []string{
+				"Authorization",
+			}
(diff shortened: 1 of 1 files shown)


sagikazarmark (approved):
Thanks!

Issue #1820, opened 2020-10-05
"web" directory copied incorrectly in Dockerfile


al45tair:
The Dockerfile was changed from

    COPY web /web

To

    COPY web .

Unfortunately, the latter copies *the contents* of the directory "web" to the current directory. i.e. the final Docker image ends up looking like this:

    / $ ls
    bin        home       mnt        root       srv        templates  usr
    dev        lib        opt        run        static     themes     var
    etc        media      proc       sbin       sys        tmp

instead of like this:

    / $ ls /
    bin    etc    lib    mnt    proc   run    srv    tmp    var
    dev    home   media  opt    root   sbin   sys    usr    web

The upshot is that when you try to use the Docker image, it terminates with

    failed to initialize server: server: failed to load web static: load web dir: directory "./web" does not exist

nabokihms:
Closed in https://github.com/dexidp/dex/pull/1821

PR #1821, opened 2020-10-05
fix: copy the "web" directory.

al45tair:
The Dockerfile was changed to do

    COPY web .

which is unfortunate because that means copy the *contents* of the directory
"web", rather than copying the directory itself which is what the author
intended.

Issues: #1820


sagikazarmark (approved):
Thanks!

PR #1822, opened 2020-10-05
Allow public clients (e.g. SPAs using implicit flow or PKCE) to have redirect URLs other than localhost

heidemn-faro:
Relates to:
* https://github.com/dexidp/dex/pull/1722 (also allow usual IP addresses)
* https://github.com/dexidp/dex/pull/1784 (PKCE)

### DIFF
--- a/server/oauth2.go
+++ b/server/oauth2.go
@@ -588,12 +588,15 @@ func (s *Server) validateCrossClientTrust(clientID, peerID string) (trusted bool
 }
 
 func validateRedirectURI(client storage.Client, redirectURI string) bool {
-	if !client.Public {
-		for _, uri := range client.RedirectURIs {
-			if redirectURI == uri {
-				return true
-			}
+	// Allow named RedirectURIs for both public and non-public clients.
+	// This is required make PKCE-enabled web apps work, when configured as public clients.
+	for _, uri := range client.RedirectURIs {
+		if redirectURI == uri {
+			return true
 		}
+	}
+	// For non-public clients, only named RedirectURIs are allowed.
+	if !client.Public {
 		return false
 	}
 
--- a/server/oauth2.go
+++ b/server/oauth2.go
@@ -595,8 +595,9 @@ func validateRedirectURI(client storage.Client, redirectURI string) bool {
 			return true
 		}
 	}
-	// For non-public clients, only named RedirectURIs are allowed.
-	if !client.Public {
+	// For non-public clients or when RedirectURIs is set, we allow only explicitly named RedirectURIs.
+	// Otherwise, we check below for special URIs used for desktop or mobile apps.
+	if !client.Public || len(client.RedirectURIs) > 0 {
 		return false
 	}
 
--- a/server/oauth2_test.go
+++ b/server/oauth2_test.go
@@ -340,6 +340,7 @@ func TestValidRedirectURI(t *testing.T) {
 				RedirectURIs: []string{"http://foo.com/bar"},
 			},
 			redirectURI: "http://foo.com/bar/baz",
+			wantValid:   false,
 		},
 		{
 			client: storage.Client{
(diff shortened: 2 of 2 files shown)


server/oauth2.go
```diff
@@ -572,12 +572,15 @@ func (s *Server) validateCrossClientTrust(clientID, peerID string) (trusted bool
...
+	// This is required make PKCE-enabled web apps work, when configured as public clients.
+	for _, uri := range client.RedirectURIs {
+		if redirectURI == uri {
+			return true
 		}
+	}
+	// For non-public clients, only named RedirectURIs are allowed.
+	if !client.Public {
 		return false
 	}
```
heidemn-faro, line 585: Our use case is to use PKCE for a web app that doesn't run on "localhost".
Therefore, we would like that `RedirectURIs` are also considered for public clients.
PKCE was definitely also intended for web apps, and you can already configure such web apps e.g. in Okta or Auth0.

The following code to allow any localhost URI, `redirectURIOOB` or `deviceCallbackURI` for public clients is still active.
An alternative could be to no longer allow these special URLs implicitly when `RedirectURIs` is non-empty.
I don't have a strong opinion if this would be better. The maintainers should decide.

tkleczek: I am curious - why don't you configure the app as confidential client if it is a web app and can keep the client secret secure?

heidemn-faro:
@tkleczek I saw your comment in an e-mail, can't find it here any more:
> I am curious - why don't you configure the app as confidential client if it is a web app and can keep the client secret secure?

What about SPAs without server component, hosted as static websites?
These can use Implicit Flow, or Auth Code + PKCE, and don't need a secret then.

Read more here: https://developer.okta.com/blog/2019/08/22/okta-authjs-pkce

tkleczek:
@heidemn-faro Yes, I realized this soon after I've written my comment and therefore deleted it.

server/oauth2.go
HEllRZA: I think the implementation is good as it is. The alternative of not supporting the special URLs, when `redirectURIs` is defined, would break currently working configurations.

HEllRZA (approved):
LGTM

@sagikazarmark I would propose to add this PR to [v2.26.0](https://github.com/dexidp/dex/milestone/18) milestone, because it enables PKCE #1784 for single page web applications.

server/oauth2.go
tkleczek: I would personally prefer the logic to only allow whitelisted redirect uris, if the list is non-empty.

This would tighten the security and would not break existing configurations that do not use RedirectURIs for public clients.
And I bet the most configurations that do currently use it for public clients operate under the false assumption that the list is actually respected (I know we did in our deployment for some time).

HEllRZA: I fear that there are configurations that may be working by accident, because currently it is possible to have the RedirectURIs wrongly configured, and it still works.

Let's see what the maintainers decide.

sagikazarmark: > I would personally prefer the logic to only allow whitelisted redirect uris, if the list is non-empty.

@tkleczek AFAICT this is not the case now, but the rest of the conditions provide backwards compatibility and limit redirect uris to `redirectURIOOB`, `deviceCallbackURI` or `localhost` for public clients which seems fine. WDYT?

tkleczek: My concern is that for SPAs allowing localhost redirects seems like an unnecessary risk.
But if we care about potential backwards compatibility issues more, then I am fine with the change as it is.

logikone:
Hey, thanks for this! Just wanted to ask quickly what the timeline is on getting this merged and releasing v2.26.0? Thanks in advance!

sagikazarmark:
I'll try to get this merged and released this week.

logikone:
> I'll try to get this merged and released this week.

@sagikazarmark awesome, thank you sir!

server/oauth2.go
heidemn-faro: > My concern is that for SPAs allowing localhost redirects seems like an unnecessary risk.

That's actually true. Consider that e.g. a malicious desktop app could obtain a token intended for https://webapp.com , since the desktop app can run a server listening on http://localhost .
If the user is already logged in at the IDP in the browser, then this could probably be done without user interaction.
The desktop app could then call the API of https://webapp.com with the token.

So I would prefer not to allow these special URLs implicitly if `RedirectURIs` is non-empty.

In the future, three additional boolean options could be added to explicitly (dis)allow each of `redirectURIOOB`, `deviceCallbackURI` or `localhost`-ish URLs. But this would definitely break backwards compatibility, and might be over-engineering.

heidemn-faro: @sagikazarmark I added more tests for these special URIs in https://github.com/dexidp/dex/pull/1822/commits/c15e2887bc262a7dae2df741c7ba06f866272272 .  
Then I made the change suggested by @tkleczek in https://github.com/dexidp/dex/pull/1822/commits/162073b33e46e502696b47ee97f871b944540114 : `if !client.Public || len(client.RedirectURIs) > 0 { return false }`

Please take another look.

sagikazarmark (approved):
LGTM, thanks!

bonifaido (approved):

PR #1832, opened 2020-10-09
Add Emacs backups to .gitignore

al45tair:
Emacs backup files should be in .gitignore, to avoid accidental check-ins.


sagikazarmark:
Thanks, but editor specific ignores should be added to your global gitignore.

Issue #1831, opened 2020-10-09
Generic database connector


al45tair:
Many of us have an existing database containing users (and groups) and would like to be able to plug Dex into it. There should be a connector that allows for this, perhaps allowing for customisable SQL queries (rather like the way the LDAP connector allows for custom LDAP queries).

sagikazarmark:
Rewriting the SQL layer is high on our priority list. We are looking at https://github.com/facebook/ent

I'm not sure how generic we can be. Supporting customizable queries is certainly a deep rabbit hole that we don't necessarily want to go down into. Supporting more storage drivers is an option, but adds an extra maintenance burden.

Postgres, Mysql, sqlite, etcd are probably generic enough, so we will stick to those for now.

al45tair:
I've actually got a PR I'm working on for this.

Issue #12, opened 2015-08-17
implicit grants


bobbyrullo:
<a href="https://github.com/sym3tri"><img src="https://avatars.githubusercontent.com/u/224222?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [sym3tri](https://github.com/sym3tri)**
_Thursday Dec 04, 2014 at 23:46 GMT_
_Originally opened as https://github.com/coreos-inc/auth/issues/128_

----

The spec requires us to support it. We need to do so or document that we intentionally diverge from the spec.

> `grant_types_supported`
> OPTIONAL. JSON array containing a list of the OAuth 2.0 Grant Type values that this OP supports. Dynamic OpenID Providers MUST support the authorization_code and implicit Grant Type values and MAY support other Grant Types. If omitted, the default value is ["authorization_code", "implicit"].

http://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

ericchiang:
Moved to #256

PR #1833, opened 2020-10-09
feat: add a generic database connector

al45tair:
Adds a generic database connector that can be used to plug Dex into an arbitrary SQL database.

Issues: #1831


al45tair:
There might still be some work to do on this (when we start actively using it, we may well find things we want to change), but I wanted to create the PR now so that you can see what we're working on and comment on it.

I think this should be quite useful to other people besides just us here at World Programming.

al45tair:
Just a note that this build failed because passlib doesn't build on 32-bit systems (see [passlib #12](https://github.com/hlandau/passlib/pull/12)). This needs fixing upstream.

sagikazarmark:
Thanks for working on this @al45tair 

As I said, I'm not sure how I feel about allowing users to customize queries from a config file, but I can see the need for a generic connector.

What do you think about a gRPC API defined by Dex that you can implement for your needs and use as a connector?

al45tair:
@sagikazarmark In terms of *this* connector, it's only really the same as the LDAP connector, except instead of LDAP queries it has SQL queries. It's also exactly what you want if you have an existing system that already has users (and perhaps groups) in an SQL database, but you'd like to start using Dex to handle the authentication for that system.

It also has the advantage over Dex's built-in password database support that Dex isn't responsible for the schema *at all*, so we don't need to worry about database migrations and so on, and obviously it supports groups as well as users.

An external connector API using gRPC would certainly be a nice way to allow people to plug in to Dex without having to actually change the source code though. If Dex had had such a thing, it's possible we might have used it and then had a component with fixed queries that actually correspond to our specific back-end database. But then I don't mind writing gRPC servers :-) I can imagine that there are people who very much would mind writing them (or just plain couldn't), but could cope with writing some SQL in a config file.

sagikazarmark:
> it's only really the same as the LDAP connector, except instead of LDAP queries it has SQL queries

LDAP is an established protocol for this purpose with a large amount of experience available in the field, whereas this PR (no offense) implements an arbitrary, SQL based solution. I guess you are right that they are similar, but for some reason the idea of writing SQL queries and column names in a config file makes me itchy. (Not to mention this will only work SQL drivers already available in Dex and I already see the PRs requesting to add Oracle and MSSQL drivers)

If we can find a way to provide a generic enough protobuf interface, I wouldn't mind hosting this as an implementation with huge exclamation marks warning users not to use it and instead implement and properly test (yeah, testing is another huge question) custom servers.

That's my (semi-professionally phrased) opinion at the moment which is why I would rather hold off on merging this PR, until we come up with at least a few alternative ideas. Change my mind.

al45tair:
> LDAP is an established protocol for this purpose with a large amount of experience available in the field, whereas this PR (no offense) implements an arbitrary, SQL based solution.

I'd say this particular use of SQL databases is actually far more common than the use of LDAP. Most database-backed websites and web apps start out with users (and groups, if they support groups) in tables in their databases; they mostly *don't* start out with LDAP, and until recently I'd say most of them wouldn't have started out with OpenID Connect support either.

LDAP is common in corporate/institutional/governmental networks, but outside of that it's rather uncommon.

I'll add that I see this kind of connector serving two purposes:

1. In some web applications, it could be used to provide a local user/group store for that application, as an alternative to the user storage built in to Dex. The benefit of that is that Dex doesn't need to worry about the schema, or about providing functions for managing users and groups, while the application can implement whatever user/group management UI makes sense without needing to satisfy any significant constraints from the Dex side.

2. In situations where a site or application wishes to transition to OpenID Connect, this could be used to connect to their existing database-backed user/group storage, both for compatibility with existing deployments and also to provide a stepping stone where Dex is used in a simple way (and OpenID Connect support to talk to it implemented) before the system is ready to allow external OpenID Connect, SAML, LDAP and so on.

> for some reason the idea of writing SQL queries and column names in a config file makes me itchy

I don't know that it should. What are you actually worrying about in that regard? What is it that you worry about with SQL queries that you don't worry about with LDAP queries? It isn't as if we're letting users who log-in to the system write SQL queries (or LDAP queries for that matter).

> Not to mention this will only work SQL drivers already available in Dex and I already see the PRs requesting to add Oracle and MSSQL drivers

Agreed, that is an issue, and coupled with the Go database library's decision to leave argument substitution to the underlying driver (thus resulting in database-specific syntax), it's potentially an annoying one (that's why I used sqlx here rather than plain database/sql, as that lets us avoid the argument substitution problem).

> I would rather hold off on merging this PR

As I said in my earlier message, I'd like to hold off anyway because it might be that we find some problems or make some useful enhancements while actually trying to use it.

I *like* the idea of having a gRPC connector interface (and something similar for the middleware thing too), but I think having a generic SQL connector is at least as useful as having the similar one for LDAP. If we can get to the bottom of why you don't like SQL queries and column names in the config, I think that would at least be an interesting place to start.

janwillies:
Even for existing ldap infrastructure I'd like to add that it is sometimes limited in capabilites (e.g. max number of groups, no api accessible, attached kerberos issues) so folks who want fine grained authorization have to build custom user-org mapping databases. This PR enables access to that information.

Seeing #1841 we might have exactly the same use case as @al45tair

al45tair:
I don't think this will get merged in its current form (mainly because of the database driver issue). I'm closing the PR; the code will still exist in the upstream branch if anyone wants it.

PR #1836, opened 2020-10-16
feat: reduce image size without apk cache

lcostea:
Reduced size by almost 2MB by not using the apk cache


sagikazarmark (approved):
Nice catch, thanks!

PR #1837, opened 2020-10-17
fix: Bump golangci-lint version and fix some linter's problems

nabokihms:
There is a naughty warning in the "Run linter" CI job logs:
```
[runner] Can't run linter goanalysis_metalinter: ST1001: failed prerequisites:...
```
Examples: [1](https://github.com/dexidp/dex/runs/1264022987?check_suite_focus=true), [2](https://github.com/dexidp/dex/runs/1247767485?check_suite_focus=true), [3](https://github.com/dexidp/dex/runs/1260081191?check_suite_focus=true)

Bumping the golangci-lint version to the actual one fixes the problem. It requires golangci-linter configuration changes because the `enable-all` option was marked as deprecated. Users have to use the `disable-all` option and manually enable linters they want.

I also fixed some style problems to enable the following linters:
- golint
- gochecknoinits
- gocritic
- ineffassign
- nakedret
- staticcheck
- varcheck

Think it might be a useful improvement.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,6 +19,7 @@ linters:
   - goimports
   - golint
   - gosimple
+  - gocritic
   - govet
   - ineffassign
   - interfacer
@@ -40,6 +41,5 @@ linters:
   # - gocyclo
   # - lll
   # - goconst
-  # - gocritic
   # - errcheck
   # - dupl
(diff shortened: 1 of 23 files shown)


sagikazarmark (approved):
Thanks @nabokihms

PR #1839, opened 2020-10-20
Allow configuration of returned auth proxy header

seuf:
This is a small PR to allow configuration of the header returned by the auth proxy.
We are using traefik as reverse proxy + traefik-forward-auth plugged on google to access our argocd interface.
But in traefik-forward-auth, the returned header after authentication is `X-Forwarded-User`.
With this feature, we can now login to argocd using the dex config :
```
connectors:
      - type: authproxy
        id: traefik
        name: traefik
        config:
          headerName: X-Forwarded-User
```
dex will fetch the header set by traefik forward auth and authenticate the user with the correct user email.

### DIFF
--- a/connector/authproxy/authproxy.go
+++ b/connector/authproxy/authproxy.go
@@ -20,13 +20,18 @@ type Config struct {
 
 // Open returns an authentication strategy which requires no user interaction.
 func (c *Config) Open(id string, logger log.Logger) (connector.Connector, error) {
-	return &callback{UserHeader: c.UserHeader, logger: logger, pathSuffix: "/" + id}, nil
+	userHeader := c.UserHeader
+	if userHeader == "" {
+		userHeader = "X-Remote-User"
+	}
+
+	return &callback{userHeader: userHeader, logger: logger, pathSuffix: "/" + id}, nil
 }
 
 // Callback is a connector which returns an identity with the HTTP header
 // X-Remote-User as verified email.
 type callback struct {
-	UserHeader string
+	userHeader string
 	logger     log.Logger
 	pathSuffix string
 }
@@ -46,13 +51,9 @@ func (m *callback) LoginURL(s connector.Scopes, callbackURL, state string) (stri
 
 // HandleCallback parses the request and returns the user's identity
 func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connector.Identity, error) {
-	userHeader := "X-Remote-User" // Default value
-	if m.UserHeader != "" {
-		userHeader = m.UserHeader
-	}
-	remoteUser := r.Header.Get(userHeader)
+	remoteUser := r.Header.Get(m.userHeader)
 	if remoteUser == "" {
-		return connector.Identity{}, fmt.Errorf("required HTTP header %s is not set", m.UserHeader)
+		return connector.Identity{}, fmt.Errorf("required HTTP header %s is not set", m.userHeader)
 	}
 	// TODO: add support for X-Remote-Group, see
 	// https://kubernetes.io/docs/admin/authentication/#authenticating-proxy
--- a/connector/authproxy/authproxy.go
+++ b/connector/authproxy/authproxy.go
@@ -15,23 +15,18 @@ import (
 // Config holds the configuration parameters for a connector which returns an
 // identity with the HTTP header X-Remote-User as verified email.
 type Config struct {
-	HeaderName string   `json:"headerName"`
-	Groups     []string `json:"groups"`
+	UserHeader string `json:"userHeader"`
 }
 
 // Open returns an authentication strategy which requires no user interaction.
 func (c *Config) Open(id string, logger log.Logger) (connector.Connector, error) {
-	if c.HeaderName == "" {
-		c.HeaderName = "X-Remote-User"
-	}
-	return &callback{headerName: c.HeaderName, logger: logger, pathSuffix: "/" + id, groups: c.Groups}, nil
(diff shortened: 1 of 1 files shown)


seuf:
Hi, Is it possible to review this ? we are using this branch in production for a few weeks now, and it works great.

connector/authproxy/authproxy.go
```diff
@@ -14,16 +14,24 @@ import (
 
 // Config holds the configuration parameters for a connector which returns an
 // identity with the HTTP header X-Remote-User as verified email.
-type Config struct{}
+type Config struct {
+	HeaderName string   `json:"headerName"`
```
sagikazarmark, line 18: I wonder if this should be called something like `UserHeaderName` or just `UserHeader`. Another common header name is `X-Forwarded-Groups` for group information.

connector/authproxy/authproxy.go
```diff
@@ -14,16 +14,24 @@ import (
...
+	Groups     []string `json:"groups"`
+}
 
 // Open returns an authentication strategy which requires no user interaction.
 func (c *Config) Open(id string, logger log.Logger) (connector.Connector, error) {
-	return &callback{logger: logger, pathSuffix: "/" + id}, nil
+	if c.HeaderName == "" {
```
sagikazarmark, line 24: Note that this actually modifies the config. I think it would be better to create a local variable for the header and fall back to the default if it's empty.

sagikazarmark (changes requested):

connector/authproxy/authproxy.go
```diff
@@ -14,16 +14,24 @@ import (
 
 // Config holds the configuration parameters for a connector which returns an
 // identity with the HTTP header X-Remote-User as verified email.
-type Config struct{}
+type Config struct {
+	HeaderName string   `json:"headerName"`
+	Groups     []string `json:"groups"`
```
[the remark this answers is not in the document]
seuf: ok, I can split in 2 PR to add also the groups configuration.

connector/authproxy/authproxy.go
seuf: ok, I've used a default value in the callback directly without modifying the config.

connector/authproxy/authproxy.go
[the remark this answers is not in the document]
sagikazarmark: You can submit it in a separate PR, but we are working on #1841 which will add this feature generally to all connectors, so I'm not sure we'll be accepting any more config changes like this.

connector/authproxy/authproxy.go
```diff
@@ -14,16 +14,19 @@ import (
...
 // identity with the HTTP header X-Remote-User as verified email.
-type Config struct{}
+type Config struct {
+	UserHeader string `json:"userHeader"`
+}
 
 // Open returns an authentication strategy which requires no user interaction.
 func (c *Config) Open(id string, logger log.Logger) (connector.Connector, error) {
-	return &callback{logger: logger, pathSuffix: "/" + id}, nil
+	return &callback{UserHeader: c.UserHeader, logger: logger, pathSuffix: "/" + id}, nil
```
sagikazarmark, line 23:
```suggestion
	userHeader := c.UserHeader
	if userHeader == "" {
		userHeader = "X-Remote-User"
	}

	return &callback{userHeader: userHeader, logger: logger, pathSuffix: "/" + id}, nil
```

connector/authproxy/authproxy.go
```diff
@@ -14,16 +14,19 @@ import (
...
 }
 
 // Callback is a connector which returns an identity with the HTTP header
 // X-Remote-User as verified email.
 type callback struct {
+	UserHeader string
```
sagikazarmark, line 29:
```suggestion
	userHeader string
```

connector/authproxy/authproxy.go
```diff
@@ -43,9 +46,13 @@ func (m *callback) LoginURL(s connector.Scopes, callbackURL, state string) (stri
 
 // HandleCallback parses the request and returns the user's identity
 func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connector.Identity, error) {
-	remoteUser := r.Header.Get("X-Remote-User")
+	userHeader := "X-Remote-User" // Default value
+	if m.UserHeader != "" {
+		userHeader = m.UserHeader
+	}
```
sagikazarmark, line 52:
```suggestion
```

connector/authproxy/authproxy.go
```diff
@@ -43,9 +46,13 @@ func (m *callback) LoginURL(s connector.Scopes, callbackURL, state string) (stri
...
+	remoteUser := r.Header.Get(userHeader)
```
sagikazarmark, line 53:
```suggestion
	remoteUser := r.Header.Get(m.userHeader)
```

connector/authproxy/authproxy.go
```diff
@@ -43,9 +46,13 @@ func (m *callback) LoginURL(s connector.Scopes, callbackURL, state string) (stri
...
 	if remoteUser == "" {
-		return connector.Identity{}, fmt.Errorf("required HTTP header X-Remote-User is not set")
+		return connector.Identity{}, fmt.Errorf("required HTTP header %s is not set", m.UserHeader)
```
sagikazarmark, line 55:
```suggestion
		return connector.Identity{}, fmt.Errorf("required HTTP header %s is not set", m.userHeader)
```

sagikazarmark (changes requested):

sagikazarmark (approved):
Thanks @seuf

PR #1846, opened 2020-10-28
feat: Add refresh token expiration and rotation settings

### LINKED ISSUE #787: server: add option to turn off refresh token rotation
Hi,

There are known problem with refresh token's usage that can be reproduced in following scenarios:
1) Client tries to refresh token and IDP successfully refreshes and generates new tokens, but client fails to accept tokens for example because of network issue or client failure.
2) Client of the IDP is some kind of User Agent, like smart device, browser or so and it makes a lot of concurrent requests to Resource Server. At some time Client determines that he need to refresh tokens and it possibly can do several refresh requests at the same time (for example in different browser tabs) and one will fail.

We have faced with such issues in production and we decided to make refresh tokens be usable several times (with some expiration period that starts when first refresh is done). This approach allows to handle issues from client side at it is not contradicts with OAuth2/OIDC1 RFC's.

### LINKED ISSUE #973: Refresh token become invalid after the same user requests authorization
It looks a refresh token become invalid after the owner of the token requests to be authorized, even if the request was sent from another machine. Is this a correct behavior?

If this is by design, how can we implement an application which can authorize the same user from multiple machines?

**Repro steps**:

1. Check out the master of the repository and build dex and example-app.
2. Add your static password on ./examples/config-dev.yaml.
3. Run `./cmd/dex/dex serve ./examples/config-dev.yaml`.
4. Run `./cmd/example-app/example-app`.
5. Open http://127.0.0.1:5555 with an browser and...
    5.1. Click "Login" button with "Request offline access" checked.
    5.2. Click "Log in with Email", enter your email and password, and click "Grant Access".
    5.3. Confirm tokens have been received.
6. Open http://127.0.0.1:5555 with another browser and...
    6.1. Log in with the same email and password as used at step 5.
7. Return to the browser (at step 5.3.) and click "Redeem refresh token" to confirm if the refresh token is valid.

**Expected results**:

Fields "Token", "Claims", and "Refresh Token" are updated.

**Actual results**:

The following message is displayed.
```
failed to get token: oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"invalid_request","error_description":"Refresh token is invalid or has already been claimed by another client."}
```

### LINKED ISSUE #1385: Refresh token changes to work with IFTTT
Hi folks,

This is a great project, so thanks for all the effort in this.  

Recently I've tried to create a small project in IFTTT that would trigger an authentication with dex running locally in our cluster.  I found two stumbling blocks to this:

First, IFTTT passes the access_token, not the id_token to our service endpoint.  I found that patching in https://github.com/dexidp/dex/pull/1315 worked here.  Is there a chance that functionality could be approved and merged?

Second, IFTTT has a validation test called "Ensure older refresh token does not expire immediately".  I made a small patch to dex that (I think) enables this.  I'm more than happy to share this if it is something the dex project would like.  

I've seen some discussing that the existing behavior is by default -- I totally understand that, but I can also see the point of a small window for re-requesting a token as described here: https://platform.ifttt.com/docs/faq#authentication

Before I package up my changes and make a PR, is this something that the project would like?

### LINKED ISSUE #1685: Refresh Token Expiry?

nabokihms:
### Description
This PR adds various settings to adjust refresh token rotation and expiration. I will describe the purpose of each option below. Some of them required to improve security and suggested by specs. On the other hand, there are options aimed to improve UX in exchange for security.

### Options
Current refresh tokens issued by dex has no lifetime limit. Even with enabled rotation, it possible (for example, for device thieves) to use stolen token endlessly if the original user never logs in again. Options below allow users to make their authentication pipeline more securable.
* **validForIfNotUsed** - it's recommended to invalidate refresh tokens if they are not used.
* **absoluteLifetime** - a stricter variant of the previous option, forces users to reauthenticate and obtain a new refresh token no matter what.

The following options are aimed to improve user experience. [RFC6819](https://tools.ietf.org/html/rfc6819#section-5.2.2.3) mention, that users may encounter problems with every-request refresh token rotation using clusterized applications (or enabling high availability mode). Dex has to provide settings to deal with them effectively to attract users and adopters.

* **disableRotation** - completely disables every-request rotation. If it's enabled, we force the user to specify one of the previous options. Costs are huge, but it's okay to use in isolated environments.
* **reuseInterval** - allows getting the same refresh token from refresh endpoint within a specified interval, but only if the user's request contains the previous refresh token. This required to effectively deal with network lags, concurrent requests, and so on. The same option exists in [auth0 refresh tokens rotation settings](https://auth0.com/docs/tokens/refresh-tokens/configure-refresh-token-rotation).

The default rotation settings will remain the same.
Closes #787
Closes #973
Closes #1385
Closes #1685


server/handlers.go
```diff
@@ -1142,22 +1155,28 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
...
 	}
 
 	lastUsed := s.now()
 	updater := func(old storage.RefreshToken) (storage.RefreshToken, error) {
-		if old.Token != refresh.Token {
-			return old, errors.New("refresh token claimed twice")
+		if s.refreshTokenPolicy.RotationEnabled() {
+			if old.Token != refresh.Token {
+				if s.refreshTokenPolicy.AllowedToReuse(old.LastUsed) && old.ObsoleteToken == refresh.Token {
+					newToken.Token = old.Token
```
tkleczek, line 1171: I would expect old.Claims and old.LastUsed to be updated in this scenario as well

server/rotation.go
```diff
@@ -177,3 +177,73 @@ func (k keyRotater) rotate() error {
...
 }
+
+type RefreshTokenPolicy struct {
+	rotateRefreshTokens bool // enable rotation
+
+	absoluteLifetime  time.Duration // interval from token creation to the end of its life
+	validIfNotUsedFor time.Duration // interval from last token update to the end of its life
+	reuseInterval     time.Duration // interval within which old refresh token is allowed to be reused
+
+	Clock func() time.Time
```
tkleczek, line 188: does it need to be exported? also I see that in other places such function is called just 'now' (see server.Server or server.Config)

server/rotation_test.go
```diff
@@ -100,3 +101,30 @@ func TestKeyRotater(t *testing.T) {
...
+	t.Run("Allowed", func(t *testing.T) {
+		r.Clock = func() time.Time { return lastTime }
+		require.Equal(t, true, r.AllowedToReuse(lastTime))
+		require.Equal(t, false, r.ExpiredBecauseUnused(lastTime))
+		require.Equal(t, false, r.CompletelyExpired(lastTime))
+	})
+
+	t.Run("Expired", func(t *testing.T) {
+		r.Clock = func() time.Time { return lastTime.Add(2 * time.Minute) }
+		time.Sleep(1 * time.Second)
```
tkleczek, line 125: we should not need to sleep in this test

server/server_test.go
```diff
@@ -648,6 +648,13 @@ func TestOAuth2CodeFlow(t *testing.T) {
 			})
 			defer httpServer.Close()
 
+			policy, err := NewRefreshTokenPolicyFromConfig(s.logger, false, "", "", "")
```
tkleczek, line 651: how about making this a part of newTestServer not to duplicate code?

storage/sql/migrate.go
```diff
@@ -174,6 +174,9 @@ var migrations = []migration{
 			alter table refresh_token
 				add column token text not null default '';`,
 			`
+			alter table refresh_token
```
tkleczek, line 177: This is not right, we should add a new migration entry, at the end of the list, containing just this statement.

tkleczek (changes requested):
I really like the PR overall, thanks!, it would be useful in our company as well.

My 2 cents:

I would design the AllowedToReuse part differently. If I understand the code correctly - with current design two successive requests to refresh token handler will completely rotate-away the old refresh token (first moves it to ObsoleteToken, the second removes completely). What I would expect is that the second rotation does not happen until the ObsoleteToken really 'expires' i.e. reuseInterval passes from the time it was created

I would also say we should be adding some more tests to server_test to verify the new token refresh scenarios.

server/handlers.go
```diff
@@ -1030,14 +1030,27 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
...
+
 	if refresh.ClientID != client.ID {
 		s.logger.Errorf("client %s trying to claim token for client %s", client.ID, refresh.ClientID)
 		s.tokenErrHelper(w, errInvalidRequest, "Refresh token is invalid or has already been claimed by another client.", http.StatusBadRequest)
 		return
 	}
 	if refresh.Token != token.Token {
-		s.logger.Errorf("refresh token with id %s claimed twice", refresh.ID)
-		s.tokenErrHelper(w, errInvalidRequest, "Refresh token is invalid or has already been claimed by another client.", http.StatusBadRequest)
+		if !s.refreshTokenPolicy.AllowedToReuse(refresh.LastUsed) || refresh.ObsoleteToken != token.Token {
```
tkleczek, line 1040: We should handle the case in which refresh.ObsoleteToken == "" which happens for refresh entries that we written by the old version of the code and disallow refresh in that case

cmd/dex/config.go
```diff
@@ -282,6 +282,9 @@ type Expiry struct {
 
 	// DeviceRequests defines the duration of time for which the DeviceRequests will be valid.
 	DeviceRequests string `json:"deviceRequests"`
+
+	// RefreshToken defines refresh tokens expiry policy
+	RefreshToken RefreshTokenExpiry `json:"refreshTokens"`
```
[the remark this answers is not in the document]
nabokihms: Yes, good catch! Fixed.

cmd/dex/serve.go
```diff
@@ -277,6 +277,18 @@ func serve(cmd *cobra.Command, args []string) error {
 		logger.Infof("config device requests valid for: %v", deviceRequests)
 		serverConfig.DeviceRequestsValidFor = deviceRequests
 	}
+	refreshTokenPolicy, err := server.NewRefreshTokenPolicyFromConfig(
```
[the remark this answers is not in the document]
nabokihms: Thought about something similar. Unfortunately, this is not possible with the current design. We have to do cross imports, which is not allowed. I will remove the `FromConfig` suffix because it's not true.

server/handlers.go
nabokihms: Every request must be responded with the same refresh token and id token. Updating claims here will change that logic. We also should avoid calling the refresh method of connectors.

Need a second opinion here.

server/rotation.go
nabokihms: Yeah, fixed.

server/server_test.go
nabokihms: Fixed.

storage/sql/migrate.go
nabokihms: Oh, yeah. Migrations... Fixed.

server/handlers.go
nabokihms: Added your case, thanks!

server/handlers.go
tkleczek: I don't quite follow. refreshConn.Refresh() is called earlier in this method regardless of the update here. And what do you mean by:
> Every request must be responded with the same refresh token and id token

server/rotation_test.go
nabokihms: Yeah, it was here by accident. Fixed.

server/handlers.go
nabokihms: The question is that maybe we don't need to call connector refresh for requests with an obsolete token? This is an open question to discuss.

server/handlers.go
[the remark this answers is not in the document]
nabokihms: Yeah. The method looks a little bit fatty :) However, in my opinion, it would be better to approve the concept first, then polish this piece of code, WDYT?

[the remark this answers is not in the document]
tkleczek: Makes perfect sense :)

nabokihms:
@tkleczek thanks a lot! Your review was more than helpful 👍

tkleczek:
We might also want to add garbage-collection of the expired tokens

server/handlers.go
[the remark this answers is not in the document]
nabokihms: I made a little refactoring (actually moved refresh handler to another file and split it into several functions).

server/handlers.go
nabokihms: Added the clause to make dex skip refreshing claims using a connector on requests with an obsolete token.

sagikazarmark:
@nabokihms I know you have been working on this for a long time, thanks for that. Let me know if you think this is ready for a review.

nabokihms:
@sagikazarmark, I think that I've finished polishing this. Please, look through the PR when you have time.

JoelSpeed (commented):
Looks great! Left a few comments and questions throughout, but no major concerns from me

cmd/dex/config.go
```diff
@@ -314,3 +317,10 @@ type Logger struct {
 	// Format specifies the format to be used for logging.
 	Format string `json:"format"`
 }
+
+type RefreshTokenExpiry struct {
```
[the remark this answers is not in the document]
nabokihms: Right now, there are only expiration settings, but I got your point. Maybe we don't want to have any unnecessary limitations here. I will rename this structure.

server/refreshhandlers.go
```diff
@@ -0,0 +1,339 @@
...
+func (s *Server) refreshWithConnector(ctx context.Context, token *internal.RefreshToken, refresh *storage.RefreshToken, scopes []string) (connector.Identity, *refreshError) {
+	var connectorData []byte
+
+	session, err := s.storage.GetOfflineSessions(refresh.Claims.UserID, refresh.ConnectorID)
+	switch {
+	case err != nil:
+		if err != storage.ErrNotFound {
+			s.logger.Errorf("failed to get offline session: %v", err)
+			// TODO: previously there was a naked return without writing anything in response
+			//   Need to ensure that everything works as expected.
```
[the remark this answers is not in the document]
nabokihms: Oh, yes. I checked this out, and it seems we need to throw an error here. This situation is not possible if dex works normally and there are no problems with storage.

examples/config-dev.yaml
```diff
@@ -77,6 +77,10 @@ telemetry:
 #   deviceRequests: "5m"
 #   signingKeys: "6h"
 #   idTokens: "24h"
+#   refreshTokens:
+#     reuseInterval: "3s"
+#     validIfNotUsedFor: "2190h"
+#     absoluteLifetime: "5000h"
```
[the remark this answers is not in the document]
nabokihms: It is related to golang and how it parses duration. The corresponding issue is here https://github.com/golang/go/issues/17767
In dex, [time.ParseDuration](https://github.com/golang/go/blob/0e85fd7561de869add933801c531bf25dee9561c/src/time/format.go#L1374)  is used to parse other time intervals, so I decided to stick with it.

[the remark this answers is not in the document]
JoelSpeed: Ok, that makes sense, can we add a comment that explains how many days they are afterwards so users don't have to do that maths themself?

JoelSpeed:
No further comments from me, other than potentially adding a comment to the dev examples that shows how many days the durations are so users don't have to work it out by themself

examples/config-dev.yaml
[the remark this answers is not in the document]
nabokihms: Fixed

nabokihms:
@JoelSpeed, thanks for the review!
I didn't find documentation about refresh tokens and tokens expiration on the dex website. Is it worth adding more information about how the refresh process works?

/cc @sagikazarmark

sagikazarmark (approved):
Thanks @nabokihms ! This is great!

justinas-b:
Hey @sagikazarmark , any ideas when we can expect to see new `dex` release including these changes?

sagikazarmark:
@justinas-b We are targeting mid-May (it depends on the etcd feature freeze).

PR #1847, opened 2020-10-28
feat: Retry Kubernetes update requests

nabokihms:
Kubernetes provides API consistency through `resourceVersion` field. Updating resource with a higher resource version than contained in storage cause conflict errors. Users may encounter such errors, which is unwanted behavior. In my opinion, it's good just to retry a failed request in case of conflict.


sagikazarmark (approved):
LGTM, thanks!

PR #1859, opened 2020-11-09
Pin build stage platform to host

sagikazarmark:


sagikazarmark:
Doesn't seem to work :\

PR #1861, opened 2020-11-11
Use constant time comparison for client secret verification

xtremerui:
**Overview**:
~~This PR adds an option to Dex config to enable enforement of client secret encryption.~~
Use subtle.ConstantTImeCompare for client secret verification to prevent a potential timing acttack

The current implementation for checking client secret is by plain text comparison, which is vulnerable to [timing attack](https://en.wikipedia.org/wiki/Timing_attack). Using constant time comparison will mitigate the threat.

Thx!


tkleczek:
I agree that using `subtle.ConstantTimeCompare ` protects against timing attack. But could you explain a little bit more how encrypting client secret gives additional safety?
Also, your PR seems not to take into consideration that public clients have no secret.

xtremerui:
@tkleczek FYI I just pushed a commit to fix a bug in the comparison that client secret in request should be hashed and client secret in dex storage should be plain text. Though I don't think your question is specific on the implementation.

I agree, not liking a password that could be used widely, a client secret is mostly specific to the client. Though I think it doesn't hurt to provide a way to hide client secret in request in case some one doesn't want client secret to be exposed anyway.

For public client that has no secret, the original logic still applies. `subtle.ConstantTimeCompare` takes empty value so it allows client secret to be empty. Not sure about what you mean here. Could you clarify please?

Thx for commenting!

tkleczek:
@xtremerui
> For public client that has no secret, the original logic still applies. subtle.ConstantTimeCompare takes empty value so it allows client secret to be empty. Not sure about what you mean here. Could you clarify please?

Not sure now, maybe I was mistaken.

> Though I think it doesn't hurt to provide a way to hide client secret in request in case some one doesn't want client secret to be exposed anyway.

I think it hurts mostly in the additional code that has to be maintained and excessive options in configuration. I don't think you have to protect client secret in transit more than standard TLS does. OIDC standard also does not mention this "extension".

Btw, you might want to split this PR in two separate ones. I think using subtle.ConstantTimeCompare is has obvious advantages so it could hopefully go in quicker

xtremerui:
> Btw, you might want to split this PR in two separate ones. I think using subtle.ConstantTimeCompare is has obvious advantages so it could hopefully go in quicker
yes I think that makes sense.

In our own use case, concourse CLI is actually a public client and we have the client secret in OSS github code. So technically this PR is also an option for us but not a must. 

I will update this PR to include only subtle.ConstanTimeCompare. Thx!

sagikazarmark (approved):
Thanks!

PR #1862, opened 2020-11-11
Improve auth flow error handling

### LINKED ISSUE #1796: Dex error-handling is not RFC-compliant
According to RFC (https://tools.ietf.org/html/rfc6749#section-4.1.2.1) most errors from Authorization Code Grant part of the authentication flow should be reported back to the client via HTTP redirect with additional query parameters set, e.g:

```
HTTP/1.1 302 Found
   Location: https://client.example.com/cb?error=access_denied&state=xyz
```

Instead, most error paths in Server's methods handleAuthorization, handleConnectorLogin, handleConnectorCallback and handleApproval (which all handle code grant) serve dex error page template in case of errors.
Moreover, in at least once instance a bare json is returned via server.tokenErrHelper method.

Also: when handling /token endpoint RFC says that "invalid_grant" error should be returned if code parameter is invalid/expired, (https://tools.ietf.org/html/rfc6749#section-5.2) whereas Dex returns "invalid_request in this case.

tkleczek:
**Overview**:
Fixes several unrelated problems related to auth flow error handling

1) Fixes a case of wrong error returned in case of unsupported grant type (according to RFC should be "unsupported_grant_type", not "invalid_grant").
2) Fixes "connector not found" error displayed in an inconsistent not-end-user-friendly way (errTokenHelper used instead of renderError)  
3) Makes errors related to invalid connector in parseAuthorizationRequest redirect back to the client (as RFC states that after client_id/redirect uri is validated, all further errors should be redirected to client)
4) Makes two type of errors returned from parseAuthorizationRequest explicit (either redirected error or displayed error)

This PR partially fixes #1796 

**Special notes for a reviewer**:

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -158,7 +158,7 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 		}
-		s.tokenErrHelper(w, errInvalidConnectorID, "Connector ID does not match a valid Connector", http.StatusNotFound)
+		s.renderError(r, w, http.StatusBadRequest, "Connector ID does not match a valid Connector")
 		return
 	}
 
@@ -187,21 +187,16 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
 	authReq, err := s.parseAuthorizationRequest(r)
 	if err != nil {
 		s.logger.Errorf("Failed to parse authorization request: %v", err)
-		status := http.StatusInternalServerError
 
-		// If this is an authErr, let's let it handle the error, or update the HTTP
-		// status code
-		if err, ok := err.(*authErr); ok {
-			if handler, ok := err.Handle(); ok {
-				// client_id and redirect_uri checked out and we can redirect back to
-				// the client with the error.
-				handler.ServeHTTP(w, r)
-				return
-			}
-			status = err.Status()
+		switch authErr := err.(type) {
+		case *redirectedAuthErr:
+			authErr.Handler().ServeHTTP(w, r)
+		case *displayedAuthErr:
+			s.renderError(r, w, authErr.Status, err.Error())
+		default:
+			panic("unsupported error type")
 		}
 
-		s.renderError(r, w, status, err.Error())
 		return
 	}
 
@@ -770,7 +765,7 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
 	case grantTypePassword:
 		s.withClientFromStorage(w, r, s.handlePasswordGrant)
 	default:
-		s.tokenErrHelper(w, errInvalidGrant, "", http.StatusBadRequest)
+		s.tokenErrHelper(w, errUnsupportedGrantType, "", http.StatusBadRequest)
 	}
 }
 
--- a/server/oauth2.go
+++ b/server/oauth2.go
@@ -29,32 +29,35 @@ import (
 
 // TODO(ericchiang): clean this file up and figure out more idiomatic error handling.
 
-// authErr is an error response to an authorization request.
 // See: https://tools.ietf.org/html/rfc6749#section-4.1.2.1
-type authErr struct {
+
+// displayedAuthErr is an error that should be displayed to the user as a web page
+type displayedAuthErr struct {
+	Status      int
+	Description string
+}
+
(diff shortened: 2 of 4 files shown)


sagikazarmark:
@tkleczek can you please rebase your PR?

tkleczek:
@sagikazarmark done, sorry for the delay

server/oauth2_test.go
```diff
@@ -27,8 +27,13 @@ func TestParseAuthorizationRequest(t *testing.T) {
 
 		queryParams map[string]string
 
-		wantErr    bool
-		exactError *authErr
```
nabokihms, line 31: With fields for different error types test looks a little bit complicated now. You need to know a lot about auth error structure to add new test cases. Can we leave the `exactError` here?

server/oauth2.go
```diff
@@ -31,26 +31,31 @@ import (
...
-
-func (err *authErr) Error() string {
+func (err *redirectedAuthErr) Error() string {
 	return err.Description
 }
 
-func (err *authErr) Handle() (http.Handler, bool) {
-	// Didn't get a valid redirect URI.
+func (err *redirectedAuthErr) Handle() (http.Handler, bool) {
+	// // Didn't get a valid redirect URI.
```
nabokihms, line 58: Double comment! :)

server/handlers.go
```diff
@@ -697,7 +695,7 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
 	case grantTypePassword:
 		s.handlePasswordGrant(w, r, client)
 	default:
-		s.tokenErrHelper(w, errInvalidGrant, "", http.StatusBadRequest)
+		s.tokenErrHelper(w, errUnsupportedGrantType, "", http.StatusBadRequest)
```
nabokihms, line 698: Some providers use the unsupported grant type error to indicate that this is a valid grant, but they do not allow the user to authorize with this grant.  However, I think for us, it is better to return "unsuppoerted_grant_type" here. Thanks!

nabokihms (commented):
I left some comments, but overall it looks good! Thanks!

sagikazarmark:
@tkleczek I know this has been in the queue for a long time, but I'd like to release it. Do you think you can take a look at @nabokihms comments?

sagikazarmark:
Moving to the next milestone

tkleczek:
@sagikazarmark  @nabokihms I rebased the PR and fixed code review issues.

server/oauth2_test.go
```diff
@@ -321,13 +315,32 @@ func TestParseAuthorizationRequest(t *testing.T) {
...
 
 			_, err := server.parseAuthorizationRequest(req)
-			if tc.wantErr {
-				require.Error(t, err)
-				if tc.exactError != nil {
-					require.Equal(t, tc.exactError, err)
-				}
-			} else {
+			if tc.expectedError == nil {
 				require.NoError(t, err)
```
nabokihms, line 319: Could you rewrite this block using the same library for assertions? It is not so important, testify of testing, but sticking to one library for the test is important.

server/oauth2_test.go
```diff
@@ -321,13 +315,32 @@ func TestParseAuthorizationRequest(t *testing.T) {
...
+				case *displayedAuthErr:
+					e, ok := err.(*displayedAuthErr)
+					if !ok {
+						t.Fatalf("%s: expected displayedAuthErr error", tc.name)
+					}
+					if e.Status != expectedErr.Status {
+						t.Errorf("%s: expected http status %v, got %v", tc.name, expectedErr.Status, e.Status)
+					}
+				default:
+					panic("unsupported error type")
```
nabokihms, line 342: Let's use `t.Fatal` here.

nabokihms (commented):
Everything looks good, thanks. I added some NIT comments.

I am going to test this PR. If everything is ok, I will mark it as approved.

nabokihms (approved):
The only problem I found is that we still see a displayed error on accessing the `/auth` endpoint with the wrong connector id. This happens because now we are parsing an auth request only after redirecting to the connector login page. 

I don't think this blocks us from accepting this PR as is. Let's merge it. We can deal with the rest later.

nabokihms:
@tkleczek thank you for the contribution!

PR #1863, opened 2020-11-12
Copy module dependencies to Docker image for CVE scanning / dependency analysis

heidemn-faro:
Signed-off-by: Martin Heide <martin.heide@faro.com>

**Overview**:
- Add go.mod + go.sum to Dex docker image.

- Inventory of Go modules for analysis of CVEs and open source licenses.
- Important for commercial projects.

**Special notes for a reviewer**:

Example inventory result from WhiteSource:
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,10 @@ USER 1001:1001
 
 COPY --from=0 /go/bin/dex /usr/local/bin/dex
 
+# Copy module dependencies for CVE scanning / dependency analysis.
+COPY go.mod go.sum                 /opt/dex/dependencies/
+COPY api/v2/go.mod api/v2/go.sum   /opt/dex/dependencies/api/v2/
+
 # Import frontend assets and set the correct CWD directory so the assets
 # are in the default path.
 COPY web web


Dockerfile
```diff
@@ -37,6 +37,9 @@ USER 1001:1001
 
 COPY --from=0 /go/bin/dex /usr/local/bin/dex
 
+# Copy module dependencies for CVE scanning / dependency analysis.
+COPY go.mod go.sum /dex-dependencies/
```
sagikazarmark, line 41: I'd prefer `/opt/dex` or something similar. Also, you'll probably need the module files for the api module.

sagikazarmark (commented):
Interesting. I'm not necessarily sure it's a good idea that analysis tools rely on source files, but it's trivial to add.

However, I think it would be better and more correct to rely on the module information compiled into the binary itself:

https://github.com/mitchellh/golicense/blob/master/module/module.go

So I'd urge the developers of those tools to rely on that instead.

sagikazarmark (changes requested):
Please see the comments above

Dockerfile
heidemn-faro: @sagikazarmark okay done:

```
$ docker run --entrypoint sh quay.io/dexidp/dex:latest -c 'find /opt/'
/opt/
/opt/dex
/opt/dex/dependencies
/opt/dex/dependencies/go.sum
/opt/dex/dependencies/go.mod
/opt/dex/dependencies/api
/opt/dex/dependencies/api/v2
/opt/dex/dependencies/api/v2/go.sum
/opt/dex/dependencies/api/v2/go.mod
```

Dockerfile
```diff
@@ -37,6 +37,10 @@ USER 1001:1001
 
 COPY --from=0 /go/bin/dex /usr/local/bin/dex
 
+# Copy module dependencies for CVE scanning / dependency analysis.
```
heidemn-faro, line 40: If preferred, this could be reduced to a single COPY, by preparing the folder structure in the 1st build stage.
However this would make the Dockerfile more complicated.

Let me know if I should change it.

sagikazarmark (approved):

Issue #1851, opened 2020-11-01
"select another login method" doesn't work after a failed login


tomqwpl:
The "select another login method" link is simply implemented using "back". Hence, if you had a failed login, "back" doesn't take you back to the connection selection page, it takes you to the password login page again. The "select another login method" link needs to do an explicit navigate to a known location, or else it needs to be disabled after a failed login.

Issue #1849, opened 2020-10-30
"select another login method" link causes database error


tomqwpl:
If you select a login method that requires the password entry dialog to be displayed, there's a link at the bottom "select another login method". If you click on this, then select a different login method you get an error screen with "database error".
In the Dex log there is "connector is already set for this auth request".
So it looks like Dex doesn't actually support changing to a different login method.

al45tair:
This is a symptom of the problem mentioned in #646, namely that the authorisation request is persisted too early.

PR #1865, opened 2020-11-16
fix: defer creation of auth request.

al45tair:
**Overview**:
Defer creation of the auth request in Dex's back-end storage until the user chooses an authentication method.

Right now, the option to change your mind and use a different authentication method won't work at all, because it finds the existing auth request and the connector ID is already set for that request (#1849).

There are also some other problems, like the auth request expiring while the user is looking at the log-in page (#646).

Additionally, deferring creation of the auth request allows us to fix the back link on the password page so it works properly if the user fails to authenticate first time (#1851).


sagikazarmark:
@al45tair can you please rebase your PR?

al45tair:
I'll try to get to this soon. Probably not going to be this week though (unless I do it at the weekend; we'll see).

server/handlers.go
```diff
@@ -178,22 +161,20 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 
 	if len(connectors) == 1 && !s.alwaysShowLogin {
 		for _, c := range connectors {
```
nabokihms, line 163: We already checked that connectors length equals 1. We can use `connectors[0]` instead of iterating.
Do you mind if I ask you to fix it? :)

server/handlers.go
```diff
@@ -211,37 +213,35 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
...
-	authReq, err := s.storage.GetAuthRequest(authReqID)
-	if err != nil {
-		s.logger.Errorf("Failed to get auth request: %v", err)
-		if err == storage.ErrNotFound {
-			s.renderError(r, w, http.StatusBadRequest, "Login session expired.")
-		} else {
-			s.renderError(r, w, http.StatusInternalServerError, "Database error.")
-		}
+	// Set the connector being used for the login.
+	if authReq.ConnectorID != "" && authReq.ConnectorID != connID {
```
nabokihms, line 217: This check is here only because we pass connector id twice? Do we still need the ConnectorID field in the AuthRequest?

server/handlers.go
```diff
@@ -164,11 +141,17 @@ func (s *Server) handleAuthorization(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Construct a URL with all of the arguments in its query
+	connURL := url.URL{
+		RawQuery: r.Form.Encode(),
```
nabokihms, line 146: I think redirecting a user to `/auth/my_connector?connector_id=my_connector` may be puzzling a bit.
Can we use only mux variables or query parameters?

server/handlers.go
```diff
@@ -211,37 +213,35 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
...
+	if err := s.storage.CreateAuthRequest(*authReq); err != nil {
+		s.logger.Errorf("Failed to create authorization request: %v", err)
+		s.renderError(r, w, http.StatusInternalServerError, "Failed to connect to the database.")
+		return
 	}
 
 	scopes := parseScopes(authReq.Scopes)
-	showBacklink := len(s.connectors) > 1
+
+	// Work out where the "Select another login method" link should go.
```
nabokihms, line 236: Does it mean that clicking the `back` button and choosing a new connector each time will create an auth request?
It leads us to a problem that by pressing F5 in a browser on a password connector logic screen, a user can create a lot of auth requests in storage, isn't it?

server/handlers.go
```diff
@@ -211,37 +213,35 @@ func (s *Server) handleConnectorLogin(w http.ResponseWriter, r *http.Request) {
...
+	backLink := ""
+	if len(s.connectors) > 1 {
+		backLinkURL := url.URL{
+			Path:     s.absPath("/auth"),
+			RawQuery: r.Form.Encode(),
```
nabokihms, line 241: Is this raw query includes connector_id?

server/handlers.go
```diff
@@ -124,38 +124,15 @@ func (s *Server) discoveryHandler() (http.HandlerFunc, error) {
...
-				return
-			}
-			status = err.Status()
-		}
+	// Extract the arguments
+	if err := r.ParseForm(); err != nil {
+		s.logger.Errorf("Failed to parse arguments: %v", err)
 
-		s.renderError(r, w, status, err.Error())
+		s.renderError(r, w, http.StatusBadRequest, "Bad query/form arguments")
```
nabokihms, line 131: The original message was `"Failed to parse request body."`. I think it is better to leave it as it was.

nabokihms (commented):
Hello, @al45tair. Thank you for sending this PR. I have some suggestions/questions regarding the code. Hope that you will find it helpful. 

Since I am a new maintainer, it will be impossible to merge this only with my approval. We will have to wait for a review from another maintainer.

server/handlers_test.go
```diff
@@ -131,182 +126,3 @@ func TestHandleInvalidSAMLCallbacks(t *testing.T) {
...
-
-	recorder2 := httptest.NewRecorder()
-	s.ServeHTTP(recorder2, createConnectorLoginRequest(connector2.ID))
-	if recorder2.Code != 500 {
-		t.Error("attempt to overwrite connector on auth request should fail")
-	}
-}
-
-// TestHandleAuthCode checks that it is forbidden to use same code twice
-func TestHandleAuthCode(t *testing.T) {
```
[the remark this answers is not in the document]
al45tair: Ah. I hadn't appreciated that.

server/handlers.go
al45tair: No problem.

server/handlers.go
al45tair: Hmmm. I think it was here before because the auth request was already created, so we needed to check that it matched what we'd already agreed on. You're right that now it really only exists because we pass the connector ID twice to the login method (once as "connector_id" and once as "connector"), though one of those is specified by the client application, while the other comes from the choose-a-login-method page. I don't think it hurts to check, though I don't see how the two could be inconsistent at this point.

I think it's still useful to store the connector ID in the AuthRequest, because we also want to check it in e.g. handleConnectorCallback()

server/handlers.go
al45tair: Removing connector_id from r.Form would seem to make sense here.

server/handlers.go
al45tair: You mean, if they go to the password page and keep hitting F5? Yes, they will create more AuthRequests in that case. I suppose we could create another endpoint for that and rather than rendering the template directly from handleConnectorLogin, do a redirect there and render it from the redirected endpoint. There's also an issue right now in that I think the password connector will end up making two AuthRequests.

FWIW, I think the F5 issue was already a problem with the previous handleAuthorization() behaviour, because that directly renders the login method selection page, having created an AuthRequest. So you could have it F5 repeatedly there too.

server/handlers.go
al45tair: Only if the user originally specified connector_id in the auth request, in which case there's only one connector here anyway. (Otherwise the back button wouldn't work. It definitely does.)

server/handlers.go
al45tair: I'll revert that part.

server/handlers_test.go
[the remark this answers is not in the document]
al45tair: I've reinstated the test.

server/handlers.go
al45tair: Fixed.

server/handlers.go
al45tair: I've added an extra endpoint to mitigate this problem.

nabokihms:
Looks good, thanks! I am going to test it and will come back to you with more feedback.

al45tair:
I should probably say: I've changed jobs. This PR was done as part of my previous employment, and because of my new employer's policies surrounding Open Source contributions, I won't be able to work on it — or any other Dex related things —any further. If further changes are needed, @tomqwpl is probably the person to talk to.

nabokihms:
@al45tair, thanks for letting us know. I can only wish you a stroke of good luck in a new place then! We appreciate the contributions you have made so far to Dex.

nabokihms (approved):
So, I tested this PR, and overall it looks good. There are still things that we can improve, but this is not an obstacle to merge.

sagikazarmark:
Thanks for your contributions @al45tair !

Good luck out there!

sagikazarmark (approved):
LGTM

srenatus:
Great to see this happen 👏 🎉

Issue #1870, opened 2020-11-23
Adding public clients without secret through API not possible


spohner:
**Expected behavior (what you expected to happen)**:
When adding a client through the gRPC API without a secret and public set to true I expected the PKCE flow to be available to my public webapp.

**Actual behavior (what actually happened)**:
Adding a client with the gRPC API creates a new secret if no secret is provided. Dex then expects a client_secret to be sent and complains about missing client_secret.  

**Steps to reproduce**:
1. Create a public client through the API with no secret.
2. Observe in the storage that a secret is created.
3. See missing client_secret in the logs after trying to authenticate since no secret is provided from the public client. 

**Environment**:
- Dex version: v2.26.0
- Storage type: SQL, Postgres

**Anything else we should know?**:
I guess the code line in question is https://github.com/dexidp/dex/blob/33e13c2aad9bb8a91abea6a2870dc178e3bd00de/server/api.go#L50 . Setting the secret to "" in the database manually after client creation allows for the correct behavior. 

**Additional information for debugging (if necessary)**:

<details><summary>Logs</summary><br><pre>
time="2020-11-23T12:15:10Z" level=info msg="missing client_secret on token request for client: xxx"
</pre></details>

PR #1871, opened 2020-11-24
Allow public clients created with API to have no client_secret

spohner:
**Overview**:
Allows public clients created with the API to have no client_secret to support the PKCE flow.

Closes: #1870. Now all clients with no client_secret created by the API gets a random secret. This adds a check if the client is public and allows for empty client_secret if that is the case.


tkleczek (approved):

cyrilzh (commented):
The code requires public client has no "secret" value (req.Client.Secret == "" && !req.Client.Public). This is inconsistent with the document in which the public client has secret value.
https://dexidp.io/docs/custom-scopes-claims-clients/#public-clients

bonifaido (approved):
LGTM

PR #1886, opened 2020-12-15
Add Flant to ADOPTERS.md

nabokihms:
#### Overview
Add the entry about Flant into ADOPTERS.md

We are ready to state, that Flant has the experience of using and adopting Dex to various Cloud and Baremetal environments. Customers of our Managed Kubernetes service authenticate through almost all connectors from the list. The topmost used are Github, Gitlab, BitbucketCloud, Atlassian Crowd, OIDC connectors.

We manage Dex configuration files through our own CRDs with an additional controller, which is based on [shell-operator](https://github.com/flant/shell-operator).

In addition, we believe in a bright future for Dex project and hope, that it will continue growing up and moving towards CNCF graduation.


sagikazarmark (approved):
Thanks! 👍 

Great to have you here @nabokihms

PR #1888, opened 2020-12-18
Added the possibility to activate lowercase for UPN-Strings

VF-mbrauer:
#### Overview
There was a need to introduce a Boolean-Flag which enables to possibility to transform the UPN (E-Mail) of the user to lowercase.
As we never now how those string are stored in Active Directory or Azure Active Directory. But they are important for mapping.

In some Active Directories we have the UPNs in various formats stored. It can be lowercase, uppercase and also it
can be a mix of them. When now sending those to DEX and then when it will be used by the RBAC of the Custer
it needs to match 1:1. But we never now how the name will be transported and how it has been originally stored.
So overcome that issue, we need a small feature which does the trick an put all character to a format where we know
how to handle. As when they are all lowercases and we now that, we just put them the way on the Rolebindings on the cluster.

Nothing special to highlight, expect that nothing will be put to LowerCase when Flag has not been set.
Otherwise the parameter keeps false and noting will change at all.

Depends, if the user want to use this feature, he needs to activate it. Once done the UPNs (E-Mail) will be send all the time
in lowercase letters. So if the user expect then to get the 1:1 from AD or AAD he will be surprised when the feature is on.

**_Configuration-Details:_**
If you put the `UpnToLowercase: true` into your yaml config and if you make it true to enable it will start conversion to LowerCase.

```release-note

connectors:
  - type: microsoft
    # Required field for connector id.
    id: microsoft
    # Required field for connector name.
    name: Microsoft
    config:
      # Credentials can be string literals or pulled from the environment.
      clientID: $MICROSOFT_APPLICATION_ID
      clientSecret: $MICROSOFT_CLIENT_SECRET
      redirectURI: http://127.0.0.1:5556/dex/callback
      UpnToLowercase: true
```


go.mod
```diff
@@ -46,6 +47,7 @@ require (
 	gopkg.in/asn1-ber.v1 v1.0.0-20181015200546-f715ec2f112d // indirect
 	gopkg.in/ldap.v2 v2.5.1
 	gopkg.in/square/go-jose.v2 v2.4.1
+	honnef.co/go/tools v0.0.0-20190523083050-ea95bdfd59fc // indirect
```
sagikazarmark, line 50: Why is this added as a dependency?

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/lib/pq v1.3.0
 	github.com/mattermost/xml-roundtrip-validator v0.0.0-20201204154048-1a8688af4cf1
 	github.com/mattn/go-sqlite3 v1.11.0
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
```
sagikazarmark, line 28: Why is this added as a dependency?

connector/microsoft/microsoft.go
```diff
@@ -51,6 +52,7 @@ type Config struct {
 	Groups               []string        `json:"groups"`
 	GroupNameFormat      GroupNameFormat `json:"groupNameFormat"`
 	UseGroupsAsWhitelist bool            `json:"useGroupsAsWhitelist"`
+	UpnToLowercase       bool            `json:"upnToLowercase"`
```
sagikazarmark, line 55: I'm not intimately familiar with microsoft/AD, but I think it would make sense to use email as a phrase here. I guess the UPN is always an email address.

sagikazarmark (changes requested):

sagikazarmark (approved):
Thanks!

PR #1893, opened 2020-12-21
feat: Add gomplate to the docker image

### LINKED ISSUE #1838: Add dockerize to the docker image
Adding dockerize will allow users to use environment variables in config files.

nabokihms:
#### Overview

* Add [gomplate](https://github.com/hairyhenderson/gomplate) to the Docker image
* Add `entrypoint.sh` script that renders the config file as a template

Closes #1838

This PR solves a long time issue in Dex, name injecting secrets into the config file (especially on Kubernetes) without actually saving those secrets in less/non-secure storage layers. By using templating (and environment variables) we can create a config file runtime (when the container starts, in the entrypoint).

Obviously, the rendered config file is still saved to ephemeral storage, so it's not all the way secure, but anyone with access to the storage layer would have other means to get the secrets anyway.

A better long-term solution for secrets would be using environment variables directly and/or secret storage solutions (like Vault) or even KMS integration.

```release-note
Add [gomplate](https://github.com/hairyhenderson/gomplate) to the Dex docker image to provide config file templating features
```


sagikazarmark (commented):
I feel like there is more we can do here:

- I'd get rid of the entrypoint entirely
- `serve` should be the default command IMHO
- `serve` requires a config, so we need some sort of default

I'm not sure I'm a fan of reading config from stdin. I guess it's there so we don't write the resulting config to ephemeral storage, but it feels like a complication without benefits: anyone with access to the storage has already access to the container itself.

At least for now, I'd just write the configuration to `/etc/dex/config.yaml`.

WDYT?

nabokihms:
Everything sounds reasonable to me. The only thing that remains is implementation. I will work on it.

Dockerfile
```diff
@@ -44,7 +48,10 @@ COPY api/v2/go.mod api/v2/go.sum   /opt/dex/dependencies/api/v2/
...
 COPY web web
+COPY examples/config.tmpl /etc/dex/config.tmpl
 
-ENTRYPOINT ["dex"]
+RUN chown -R 1001:1001 /etc/dex/
+USER 1001:1001
 
-CMD ["version"]
+ENTRYPOINT ["dockerize", "-no-overwrite", "-template", "/etc/dex/config.tmpl:/etc/dex/config.yaml"]
+CMD ["dex", "serve", "/etc/dex/config.yaml"]
```
heidemn, line 57: What about moving `"dex"` to the end of ENTRYPOINT, to avoid a breaking change for current users of the Docker image?

```
ENTRYPOINT ["dockerize", "-no-overwrite", "-template", "/etc/dex/config.tmpl:/etc/dex/config.yaml", "dex"]
CMD ["serve", "/etc/dex/config.yaml"]
```

heidemn:
Once my PR https://github.com/dexidp/dex/pull/1902 is merged, most Dockerize users will probably want to set `DEX_EXPAND_ENV = false`, to avoid another unexpected level of env expansion.

Dockerfile
nabokihms: Seems reasonable. Thanks!

sagikazarmark: Personally, I never like when an application is the entrypoint. It just ties the user's hands. Although it is a breaking change, but I don't think it'll affect most users. We could think of some sort of deprecation for this change to warn users in advance, but I'm not sure how effective it would be.

examples/config.tmpl
```diff
@@ -0,0 +1,44 @@
...
+
+oauth2:
+  responseTypes: {{ default .Env.DEX_OAUTH2_RESPONSE_TYPES "[code]" }}
+  skipApprovalScreen: {{ default .Env.DEX_OAUTH2_SKIP_APPROVAL_SCREEN "false" }}
+  alwaysShowLoginScreen: {{ default .Env.DEX_OAUTH2_ALWAYS_SHOW_LOGIN_SCREEN "false" }}
+{{- if .Env.DEX_OAUTH2_PASSWORD_CONNECTOR }}
+  passwordConnector: {{ .Env.DEX_OAUTH2_PASSWORD_CONNECTOR }}
+{{- end }}
+
+enablePasswordDB: {{ default .Env.DEX_ENABLE_PASSWORD_DB "true" }}
```
heidemn, line 44: Question: What is the best way to pass an env variable that may contain YAML special characters like `'"`?
Is there a better solution than manually escaping single quotes?
`bindPW: '{{ replace .Env.DEX_LDAP_BINDPW "'" "''" }}'`

Since Dockerize uses `text/template`, I guess it's not aware of YAML syntax.

sagikazarmark: Good question. I think it should be raised in the dockerize project. Helm has a `quote` filter that does exactly that.

Also, I see open issues in the project asking to add the sprig template library.

sagikazarmark (commented):
I've looked into dockerize a bit more and I'm not sure it's gonna be good for us. The author doesn't answer to issues/PRs. I pinged him, let's if we get an answer.

Dockerfile
```diff
@@ -33,7 +33,11 @@ WORKDIR /
 # OpenSSL is required so wget can query HTTPS endpoints for health checking.
 RUN apk add --no-cache --update ca-certificates openssl
 
-USER 1001:1001
+# Install dockerize to provide the way of config parametrization
+ENV DOCKERIZE_VERSION v0.6.1
+RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
+    && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
```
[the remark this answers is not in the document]
nabokihms: The kind of a problem we can't solve on our own. Looks like we need to search for a better solution.

Dockerfile
nabokihms: I hardly believe that adding a deprecation warning will help. Mentioning such a change in the RM should be enough. However, for me, it looks better to leave `dex` at the entry point just because changing behavior is not required by this PR.

Dockerfile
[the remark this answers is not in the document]
heidemn: The [templating code](https://github.com/jwilder/dockerize/blob/master/template.go) of Dockerize is only 200 lines, since it's basically Go's `text/template` with some helper functions added.
On first sight, only other feature from Dockerize that could be helpful for Dex is waiting for other dependencies (`-wait` , `-timeout` & co). And that can usually be solved with a crash loop.

Would it be an option to glue together ourselves Go templating with helpers from a well-maintained library like https://github.com/Masterminds/sprig in Dex? 
Could work like this for the user:
`$ dex serve config.yaml`
`$ dex serve --template config.tmpl`

Sprig already includes `env` and `default` [functions](http://masterminds.github.io/sprig/), similar to Dockerize.

nabokihms:
@sagikazarmark, I replaced dockerize with gomplate in this PR. Some concerns, problems, tasks to do:
* entrypoint is back because gomplate can't work as an entrypoint
* TARGETVARIANT is empty for arm/v6 and arm/v7 during the build
* gomplate binary size is 36.7M 
* config template is located in /etc/dex, and directory is shared for nobody user
* need to change description an title of the PR

Please take a look if you have time 🙏

entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+
+### Usage: /entrypoint.sh <command> <args>
+set -e
+command=$1
+
+if [ "$command" == "serve" ]; then
+  file="$2"
+  gomplate -f "$file" -o "/etc/dex/config.yaml";
```
sagikazarmark, line 9: This is a bit problematic. Ideally, we want users to use this mount point going forward (eg. we could add some sort of config discovery, so manually choosing a config file becomes optional and `/etc/dex/config.yaml` is loaded automatically). This command will simply overwrite that file all the time.

In addition, they might also mount `/etc/dex` (eg. to mount certs), so it would actually become readonly, meaning we can't write the rendered template there.

`/tmp` sounds like a potential destination, `/var/run`/`/run` as well, but I'm not sure how we should tackle this problem yet.

sagikazarmark (changes requested):
Thanks @nabokihms !

This looks like a good approach.

I had a few comments, can you please take a look at them?

entrypoint.sh
nabokihms: /tmp sounds good for me

examples/config-example.tmpl
```diff
@@ -0,0 +1,54 @@
...
+  responseTypes: {{ getenv "DEX_OAUTH2_RESPONSE_TYPES" "[code]" }}
+  skipApprovalScreen: {{ getenv "DEX_OAUTH2_SKIP_APPROVAL_SCREEN" "false" }}
+  alwaysShowLoginScreen: {{ getenv "DEX_OAUTH2_ALWAYS_SHOW_LOGIN_SCREEN" "false" }}
+{{- if getenv "DEX_OAUTH2_PASSWORD_CONNECTOR" "" }}
+  passwordConnector: {{ .Env.DEX_OAUTH2_PASSWORD_CONNECTOR }}
+{{- end }}
+
+enablePasswordDB: {{ getenv "DEX_ENABLE_PASSWORD_DB" "true" }}
+
+connectors:
```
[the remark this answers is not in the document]
nabokihms: Yeah, firstly, I thought the same way. However, it works :)

examples/config-example.tmpl
```diff
@@ -0,0 +1,54 @@
...
+
+storage:
+  type: sqlite3
+  config:
+    file: {{ getenv "DEX_STORAGE_SQLITE3_CONFIG_FILE" "/etc/dex/dex.db" }}
+
+web:
+{{- if getenv "DEX_WEB_HTTPS" "" }}
+  https: {{ .Env.DEX_WEB_HTTPS }}
+  {{- if getenv "DEX_WEB_TLS_KEY" }}
```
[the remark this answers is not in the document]
nabokihms: Yes, this one is better. Fixed.

entrypoint.sh
```diff
@@ -0,0 +1,13 @@
...
+  exec dex serve "/etc/dex/config.yaml"
+else
```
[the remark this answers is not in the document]
nabokihms: I reworked the entrypoint script a little.

examples/config-example.tmpl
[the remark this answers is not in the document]
nabokihms: Maybe it will be better to add one more variable. Let's call it `DEX_CONNECTORS_ADDITIONAL` for now. 
Users will be able to specify custom connectors via it. Example:

Template:
```
connectors:
{{- if getenv "DEX_CONNECTORS_ENABLE_MOCK" }}
- type: mockCallback
  id: mock
  name: Example
{{- end }}
{{ getenv "DEX_CONNECTORS_ADDITIONAL" | jsonArray | toYAML }}
```
ENV:
```
DEX_CONNECTORS_ENABLE_MOCK=yes
DEX_CONNECTORS_ADDITIONAL='[{"type":"mockCallback","id":"secondMock","name":"Example2"}]'
```
Result:
```yaml
connectors:
- type: mockCallback
  id: mock
  name: Example
- id: secondMock
  name: Example2
  type: mockCallback
```

Dockerfile
```diff
@@ -29,9 +29,22 @@ FROM alpine:3.13.0
...
 
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETVARIANT
+ARG GOMPLATE_VERSION=v3.9.0
+
+RUN wget -O /usr/local/bin/gomplate \
+  "https://github.com/hairyhenderson/gomplate/releases/download/${GOMPLATE_VERSION}/gomplate_${TARGETOS:-linux}-${TARGETARCH:-amd64}${TARGETVARIANT}" \
+  && chmod +x /usr/local/bin/gomplate
+
```
sagikazarmark, line 40: I meant a completely separate build stage (before the final stage), so it can be cached separately from all the other stages (not that we utilize image cache today, but I definitely want to)

docker-entrypoint.sh
```diff
@@ -0,0 +1,27 @@
...
+### Usage: /docker-entrypoint.sh <command> <args>
+command=$1
+
+case "$command" in
+  serve)
+    for file_candidate in $@ ; do
+      if test -f "$file_candidate"; then
+        tmpfile=$(mktemp /tmp/dex.config.yaml-XXXXXX)
+        gomplate -f "$file_candidate" -o "$tmpfile"
+        echo "config rendered successfully into the tmp file ${tmpfile}"
```
sagikazarmark, line 12: Generally I don't think writing output from the entrypoint is not a good idea. If your logging system expects JSON logs this can mess with your log parser/analyzer and it doesn't really have any added value IMHO.

nabokihms: I was not 100% sure about adding this `echo`. The main reason for it is that in case of an error in config Dex returns an error about some `/tmp/dex-config.yaml-ABCDEFG` file. This can be confusing for users.

Ok, let's get rid of this `echo`.

docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh -e
+
+### Usage: /docker-entrypoint.sh <command> <args>
+command=$1
+
+case "$command" in
+  serve)
+    for file_candidate in $@ ; do
+      if test -f "$file_candidate"; then
```
heidemn, line 9: I wonder if it would be better to use e.g. `serve --template TEMPLATE ...`  instead of interpreting any existing file as a template.
This would make it more explicit, and could avoid potential esoteric bugs.

docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+        args="${args} ${file_candidate}"
+      fi
+    done
+    exec dex $args
+    ;;
+  --help|-h|version)
+    exec dex $@
+    ;;
+  *)
+    exec $@
```
heidemn, line 24: I like this option, allowing it bypass the templating by adding `dex` in front of the arguments. 👍 
Maybe mention it in the "Usage" comment, so it's more obvious.

docker-entrypoint.sh
nabokihms: The advantage of the implicit approach is avoiding unnecessary complications.

Thinking about this, adding new flags to the entrypoint has its pros and cons. Let's talk more about the disadvantages of using a flag:
* It's hard to document them.
* Users can't see them through help.
* Interference of the entrypoint is too big.
* The serve command has additional flags for addresses to listen on, which leads us to parse flags two times (with the entrypoint and with dex).

Maybe we can add a new *command* instead (something like "serve-with-template"), or use some env variable? The question is open.

heidemn: You're right with the disadvantages. 
Then I would suggest to either leave it as is, or to use `serve(-with|-from)-template`

config.docker.yaml
```diff
@@ -0,0 +1,46 @@
+issuer: {{ getenv "DEX_ISSUER" "http://127.0.0.1:5556/dex" }}
+
```
heidemn, line 2: Maybe add a comment that this config file is only an example / for development purposes?

And maybe mention in another comment that for escaping "unfriendly" input like passwords, this function could be used for escaping? https://docs.gomplate.ca/functions/strings/#strings-squote (Sorry haven't tested it yet, but I'm assuming that such escaping is not done automatically by Gomplate)

docker-entrypoint.sh
heidemn: You should also adapt the section "Does this PR introduce a user-facing change?":

* By default, Gomplate is now applied on the config file for command `serve ...`.
* To opt out, one can either adapt the command `serve ...` -> `dex serve ...`, or override the entrypoint.

nabokihms: Yes, you are right! Whole description has to be rewritten.

Dockerfile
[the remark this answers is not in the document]
mouchar: FYI, if you want to use dockerize, I recommend using its fork https://github.com/powerman/dockerize
This fork was created by an active contributor who was not satisfied with Jwilder's project management, communication, and so on.
Since then, a lot of improvements were done in Powerman's fork, incl. arm builds, for example.
The original dockerize seems to be an abandoned project.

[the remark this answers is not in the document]
sagikazarmark: I'm familiar with that fork, but I'm not sure I'm quite happy with the democratized maintenance model to implements:

> Everyone who has contributed to the project may become a collaborator - just ask for it in PR comments after your PR has being merged.

Also, we really don't need all that functionality, so I'm happy with a simple templating solution. My only concern with gotemplate is its size.

docker-entrypoint.sh
sagikazarmark: I'd really like to avoid adding templating to the Dex binary.

Dockerfile
```diff
@@ -32,11 +44,16 @@ RUN apk add --no-cache --update ca-certificates openssl
 RUN mkdir -p /var/dex
 RUN chown -R 1001:1001 /var/dex
 
+RUN mkdir -p /etc/dex
+COPY config.docker.yaml /etc/dex/config.docker.yaml
```
sagikazarmark, line 48:
```suggestion
COPY config.docker.yaml /etc/dex/
```

Dockerfile
```diff
@@ -20,6 +20,18 @@ COPY . .
 
 RUN make release-binary
 
+FROM alpine:3.13.1 AS gomplate
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETVARIANT
+ARG GOMPLATE_VERSION=v3.9.0
```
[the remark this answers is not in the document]
mouchar: IMHO `ARG` is correct in this case - you need this value at build time, not at run time.
there's no way how to parametrize `ENV` values during `docker build`.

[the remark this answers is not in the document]
sagikazarmark: I don't think we want to parametrize it. Do we?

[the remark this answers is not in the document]
mouchar: Not sure. Using `ENV` basically hard-codes its value. If it's not necessary to modify this value at build time, there's no difference between `ARG` and `ENV`.

On the other hand, `ARG` gives you the possibility to quickly test different gomplate version without modifying the Dockerfile: `docker build --build-arg  GOMPLATE_VERSION=v4.0.0 .`
If it is not required, I'm happy with `ENV` here :)

Dockerfile
```diff
@@ -20,6 +20,19 @@ COPY . .
...
 
+FROM alpine:3.13.1 AS gomplate
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETVARIANT
+
+ENV GOMPLATE_VERSION=v3.9.0
+
+RUN wget -O /usr/local/bin/gomplate \
```
sagikazarmark, line 31: Maybe use the slim version instead for a smaller binary: https://github.com/hairyhenderson/gomplate/releases

config.docker.yaml
sagikazarmark: I think that makes sense. Let's add a comment

sagikazarmark (changes requested):
I think this is good overall, thanks for working on this @nabokihms 

There are two comments left, can you please take a look at those? Thanks!

Dockerfile
sagikazarmark: Actually, does this work with every arm version? arm64, armv7? (The first one is GOARCH, the latter is variant AFAIK)

docker-entrypoint.sh
nabokihms: An example of the solution described by heidemn can be found here https://github.com/dexidp/dex/pull/1965. We closed that PR in favor of #1893.

sagikazarmark (approved):
Thanks! I think this is a great improvement!

nabokihms:
I have no time to change the description yet. Hope that I will be able to do it tonight.

Dockerfile
nabokihms: Yes, it does! I have tested with armv7 and arm64.

sagikazarmark: Awesome, thanks!

sagikazarmark:
No worries, I've updated the PR description.

PR #1898, opened 2020-12-28
docs: Correct tlsClientCA example reference

parrotmac:
#### Overview

Update path reference in documentation example.

The example path probably doesn't exist, and even if it does, the filename is pretty much certainly wrong.

```release-note
NONE
```


sagikazarmark (approved):
Thanks!

PR #1899, opened 2020-12-28
update example/k8s/dex.yaml

lcc3108:
#### Overview

update example/k8s/dex.yaml

example/k8s/dex.yaml is too old.
api version is not valid in lastest kubernetes cluster.
The service account may not have a default namespace when executing a kubectl command

NONE

NONE


sagikazarmark (approved):
Thanks!

PR #1901, opened 2020-12-29
fix: update dependencies to avoid some CVEs.

al45tair:
Updated hcsshim and Gorilla dependencies to avoid some CVEs.

I'm also seeing [Nancy](https://github.com/sonatype-nexus-community/nancy) report CVEs for golang/github.com/coreos/etcd@3.3.10 and golang/github.com/opencontainers/runc@0.1.1, but I think those relate to etcd and runc themselves (rather than our client side usage).

I would update the dependencies for etcd and runc, which would get rid of the CVE warnings from Nancy, but updating etcd is not possible at the moment because their go.mod is broken and updating runc appears to have some problems too. Since the CVEs don't really apply to Dex's use of them anyway, I don't see a reason to care too much about them at the moment.


sagikazarmark (approved):
Thanks!

etcd 3.5 will fix those module issues and we'll be able to rely on the client modules only.

Issue #1876, opened 2020-12-02
Allow to disable env variable expansion by setting DEX_EXPAND_ENV = false


heidemn-faro:
**Is your feature request related to a problem? Please describe.**
This is both a bug and a feature request.
The current way to expand env variables in the Dex config is error-prone, since it happens before the YAML parsing.
Therefore, weird hacks are needed e.g. to support `$` characters in the LDAP `bindPW`.
We're currently using a workaround there by defining an env variable `${DOLLAR}` with value `$`.
https://github.com/dexidp/dex/issues/1051
https://github.com/dexidp/dex/issues/1382

**Describe the solution you'd like to see**
Add a toplevel config variable `expandEnv` which defaults to `true` (for compatibility with existing config files), but can be set to `false` by Dex admins who don't need variable expansion at all.

**Describe alternatives you've considered**
https://github.com/dexidp/dex/issues/1838

**Additional context**
I can make this change, but @sagikazarmark or @bonifaido can you tell me if such a PR would be accepted?

**Current implementation:**

These parts would be adapted in the PR.

```go
func (s *Storage) UnmarshalJSON(b []byte) error {
	// ...
	if len(store.Config) != 0 {
		data := []byte(os.ExpandEnv(string(store.Config)))
		if err := json.Unmarshal(data, storageConfig); err != nil {
			return fmt.Errorf("parse storage config: %v", err)
		}
	}
	// ...
}

func (c *Connector) UnmarshalJSON(b []byte) error {
	// ...
	if len(conn.Config) != 0 {
		data := []byte(os.ExpandEnv(string(conn.Config)))
		if err := json.Unmarshal(data, connConfig); err != nil {
			return fmt.Errorf("parse connector config: %v", err)
		}
	}
	// ...
}

```

sagikazarmark:
How about an env var instead that can disable this behavior? Then you don't have to parse the configuration upfront. Something like `DEX_EXPAND_ENV`.

The config layer needs a broader refactor in my opinion, but adding an env var like this would probably improve the situation in the short term.

heidemn-faro:
@sagikazarmark an env variable is also fine for me. Then I'll implement it like this. 👍

heidemn-faro:
Update:
`DEX_EXPAND_ENV` = falsy value -> don't expand.
Unset, or truthy value -> expand.

PR #1902, opened 2020-12-29
Allow to disable os.ExpandEnv for storage + connector configs by env variable DEX_EXPAND_ENV = false

heidemn-faro:
#### Overview

Closes https://github.com/dexidp/dex/issues/1876

See issue + release note below.

@sagikazarmark I guess this one is for you, since you commented on the issue.

```release-note
By setting the env variable `DEX_EXPAND_ENV` to `false`, the user can disable env expansion in storage and connector configs.
This is an opt-in change: Env expansion is still enabled by default.

It can be useful to disable env expansion to avoid workarounds e.g. if you use templating to generate dex-config.yaml, and some of your values (e.g. the LDAP bindPW) may include dollar characters.
```

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"golang.org/x/crypto/bcrypt"
@@ -181,6 +182,19 @@ var storages = map[string]func() StorageConfig{
 	"mysql":      func() StorageConfig { return new(sql.MySQL) },
 }
 
+// isExpandEnvEnabled returns if os.ExpandEnv should be used for each storage and connector config.
+// Disabling this feature avoids surprises e.g. if the LDAP bind password contains a dollar character.
+// Returns false if the env variable "DEX_EXPAND_ENV" is a falsy string, e.g. "false".
+// Returns true if the env variable is unset or a truthy string, e.g. "true", or can't be parsed as bool.
+func isExpandEnvEnabled() bool {
+	enabled, err := strconv.ParseBool(os.Getenv("DEX_EXPAND_ENV"))
+	if err != nil {
+		// Unset, empty string or can't be parsed as bool: Default = true.
+		return true
+	}
+	return enabled
+}
+
 // UnmarshalJSON allows Storage to implement the unmarshaler interface to
 // dynamically determine the type of the storage config.
 func (s *Storage) UnmarshalJSON(b []byte) error {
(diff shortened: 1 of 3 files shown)


cmd/dex/config.go
```diff
@@ -181,6 +181,14 @@ var storages = map[string]func() StorageConfig{
 	"mysql":      func() StorageConfig { return new(sql.MySQL) },
 }
 
+// isExpandEnvEnabled returns if os.ExpandEnv should be used for each storage and connector config.
+// Disabling this feature avoids surprises e.g. if the LDAP bind password contains a dollar character.
+// True by default, to avoid a breaking change.
+// False if the env variable "DEX_EXPAND_ENV" is set to "0".
+func isExpandEnvEnabled() bool {
+	return os.Getenv("DEX_EXPAND_ENV") != "0"
```
sagikazarmark, line 189: How about using [ParseBool](https://golang.org/pkg/strconv/#ParseBool) instead? Then a whole bunch of other boolean-like values can be accepted. `false/true` comes to mind as better alternatives to 0/1 and I'd also use those in the documentation instead of 0/1.

Obviously this means that you have to look up the env first.

Also, thinking about the variable name: since this is true by default, I wonder if  it would make sense to use `DEX_NO_EXPAND_ENV` instead.

heidemn: > How about using ParseBool instead?

Sounds good, will do that.

> I wonder if it would make sense to use DEX_NO_EXPAND_ENV instead.

I'm not a big fan of this, for the following reasons:
* Consider `DEX_NO_EXPAND_ENV=false`, which is a double negation. I wanted to avoid that case.
* In case that we later change the default to "don't expand", we would have to change back the variable name `DEX_NO_EXPAND_ENV` -> `DEX_EXPAND_ENV`.

E.g. the [OpenLDAP Docker image](https://github.com/osixia/docker-openldap) also uses `LDAP_...=false` instead of `LDAP_NO_...=true`
But if you insist, I can still change it.

heidemn-faro:
@sagikazarmark please take another look, and let me know if further changes are required.

cmd/dex/config.go
```diff
@@ -181,6 +182,20 @@ var storages = map[string]func() StorageConfig{
...
+// Returns true if the env variable "DEX_EXPAND_ENV" is unset or a truthy string, e.g. "true".
+// Returns false if the env variable is a falsy string, e.g. "false".
+// Returns an error if the env variable can't be parsed into a bool.
+func isExpandEnvEnabled() (bool, error) {
+	dexExpandEnv, exists := os.LookupEnv("DEX_EXPAND_ENV")
+	if !exists {
+		// Default, for downwards-compatibility: DEX_EXPAND_ENV = true
+		return true, nil
+	}
+	return strconv.ParseBool(dexExpandEnv)
```
sagikazarmark, line 196: The reason I thought a negated variable name might be better is because variable expansion should be true by default, even if the content of the variable is considered invalid.

I agree it sounds bad, but I think we can keep `DEX_EXPAND_ENV`: I think we should handle the error returned by `ParseBool`: if there is an error, it should return true, otherwise return the parsed value. I think it's okay to silently fall back to true when the value of the variable is empty for example or some gibberish.

heidemn-faro: Oh no, then I spent too much effort on the tests etc.... But you're right, silently using the default seems fine.

heidemn-faro:
@sagikazarmark please take a 3rd look :-) Simplified as requested.

sagikazarmark (approved):
Thanks @heidemn

heidemn:
Thanks for reviewing @sagikazarmark , and happy new year!

PR #1904, opened 2020-12-30
feat: ldap cluster connector

### LINKED ISSUE #1548: Q: Fault tolerant LDAP connection?
Is it possible to configure dex in a way that is connecting to one LDAP server and in case that is not reachable fail over to a second LDAP server?

I.e. in Active Directory environments fail-over is required if one domain controller gets updated / rebooted - during that period of time the LDAP/AD clients need to fail over to a second one which is providing the same data for fault tolerance and scalability.

phiremande:
#### Overview

This changes implements a new ldapcluster connector, which helps configure multiple ldap servers and requests are tried in round robin fashion to these servers with an affinity to last active server.
https://github.com/dexidp/dex/issues/1548

```release-note
Introduces a new connector 'ldapcluster', which helps configure a cluster of ldap servers.  This can be used to provide fault tolerant LDAP connection.
```


scottd018:
+1 to this MR.  Have an enterprise customer that wishes to use Dex but has concerns with lack of redundancy provided by the connector.

sagikazarmark:
Thanks for the contribution. TBH, I don't think this is the best way to tackle the problem. I think it should be handled on the connection/LDAP connection level.

I asked about an existing solution in the LDAP client repo. Let's see what they answer.

stefanmcshane:
Hi All,

We are discussing expanding the reach of the ldap library, and adding support for clustering connections. 

Any ideas or usecases you have would be appreciated on go-ldap/ldap#314

Thanks,
Stefan

PR #1905, opened 2020-12-30
Add Apple connector

mfmarche:
#### Overview

This pull request introduces the apple connector. A new connector was introduced due differences with the standard OIDC connector, namely:
 - uses response_mode=form_post for /auth/authorize
 - secret is generated from a private key. I added secret generation automatically which allows short duration keys and automatic creation

Many thanks to https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple for the initial work.

Allows OIDC signin with Apple ID.

Apple secrets work differently than other oidc connectors in that the secret is generated from a private key, provided by the apple developer portal. The secret is a JWT and the claims need to be provisioned in the connector configuration section.


sagikazarmark:
Thanks a lot for the PR @mfmarche!

Can you please post some detailed steps to test this connector? If we accept it to the core, we need a way to test it.

Also, FYI we are working on a way to allow external connectors to be registered in Dex without making them part of the core, so there is a change we won't accept this change for a while. Thank you for understanding!

mfmarche:
Hi @sagikazarmark , It seems the documentation is in a different repo now. I will post here for now how to use it.

The following configuration items are need in the connector configuration (in addition to the other typical oidc config like name, redirectURL, etc)

- clientID
- team_id
- key_id
- private_key_file

These items all come from the apple dev portal, which https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple does a great job explaining with pictures how and where to get them.

PR #1908, opened 2021-01-05
Fix the etcd PKCE AuthCode deserialization

bnu0:
#### Overview

This change fixes a bug in the PKCE implementation when using `etcd` as the storage engine. As of v2.27.0, this combination is broken and a `code_verifier` cannot be used to request a token. 

Specifically, you will get an errant error `No PKCE flow started. Cannot check code_verifier` error when requesting `/token` despite the fact that a `code_challenge` and `code_challenge_method` were indeed provided to `/auth`.

This PR also adds a PKCE AuthCode round-trip check to the storage conformance tests. Indeed that test fails on v2.27, and now it passes 🙏 .

- In `CreateAuthCode`, the etcd storage implementation uses `fromStorageAuthCode()` to convert a `storage.AuthCode` into the etcd variant.
- In `GetAuthCode`, however, the implementation attempts to directly unmarshal the json into the `storage.AuthCode` without the etcd intermediate.
- This is wrong! Specifically, the PKCE data is inline in the etcd `AuthCode` struct, but is nested in the `storage.AuthCode` struct.
- So as a result, it kind of works because most fields line up, but the `CodeChallenge` is always the empty string.
- Therefore, `handlers.go:820` always gets `codeChallengeFromStorage = ""` and falls into the error condition.
- This PR adds the `toStorageAuthCode` equivalent to correctly deserialize, and uses it. 

Enjoy! 💖

```release-note
NONE
```

### DIFF
--- a/storage/conformance/conformance.go
+++ b/storage/conformance/conformance.go
@@ -185,6 +185,10 @@ func testAuthCodeCRUD(t *testing.T, s storage.Storage) {
 		Expiry:        neverExpire,
 		ConnectorID:   "ldap",
 		ConnectorData: []byte(`{"some":"data"}`),
+		PKCE: storage.PKCE{
+			CodeChallenge:       "12345",
+			CodeChallengeMethod: "Whatever",
+		},
 		Claims: storage.Claims{
 			UserID:        "1",
 			Username:      "jane",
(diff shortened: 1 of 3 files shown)


kirooshu (approved):
LGTM

bnu0:
CC @HEllRZA @sagikazarmark as participants in the original PKCE implementation over in #1784

sagikazarmark (approved):
LGTM, thanks!

PR #1912, opened 2021-01-11
Support setting the prompt type for the Microsoft connector

### LINKED ISSUE #1911: Add an option for setting prompt= on the Microsoft connector
When logging in with the Microsoft connector, it doesn't prompt for which account to use, sometimes resulting in weird failures since we are using our tenant for login. (Additionally, we would prefer it if the login flow for Microsoft prompted again after logging out in Dex.)

Like other connectors, the `prompt=` parameter should be configurable.

It might be possible to just provide an override for the authorisation URL, though this seems pretty heavy handed.

#### Additional context
I'm happy to put in a PR! It seems pretty straight-forward.

ricky26:
Signed-off-by: Erica Taylor <ricky@wellplayed.games>

#### Overview
This PR adds a `promptType` field to the configuration of the Microsoft connector.
Closes #1911.

This allows users of Dex to specify which kind of prompt the Microsoft connector should display. Previously this used the default value which didn't prompt unless the user was signed out of Microsoft.

```release-note
The prompt type which the Microsoft connector uses can now be configured with the `promptType` setting.
Valid values are `login`, `none`, `consent` and `select_account`.
```

### DIFF
--- a/connector/microsoft/microsoft.go
+++ b/connector/microsoft/microsoft.go
@@ -53,6 +53,10 @@ type Config struct {
 	GroupNameFormat      GroupNameFormat `json:"groupNameFormat"`
 	UseGroupsAsWhitelist bool            `json:"useGroupsAsWhitelist"`
 	EmailToLowercase     bool            `json:"emailToLowercase"`
+
+	// PromptType is used for the prompt query parameter.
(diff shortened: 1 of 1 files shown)


sagikazarmark (approved):

sagikazarmark:
Thanks!

PR #1913, opened 2021-01-12
chore: add frontend section to dev config

nabokihms:
#### Overview
* Add frontend section to `examples/config-dev.yaml`
* Fix stale code commentaries
* Minor fixes dark theme colors

The initial intention of this PR was to add a frontend settings section to ease fronted development. Playing with the example-app and frontend settings, I found outdated commentaries and possible improvements for dark theme colors. To avoid spamming with PRs, the fixes were included in this one.

Nothing that should be added to release notes.

### DIFF
--- a/examples/config-dev.yaml
+++ b/examples/config-dev.yaml
@@ -52,11 +52,10 @@ web:
 
 # Configuration for dex appearance
 # frontend:
-  # issuer: Dex
-  # issuerUrl: http://localhost:5556
-  # logoURL: theme/logo.png
-  # dir: web/
-  # theme: light
+#   issuer: dex
+#   logoURL: theme/logo.png
+#   dir: web/
+#   theme: light
 
 # Configuration for telemetry
 telemetry:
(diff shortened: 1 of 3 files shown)


sagikazarmark (commented):
Can you include a few screenshots highlighting the changes? Thanks!

nabokihms:
Without changes:
Снимок экрана 2021-01-12 в 19 10 47
Снимок экрана 2021-01-12 в 19 11 03

With changes:
Снимок экрана 2021-01-12 в 19 12 00
Снимок экрана 2021-01-12 в 19 11 49

Now dark theme approval screen looks more like the light theme one (with a colored button). And I also changed the form fields background color to dark-gray.

examples/config-dev.yaml
```diff
@@ -50,6 +50,14 @@ web:
   # tlsCert: /etc/dex/tls.crt
   # tlsKey: /etc/dex/tls.key
 
+# Configuration for dex appearance
+# frontend:
+  # issuer: Dex
+  # issuerUrl: http://localhost:5556
+  # logoURL: theme/logo.png
```
sagikazarmark, line 57: Are `Url` and `URL` case sensitive fields? I think we should be more consistent about naming config parameters. If it's case sensitive and the fields are named differently...that's life, otherwise let's use the same casing (honestly, don't care which one)

sagikazarmark (approved):
Thanks @nabokihms

examples/config-dev.yaml
nabokihms: They are both case-insensitive, I think.

nabokihms: Oh, I just accidentally copied the issuerUrl option from somewhere. It turns out that it doesn't exist.

sagikazarmark:
Thanks!

PR #1915, opened 2021-01-13
Update CODE_OF_CONDUCT.md

nabokihms:
#### Overview
* Update CODE_OF_CONDUCT.md

The CNCF code of conduct version has several advantages:
* Well-maintained
* Multilingual
* Based on the same Contributor Covenant version (1.2.0)

Because of joining the CNCF, it would be better to start following its code of conduct.

Nothing is changed in the application.


sagikazarmark (commented):
Thanks! I think we should move it to the .github directory to keep the root clean.

sagikazarmark (approved):
Thanks!

PR #1946, opened 2021-01-15
Enable unparam, prealloc, sqlclosecheck linters

nabokihms:

#### Overview

* Enable some useful linters
* Fix errors

We need this to make (and keep) code cleaner.

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -291,9 +291,7 @@ func (d dexAPI) ListRefresh(ctx context.Context, req *api.ListRefreshReq) (*api.
 		if err == storage.ErrNotFound {
 			// This means that this user-client pair does not have a refresh token yet.
 			// An empty list should be returned instead of an error.
-			return &api.ListRefreshResp{
-				RefreshTokens: nil,
-			}, nil
+			return &api.ListRefreshResp{}, nil
 		}
 		d.logger.Errorf("api: failed to list refresh tokens %t here : %v", err == storage.ErrNotFound, err)
 		return nil, err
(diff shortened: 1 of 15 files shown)


server/api.go
```diff
@@ -286,20 +286,20 @@ func (d dexAPI) ListRefresh(ctx context.Context, req *api.ListRefreshReq) (*api.
...
 
-	var refreshTokenRefs []*api.RefreshTokenRef
 	offlineSessions, err := d.s.GetOfflineSessions(id.UserId, id.ConnId)
 	if err != nil {
 		if err == storage.ErrNotFound {
 			// This means that this user-client pair does not have a refresh token yet.
 			// An empty list should be returned instead of an error.
 			return &api.ListRefreshResp{
-				RefreshTokens: refreshTokenRefs,
+				RefreshTokens: nil,
```
sagikazarmark, line 295: Can't we just remove this line?

server/server.go
```diff
@@ -270,10 +268,10 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
 		}
 
 		instrumentHandlerCounter = func(handlerName string, handler http.Handler) http.HandlerFunc {
```
sagikazarmark, line 270: Not the scope of this PR, but I wonder if this should return an `http.Handler` instead.

nabokihms: I think it returns http.HandlerFunc because of the way `httpsnoop` works. For me, it looks better to fix it in another PR with server refactoring.

sagikazarmark (approved):
Thanks!

PR #1948, opened 2021-01-18
Add Cache-control headers to token responses

nabokihms:

#### Overview
* Add headers
* Add tests

https://tools.ietf.org/html/rfc6749#section-5.1
> The authorization server MUST include the HTTP "Cache-Control"
   response header field [RFC2616] with a value of "no-store" in any
   response containing tokens, credentials, or other sensitive
   information, as well as the "Pragma" response header field [RFC2616]
   with a value of "no-cache".

It is also ok (and better) to add Cache-Control header to device authorization response as in the example in https://tools.ietf.org/html/rfc8628#section-3.2

Cache headers are required to pass the OpenID certification.

```release-note
Add "Cache-control: no-store" and "Pragma: no-cache" headers to token responses
```

### DIFF
--- a/server/deviceflowhandlers.go
+++ b/server/deviceflowhandlers.go
@@ -140,6 +140,10 @@ func (s *Server) handleDeviceCode(w http.ResponseWriter, r *http.Request) {
 			PollInterval:            pollIntervalSeconds,
 		}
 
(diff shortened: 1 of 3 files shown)


sagikazarmark (approved):
Thanks!

PR #1949, opened 2021-01-18
Use constants to form oauth2 error responses

nabokihms:

#### Overview
* Used a constant as the first argument of newErr method
* Replaced error state "invalid_requests" with "invalid_request" in case of receiving no response_type
* Added a test

According to https://tools.ietf.org/html/rfc6749#section-3.1.1, the state `invalid_requests` is invalid (and it looks like a typo). 
To avoid similar errors in the future, this PR replaces all state arguments with constants.

Correct error on empty response_type is required to pass the OpenID certification.

The only user-facing change is replacing invalid_requests with invalid_request.

### DIFF
--- a/server/oauth2.go
+++ b/server/oauth2.go
@@ -485,13 +485,13 @@ func (s *Server) parseAuthorizationRequest(r *http.Request) (*storage.AuthReques
 		}
 	}
 	if !hasOpenIDScope {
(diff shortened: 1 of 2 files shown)


bonifaido (approved):
LGTM!

sagikazarmark (approved):

PR #1951, opened 2021-01-20
fix: update auth methods and claims from discovery endpoint

nabokihms:

#### Overview
* Update claims_supported
* Update token_endpoint_auth_methods_supported

Dex supports sending client credentials via post but doesn't say about it on discovery. The same about claims.

Adding `client_secret_post` to auth metods is required to pass Basic OpenID certification tests.

```release-note
Discovery endpoint contains updated claims and auth methods.
```


sagikazarmark (approved):
Thanks!

PR #1952, opened 2021-01-20
fix: return invalid_grant error for invalid or expired auth codes

nabokihms:

#### Overview
Change returning error for invalid auth codes.

The https://tools.ietf.org/html/rfc6749#section-5.2 says that we MUST return invalid_grant error for invalid auth codes.

Returning correct (according to RFC) errors is required to pass OpenID certification tests.

```release-note
Return invalid_grant error if auth code is invalid or expired.
```

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -811,7 +811,7 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 			s.logger.Errorf("failed to get auth code: %v", err)
 			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
 		} else {
-			s.tokenErrHelper(w, errInvalidRequest, "Invalid or expired code parameter.", http.StatusBadRequest)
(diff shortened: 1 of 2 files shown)


server/handlers.go
```diff
@@ -811,7 +811,7 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 			s.logger.Errorf("failed to get auth code: %v", err)
 			s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
 		} else {
-			s.tokenErrHelper(w, errInvalidRequest, "Invalid or expired code parameter.", http.StatusBadRequest)
+			s.tokenErrHelper(w, errInvalidGrant, "Invalid or expired code parameter.", http.StatusBadRequest)
```
sagikazarmark, line 814: I wonder if we should check if the code is missing/empty first (and return `invalid_request` if so). WDYT?

nabokihms: Looks like your assumption is right, but I need to check it first.

sagikazarmark (approved):
Thanks!

PR #1956, opened 2021-01-22
fix: unsupported request parameter error

nabokihms:

#### Overview
Check the "request" in auth request and throw an error if found. 

According to the [specification](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests), an error is added. There is a [specification for error responses](https://openid.net/specs/openid-connect-core-1_0.html#AuthError).

Sending an error is required to pass OpenID certification tests.  The previous test suite allows completely ignoring this parameter, but after migrating to the new suit treating the request parameter correctly becomes mandatory.

https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests

```release-note
Return an error to auth requests with the "request" parameter.
```

### DIFF
--- a/server/oauth2.go
+++ b/server/oauth2.go
@@ -95,6 +95,7 @@ const (
 	errUnauthorizedClient      = "unauthorized_client"
 	errAccessDenied            = "access_denied"
 	errUnsupportedResponseType = "unsupported_response_type"
+	errRequestNotSupported     = "request_not_supported"
 	errInvalidScope            = "invalid_scope"
 	errServerError             = "server_error"
 	errTemporarilyUnavailable  = "temporarily_unavailable"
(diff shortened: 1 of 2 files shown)


server/oauth2.go
```diff
@@ -453,6 +454,12 @@ func (s *Server) parseAuthorizationRequest(r *http.Request) (*storage.AuthReques
 		return &authErr{state, redirectURI, typ, fmt.Sprintf(format, a...)}
 	}
 
+	// dex doesn't support request parameter and must return request_not_supported error
+	// https://openid.net/specs/openid-connect-core-1_0.html#6.1
+	if q.Get("request") != "" {
+		return nil, newErr(errRequestNotSupported, "Server does not support request parameter.")
```
sagikazarmark, line 460: Do I recall correctly that the first implementation involved some redirections? I briefly looked at that version and I didn't really understand it.

nabokihms: I realized that this Dex already does the necessary redirect if an error is returned from the `parseAuthorizationRequest` method.

sagikazarmark (approved):
Thanks!

Issue #1954, opened 2021-01-21
Docker volume support for storing sqlite3 database


mouchar:
#### Is your feature request related to a problem?
When I want to start using Dex IdP docker image, the easiest storage driver to configure is the sqlite3. But the only writable directories for user `1001` are `/tmp` and `/var/tmp`. I'd like to store sqlite3 datafile to docker volume to keep changes persistent between multiple Dex container incarnations.

Unfortunately, Docker doesn't allow to set volume ownership/permissions explicitly, but it inherits ownership from the directory, where the volume is being mounted.

#### Describe the solution you'd like to see
Create a writeable directory in Dockerfile, optionally assigning it a mount point using `VOLUME` instruction.

```dockerfile
RUN mkdir /data && chown -R 1001 /data
VOLUME /data

# existing line
USER 1001:1001
```

User can run container from image like:
```
docker run --rm -v /path/to/dex.yaml:/dex.yaml -v dex-data:/data dexidp/dex:v2.27.0 serve /dex.yaml
```
(assuming the dex.yaml points sqlite3 storage file to `/data/dex.db`)

Any changes (users, clients added via gRPC) will be kept in `dex-data` volume.

#### Describe alternatives you've considered
There are not many alternatives, except for building own image based on `dexidp/dex` image that contains this writable directory. Of course, users could also use another storage backend (like Postgres DB), but it is a more complex solution that could deter faint-hearted souls.

nabokihms:
Hello!

With docker for mac, it is possible to simply mount a folder without changing anything in the docker file.
```bash
docker run --rm -v "$(pwd)/dex-data:/examples" -v "$(pwd)/examples/config-dev.yaml:/etc/dex/config.yaml" dexidp/dex:v2.27.0 serve  /etc/dex/config.yaml
```

However, using a Linux environment, you need extra steps.
```bash
data_dir=$(pwd)/dex-data

mkdir -p "$data_dir"
chmod 755 "$data_dir"
chown 1001 "$data_dir"
docker run --rm -v "$(pwd)/dex-data:/examples" -v "$(pwd)/examples/config-dev.yaml:/etc/dex/config.yaml" dexidp/dex:v2.27.0 serve  /etc/dex/config.yaml
```

It is an example of how you can avoid changing the docker file.

mouchar:
Hello Maksim, thank you for your quick response. I'm working on Linux, so it is possible to use the workaround from your proposal based on bind mount. There are other issues with this solution, for example:
[...]

PR #1957, opened 2021-01-23
Add new configuration examples

sagikazarmark:
#### Overview

New configuration examples for development purposes.

This PR adds two new configuration examples to the repo:

- `config.yaml.dist` - This serves as a dummy, empty config demonstrating most of the available options. Whenever making config option changes in code, this file should be updated to reflect those changes. Not all options are detailed here though: connector, client and password changes should be documented instead.
- `config.dev.yaml` - This file serves as a template for development purposes. It's minimal and contains only the necessary config options. It should be used instead of `examples/config-dev.yaml` (easier to find in the repo root, we can add it to the `Makefile` at some point)

`config.yaml.dist` can serve as a default configuration mentioned in https://github.com/dexidp/dex/issues/1954#issuecomment-765717418

```release-note
NONE
```

### DIFF
--- a/config.dev.yaml
+++ b/config.dev.yaml
@@ -1,0 +1,35 @@
+issuer: http://127.0.0.1:5556/dex
+
+storage:
+  type: sqlite3
+  config:
+    file: var/sqlite/dex.db
+
+web:
+  http: 127.0.0.1:5556
+
+telemetry:
+  http: 127.0.0.1:5558
+
+grpc:
+  addr: 127.0.0.1:5557
+
+staticClients:
+  - id: example-app
+    redirectURIs:
+      - 'http://127.0.0.1:5555/callback'
+    name: 'Example App'
+    secret: ZXhhbXBsZS1hcHAtc2VjcmV0
+
+connectors:
+  - type: mockCallback
+    id: mock
+    name: Example
+
+enablePasswordDB: true
+
+staticPasswords:
+  - email: "admin@example.com"
+    hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
+    username: "admin"
+    userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,3 +16,6 @@ indent_size = 2
 
 [{Makefile,*.mk}]
 indent_style = tab
+
+[config.yaml.dist]
+indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,5 +17,5 @@ indent_size = 2
 [{Makefile,*.mk}]
 indent_style = tab
 
-[config.yaml.dist]
+[{config.yaml.dist,config.dev.yaml}]
 indent_size = 2
--- a/config.yaml.dist
+++ b/config.yaml.dist
@@ -1,0 +1,131 @@
+# The base path of Dex and the external name of the OpenID Connect service.
+# This is the canonical URL that all clients MUST use to refer to Dex. If a
+# path is provided, Dex's HTTP service will listen at a non-root URL.
+issuer: http://127.0.0.1:5556/dex
+
+# The storage configuration determines where Dex stores its state.
+# Supported options include:
+#   - SQL flavors
+#   - key-value stores (eg. etcd)
+#   - Kubernetes Custom Resources
+#
+# See the documentation (https://dexidp.io/docs/storage/) for further information.
+storage:
+  type: memory
+
+  # type: sqlite3
+  # config:
+  #   file: /var/dex/dex.db
+
+  # type: mysql
+  # config:
+  #   host: 127.0.0.1
+  #   port: 3306
+  #   database: dex
+  #   user: mysql
+  #   password: mysql
+  #   ssl:
+  #     mode: "false"
+
+  # type: postgres
+  # config:
+  #   host: 127.0.0.1
+  #   port: 5432
+  #   database: dex
+  #   user: postgres
+  #   password: postgres
+  #   ssl:
+  #     mode: disable
+
+  # type: etcd
+  # config:
+  #   endpoints:
+  #     - http://127.0.0.1:2379
+  #   namespace: dex/
+
+  # type: kubernetes
+  # config:
+  #   kubeConfigFile: $HOME/.kube/config
+
+# HTTP service configuration
+web:
+  http: 127.0.0.1:5556
+
+  # Uncomment to enable HTTPS endpoint.
+  # https: 127.0.0.1:5554
+  # tlsCert: /etc/dex/tls.crt
+  # tlsKey: /etc/dex/tls.key
+
+# Dex UI configuration
+# frontend:
+#   issuer: dex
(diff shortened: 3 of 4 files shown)


config.dev.yaml
```diff
@@ -0,0 +1,35 @@
+issuer: http://127.0.0.1:5556/dex
+
+storage:
+  type: memory
```
mouchar, line 4:
```suggestion
  type: sqlite3
```

sagikazarmark: I changed it to memory on purpose. This is supposed to be a dummy config. Without the proper volume mapping changing it to sqlite works the same way.

config.dev.yaml
```diff
@@ -0,0 +1,35 @@
...
+  config:
+    file: var/sqlite/dex.db
```
mouchar, line 6: this path doesn't correspond to `/var/dex` from #1958

sagikazarmark: This config is for local development, relative to the project directory.

mouchar: The memory storage backend doesn't have any config options, so it is confusing to keep sqlite3-specific settings here.

So I recommend removing `config` completely if the `storage.type` is set to `memory`.

sagikazarmark: Damn, I missed that all your comments were related to the `config.dev.yaml` file. 😄 

My original intention was to use the memory storage in the default config and sqlite in the dev config. I'll fix that in a bit.

sagikazarmark:
@mouchar thanks for the sanity check. Should be good now.

config.dev.yaml
mouchar: No problem, but keep in mind that directory `var/sqlite/` doesn't exist, so the Dex will fail to start if you use this config file (both locally and in docker). That was my original proposal to put `/var/dex/dex.db` here so it would work in docker, at least.
Maybe other developers could express their opinion if this change is acceptable? I really don't want to break their workflow...

If your intention with this file was to allow instantly running dex Docker image, you should put `file: /var/dex/dex.db` here.
But if you prefer running dex locally, then having something like `file: dex.db` is better.

Or, a more hackish approach: add empty file `var/dex/.keep` to git and use `file: var/dex/dex.db`. It will run both in docker and locally, provided you start local dex from the git repo root directory. :wink:

sagikazarmark: This file is intended for local development, so Dex will most probably run locally (that's how we develop Dex today). So for this file, I don't really care much about making it work in docker.

As for the non-existent path: I haven't really thought about it. Ideally, this file would be copied to `config.yaml` (so that developers can change it as needed) by some `make up` or similar command, which could also create the var directory?

Or perhaps I should just revert it to `dex.db`? I don't know what others think, but it physically hurts me to see random files hanging around in the project root.

bonifaido (approved):

config.dev.yaml
jawabuu: Hey @mouchar @sagikazarmark I cam e across this while working on a different issue.
I'm currently deploying dexidp in docker-compose. I am using the  manifests in this issue.
https://github.com/traefik/traefik/issues/7814
Let me know if this could enhance dev experience for running dexidp with docker.

sagikazarmark: @jawabuu Thanks! I'm not entirely sure what I should be looking at though. We are in the middle of improving our dev environment, so things will change anyway.

mouchar: From a developer experience perspective, the major drawback of Dex in the Docker deployment I can see is the need for an external config file. For quick tests and prototypes, running `docker run dex` is simply not enough, you need to get an example config file from the git repo, or write your own from scratch. 12-factor app methodology recommends configuring apps using the environment. Environment cfg is currently very limited (according to `git grep os.Getenv`). Having the possibility to run the container straight from an image (without the need to download and bind-mount a config file) would be great, especially for the docker/docker-compose use case.
On the other hand, running Dex in K8S with external config is fine - proper tools already exist in the K8S ecosystem that will help you to generate a Dex config file in advance.

Just my 2¢ based on my experience with Dex (local run, Docker, Docker Compose, and Kubernetes+Helm).

sagikazarmark: @mouchar I agree, configuring Dex is a bit cumbersome today.

#1965 should help with using environment variables. There are also flags now for a few options (where it makes sense): addresses mainly (http, https, grpc, telemetry).

This PR should also improve the default configuration so that you can run Dex without a config file, albeit it might not be useful for anything.

jawabuu: Hey @mouchar @sagikazarmark 
Thank you for getting back to me.

In my opinion having a docker-compose file & a default config file in the repo that can consistently & reliably bring up dexidp and quickly demonstrate it's features provides an unmatched developer experience.
That is, as long as you have docker just clone the example & run
`docker-compose up -d`
that's it 🙂
As this is a dev environment running locally, bind mounts are expected and actually preferred to allow one to play around with the different config options.
Additionally, the docker-compose should work hand in hand with existing examples.
You could have the following structure,
`examples/local`
`examples/docker-compose`
`examples/[anything]`
to give you flexibility in future.
The `docker-compose` directory will have at least the following files
`examples/docker-compose/docker-compose.yml`
`examples/docker-compose/config.yml`
Let me post the files to demonstrate

jawabuu: #### Structure
```
+---compose
|       dex-config.yml
|       docker-compose.yml
```
####  dex-config.yml
```dex-config.yml
issuer: http://localhost:5556
storage:
  type: memory
web:
  http: 0.0.0.0:5556
telemetry:
  http: 0.0.0.0:5558
logger:
  level: "debug"
staticClients:
- name: 'Example'
  id: example-app
  secret: ZXhhbXBsZS1hcHAtc2VjcmV0
  redirectURIs:
  - 'http://localhost:6555/callback'
  - 'http://dex-app:6555/callback'
connectors:
- type: mockCallback
  id: mock
  name: Example
enablePasswordDB: true
staticPasswords:
- email: "admin@example.org"
  # bcrypt hash of the string "password"
  username: "admin"
```
####  docker-compose.yml
``` docker-compose.yml
version: '2.1'
networks:
  dex:
services:  
  dex:
    image: ghcr.io/dexidp/dex:v2.27.0
    healthcheck:
      test: ["CMD", "wget", "-qO-", "-t1", "http://localhost:5556/healthz"]
      interval: 30s
      timeout: 60s
      retries: 5
    command: ["serve", "/etc/config.yml"]
    ports:
      - 5556:5556
    volumes:
      - "./dex-config.yml:/etc/config.yml:ro"
    networks:
      dex:
        aliases:
          - localhost
  
  dex-app:
    image: obitech/dex-example-app
    command: ["--debug","--listen=http://dex-app:6555","--redirect-uri=http://localhost:6555/callback","--issuer=http://localhost:5556"]
    ports:
      - 6555:6555
    networks:
      - dex
    depends_on:
      dex:
        condition: service_healthy

```

Above is the most basic example that you can test out.

jawabuu: @mouchar @sagikazarmark 
If acceptable, you may have a more advanced setup that includes e.g. a mysql or postgres service and an example `.env` file to demonstrate env variable features.
This is simple enough to do.
You should probably also host the `dex-example-app` image yourselves.

PR #1963, opened 2021-01-26
feat: graceful shutdown

### LINKED ISSUE #1942: Graceful shutdown
Dex shutdown is not graceful at the moment.

Use https://github.com/oklog/run in the serve command

#### Additional context
Fixes #1523
Fixes #1541

nabokihms:

#### Overview
Add graceful shutdown mechanism to all servers:
* HTTP
* HTTPS
* Telemetry
* grpc

Closes #1942

```release-note
Graceful shutdown.
```

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -10,11 +10,14 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/ghodss/yaml"
 	grpcprometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
+	"github.com/oklog/run"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
@@ -66,6 +69,28 @@ func commandServe() *cobra.Command {
 	flags.StringVar(&options.grpcAddr, "grpc-addr", "", "gRPC API address")
 
 	return cmd
+}
+
+func listenAndShutdownGracefully(logger log.Logger, gr *run.Group, srv *http.Server, name string) error {
+	l, err := net.Listen("tcp", srv.Addr)
+	if err != nil {
+		return fmt.Errorf("listening (%s) on %s: %v", name, srv.Addr, err)
+	}
+
+	gr.Add(func() error {
+		logger.Infof("listening (%s) on %s", name, srv.Addr)
+		return srv.Serve(l)
+	}, func(err error) {
+		logger.Debugf("starting gracefully shutdown (%s)", name)
+		if err := l.Close(); err != nil {
+			logger.Errorf("gracefully close (%s) listener: %v", name, err)
+		}
+
+		if err := srv.Shutdown(context.Background()); err != nil {
+			logger.Errorf("gracefully shutdown (%s): %v", name, err)
+		}
+	})
+	return nil
 }
 
 func runServe(options serveOptions) error {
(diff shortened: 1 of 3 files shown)


cmd/dex/serve.go
```diff
@@ -327,35 +352,51 @@ func runServe(options serveOptions) error {
...
+			if err := grpcListener.Close(); err != nil {
+				logger.Errorf("failed to gracefully close (grpc) listener: %v", err)
+			}
+			grpcSrv.GracefulStop()
+		})
 	}
 
-	return <-errc
+	sig := make(chan os.Signal)
+	gr.Add(func() error {
```
sagikazarmark, line 387: The library has a builtin signal actor, maybe we could use it.

cmd/dex/serve.go
```diff
@@ -68,6 +71,28 @@ func commandServe() *cobra.Command {
...
+	}
+
+	gr.Add(func() error {
+		logger.Infof("listening (%s) on %s", name, srv.Addr)
+		return srv.Serve(l)
+	}, func(err error) {
+		logger.Debugf("starting gracefully shutdown (%s)", name)
+		if err := l.Close(); err != nil {
+			logger.Errorf("gracefully close (%s) listener: %v", name, err)
+		}
```
sagikazarmark, line 87: AFAIK the http server takes care of this, so you don't need to call close on the listener manually.

cmd/dex/serve.go
```diff
@@ -68,6 +71,28 @@ func commandServe() *cobra.Command {
...
+
+		if err := srv.Shutdown(context.Background()); err != nil {
+			logger.Errorf("gracefully shutdown (%s): %v", name, err)
+		}
+	})
+	return nil
```
sagikazarmark, line 93:
```suggestion

	return nil
```

cmd/dex/serve.go
```diff
@@ -327,35 +352,51 @@ func runServe(options serveOptions) error {
...
+		}
+
+		gr.Add(func() error {
+			logger.Infof("listening (grpc) on %s", c.GRPC.Addr)
+			return grpcSrv.Serve(grpcListener)
+		}, func(err error) {
+			logger.Debugf("starting gracefully shutdown (grpc)")
+			if err := grpcListener.Close(); err != nil {
+				logger.Errorf("failed to gracefully close (grpc) listener: %v", err)
+			}
```
sagikazarmark, line 381: Same as above

cmd/dex/serve.go
```diff
@@ -302,21 +327,21 @@ func runServe(options serveOptions) error {
...
 
-	errc := make(chan error, 3)
+	var gr run.Group
 	if c.Telemetry.HTTP != "" {
-		logger.Infof("listening (http/telemetry) on %s", c.Telemetry.HTTP)
-		go func() {
-			err := http.ListenAndServe(c.Telemetry.HTTP, telemetryServ)
-			errc <- fmt.Errorf("listening on %s failed: %v", c.Telemetry.HTTP, err)
-		}()
+		telemetrySrv := &http.Server{Addr: c.Telemetry.HTTP, Handler: telemetryServ}
```
sagikazarmark, line 332: I think `defer httpServer.Close()` should be called on each server in the main func as a general cleanup step and as a last resort to graceful shutdown.

cmd/dex/serve.go
```diff
@@ -68,6 +71,28 @@ func commandServe() *cobra.Command {
...
+	gr.Add(func() error {
+		logger.Infof("listening (%s) on %s", name, srv.Addr)
+		return srv.Serve(l)
+	}, func(err error) {
+		logger.Debugf("starting gracefully shutdown (%s)", name)
+		if err := l.Close(); err != nil {
+			logger.Errorf("gracefully close (%s) listener: %v", name, err)
+		}
+
+		if err := srv.Shutdown(context.Background()); err != nil {
```
sagikazarmark, line 89: What happens if there is an active connection for an indefinite amount of time? Will this wait indefinitely? If so, we should probably add a timeout here.

sagikazarmark (changes requested):
Thanks @nabokihms !

I had a few requests, but it's great to finally have graceful shutdown in Dex!

cmd/dex/serve.go
nabokihms: My thought was about splitting listener closing error and server shutdown error. Maybe it is excessive.

cmd/dex/serve.go
nabokihms: Fixed. The only thing that I dislike about helper is that I can't log the received signal immediately.

cmd/dex/serve.go
nabokihms: Listener closing now is a duty of a server.

cmd/dex/serve.go
nabokihms: Yes, good catch. I hardcoded a one-minute timeout here. Please tell me if you want to make this configurable or increase the value.

cmd/dex/serve.go
```diff
@@ -68,6 +70,27 @@ func commandServe() *cobra.Command {
...
+	}
+
+	gr.Add(func() error {
+		logger.Infof("listening (%s) on %s", name, srv.Addr)
+		return srv.Serve(l)
+	}, func(err error) {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+		defer cancel()
+
+		logger.Debugf("starting gracefully shutdown (%s)", name)
```
sagikazarmark, line 86: nit

```suggestion
		logger.Debugf("starting graceful shutdown (%s)", name)
```

cmd/dex/serve.go
```diff
@@ -68,6 +70,27 @@ func commandServe() *cobra.Command {
...
+		if err := srv.Shutdown(ctx); err != nil {
+			logger.Errorf("gracefully shutdown (%s): %v", name, err)
```
sagikazarmark, line 88: nit

```suggestion
			logger.Errorf("graceful shutdown (%s): %v", name, err)
```

cmd/dex/serve.go
```diff
@@ -328,34 +355,44 @@ func runServe(options serveOptions) error {
...
+		if c.GRPC.Reflection {
+			logger.Info("enabling reflection in grpc service")
+			reflection.Register(grpcSrv)
+		}
+
+		gr.Add(func() error {
+			logger.Infof("listening (grpc) on %s", c.GRPC.Addr)
+			return grpcSrv.Serve(grpcListener)
+		}, func(err error) {
+			logger.Debugf("starting gracefully shutdown (grpc)")
```
sagikazarmark, line 383: nit

```suggestion
			logger.Debugf("starting graceful shutdown (grpc)")
```

sagikazarmark (changes requested):
A couple small things, but overall 👍 Thanks!

sagikazarmark (approved):
Thanks @nabokihms !

PR #1965, opened 2021-01-28
feat: gotemplate formatting for Dex config file

nabokihms:

#### Overview
* read config file as a template and execute it
* add [sprig function](http://masterminds.github.io/sprig/) support
* provide template configuration example

An alternative of #1893
Instead of adding an external tool (dockerize), this PR introduces another approach.

```release-note
Add go template formatting support for Dex config file.
```

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -1,13 +1,18 @@
 package main
 
 import (
+	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strconv"
 	"strings"
+	"text/template"
 
+	"github.com/Masterminds/sprig/v3"
+	"github.com/ghodss/yaml"
 	"golang.org/x/crypto/bcrypt"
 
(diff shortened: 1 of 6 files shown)


sagikazarmark:
I'm **very** much against this approach. This shouldn't be Dex's responsibility. Dockerize indeed seems to be a dead end, so I see two ways ahead:

- Fork/rewrite dockerize: the community needs a project like this IMO anyway
- Try [gotemplate](https://github.com/hairyhenderson/gomplate) (suggested in https://github.com/dexidp/dex/issues/1838#issuecomment-719424256): we don't use anything other than template generation.

nabokihms:
Just an example, we can close this PR after figuring out the right way.

nabokihms:
Personally, I'm more into gomplate, than rewriting dockerize (or rewriting anything).
* arm packages
* Crypto functions
* YAML formatting function
* Solid doccumentation

sagikazarmark:
I think we can close this in favor of #1893

PR #1980, opened 2021-02-11
Refactor health checks

### LINKED ISSUE #1941: Refactor health checks

Since the health check actually relies on the state of underlying storage, it's not suitable for both liveness and readiness check in Kubernetes.

Add separate liveness and readiness health checks.

Move health checks to the telemetry server, because it's only needed internally.

Liveness check should always just return "ok", readiness check should be what the health check is at the moment.

Consider using [go-sundheit](https://github.com/AppsFlyer/go-sundheit) (beware that it imports a couple heavy/unnecessary dependencies). It's also in line with #1386 and #1397

#### Additional context

#853 points out that the current health check causes Dex to fail when the underlying storage is unavailable. Separating liveness and readiness checks would definitely be an improvement, but might not be enough to improve that situation. Although readiness checks should clearly signal when a pod is unable to serve traffic (and in case of a failing/unavailable storage that is the case), Dex is not HA at the moment (there are some GC jobs internally that should not run in parallel), so taking the single pod out of traffic would make Dex become completely unavailable.

sagikazarmark:
#### Overview

Improve health checks by separating liveness and readiness checks

Closes #1941

```release-note
NONE
```

### DIFF
--- a/storage/health.go
+++ b/storage/health.go
@@ -1,0 +1,29 @@
+package storage
+
+import (
+	"fmt"
+	"time"
+)
+
+// NewCustomHealthCheckFunc returns a new health check function.
+func NewCustomHealthCheckFunc(s Storage, now func() time.Time) func() (details interface{}, err error) {
+	return func() (details interface{}, err error) {
+		a := AuthRequest{
+			ID:       NewID(),
+			ClientID: NewID(),
+
+			// Set a short expiry so if the delete fails this will be cleaned up quickly by garbage collection.
+			Expiry: now().Add(time.Minute),
+		}
+
+		if err := s.CreateAuthRequest(a); err != nil {
+			return nil, fmt.Errorf("create auth request: %v", err)
+		}
+
+		if err := s.DeleteAuthRequest(a.ID); err != nil {
+			return nil, fmt.Errorf("delete auth request: %v", err)
+		}
+
+		return nil, nil
+	}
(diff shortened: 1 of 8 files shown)


storage/health.go
```diff
@@ -0,0 +1,29 @@
+package storage
+
+import (
+	"fmt"
+	"time"
+)
+
+// NewCustomHealthCheckFunc returns a new health check function.
+func NewCustomHealthCheckFunc(s Storage, now func() time.Time) func() (details interface{}, err error) {
```
bonifaido, line 9: So if I'm correct, the health check creates and deletes an auth request every time it hits the check endpoint?

sagikazarmark: It does. It's not optimal IMO, but this PR is about moving the health check to the telemetry endpoint.

Most storage solutions implement some sort of Ping under the hood, so we should probably use that in the future.

sagikazarmark: Sorry, not quite. The health check runs in the background with a predefined interval, the health endpoint just reports back the last result. There is no behavior change on this front.

But creating dummy auth requests is probably not a good idea. I'll try to address that in a followup PR, but I need this change for the helm chart.

bonifaido: No problem, I'm okay with this right now.

bonifaido (approved):

PR #1991, opened 2021-02-12
fix: close storage on shutdown

### LINKED ISSUE #1523: Graceful shutdown should be supported

nabokihms:
#### Overview
Call the `Close` method on storage in case of graceful shutdown

Closes #1523

```release-note
Close storage on graceful shutdown
```

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -193,6 +193,8 @@ func runServe(options serveOptions) error {
 	if err != nil {
 		return fmt.Errorf("failed to initialize storage: %v", err)
 	}
+	defer s.Close()
+
 	logger.Infof("config storage: %s", c.Storage.Type)
 
 	if len(c.StaticClients) > 0 {


cmd/dex/serve.go
```diff
@@ -193,6 +202,8 @@ func runServe(options serveOptions) error {
 	if err != nil {
 		return fmt.Errorf("failed to initialize storage: %v", err)
 	}
+	defer closeStorage(logger, s)
```
sagikazarmark, line 205: I kinda think `defer s.Close()`  should be enough for now. Failure to close a storage is not actionable, we can't do anything with the fact that it failed.

nabokihms: Oh, ok. I will simplify it.

Incorrectly closing storage can affect the database, e.g., the connection may be left open. I thought this message can help with debugging problems on the storage side.

sagikazarmark (approved):

PR #1997, opened 2021-02-15
Rewrite LDAP tests to use a single server instance

sagikazarmark:
Signed-off-by: Mark Sagi-Kazar <mark.sagikazar@gmail.com>

#### Overview

Rewrite LDAP tests to rely on a single server instance

#### What this PR does / why we need it

With this change, LDAP tests won't launch containers to run integration tests against.
Instead, the ldap server running in the docker compose setup can be used.

#### Special notes for your reviewer

LDAP definitions changed substantially.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #2009, opened 2021-02-20
fix: do not run LDAP tests locally by default

nabokihms:

#### Overview
Explicitly launch LDAP tests

After merging #1997, it became impossible to use `make test` or `go test ./...` locally without a mock LDAP server. 

```release-note
NONE
```


sagikazarmark:
I'd rather keep this closer to the storage conformance tests. I think in those tests we skip if the host is not defined: https://github.com/dexidp/dex/blob/master/storage/sql/config_test.go#L220

nabokihms:
Ok, good for me. I will use it.

sagikazarmark:
Maybe in a separate PR, we should add a shell script/env file/whatever so that developers can easily run the conformance tests against the docker-compose environment without having to manually specify env vars.

sagikazarmark (approved):

Issue #1953, opened 2021-01-20
device auth flow token endpoint is not discoverable


justenwalker:
#### Is your feature request related to a problem?

#1706 implements device auth flow, which is a great feature to have. However, the token endpoint is not discoverable.

In order to exchange a `device_code` to get the authorization tokens, you have to know out-of-band to contact the `/device/token` endpoint. 

#### Describe the solution you'd like to see

It would be helpful to instead enable the `/token` endpoint as indicated by the `token_endpoint` key in the `/.well-known/openid-configuration` to handle device code authorization as well. 

In order to keep backwards compatibility, the `/device/token` endpoint can still be present, but the `/token` endpoint should detect the `grant_type = urn:ietf:params:oauth:grant-type:device_code` parameter and delegate to that same handler.

#### Describe alternatives you've considered

There isn't really an alternative beside some out-of-band way of knowing you are communicating with Dex as opposed to another OpenID connect provider.

nabokihms:
Hello! Thank you for this issue. I personally like it!
There is a link to the corresponding RFC entry - https://tools.ietf.org/html/rfc6749#section-3.2

PR #2010, opened 2021-02-20
fix: use /token endpoint to get tokens with device flow

nabokihms:

#### Overview
[rfc6749#section-3.2](https://tools.ietf.org/html/rfc6749#section-3.2)
* Make `/token` endpoint responding to requests with device auth grant
* Allow only POST requests `/token` endpoint
* Do not search for a client in storage if an invalid grant type is provided

Fixes https://github.com/dexidp/dex/issues/1953 
As was mentioned in the issue above, the main problem that the token endpoint is not discoverable.

This MR leaves the possibility to get a token from the `/device/token` endpoint. 

There are no breaking changes, but it will be better to avoid using /device/token endpoint in the feature.

```release-note
Make `/token` endpoint responding to requests with device auth grant. The `/device/token` endpoint becomes DEPRECATED.
```

### DIFF
--- a/server/deviceflowhandlers.go
+++ b/server/deviceflowhandlers.go
@@ -152,6 +152,9 @@ func (s *Server) handleDeviceCode(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *Server) handleDeviceTokenGrant(w http.ResponseWriter, r *http.Request) {
+	s.logger.Warn(`Request to the deprecated "/device/token" endpoint was received.`)
+	s.logger.Warn(`The "/device/token" endpoint will be removed in a future release.`)
+
 	w.Header().Set("Content-Type", "application/json")
 	switch r.Method {
 	case http.MethodPost:
--- a/server/deviceflowhandlers.go
+++ b/server/deviceflowhandlers.go
@@ -151,9 +151,8 @@ func (s *Server) handleDeviceCode(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (s *Server) handleDeviceTokenGrant(w http.ResponseWriter, r *http.Request) {
-	s.logger.Warn(`Request to the deprecated "/device/token" endpoint was received.`)
-	s.logger.Warn(`The "/device/token" endpoint will be removed in a future release.`)
+func (s *Server) handleDeviceTokenDeprecated(w http.ResponseWriter, r *http.Request) {
+	s.logger.Warn(`The deprecated "/device/token" endpoint was called. It will be removed, use "/token" instead.`)
 
 	w.Header().Set("Content-Type", "application/json")
 	switch r.Method {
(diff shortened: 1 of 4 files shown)


sagikazarmark:
Should we deprecate the `/device/token` endpoint?

nabokihms:
Yes, I think so. I left a to-do but did nothing with deprecation. Maybe adding some sort of warning on endpoint call will be good enough.
"Some of Dex clients called the deprecated /device/token endpoint. It will be removed in a future release."

Something like this.

sagikazarmark:
For me, deprecation would mean

- a comment in the code somewhere
- potentially a warning log message
- a note in the release notes

server/deviceflowhandlers.go
```diff
@@ -151,7 +151,10 @@ func (s *Server) handleDeviceCode(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (s *Server) handleDeviceToken(w http.ResponseWriter, r *http.Request) {
+func (s *Server) handleDeviceTokenGrant(w http.ResponseWriter, r *http.Request) {
+	s.logger.Warn(`Request to the deprecated "/device/token" endpoint was received.`)
+	s.logger.Warn(`The "/device/token" endpoint will be removed in a future release.`)
```
sagikazarmark, line 156: Can we make this a single log line? And maybe make the message a bit more concise?

server/server_test.go
```diff
@@ -1583,7 +1583,7 @@ func TestOAuth2DeviceFlow(t *testing.T) {
 
 			// Hit the Token Endpoint, and try and get an access token
 			tokenURL, _ := url.Parse(issuer.String())
-			tokenURL.Path = path.Join(tokenURL.Path, "/device/token")
+			tokenURL.Path = path.Join(tokenURL.Path, "/token")
```
sagikazarmark, line 1586: Can we add a test with the old path to make sure it's still functional?

server/deviceflowhandlers.go
```diff
@@ -151,7 +151,10 @@ func (s *Server) handleDeviceCode(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (s *Server) handleDeviceToken(w http.ResponseWriter, r *http.Request) {
+func (s *Server) handleDeviceTokenGrant(w http.ResponseWriter, r *http.Request) {
```
sagikazarmark, line 154: I think this name is a bit confusing and ambiguous with the `handleDeviceToken` name. Maybe call it something like `deprecatedDeviceTokenHandler` or `handleDeviceTokenDeprecated`.

sagikazarmark (changes requested):

sagikazarmark (approved):
LGTM, thanks

justenwalker:
@sagikazarmark Anything preventing this from being merged/closed?

PR #2018, opened 2021-02-25
Fix HTTPS server listener regression

nabokihms:

#### Overview
* New abstraction to run and shutdown HTTP server gracefully
* Add optional support for the  TLS encryption

Fix of the regression of https://github.com/dexidp/dex/pull/1963
The HTTPS server of dex built from the master branch listens on a port without TLS encryption.

It is the fix for the previous PR with a new feature, so we should not add anything new to the release message.

```release-note
NONE
```

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -73,22 +73,49 @@ func commandServe() *cobra.Command {
 	return cmd
 }
 
-func listenAndShutdownGracefully(logger log.Logger, gr *run.Group, srv *http.Server, name string) error {
-	l, err := net.Listen("tcp", srv.Addr)
+type serverRunner struct {
+	name string
+	srv  *http.Server
+
+	tlsCrt string
+	tlsKey string
+
+	logger log.Logger
+}
+
+func newServerRunner(name string, srv *http.Server, logger log.Logger) *serverRunner {
+	return &serverRunner{name: name, srv: srv, logger: logger}
+}
+
+func (s *serverRunner) WithTLS(crt, key string) *serverRunner {
+	s.tlsCrt = crt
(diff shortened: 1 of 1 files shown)


cmd/dex/serve.go
```diff
@@ -73,22 +73,49 @@ func commandServe() *cobra.Command {
 	return cmd
 }
 
-func listenAndShutdownGracefully(logger log.Logger, gr *run.Group, srv *http.Server, name string) error {
-	l, err := net.Listen("tcp", srv.Addr)
+type serverRunner struct {
```
sagikazarmark, line 76: TBH I'm not sure I like this structure. It complicates the code IMHO, decreases visibility while it only spares a few lines of code.

Here is how I usually do it: https://github.com/sagikazarmark/modern-go-application/blob/master/cmd/modern-go-application/main.go#L194-L214

It does come with some duplications, but keeps the code more readable in my opinion.

WDYT?

sagikazarmark (commented):
I don't fully agree with this way. Also, I don't see the actual bug what caused the HTTPS server to stop working. Is it because we always called `Serve` and never `ServeTLS`?

cmd/dex/serve.go
nabokihms: 1. About the bug: Yes, it is caused by not wrapping the listener with TLS. `ServeTLS` solves the problem.
2. Ok, I will try to simplify it. My major intention was to hide some common logic inside a black-box, e.g., logging, listener creation. However, I have no objection to try other approaches.

sagikazarmark: See #2019

sagikazarmark:
Fixed in #2019

Issue #1999, opened 2021-02-15
Update Kubernetes storage tests dependencies


nabokihms:
#### Is your feature request related to a problem?
It looks like dex is tested with an outdated version of the Kubernetes API server.
```bash
curl -L https://storage.googleapis.com/k8s-c10s-test-binaries/kube-apiserver-Linux-x86_64 > bin/test/kube-apiserver
chmod +x bin/test/kube-apiserver
```
```bash
bin/test/kube-apiserver --version
> Kubernetes v1.10.0-alpha.1
```
#### Describe the solution you'd like to see
This is not required right now, but we need to figure out how to update Kubernetes storage test dependencies. I think it is ok to switch tests to run on a local minimal Kubernetes installation, e.g., kind, minikube.
Another possible option is rewriting the Kubernetes storage layer with the client-go Kubernetes library and using its fakeclient for mocking.

#### Additional context
Slightly related to https://github.com/dexidp/dex/issues/1895

PR #2025, opened 2021-03-02
feat: create CRDs as apiextensions.k8s.io/v1

### LINKED ISSUE #1895: Migrate Kubernetes storage CRDs API version to apiextensions.k8s.io/v1
Starting from Kubernetes 1.22, API version `apiextensions.k8s.io/v1beta1` [will no longer be available](https://github.com/kubernetes/kubernetes/issues/82022).

Version `apiextensions.k8s.io/v1` is only available in Kubernetes >=1.16. We need some kind of mechanism to detect the Kubernetes version and deploy crds based on it.

Migrate to `apiextensions.k8s.io/v1` with losing the support of Kubernetes clusters 1.16>= 

It's not a big deal, though. Users with outdated Kubernetes clusters can turn off automatic crds creation and deploy them manually.

#### Additional context
There is a [link to the doc](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions) that contains some nuances about migration.

nabokihms:

#### Overview
* add Kubernetes version discovery mechanism
* create crds as apiextensions.k8s.io/v1 for kubernetes clusters >= 1.16

Fixes #1895

There is no test that proves crds creation in kubernetes clusters >= 1.10 works (related to #1999).

It is not a breaking change.
```release-note
kubernetes storage: create custom resource definitions as apiextensions.k8s.io/v1
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.16
 require (
 	entgo.io/ent v0.8.0
 	github.com/AppsFlyer/go-sundheit v0.4.0
+	github.com/Masterminds/semver v1.5.0
 	github.com/Masterminds/sprig/v3 v3.2.2
 	github.com/beevik/etree v1.1.0
 	github.com/coreos/go-oidc/v3 v3.0.0
--- a/go.sum
+++ b/go.sum
@@ -51,6 +51,8 @@ github.com/DATA-DOG/go-sqlmock v1.5.0/go.mod h1:f/Ixk793poVmq4qj/V1dPUg2JEAKC73Q
 github.com/Knetic/govaluate v3.0.1-0.20171022003610-9aa49832a739+incompatible/go.mod h1:r7JcOSlj0wfOMncg0iLm8Leh48TZaKVeNIfJntJ2wa0=
 github.com/Masterminds/goutils v1.1.1 h1:5nUrii3FMTL5diU80unEVvNevw1nH4+ZV4DSLVJLSYI=
 github.com/Masterminds/goutils v1.1.1/go.mod h1:8cTjp+g8YejhMuvIA5y2vz3BpJxksy863GQaJW2MFNU=
+github.com/Masterminds/semver v1.5.0 h1:H65muMkzWKEuNDnfl9d70GUjFniHKHRbFPGBuZ3QEww=
+github.com/Masterminds/semver v1.5.0/go.mod h1:MB6lktGJrhw8PrUyiEoblNEGEQ+RzHPF078ddwwvV3Y=
 github.com/Masterminds/semver/v3 v3.1.1 h1:hLg3sBzpNErnxhQtUy/mmLR2I9foDujNK030IGemrRc=
 github.com/Masterminds/semver/v3 v3.1.1/go.mod h1:VPu/7SZ7ePZ3QOrcuXROw5FAcLl4a0cBrbBpGY/8hQs=
 github.com/Masterminds/sprig/v3 v3.2.2 h1:17jRggJu518dr3QaafizSXOjKYp94wKfABxUmyxvxX8=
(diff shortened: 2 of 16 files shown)


sagikazarmark:
Needs rebase.

I think we don't have to keep supporting Kubernetes 1.10. Supporting the officially supported Kubernetes versions, minus a couple versions as we see fit should be more than enough in my opinion.

I would also consider removing the CRD installation from Dex (or at least make it optional). CRDs are better installed by the user or the Helm chart IMHO.

These aren't blockers of this PR, but I think it's a good conversation starter.

My only concern about this PR is that we need to make sure, that we have an upgrade path from older CRD versions. If that's a `kubectl apply -f` command, that's perfectly fine, but we need that documented somewhere.

nabokihms:
Actually, we do not need to upgrade previously created CRDs. Kubernetes does it for us. If CRDs were deployed earlier, it is fine. We will be able to access them using old and new API versions (one apiVersion is called storage version, the others are called served versions). 
We will face the problem only in Kubernetes >=1.22 because the API version we use to create them will be removed.

I like the idea to add the flag or option to prevent CRDs creation by Dex, as it is in other Kubernetes operators/controllers. In Flant, we also want to manage CRDs ourselves.

nabokihms:
Kubernetes 1.22 is scheduled for Aug 4
> [Wednesday, August 4th: Week 15 - Kubernetes v1.22.0 released](https://www.kubernetes.dev/resources/release/)

sagikazarmark:
Let's schedule 2.30 for July. I'd like to get 2.29 released ASAP.

omesser:
@sagikazarmark  - Hey there,
Any estimated july release date for the 2.30 with this change?
The CRD incompatibility is of course a showstopper for using newer k8s versions (>=1.19)

sagikazarmark:
@omesser there is no scheduled date yet.

> The CRD incompatibility is of course a showstopper for using newer k8s versions (>=1.19)

Is it? I thought it will only get removed in 1.22

@nabokihms I think this is good to go. A couple questions:

> It is not a breaking change.

Are you sure? I think this will make Dex uninstallable on k8s <1.16. If that's the case, I think it would make sense to at least document that fact.

How are upgrades going to work?

nabokihms:
1. I was choosing between making a breaking change or saving compatibility. Eventually, I took the second variant. For Kubernetes clusters <1.16, Dex will deploy CRDs as it does right now. For higher versions, Dex will use the stable API. 

2. About upgrade process:
* Dex will detect the suitable API based on the Kubernetes version to work with CRDs.
* Dex checks CRDs in the cluster with detected API version.
NOTE: If CRDs were created through the old API and we want to get them through the new API, Kubernetes will convert them to the desired version on-flight and respond with converted objects.
* If CRDs are available, Dex does nothing.
* If there are no CRDs, Dex will create them.

Therefore no upgrade is required, and no breaking changes are provided!

sagikazarmark (approved):
Perfect! I looked at the implementation and it looks good to me!

omesser:
@sagikazarmark  - You're right, apologies, I've seen some contradicting info there, there's this from 1.16 stating it would be removed by 1.19, for example (and a lot of other products took it that way):
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#deprecations-and-removals
But later release notes support that v1beta is still served by 1.19 (while it's *deprecated* it should still be served), and will only be completely removed in 1.22:
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122
https://github.com/kubernetes/kubernetes/pull/90673
so, yeah.
I'm gonna test to see if it's indeed ok on k8s 1.19, and this will make this not as immediate 🔥

PR #2026, opened 2021-03-02
chore: warning about deprecated LDAP groupSearch fields

nabokihms:

#### Overview
* print warning message if deprecated fields are used
* calculate user matches only on the connector opening

We cannot remove deprecated fields without letting users know about our intentions.

Users will see a warning message in dex logs if they have deprecated fields in their LDAP connector config.

```release-note
ldap: deprecation warning about `groupSearch.userAttr` and `groupSearch.groupAttr`
```

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -192,7 +192,7 @@ func userMatchers(c *Config, logger log.Logger) []UserMatcher {
 		return c.GroupSearch.UserMatchers
 	}
 
-	logger.Warn(`ldap: fields groupSearch userAttr/groupAttr are deprecated, use groupSearch.userMatchers instead.`)
+	log.Deprecated(logger, `LDAP: use groupSearch.userMatchers option instead of "userAttr/groupAttr" fields.`)
 	return []UserMatcher{
 		{
 			UserAttr:  c.GroupSearch.UserAttr,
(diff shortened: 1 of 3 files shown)


connector/ldap/ldap.go
```diff
@@ -187,11 +187,12 @@ func parseScope(s string) (int, bool) {
 // Function exists here to allow backward compatibility between old and new
 // group to user matching implementations.
 // See "Config.GroupSearch.UserMatchers" comments for the details
-func (c *ldapConnector) userMatchers() []UserMatcher {
+func userMatchers(c *Config, logger log.Logger) []UserMatcher {
 	if len(c.GroupSearch.UserMatchers) > 0 && c.GroupSearch.UserMatchers[0].UserAttr != "" {
 		return c.GroupSearch.UserMatchers
 	}
 
+	logger.Warn(`ldap: fields groupSearch userAttr/groupAttr are deprecated, use groupSearch.userMatchers instead.`)
```
sagikazarmark, line 195: I think we should come up with a standard format for deprecation warnings. It would make searching for deprecations in the code much more easier and would make filtering easier for consumers.

WDYT?

nabokihms: Good point. I will think about it. Maybe we can even add a function to format deprecation messages.

sagikazarmark (approved):

PR #2034, opened 2021-03-10
feat: move integration tests to docker-compose

nabokihms:

#### Overview
* Provide script to run e2e tests locally
* Move services from github-actions to docker-compose

Users and developers have no opportunity to run integration tests locally. CI system is not always an option for external PRs.

This PR aims to help developers testing their features locally. 
```release-note

```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,38 +13,6 @@ jobs:
     env:
       GOFLAGS: -mod=readonly
 
-    services:
-      postgres:
-        image: postgres:10.8
-        ports:
-          - 5432
-        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
-
-      mysql:
-        image: mysql:5.7
-        env:
-          MYSQL_ROOT_PASSWORD: root
-          MYSQL_DATABASE: dex
-        ports:
-          - 3306:3306
-        options: --health-cmd "mysql -proot -e \"show databases;\"" --health-interval 10s --health-timeout 5s --health-retries 5
-
-      etcd:
-        image: gcr.io/etcd-development/etcd:v3.2.9
-        ports:
-          - 2379
-        env:
-          ETCD_LISTEN_CLIENT_URLS: http://0.0.0.0:2379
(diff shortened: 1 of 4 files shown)


sagikazarmark:
Let's hold off on this one for a little.

I agree that we need to improve local environments, I'm not 100% sure that replacing the CI config with docker compose is a good idea.

I'd like to explore a few other options first.

nabokihms:
Ok, no problem. I thought that this is a kind of annoying issue which is worth to be fixed.

Some points to move services to docker-compose:
1. it's already written in the docker-compose format
2. docker-compose is a common tool for most developers
3. shell script makes it possible to run tests only for certain storages or connectors

I would like to have something like this as a dev environment.

sagikazarmark:
I agree with the sentiment, less so with the implementation:

- using GHA's "native" container integration sounds "safer" to me (if that makes sense). I'd probably raise an issue somewhere to support docker-compose for services (although I'm not sure if that's feasible since the checkout is part of the build process)
- can we use defaults in tests instead of requiring to set the env vars locally? Then instead of using env var for deciding whether integration tests should be ran, we can use this pattern: https://github.com/banzaicloud/pipeline/blob/da1f66e9e62d5016c6f78dc62a8fd21b80787d99/internal/helm/integration_test.go#L49

The second point should simplify things a lot in my opinion. You can just call `go test -run ^TestIntegration$ ./storage/whatever` and it'll work.

WDYT?

nabokihms:
I'm certainly ok with the second part about flag/flags. It is a lot better than `export DEX_IS_RUN_TEST_FOR_XXX=1`. I worry about the first part, which is connected to run dependencies.

1. I did not manage to find any information about which is better to use GH services or docker-compose other than with docker-compose you can mount dependencies from the repo or even build your docker image before tests.
2. Docker-compose is convenient in some cases, but we need to think about Kubernetes tests dependencies as well https://github.com/dexidp/dex/issues/1999
3. Supporting GitHub services and something else may be overhead.

So, I think flags allow us to exclude shell-script, but how do I know which dependence I need to run for the integration test?

Issue #2040, opened 2021-03-15
Carry connectorData state between LoginURL() and HandleCallback()


candlerb:
#### Is your feature request related to a problem?
When making a Vault connector (#2039), I found that I needed to maintain some state between the calls to `LoginURL()` and `HandleCallback()` for a given authentication request.

#### Describe the solution you'd like to see
The `AuthReq` connector already carries a `ConnectorData []byte`.  I would like to be able to return this from `LoginURL()`, and get the current value passed to `HandleCallback()` for the given auth request.

#### Describe alternatives you've considered
As a workaround I used a `sync.Map` to store the necessary state, but this is only local to a single process so doesn't work when Dex is scaled across multiple frontends.

#### Additional context
N/A

nabokihms:
Fixed in https://github.com/dexidp/dex/releases/tag/v2.45.0

PR #2041, opened 2021-03-15
Return connector state from LoginURL() and pass it to HandleCallback()

candlerb:
#### Overview

Additional return value and argument in connector API:
* `LoginURL()` returns `connectorData []byte`
* `HandleCallback()` receives `connectorData []byte`

When writing a connector for Vault (#2039), I found that it was necessary to maintain state between the `LoginURL()` and `HandleCallback()` calls.  This is because Vault provides its own OIDC callback `state` which must be carried forward.

Dex already maintains `ConnectorData []byte` in the authReq structure, but currently does not make this available to `LoginURL()` and `HandleCallback()`

Closes: #2040

Minimal changes made to existing connectors to return a nil slice from `LoginURL()`, and ignore the slice passed in to `HandleCallback()`

This change would likely be exposed as an optional field for external connectors, when they are implemented.

```release-note
NONE
```

### DIFF
--- a/connector/authproxy/authproxy.go
+++ b/connector/authproxy/authproxy.go
@@ -37,20 +37,20 @@ type callback struct {
 }
 
 // LoginURL returns the URL to redirect the user to login with.
-func (m *callback) LoginURL(s connector.Scopes, callbackURL, state string) (string, error) {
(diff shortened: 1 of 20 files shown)


nabokihms:
Fixed by https://github.com/dexidp/dex/pull/3777

Issue #918, opened 2017-04-21
Groups returned in token should be prefixed


foobarto:
Following example is GitHub specific but I believe it applies to all connectors.

Let's consider usecase where Dex is used with GitHub to provide authentication for Kubernetes 1.6+.

Currently the `groups` value in the token returned to the client contains list of GitHub teams for specified organization. This allows anyone with high enough permissions on github org to create a team called `system:masters` (or similar) which would then grant anyone in that github team cluster administrator privileges. There's a lot of groups that would grant sensitive access to a Kubernetes cluster like that.

Proposed solution would be to prefix the values returned in `groups` inside the token with connector name and possibly with connector specific value (ie. in case of GitHub it would be the org name).

For example:
```
"groups": ["github:myorg:myteam"]
```

vs what is returned currently:
```
"groups": ["myteam"]
```

rithujohn191:
@foobarto Thank you for your suggestion. This would be a great value add for all the connectors we have. Here are some initial thoughts about implementing this feature:
* This feature is something we wont be able to turn on my default since we have no way of making it backward compatible.
* We can probably add a flag that the user can specify which can enable this.

candlerb:
Note for the given use case: kubernetes now [supports](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server) a flag `--oidc-groups-prefix` which applies a prefix to the group names from the IDP.

However I still think it would be good to qualify the group names with the IDP name.  Upstreams may have overlapping group names which you might not want to handle in the same way - e.g. you might treat `upstream1:admins` differently to `upstream2:admins`.

candlerb:
BTW, isn't it also possible that two different IDPs both return `"sub": "123456"` ?  If so, being able to prefix the `sub` (as well as `groups`) could be worthwhile.

puja108:
We just realized that we might run into this issue as we have an upstream on which we have a group that has very high privileges (within K8s) and other upstreams should never be able to get those. However, if they guess our high privilege group claim they could theoretically create a group in their IDP to escalate their privileges to that.

I could also imagine in multi-tenant environments where tenants are using different upstreams, the case of overlapping group names like @candlerb mentioned above being a big issue.

I think a solution would need to be per upstream as it needs to set different prefixes otherwise the problem persists (like when setting prefix in k8s itself), maybe could be included in configs, but I fear that means touching every connector?

[...]

PR #2051, opened 2021-03-19
Allow using connector ID as oidc groups prefix

corest:
#### Overview

Adds prefixes for OIDC groups. I'm not sure if we want to implement this on a different level (middleware?).
But I've been using this for some time and it works fine for my use-case, which is mostly about the security side, not UX.
E.g. it is not possible to configure the exact OIDC prefix with this, but as connector IDs are unique - it makes OIDC groups unique as well. I'm open for discussion if this approach is not how the community would like to solve this. For me, it works and closes the issue https://github.com/dexidp/dex/issues/918, therefore, sharing this solution here.

Usage: "Closes https://github.com/dexidp/dex/issues/918"

```release-note
This change allows enabling OIDC groups prefixes with connector IDs values. Having this makes it impossible for groups from different connectors to impersonate other group access by using the same group names.
```

This feature can be enabled with `oidcGroupsPrefix: true`. If this option is missing, it will be disabled by default. So it doesn't require any changes if you're updating from the old version

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -32,6 +32,9 @@ type Config struct {
 
 	Frontend server.WebConfig `json:"frontend"`
 
+	// If enabled, add connector ID as prefix for groups from auth response.
+	OIDCGroupsPrefix bool `json:"oidcGroupsPrefix,omitempty"`
+
 	// StaticConnectors are user defined connectors specified in the ConfigMap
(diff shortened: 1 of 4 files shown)


sagikazarmark:
Thanks @corest!

We might put a hold on it in favor of #1635, but it's definitely a good use case.

corest:
Ok then. Closing

PR #2057, opened 2021-03-22
Upgrade protobuf in internal codec

sagikazarmark:
#### Overview

Upgrade protobuf in an internally used package.

It's the first step in upgrading protobuf and gRPC to the latest version.

In theory, this codec is only used internally (ie. nothing decodes the protobuf message outside of Dex), so it should be fully backward compatible.

```release-note
NONE
```


justaugustus (approved):
LGTM!

PR #2084, opened 2021-04-13
To enable the anonymous ldap bind

YasuhiroABE:
#### Overview

The latest code causes the error when we use the anonymous bind of the ldap connector.
It seems as of the switch to the go-ldap/ldap/v3 library.

This PR enables the anonymous bind when the "anonymousBind" parameter is set to true in the configuration file.

I checked the current issues, but nobody has an issue against the anonymous bind.
However, I'm sure it is the incompatibility against the past release.

```release-note
The "anonymousBind" ldap configuration parameter was introduced to turn on the anonymous bind. 
```


sagikazarmark (approved):
Looks good to me.

I'd like to hear some other opinions about the config name though.

/cc @nabokihms @JoelSpeed @bonifaido @justaugustus

connector/ldap/ldap.go
```diff
@@ -330,12 +334,17 @@ func (c *ldapConnector) do(_ context.Context, f func(c *ldap.Conn) error) error
 	}
 	defer conn.Close()
 
-	// If bindDN and bindPW are empty this will default to an anonymous bind.
-	if err := conn.Bind(c.BindDN, c.BindPW); err != nil {
-		if c.BindDN == "" && c.BindPW == "" {
+	if c.AnonymousBind {
```
nabokihms, line 337: Do we need to add a new option here? I mean, according to the commentary 
> // If bindDN and bindPW are empty this will default to an anonymous bind.

Dex should use an unauthenticated bind if the `binddn` and `bindpw` options are missing.

Maybe it is good to leave things that way and only change a clause?
```go
if c.BindDN == "" && c.BindPW == "" {
  if err := conn.UnauthenticatedBind(""); err != nil {			
    return fmt.Errorf("ldap: initial anonymous bind failed: %v", err)		
  }
}
...
```

About naming, both `UnauthenticatedBind` and `AnonymousBind` look fine to me.

sagikazarmark: Hm, makes sense to me.

sagikazarmark:
@YasuhiroABE can you please take a look at the comments?

PR #2087, opened 2021-04-16
fix: follow semantic import versioning

nabokihms:

#### Overview
Add a `/v2` suffix to the Dex module path.

All go projects SHOULD follow [semantic import versioning](https://github.com/golang/go/wiki/Modules#semantic-import-versioning).

The motivation to open this PR is to:
* make `goproxy` be able to detect new versions of dex
* fix `go get` command
* renew info on [pkg.go.dev/github.com/dexidp/dex](https://pkg.go.dev/github.com/dexidp/dex)

```release-note
Switch to `semantic import versioning` by adding `/v2` suffix to the Dex module path.
```


sagikazarmark:
I've considered this before, but I have a number of arguments against it:

- We need a v3 to introduce the new path, otherwise it's a breaking change leading to all kinds of issues. Dex is actually imported by some projects (sadly) and this will surely break those instances.
- Since this is an application, there aren't any benefits from implementing semantic import versioning. It's not supposed to be imported (despite the first point), so there is no point in embracing semantic import versioning. (As a result, pkg.go.dev and go proxy support are not real benefits IMHO)

We should probably consider this in v3 instead, but even there I'm not sure it's necessary. Sadly, there is no official recommendation for applications.

**tl;dr** 👎 from me, at least for v2.

nabokihms:
Good explanation, thanks. I thought about same obstacles.
I would like to do two things then:
1. Open an issue (or maybe we already have one?)
2. Remove the link to pkg.go.dev from the Readme

sagikazarmark:
Sounds good to me!

nabokihms:
We will consider the subject of this PR later. Closing it for now.

drauschenbach:
I would vote to reopen this issue. My renovate CI automation keeps trying to get me to "upgrade" to Dex 0.6.1 (Sep 2016???). And I have to import the Dex package using a Git hash instead of a tag, which is a bookkeeping burden to know what version I am on.

nabokihms:
Could you please write it to the linked issue? We opened it to track the progress and collect suggestions, proposals, also votes.

It seems that this particular PR does not worth being reopened, yet we still may consider implementing this feature.

https://github.com/dexidp/dex/issues/2088

PR #2091, opened 2021-04-19
chore: test Kubernetes storage with KinD

nabokihms:

#### Overview

This is a POC of switching from the old Kubernetes testing framework to [KinD](https://kind.sigs.k8s.io/).

Closes https://github.com/dexidp/dex/issues/1999

```release-note

```


sagikazarmark:
I definitely like the way this is going.

Question: how can we run the tests locally?
I mean can we get some instructions or scripts somewhere?
Eg. `make up` create kind cluster and export the kube config in a local directory, `make down` kill the cluster?

In general, I'd like to improve development and testing documentation. This sounds like a good first step.

nabokihms:
To run tests locally, we need to follow these steps (what I did to test this feature):
1. Run kind with `kind create`.
2. Export the env variable with a kubeconfig path
3. Run tests

Ok, put it all in a script seems reasonable. I will also add an entry to the test documentation.

nabokihms:
I have added three  instructions to the make file:
```
kind-up - up kind cluster and save kubeconfig to tmp (to avoid user's kubeconig modifications)
kind-tests - run tests with previously saved kubeconfig
kind-down - delete kind cluster and kubeconfig
```

Makefile
```diff
@@ -26,10 +26,10 @@ PROTOC_VERSION = 3.15.6
 PROTOC_GEN_GO_VERSION = 1.26.0
 PROTOC_GEN_GO_GRPC_VERSION = 1.1.0
 
-generate:
-	@go generate $(REPO_PATH)/storage/ent/
+KIND_NODE_IMAGE = "kindest/node:v1.19.7@sha256:a70639454e97a4b733f9d9b67e12c01f6b0297449d5b9cbbef87473458e26dca"
+KIND_TMP_DIR = "/tmp/dex-kind-kubeconfig"
```
sagikazarmark, line 30: Can we make this a project local tmp dir instead?

.github/workflows/ci.yaml
```diff
@@ -57,6 +57,12 @@ jobs:
       - name: Start services
         run: docker-compose -f docker-compose.test.yaml up -d
 
+      - name: Create kind cluster
+        uses: helm/kind-action@v1.0.0
```
sagikazarmark, line 61: There is a new action version

.github/workflows/ci.yaml
```diff
@@ -57,6 +57,12 @@ jobs:
...
+        with:
+          version: v0.10.0
```
sagikazarmark, line 63: There is a new kind version

.github/workflows/ci.yaml
```diff
@@ -57,6 +57,12 @@ jobs:
...
+          node_image: kindest/node:v1.19.7@sha256:a70639454e97a4b733f9d9b67e12c01f6b0297449d5b9cbbef87473458e26dca
```
sagikazarmark, line 64: It's probably fine for now, but we should probably move kube test running to a separate job and use a test matrix to run tests on multiple kube versions. Can you open an issue for that?

Also, see the helm-chart repo for an example.

sagikazarmark (changes requested):
Couple minor things, but overall 👍

.github/workflows/ci.yaml
nabokihms: Thanks! Yeah, that is a good idea. I thought that we could start with something simple.

Makefile
nabokihms: Previously, we used `ioutil.TempFile("", "dex-kubeconfig-*")` to generate a tmp file for the config, which was calling `os.TempDir()` undercover. Although, it makes sense to me to move it closer. I will do it.

sagikazarmark (approved):
LGTM

PR #2095, opened 2021-04-23
Vault signer

### LINKED ISSUE #1879: Allow using Hashicorp Vault to sign JWTs/JWKs
**Is your feature request related to a problem? Please describe.**
It would be great if dex didn't have to store signing keys in a database in plaintext.

**Describe the solution you'd like to see**
Allowing to use Hashicorp Vault as a signing mechanism would improve this from a security perspective. As vault never gives out the signing key itself. Dex would issue a request to Vault to sign a payload and Vault would return the signed payload. Similarly for verifying the signature.

**Describe alternatives you've considered**
The current approach.

**Additional context**
Primarily opening this to see if I understand the possibility for this correctly. I don't currently have time to implement this, but I may in the future. If anyone is keen on implementing this now though, don't hesitate :slightly_smiling_face: .

whs:
#### Overview

Allow using Vault to sign JWT

closes #1879

1. Split signer into an interface
2. Make the existing storage drivers a signer
3. Make a new signer driver for Vault

```release-note
Dex now support signing JWTs using key stored in Vault. See examples/config-dev.yaml in the signer section for setup.
```


sagikazarmark:
Thank you @whs for your contribution!

It looks great, but it's also massive in size, so it might take some time to properly review it.

signer/vault/config.go
```diff
@@ -0,0 +1,36 @@
...
+type Config struct {
+	Address string `json:"address" yaml:"address"`
+	// TransitMount is the path to transit storage engine
+	TransitMount string `json:"mount" yaml:"mount"`
+	KeyName      string `json:"key" yaml:"key"`
+}
+
+func (c Config) Open(logger log.Logger, tokenValid time.Duration, rotationPeriod time.Duration) (signer.Signer, error) {
+	client, err := vault.NewClient(&vault.Config{
+		AgentAddress: c.Address,
```
bonifaido, line 21: This means that using this feature requires Vault Agent, right?

whs: It's not actually requiring Vault agent - both Vault and Vault agent has the same API. The automated test also run on Vault directly.

However, there's no authentication support in this PR - at best you can supply VAULT_TOKEN which Vault SDK read from. Therefore for most users I'd imagine they'd need to use Vault agent to provide an authenticating proxy.

I'm not sure why upstream use this parameter name instead of just Vault server, but using it would affirm the intention that authentication won't be supported in Dex.

bonifaido: Thanks for the explanation. There are some Vault clients that can auto-auth to Vault based on some underlying auth mechanism like AWS/GCP IAM, Kubernetes Service accounts like: https://github.com/banzaicloud/bank-vaults/blob/master/pkg/sdk/vault/client.go#L177

Dentrax:
Any progress on this? @whs :eyes:

whs:
I probably won't work on this PR any more as we moved our primary signer to a SaaS solution. Feel free to adopt this PR

whs:
Replaced by #4512

Issue #1944, opened 2021-01-15
Relocate the version package to cmd


sagikazarmark:
#### Is your feature request related to a problem?
Having a separate version package might encourage developers to use it and creates potentially unwanted dependencies between different parts of the application.

#### Describe the solution you'd like to see
Move the version functionality to the main package and inject the version wherever necessary.

#### Additional context
Apart from the version command, only the API uses it directly.

PR #2107, opened 2021-05-04
Discard package "version"

nabokihms:

#### Overview
* Delete "version" package
* Inject version variable in two places

Closes https://github.com/dexidp/dex/issues/1944

I thought about this issue. It seems that injecting the version twice  (during the build stage) as good as injecting it once and then importing it two times. There is no possibility to import the version anymore, though (which was our goal).

```release-note
Discard package "version."
```

### DIFF
--- a/cmd/dex/version.go
+++ b/cmd/dex/version.go
@@ -5,17 +5,9 @@ import (
 	"runtime"
 
 	"github.com/spf13/cobra"
-
-	"github.com/dexidp/dex/server"
 )
 
 var version = "DEV"
-
-// nolint:gochecknoinits
-func init() {
-	// inject version for API endpoint
-	server.Version = version
-}
 
 func commandVersion() *cobra.Command {
 	return &cobra.Command{
--- a/cmd/dex/version.go
+++ b/cmd/dex/version.go
@@ -5,9 +5,9 @@ import (
 	"runtime"
 
 	"github.com/spf13/cobra"
-
-	"github.com/dexidp/dex/version"
 )
+
+var version = "was not built properly"
 
 func commandVersion() *cobra.Command {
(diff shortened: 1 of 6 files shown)


cmd/dex/version.go
```diff
@@ -5,18 +5,18 @@ import (
 	"runtime"
 
 	"github.com/spf13/cobra"
-
-	"github.com/dexidp/dex/version"
 )
 
+var version = "was not built properly"
```
sagikazarmark, line 10: I'd much rather use something like `unknown`, `dev`, `master` or something like these (maybe capitalized?)

server/api.go
```diff
@@ -29,6 +28,9 @@ const (
 	upBoundCost = 16
 )
 
+// version is the version of the Dex server. It is injected during build stage.
+var version = "was not built properly"
```
sagikazarmark, line 32: I'm wondering if injecting the version instead would make more sense.

sagikazarmark (commented):
Thanks @nabokihms !

server/api.go
nabokihms: We can inject the version variable at build-time only to the main package and then do something like

cmd
```go
import "github.com/dexidp/dex/server"

var version = "DEV"

func init() {
  server.Version = version
}
```

server
```go
var Version = "DEV"
```
Is it ok? It looks like it will work as we want.

sagikazarmark: I rather meant injecting into the server.

Something like this:

```go
NewAPI(storage, version, logger)
```

sagikazarmark: Not sure if that's a good idea, but global stuff always makes me icky.

sagikazarmark (approved):

PR #2121, opened 2021-05-14
feat: Add ent-based postgres storage

nabokihms:

#### Overview
* Add constructor for Postgres storage
* Tests for Postgres storage bootstrapping
* Connect conformance tests

Related #1864

Composing DSN method and storage structure are mostly copypasted from the "raw SQL" storage. I think they are good enough to continue using them.

```release-note
Add an experimental ent-based postgres storage.
```

### DIFF
--- a/storage/ent/postgres.go
+++ b/storage/ent/postgres.go
@@ -1,0 +1,155 @@
+package ent
+
+import (
+	"context"
+	"crypto/sha256"
+	"database/sql"
+	"fmt"
+	"net"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	// Register postgres driver.
+	_ "github.com/lib/pq"
+
+	entSQL "entgo.io/ent/dialect/sql"
+	"github.com/dexidp/dex/pkg/log"
+	"github.com/dexidp/dex/storage"
+	"github.com/dexidp/dex/storage/ent/client"
+	"github.com/dexidp/dex/storage/ent/db"
+)
+
+// nolint
+const (
+	// postgres SSL modes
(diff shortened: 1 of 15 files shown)


storage/ent/postgres.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if port != "" {
+		addParam("port", port)
+	}
+
+	if p.User != "" {
+		addParam("user", dataSourceStr(p.User))
+	}
+
+	if p.Password != "" {
```
sagikazarmark, line 122: I wonder if empty passwords worked before. If they did, this is going to be a break once we replace our previous implementation with ent.

sagikazarmark (approved):
Overall it looks good to me. One question regarding passwords.

Is this something we want to include in 2.29 or delay it to 2.30?

Once etcd 3.5 is tagged, I want to release a new Dex version.

nabokihms:
We can left it to 2.30 for sure.

storage/ent/postgres.go
nabokihms: I thought I copied the whole function from the raw SQL storage realization. I will check this.

nabokihms: It looks like [this code](https://github.com/dexidp/dex/blob/47b0a2bdf9ef7a3e981f7dba8e2433674db4ab3e/storage/sql/config.go#L167-L173) is equal to the code from the PR
```go
drv, err := entSQL.Open("postgres", p.dsn())	
if err != nil {
  return nil, err
}
```

I did a little testing, and both env-based storage and raw SQL storage works alike.

PR #2135, opened 2021-05-21
feat: add a "Forgot password" link to the password login page

al45tair:
#### Overview

Adds support for a "Forgot password" link to the password.html template, as well as configuration options to set both the link and the text associated with it.

Users may well forget their passwords, and it's useful to be able to put a suitable link onto the login page in that case. Obviously this needs to be configurable on a per-connector basis, because who they need to talk to exactly will vary depending on the connector being used.

Yes.

```release-note
For connectors that support password login, you can now specify a
link that users can click if they have forgotten their
password. Both the target URL and the text of the link can be
customised through configuration options on the supported
connectors (right now, Atlassian Crowd, Keystone and LDAP) using
the "forgotLink" and "forgotText" options in the YAML.
```


al45tair:
This hasn't gone anywhere in a year, and there's no way I can resolve the conflicts as my employer has strict rules about Open Source contributions. I'm going to close it for now.

PR #2146, opened 2021-05-27
Allow upstream OIDC provider to omit name claim

### LINKED ISSUE #2145: Allow upstream OIDC provider to omit name claim
Our upstream OIDC provider supports what amount to guest accounts, which have an identifier/email address but do not have a `name` claim. The OIDC connector enforces that `name` (or alternative claim supplied by `userNameKey`) be present. The OIDC [standard claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) do not have this requirement.

We would like to be able to configure the OIDC connector to allow users without a `name` claim. Ideally, the claims delivered to the client would omit `name`, like the upstream provider does, per [spec recommendation](https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse):

>For privacy reasons, OpenID Providers MAY elect to not return values for some requested Claims.
>
> If a Claim is not returned, that Claim Name SHOULD be omitted from the JSON object representing the Claims; it SHOULD NOT be present with a null or empty string value.

Our upstream provider is based on Shibboleth, which does support SAML, but we are encouraging OIDC adoption for consistency and ease of use (largely based on Dex being straightforward to set up). In looking over the SAML connector, we note that it is unmaintained and documents some known risks. An option that improves spec conformance for the OIDC connector seems preferable.

botimer:
#### Overview

This PR adds a config option to the OIDC connector to require or allow omission of the user's name claim from the upstream provider.

Closes #2145

(The scenario is well-described there.)

```release-note
The OIDC connector has a new boolean config option, UserNameOptional. The default behavior is unchanged, requiring the name claim.
```


botimer:
In addition to the added automated test, I have verified this patch with our local configuration and a test app (using Sinatra, OmniAuth, and the omniauth_openid_connect plugin). The Helm chart made it easy to use a custom image and add the new option to the connector config.

As hoped, the users can authenticate and the name claim is absent in the final token/info at the app, the same as presented to Dex when acting as RP. I was not entirely sure whether the core of Dex would expect the name, omit the claim, or send a blank one, but it omitted, which I believe to be correct per the specification recommendations. Thanks!

PR #2147, opened 2021-05-28
Allow a user to overwrite Auth and Token URL

rikatz:
Signed-off-by: Ricardo Katz <rkatz@vmware.com>

#### Overview

This change adds two new configuration fields in OIDC Connector: authURL and tokenURL.

This change adds two new configuration fields in OIDC Connector: authURL and tokenURL.

Dex uses the go-oidc library, which does auto-discovery based in .well-known/openid-configuration from the issuer configured.

Happens that in some cases the issuer is not the same as Auth and Token URL. In this case, is desired that the user can overwrite those configurations (instead of relying into the auto discovered by go-oidc) allowing the correct login page redirection.

In some cases UserInfoURL and JWKSURL should also be overwritable. The main issue is that, opposite to Auth and Token URL, go-oidc does not support overwriting UserInfo and JWKS but this might appear as a need in future

```release-note
Add authURL and tokenURL configurations in OIDC connector
```

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -44,6 +44,12 @@ type Config struct {
 	// InsecureEnableGroups enables groups claims. This is disabled by default until https://github.com/dexidp/dex/issues/1065 is resolved
 	InsecureEnableGroups bool `json:"insecureEnableGroups"`
 
+	// AuthURL provides a way to user overwrite the Auth URL from .well-known/openid-configuration
+	AuthURL string `json:"authURL"`
+
+	// TokenURL provides a way to user overwrite the Token URL from .well-known/openid-configuration
+	TokenURL string `json:"tokenURL"`
+
 	// GetUserInfo uses the userinfo endpoint to get additional claims for
 	// the token. This is especially useful where upstreams return "thin"
 	// id tokens
@@ -108,6 +114,14 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 	}
 
 	endpoint := provider.Endpoint()
+
+	if c.AuthURL != "" {
+		endpoint.AuthURL = c.AuthURL
+	}
+
+	if c.TokenURL != "" {
+		endpoint.TokenURL = c.TokenURL
+	}
 
 	if c.BasicAuthUnsupported != nil {
 		// Setting "basicAuthUnsupported" always overrides our detection.


scottcrawford03:
+1 to this change

nabokihms:
@rikatz sorry, I missed this PR.

The subject makes sense. For example, the oauth2 proxy allows you to override detected URLs or even make requests without detection. Even Google [has its hacks](https://github.com/coreos/go-oidc/blob/15b94d97d90c9455981dbdd90a021d34254cfdaf/oidc/verify.go#L238) inside of the go-oidc library. I believe that we can merge this PR if you are still willing to work on it.

Could you tell us a little more about your use case?

Are the discovery endpoint and the token endpoint served by the same IDP but in different domains, or maybe you want to send a code to get the token to a different service?

rikatz:
Hi @nabokihms and thanks for the response :)

So yes, this is exactly the case. The Discovery endpoint is served from a different domain from the auth and token endpoint. I know, for example Keycloak also supports overwriting this.

I was doing some researches about that, and actually this is allowed (when dealing with "multi host" discovery endpoints).

I will rebase here to remove the go.sum conflicts, but feel free to reach me about any other question.

Thank you very much!

rikatz:
done :)

connector/oidc/oidc.go
```diff
@@ -44,6 +44,12 @@ type Config struct {
 	// InsecureEnableGroups enables groups claims. This is disabled by default until https://github.com/dexidp/dex/issues/1065 is resolved
 	InsecureEnableGroups bool `json:"insecureEnableGroups"`
 
+	// AuthURL provides a way to user overwrite the Auth URL from .well-known/openid-configuration
+	AuthURL string `json:"authURL"`
+
+	// TokenURL provides a way to user overwrite the Token URL from .well-known/openid-configuration
```
nabokihms, line 50: Could you also add these options to the [documentation](https://github.com/dexidp/website/blob/main/content/docs/connectors/oidc.md)?

nabokihms (changes requested):
There is a couple of thoughts about this PR. As for me, everything looks good.

m44soroush:
hey guys, any updates on this issue?
we need this feature to overwrite auth and token URL.
It seems this change haven't merged to the main branch.

PR #2178, opened 2021-06-17
set readinessProbe to https

noesberger:
Fix the error
Readiness probe failed: Get "http://100.105.5.5:5556/healthz": dial tcp 100.105.5.5:5556: connect: connection refused
Client sent an HTTP request to an HTTPS server.

```release-note

```


sagikazarmark (approved):
Thanks! I'm gonna let it in, but we should probably get rid of this example: we have a helm chart now.

PR #2199, opened 2021-06-30
handlePasswordGrant: insert connectorData into OfflineSession

hensur:
#### Overview

Using dex with the kubernetes oidc client results in the creation
of an Offline Session with empty ConnectorData.
This results in a working login, however, the ldap connector will fail once the client attempts to refresh its token.

This change will insert the ConnectorData from the initial Login
into the OfflineSession, as already done in `handlePasswordLogin`.

Not doing this results in failures to refresh a token.
I've observed this with the ldap connetor, other connectors might
be affected as well.

The ldap connector will fail with this error message:
`failed to refresh identity: ldap: failed to unmarshal internal data: unexpected end of JSON input`

The `Refresh` method of the ldap connector will first try to unmarshal
the ConnectorData, which was empty, as dex wouldn't
store it in storage after a successful login.

While this seems to fix the issue, I'm not sure if it is desired to save the ConnectorData in `handlePasswordGrant` as done in `handlePasswordLogin`.

```release-note
NONE
```

### DIFF
--- a/server/handlers_test.go
+++ b/server/handlers_test.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
+	"path"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


nabokihms:
Hello @hensur. Thanks, it looks good. I wonder if we can write a test for this?

hensur:
Hey @nabokihms!
I had some time to look into this and came up with a test for this. It won't pass without my changes.
I took a lot of inspiration from the refresh handler tests, as I'm not really familiar with the dex code base :)

server/handlers_test.go
```diff
@@ -227,3 +229,75 @@ func TestHandleAuthCode(t *testing.T) {
...
+"username": "test",
+"password": "test"
+}`),
+	}
+
+	err = s.CreateConnector(c1)
+	require.NoError(t, err)
+}
+
+func TestConnectorDataNotEmpty(t *testing.T) {
```
nabokihms, line 259: Let's give this test a more specific name like the `TestPasswordConnectorDataNotEmpty`.

nabokihms (approved):
Approve in advance. Besides the test function name, everything looks good to me.

server/handlers_test.go
hensur: I renamed the test :)

nabokihms:
Thanks for your contribution @hensur !

PR #2218, opened 2021-07-27
Demonstrate use of `htpasswd` for bCrypt in `staticPasswords`

jglick:
#### Overview

Include an example command using `htpasswd` to bCrypt passwords in `staticPasswords` sections in example config files.

The former text

> bcrypt hash of the string "password"

is rather vague and not helpful if you are attempting to add your own static password here. I had to Google tools for running bCrypt from the command line, and found by trial and error that the default difficulty level of 5 is rejected by Dex (requires 10).

N/A

Just examples, so I think not.

```release-note
NONE

```

### DIFF
--- a/examples/config-dev.yaml
+++ b/examples/config-dev.yaml
@@ -141,7 +141,7 @@ enablePasswordDB: true
 # If this option isn't chosen users may be added through the gRPC API.
 staticPasswords:
 - email: "admin@example.com"
-  # bcrypt hash of the string "password"
+  # echo password | htpasswd -BinC 10 admin | cut -d: -f2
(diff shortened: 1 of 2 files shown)


examples/config-dev.yaml
```diff
@@ -141,7 +141,7 @@ enablePasswordDB: true
 # If this option isn't chosen users may be added through the gRPC API.
 staticPasswords:
 - email: "admin@example.com"
-  # bcrypt hash of the string "password"
```
nabokihms, line 144: Let's leave the original comment here. We need to clarify what the command you added does.

(This comment is valid for both files.)

nabokihms (approved):
Looks good.  The only obstacle is signing DCO.

PR #2233, opened 2021-08-11
Add claimMapping enforcement

### LINKED ISSUE #2162: Allow OIDC connector to remap a claim even if the original claim is not available
There is a problem, that users cannot map custom OIDC token claim to the dex claim.
For example, the original token has two claims:
```json
{
 "email": "personal@gmail.com",
 "preferred_email": "work@mycompany.com",
  ...
}
```
Apparently, a user wants to use `preferred_email` as an email claim in Dex but has no opportunity to do it, because the email claim is present in the token.

See corresponding code lines:
https://github.com/dexidp/dex/blob/master/connector/oidc/oidc.go#L306-L310

Allow users to remap claims even if the standard one exists in the token. There are many OIDC providers that allow claims customization. I see no objections to allowing users to manage claims as they want. 

Sadly, making this strategy enabled by default is a breaking change.

Middlewares? :)

#### Additional context
Was discussed here https://github.com/dexidp/dex/discussions/2158#discussioncomment-834377

Happy2C0de:
#### Overview

Fixes #2162
This PR introduces an `enforce` option to specify if a user would like to enforce the values set in `claimMapping`. For example, the token:
```
{
"email": "test@example.com",
"preferred_email": "foo@bar.com"
}
```
Currently, the user is not able to use the `preferred_email` claim because the token contains the default `email` claim.

This change is very handy and assures backwards compatibility by introducing an `enforce` option. Configuration would look like:
```
claimMapping:
  enforce: true
  email: "preferred_email"
```
Same holds for `preferred_username` and `groups` options within `claimMapping`.

Closes #2162

OIDC connector tests were adjusted and expanded. Restructured the `claimMapping` struct to a `type` in order to simplify the configuration.

```release-note
Adds the enforcement of claimMappings
```


sagikazarmark:
The proper long-term solution for this is using middleware. I have two reservations with this PR:

- the claimMapping options should remain what it is: a map. The enforce option needs to live outside of it. It's not necessarily clean from a config structure perspective, but keeps the mapping config easier to comprehend IMHO.
- Enforce is not necessarily a good word here. Override, overwrite sound better.

Happy2C0de:
@sagikazarmark Thanks for checking this PR. 👍 

> the claimMapping options should remain what it is: a map. The enforce option needs to live outside of it. It's not necessarily clean from a config structure perspective, but keeps the mapping config easier to comprehend IMHO.

Sounds reasonable. I could restructure the PR but will you merge and accept it? I would:
- Rename `claimMapping.enforce` to`overrideClaimMapping bool`
- Take this option out of the ClaimMapping type
 
I see this PR as a bridge between the current limitations and your suggestion with the middleware. I assume the latter implementation will require some time to be implemented.

sagikazarmark:
@Happy2C0de yes, I'd be inclined to merge it until we have a better solution in place.

Happy2C0de:
@sagikazarmark I did the expected changes and rerun the test. Looks good from my point of view.

connector/oidc/oidc.go
```diff
@@ -56,16 +56,24 @@ type Config struct {
...
+	// i.e. if there are 'email' and `preferred_email` claims available, by default Dex will always use the `email` claim independent of the ClaimMapping.EmailKey.
+	// This setting allows you to override the default behavior of Dex and enforce the mappings defined in `claimMapping`.
+	OverrideClaimMapping bool `json:"overrideClaimMapping"` // defaults to false
 
-		// Configurable key which contains the email claims
-		EmailKey string `json:"email"` // defaults to "email"
+	ClaimMapping ClaimMapping `json:"claimMapping"`
+}
+
+type ClaimMapping struct {
```
sagikazarmark, line 67: Can you please revert the changes that are not relevant to this feature? An inline struct might not be the nicest solution, but it should disappear eventually anyway.

If (for some reason) this is an important change, can you please make it a separate commit so that the refactor clearly separates from the feature?

connector/oidc/oidc.go
```diff
@@ -304,9 +315,12 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 	var email string
 	emailKey := "email"
 	email, found = claims[emailKey].(string)
-	if !found && c.emailKey != "" {
-		emailKey = c.emailKey
+	if (!found || c.overrideClaimMapping) && c.claimMapping.EmailKey != "" {
+		emailKey = c.claimMapping.EmailKey
 		email, found = claims[emailKey].(string)
+		if !found {
```
sagikazarmark, line 321: If overriding is enabled and the claim is empty, BUT the original claim is there, this will fail.

Is that the desirable behavior? I can argue for both.

I think we are seriously hitting the limit of declarative configuration with this as both behavior might be desirable, but adding more options would make this even more complicated.

Can you add a test case for this to verify the behavior and detect possible future "regressions"?

Is email the only affected claim?

connector/oidc/oidc_test.go
```diff
@@ -92,6 +93,25 @@ func TestHandleCallback(t *testing.T) {
...
+			claimMapping: ClaimMapping{
+				EmailKey: "custommail",
+			},
+			expectUserID:       "subvalue",
+			expectUserName:     "namevalue",
+			expectedEmailField: "customemailvalue",
+			token: map[string]interface{}{
+				"sub":            "subvalue",
+				"name":           "namevalue",
+				"mail":           "emailvalue",
```
sagikazarmark, line 110:
```suggestion
				"email":           "emailvalue",
```

I think this should be `email` (potential copy-paste error from the above custom claim example?)

sagikazarmark (changes requested):
Thanks @Happy2C0de !

There are a couple things that needs to be cleared before we can merge it, but overall, it looks good to me.

connector/oidc/oidc.go
Happy2C0de: Done

connector/oidc/oidc_test.go
Happy2C0de: Done

connector/oidc/oidc.go
Happy2C0de: I agree to that. I changed my implementation such that it applies all the properties given in the ClaimMapping if they are present.
The check for `&& c.emailKey != ""` checks that the override only applies for values that are defined.

I think this resolves your input.

Happy2C0de:
@sagikazarmark 
I did the requested changes and I simplified the merge in order to push this PR further. Mainly, I agree to your inputs and tried to make everything clear.
Thanks for checking it!

connector/oidc/oidc.go
sagikazarmark: I think you misunderstood my comment. I'm talking about the _claim_ not the claim mapping. What happens, if there is an email claim with a valid email, but there is no valid claim with the override value (ie. it's missing). In that case, the perfectly valid email is going to overridden by an empty value.

While I can argue that the point of this config is to _always_ override, it can easily lead to problems if the claim turns out to be empty.

So what I suggested is to add a test that verifies that this is the current behavior, so if anything changes we will know. I'm not sure we can do much more with the current config structure. We could add a check to fall back to the original value (if any) if the override claim is empty.

sagikazarmark:
@Happy2C0de I think you misunderstood one of my comments. I tried to clarify in a follow-up comment.

connector/oidc/oidc.go
Happy2C0de: Yes, but that's the expected behaviour. If you configure that a claim overrides the `email` claim, then I would expect that there is no fallback. Why should it? This fallback would be very intransparent to the user.

This is the responsibility of the engineer who configures Dex to assure that first, the claim is present within the defined scopes, second that the customised claim contains the necessary information. Isn't it?

**About the test**
Mainly, you would like to catch:
```
"email": "foo.bar@example.com",
```
with configuration:
```
overrideClaimMapping: true,
emailKey:             "custommail",
```
results in `"foo.bar@example.com"`.
But I think that's wrong. It depends on the scope as well.

At the moment, my implementation results in an error if the hasEmailScope is true.
https://github.com/dexidp/dex/blob/e472fe668e81b7d708c990c056e398412a2dba51/connector/oidc/oidc.go#L313
And it default to "", if hasEmailScope is false. That's probably not good because it shouldn't depend on the `email` scope.

This is what I could improve. Mainly, I have to change the check:
```
if !found && ( hasEmailScope || c.emailKey != "" ) {
```
This would assure that in either case, the error is thrown what I prefer because obviously there is going something wrong. But that's an issue that you already have in the code. What if no `email` scope is present and `emailKey` is set but its claim is not present. No error will be thrown and "" is returned.

What's your preferred way? I would go with throwing an error because something doesn't work as expected. Accordingly, I would change the if condition.

sagikazarmark: I'd agree that the more natural expectation here is to **not** fall back to the original claim. However, due to the fact that this "feature" is controlled by a set of different config options, I'm concerned that some day someone would change the implementation leading to random failures.

As I mentioned above, my concern is about _how_ this behavior is configured, not what the actual behavior is (this is an edge case after all). And this is why I suggested adding a test case that make sure that whenever someone attempts to change this behavior, there is a safe guard that stops us. That's what tests are for after all.

IMO adding a new test case that verifies that in case overriding is forced AND the claim is empty AND the original claim is not, we return an empty value is trivial.

Happy2C0de: Hi @sagikazarmark 
I agree to adding a test, I was just not sure what the preferred implementation should be, error or empty string. As you mentioned, I implemented the empty string return for this edge case and added a test `overrideWithMissingCustomEmailClaim` that takes care of that. :)

Thanks for the discussion!

connector/oidc/oidc.go
```diff
@@ -304,9 +311,13 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
 	emailKey := "email"
 	email, found = claims[emailKey].(string)
-	if !found && c.emailKey != "" {
+	if (!found || c.overrideClaimMapping) && c.emailKey != "" {
 		emailKey = c.emailKey
 		email, found = claims[emailKey].(string)
+		if !found && c.overrideClaimMapping {
+			// If override is enabled but claim was not found, empty string is preferred over fallback.
+			email, found = "", true
+		}
```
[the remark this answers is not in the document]
Happy2C0de: Removed

connector/oidc/oidc_test.go
```diff
@@ -92,6 +93,40 @@ func TestHandleCallback(t *testing.T) {
...
+			token: map[string]interface{}{
+				"sub":            "subvalue",
+				"name":           "namevalue",
+				"email":          "emailvalue",
+				"custommail":     "customemailvalue",
+				"email_verified": true,
+			},
+		},
+		{
+			name:                 "overrideWithMissingCustomEmailClaim",
```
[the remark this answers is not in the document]
Happy2C0de: @sagikazarmark 
Thanks for the review. I apologise the late reply.

I removed the test. I first thought that I could try to test against the error message but the current test cases are not handling errors either. This would require a bigger restructuring of the testing.

I hope the current changes meet your expectations.

Thanks again for checking!

sagikazarmark:
@Happy2C0de please rebase your PR so that CI can be triggered. Thanks!

sagikazarmark (approved):

Happy2C0de:
@sagikazarmark done :)

sagikazarmark (approved):
Perfect, thank you!

PR #2234, opened 2021-08-11
Return valid JWT access token from password grant

enj:
#### Overview

This change updates the password grant handler to issue a valid JWT access token instead of just returning a random value as the access token.  This makes it possible to use the access token against the user info endpoint.

Signed-off-by: Monis Khan <mok@vmware.com>

Access tokens returned from the password grant should be valid to use against the user info endpoint.

N/A

```release-note
Valid access tokens are now returned from the password grant flow.
```


sagikazarmark:
Makes sense on first sight. I'm not sure why we haven't done this in the first place. I'll have to dig in a little to understand that before merging it.

enj:
@sagikazarmark @nabokihms friendly ping 🙂

nabokihms (approved):
As far as I can get, the `newAccessToken` method was added along with the user info endpoint support. However, password grant support was added way before this. All this looks like we just forgot about password grant flow. I have seen similar mistakes before.

No objections to merging this, good catch.

P. s. Maybe we could share the code between the password handler and the code handler to reduce the chances of this happening again.

PR #2265, opened 2021-09-03
Add parametrization of grant type supported in discovery endpoint

ariary:
Signed-off-by: ariary <ariary9.2@hotmail.fr>

#### Overview

Modify the discovery endpoint result. The grant type supported are no longer fully "hard-coded". 

The discovery endpoint is used to  help client application to know how dex is configured for the sign-in flow.
Currently the grant types supported were hard-coded, which does not always match the dex configuration.

The aim of this PR is to add `password` in the grant type supported list if it is the case 

```release-note
NONE

```

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -94,7 +94,6 @@ func (s *Server) discoveryHandler() (http.HandlerFunc, error) {
 		UserInfo:          s.absURL("/userinfo"),
 		DeviceEndpoint:    s.absURL("/device/code"),
 		Subjects:          []string{"public"},
-		GrantTypes:        []string{grantTypeAuthorizationCode, grantTypeRefreshToken, grantTypeDeviceCode},
 		IDTokenAlgs:       []string{string(jose.RS256)},
 		CodeChallengeAlgs: []string{codeChallengeMethodS256, codeChallengeMethodPlain},
 		Scopes:            []string{"openid", "email", "groups", "profile", "offline_access"},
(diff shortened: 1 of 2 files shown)


nabokihms (approved):
Looks good, thanks. I requested an additional review from Mark for a second opinion.

nabokihms:
@ariary some minor fixes are required according to the linter

ariary:
@nabokihms yes, should be ok now

ariary:
@sagikazarmark friendly ping 🙂 is this ok for you?

server/handlers.go
```diff
@@ -110,6 +109,9 @@ func (s *Server) discoveryHandler() (http.HandlerFunc, error) {
 	}
 	sort.Strings(d.ResponseTypes)
 
+	d.GrantTypes = s.supportedGrantTypes
+	sort.Strings(d.GrantTypes)
```
sagikazarmark, line 113: Since this modifies the original slice, I think the sorting should rather be done where the slice is created. Consumers should treat it as read-only.

Also, the sorting called in the handler will be done every time the discovery endpoint is called.

Any reason why it should be kept here?

sagikazarmark (changes requested):
I had one tiny comment, otherwise it looks good to me!

server/handlers.go
ariary: @sagikazarmark change made! 

It was here only imitate the treatment of  `responseTypes`

sagikazarmark: Hm, yeah, I think that shouldn't be there either. But let's not block this anymore.

PR #2268, opened 2021-09-08
Resolves #2111 Option to fetch transitive group membership

### LINKED ISSUE #2111: Support nested groups for google
GKE is able to support permissions via nested groups. However, when using dex with the google connector the only groups that appear are ones directly connected to a user.

We have inconsistent permissions using various tools. Those with dex do not work where as `kubectl` does.

I don't know a lot about the topic, but when I looked into it I think that the api for `admin.google.com` doesn't actually support nested groups, but I believe the GCP IAM API does. I'm not sure if this would mean a separate connector or one could be added to google.

We're currently trying to resolve this with work arounds that will add people to the groups they already belong to via another group. It's not optimal, however.

snuggie12:
All credit goes to @santisaez as original author (special shout out to @jasonhadi for finding this PR.) This is a copy of #2077 plus review feedback which changed variable names and added a comment or two. Below is original summary:

Overview

Google Workspace groups (aka G Suite) have support for nesting groups, that is, add a group to another group. This is useful to automatically add members from a small team (backend, for example) to a larger department (engineering, for example).

A user in Google Workspace can have both direct and indirect group memberships. In the example above, "backend" is a direct group and "engineering" is an indirect group membership. The Google Admin SDK API only returns direct group memberships, and thus dex only returns direct memberships.
What this PR does / why we need it

This PR extends the work done by @JoelSpeed in #1185 and adds an option to fetch the indirect group memberships from Google Workspace, aka G Suite. This feature will be useful to fetch all groups a user belongs to, both direct and indirect.
Special notes for your reviewer

cc/ @JoelSpeed because they developed the feature to fetch the groups from Google
Does this PR introduce a user-facing change?

google: groupsIndirectFetch option added to fetch indirect group memberships from Google


snuggie12:
@JoelSpeed are you still the correct person to review?

snuggie12:
@nabokihms Any ETA on when someone might review this?

nabokihms:
I will try to take a look till the next week.

nabokihms:
I assume that rebasing is required because GitHub workflow files have been changed in the master branch.

PR #2278, opened 2021-09-17
refactor: move from io/ioutil to io and os package

Juneezee:
#### Overview

The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). This PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

> The io/ioutil package has turned out to be a poorly defined and hard to understand collection of things. -- https://golang.org/doc/go1.16#ioutil

> Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. -- https://go.dev/blog/package-names

The Golang team has decided to deprecate the `io/ioutil` package and recommends new code to use the new definitions. To avoid inconsistency in the future, it is a good idea to move all current usage of `io/ioutil` functions to the new definitions.

No breaking changes.

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,6 +2,13 @@ run:
     timeout: 2m
 
 linters-settings:
+    depguard:
+        list-type: blacklist
+        include-go-root: true
+        packages:
+            - io/ioutil
+        packages-with-error-message:
+            - io/ioutil: "The 'io/ioutil' package is deprecated. Use corresponding 'os' or 'io' functions instead."
     gci:
         local-prefixes: github.com/dexidp/dex
     goimports:
@@ -13,6 +20,7 @@ linters:
     enable:
         - bodyclose
         - deadcode
+        - depguard
         - dogsled
         - exhaustive
         - exportloopref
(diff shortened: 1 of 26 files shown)


nabokihms (changes requested):
Good catch, thanks!

I think that to consolidate what you have done there, we need to setup a depguard linter to prevent others from using the `ioutil` package.

It is possible by adding the following block to the .golangci.yml in the repo root:
```yaml
linters-settings:
  depguard:
    list-type: blacklist
    include-go-root: true
    packages:
    - io/ioutil
    packages-with-error-message:
    - io/ioutil: "The 'io/ioutil' package is deprecated. Use corresponding 'os' or 'io' functions instead."
```
and enabling the depguard linter.

Juneezee:
@nabokihms Hi! Thanks for reviewing the pull request. I have added the depguard linter. Please take a look again. Thanks

nabokihms:
You should also enable the `depguard` linter in the `linters.enable` section (it is commented right now).

Juneezee:
@nabokihms Sorry, my bad. depguard should be enabled now.

sagikazarmark (approved):
LGTM, thanks!

nabokihms (approved):

PR #2284, opened 2021-09-23
Support for multiple refresh tokens per user

vinod-trilio:
Closes https://github.com/dexidp/dex/issues/981

This PR introduces `multipleTokens` option under `expiry.refreshTokens` in config to configure multi refresh tokens per user.

**Details**

- `multipleTokens.allow` option defaults to false and in that case this PR does not change any behavior.
- `multipleTokens.maximumCount` option specifies maximum refresh tokens per user which defaults to `50` - 
- `multipleTokens.replacementPolicy` specifies then old token deletion/replacement policy if number of tokens issued crossed specified `multipleTokens.maximumCount` which defaults to LRU.
- When `multipleTokens.allow` is true, Dex skips to delete a refresh token in issuing an id token.
- To minimize changes in storage layer, `ListRefresh` and `RevokeRefresh` gRPC API use `Storage.ListRefreshTokens` in `multipleTokens.allow` mode. This can be heavy but num of refresh tokens not is expected very high (same as num of id tokens).
- `storage.OfflineSessions.Refresh` contains the latest-issued refresh token.

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -342,9 +342,16 @@ type Logger struct {
 	Format string `json:"format"`
 }
 
+type MultipleRefreshToken struct {
+	Allow             bool   `json:"allow"`
+	MaximumCount      int    `json:"maximumCount"`
+	ReplacementPolicy string `json:"replacementPolicy"`
+}
+
 type RefreshToken struct {
-	DisableRotation   bool   `json:"disableRotation"`
-	ReuseInterval     string `json:"reuseInterval"`
-	AbsoluteLifetime  string `json:"absoluteLifetime"`
-	ValidIfNotUsedFor string `json:"validIfNotUsedFor"`
+	DisableRotation   bool                 `json:"disableRotation"`
+	MultipleTokens    MultipleRefreshToken `json:"multipleTokens"`
+	ReuseInterval     string               `json:"reuseInterval"`
+	AbsoluteLifetime  string               `json:"absoluteLifetime"`
+	ValidIfNotUsedFor string               `json:"validIfNotUsedFor"`
 }
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -310,7 +310,9 @@ func runServe(options serveOptions) error {
 		c.Expiry.RefreshTokens.ValidIfNotUsedFor,
 		c.Expiry.RefreshTokens.AbsoluteLifetime,
 		c.Expiry.RefreshTokens.ReuseInterval,
-	)
+		c.Expiry.RefreshTokens.MultipleTokens.Allow,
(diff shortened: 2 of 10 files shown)


mbd-dbc-dk:
Would be great to have this. Kind of a nuissance as it is now.

Elentary:
@nabokihms will you be able to have a look on this or assign another person ?

nabokihms:
@Elentary Before implementation, we'd like to hear more about use cases, like when the limitation of a single token per user is an obstacle.

Elentary:
> @Elentary Before implementation, we'd like to hear more about use cases, like when the limitation of a single token per user is an obstacle.

I’d happy to. We use Dex for k8s auth via OIDC with Google as IdP (k8s talks with Dex, and Dex use Google to login users).
In this scenario you have a refresh token from Dex written in your kubeconfig. If you use several machines for development, you had to copy your kubeconfig there, because you can’t receive another credentials from Dex for the same Google email.
The problem is when the id token expires. After this moment the next k8s query from machine A will refresh your id token, get a new refresh token and invalidate the old one. So, as the old token was invalidated, the other machine B won’t be able to refresh its id token, and you have to manually copy kubeconfig again and again.

Elentary:
@nabokihms do you need any other actions or info ?

mbd-dbc-dk:
We have similar issues. Developers obviously need tokens on their laptops, but often also need to run tests, automations, etc, from other machines. This is a major nuisance as it is is now. 

Also: Developers often does not have the knowledge about this, so their tokens may stop working on their laptops, due to this, and then they file support issues with out IT department, etc.

Elentary:
@sagikazarmark @nabokihms could you take a look, please ?

Elentary:
@nabokihms @sagikazarmark @justaugustus gentle remind on this, we are pinging for 5 months without any response from the team

ahmedwaleedmalik:
Hey guys, do we have an update on this PR? Is it stalled because this is not an acceptable change or simply due to resource shortage for reviewing this?

matpen-wi:
> @Elentary Before implementation, we'd like to hear more about use cases, like when the limitation of a single token per user is an obstacle.

Hi,
we also would like to request support for multiple sessions for a ‘client-user’ pair. Our setup involves using Dex in combination with oauth2-proxy, which acts as the OAuth2 client for Dex and provides browser sessions secured by session cookies. However, we encounter issues when the same user logs in from different browsers. From Dex’s perspective, it’s the same client, and it doesn’t allow multiple sessions in parallel. Therefore, it would be extremely helpful if Dex could support multiple sessions for the same client-user pair.

Thank you!

Elentary:
@nabokihms @sagikazarmark @justaugustus

Guys, gentle-gentle-gentle ping, please

PR #2291, opened 2021-10-05
Adding an oidcConnector config to decide whether to validate that the Connector Callback must contain the same host as the issuer

acohen4:
#### Overview

Adding an oidcConnector config to decide whether to validate that the Connector Callback must contain the same host as the issuer.

This is useful in the case where the Dex Issuer may be used in a network configuration where it's accessible to dependent client services, but inaccessible to external users. We encounter this in localhost deployments on kubernetes where an in-cluster Dex Issuer address is used by k8s service, but the upstream IDP is configured to redirect to a Dex address that is accessible to the user (i.e. localhost).

It adds a configuration option. If not specified, the existing behavior is preserved.

```release-note
Add optional oidc connector config: `insecureSkipIssuerCallbackDomainCheck` to help with some network configurations.
```

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -44,6 +44,9 @@ type Config struct {
 	// InsecureEnableGroups enables groups claims. This is disabled by default until https://github.com/dexidp/dex/issues/1065 is resolved
 	InsecureEnableGroups bool `json:"insecureEnableGroups"`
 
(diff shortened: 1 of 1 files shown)


acohen4:
Hi @JoelSpeed @nabokihms, who's a good person for me to have review this PR?
Thanks!

chancez:
Wouldn’t it be better if you could configure multiple issuer urls instead?

PR #2300, opened 2021-10-18
fix: do not update offlinesession lastUsed field if refresh token was not updated

### LINKED ISSUE #2187: Concurrent token issuing for the same user experiences high failure rate
#### Expected behavior (what you expected to happen)

When multiple concurrent requests for the same user occur, concurrent
modifications in the backing store (ETCD) should not fail the loosing updates.

#### Actual behavior (what actually happened)

Requests are returned with 500 and log message has similar lines to following:

> failed to update offline session: failed to update key=\\"offline_session/00uz6fj623xee6srn0h7|okta
": concurrent conflicting update happened

#### Steps to reproduce

1. Submit multiple concurrent requests for the same user

#### Environment

- Dex version: 2.28.1
- Storage type: ETCD
- Installation type (official docker image, helm chart, etc.): Kubernetes

#### Anything else we should know?

Multiple locations in our application will exchange the same refresh token at
the same time causing this high concurrency. Additionally the compaction time
in ETCD can cause the requests to artificially group with compaction is
happening.

For our use case, updating the LastUsed field so frequently is not needed. We
can take updating it every few requests and thus completely skip updating the
offline session and refresh object if nothing else has changed.

nabokihms:

#### Overview

As for now, Dex updates the lastUsed field of OfflineSession on every refresh token touching. It causes unnecessary conflict errors for etcd and Kubernetes storages.

Closes #2187

```release-note
Do not update OfflineSession lastUsed field on allowed refresh token reusing
```

### DIFF
--- a/server/refreshhandlers.go
+++ b/server/refreshhandlers.go
@@ -227,16 +227,13 @@ func (s *Server) updateRefreshToken(token *internal.RefreshToken, refresh *stora
 
 	lastUsed := s.now()
 
-	rerr := s.updateOfflineSession(refresh, ident, lastUsed)
-	if rerr != nil {
-		return nil, rerr
-	}
-
 	refreshTokenUpdater := func(old storage.RefreshToken) (storage.RefreshToken, error) {
 		if s.refreshTokenPolicy.RotationEnabled() {
 			if old.Token != token.Token {
 				if s.refreshTokenPolicy.AllowedToReuse(old.LastUsed) && old.ObsoleteToken == token.Token {
 					newToken.Token = old.Token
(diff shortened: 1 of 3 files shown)


nabokihms:
I have never thought about not updating the last used field at all 😅 . Probably, it would be a breaking change. To avoid it, if a token is not updated, we will not update the offline session. Thus the refresh token `lastUsed` field and offline session `lastUsed` field will be in sync.

P.S. A link between offline sessions and refresh tokens is the most complicated part of the Dex database schema.
We need to keep these entities in sync without having transactions for kubernetes storage and etcd storage.

PR #2305, opened 2021-10-23
fix web static file path slash error for win platform

copperyp:
I test dex on Win10 platform, i found it dont launch success, I try to solve this problem, found slash convert is be necessary.

It should also be able to run on Linux.

```release-note

```


sagikazarmark:
Thanks @copperyp

PR #2344, opened 2021-12-05
fix: return invalid_grant error on claiming token of another client

nabokihms:

#### Overview

Change the error returned. Add tests.

According to [rfc6749](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2), it is required to return the `invalid_grant` error if the refresh token from the request was issued for another client.

It is required to pass the OIDC conformance test:
> Attempting to use refresh_token issued to client 2 with client 1: CheckErrorFromTokenEndpointResponseErrorInvalidGrant: 'error' field has unexpected value

image

```release-note
Return the invalid_grant error on claiming token of another client.
```

### DIFF
--- a/server/refreshhandlers.go
+++ b/server/refreshhandlers.go
@@ -76,7 +76,9 @@ func (s *Server) getRefreshTokenFromStorage(clientID string, token *internal.Ref
 
 	if refresh.ClientID != clientID {
 		s.logger.Errorf("client %s trying to claim token for client %s", clientID, refresh.ClientID)
-		return nil, invalidErr
+		// According to https://datatracker.ietf.org/doc/html/rfc6749#section-5.2 Dex should response with the
+		//  invalid grant error if token has already been claimed by another client.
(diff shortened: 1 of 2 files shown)


server/refreshhandlers.go
```diff
@@ -76,7 +76,9 @@ func (s *Server) getRefreshTokenFromStorage(clientID string, token *internal.Ref
 
 	if refresh.ClientID != clientID {
 		s.logger.Errorf("client %s trying to claim token for client %s", clientID, refresh.ClientID)
-		return nil, invalidErr
+		// According to https://datatracker.ietf.org/doc/html/rfc6749#section-5.2 Dex should response with the
```
sagikazarmark, line 79:
```suggestion
		// According to https://datatracker.ietf.org/doc/html/rfc6749#section-5.2 Dex should respond with an
```

sagikazarmark (approved):
Thanks @nabokihms

PR #2346, opened 2021-12-08
grpc: Add gRPC logging

kellyma2:
This adds logging for gRPC requests in order to make debugging easier.


go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/go-sql-driver/mysql v1.6.0
 	github.com/gorilla/handlers v1.5.1
 	github.com/gorilla/mux v1.8.0
+	github.com/grpc-ecosystem/go-grpc-middleware v1.0.0
```
sagikazarmark, line 19: Can you please upgrade to the latest version?

cmd/dex/serve.go
```diff
@@ -117,6 +117,9 @@ func runServe(options serveOptions) error {
 		return fmt.Errorf("failed to register gRPC server metrics: %v", err)
 	}
 
+	grpcLogEntry := logrus.NewEntry(logger)
```
sagikazarmark, line 120: I wonder if we should mark grpc logs somehow. Like adding a field called subsystem/component/channel.

@nabokihms what do you think?

sagikazarmark (approved):

sagikazarmark (changes requested):
Oops

cmd/dex/serve.go
kellyma2: bump

kellyma2:
> Oops

Looks like this needs to be resolved in some way?

sagikazarmark:
Sorry, oops was only about me accidentally approving the PR with pending comments.

sagikazarmark:
I guess we can merge this, but we will probably drop logrus as a logger soon.

kellyma2:
@sagikazarmark I still see '1 change requested' and it points me at your 'oops' comment.  

I'm happy to update this to not use logsrus if the alternative is already there?

nabokihms:
We migrated to log/slog. I'm going to close this PR since it uses the logrus library. Feel free to open a new PR if you still need this feature.

PR #2361, opened 2021-12-28
feat: external connectors

### LINKED ISSUE #1907: External connectors
The main problem that the Dex repo contains a lot of different connectors with various features, purposes, original authors, tests coverage, etc.

It makes Dex maintainers support a lot of custom logic that they are unfamiliar with. 

The thing to start with is writing solid design documentation (an enhancement proposal, you may say).

Nowadays, the most popular solution to this kind of problem is to develop a GRPC API interface. Connectors will be distributed as separate programs, which implement this interface (with the separate codebase, tests, maintainers, etc.)

Examples: terraform providers, Kubernetes CSI. 

If we want to move Dex forward, there are not so many good alternatives. It is possible to limit the number of connectors in the core repo and stop merging PR with new connectors until stabilizing already accepted ones. However, this doesn't look like a prospective way.

#### Additional context
The initial idea and its benefits were previously described [in a slack thread](https://kubernetes.slack.com/archives/C011URMR41W/p1608459639239300?thread_ts=1608386492.238900&cid=C011URMR41W).

nabokihms:

#### Overview

This is the very first naive gRPC-based implementation of external connectors support. Two connector types have been added: external password connector and external callback connector.

With these changes, it becomes possible to write custom connectors utilizing gRPC API or the SDK for Golang (I have added examples for both types).

Closes #1907

* Dex communicates with external connectors only via localhost because of security concerns.

* The `external-gitlab` example of the callback external connector is a copy-paste of the in-tree gitlab connector just to see how easy it will be to convert old connectors/write new ones. However, I'm highly concerned about merging it to master. It seems a lot better to develop some brand new connector in a separate repository.

```release-note
Add external connectors support.
```


sagikazarmark:
Thanks @nabokihms for working on this!

Without trying to be the bad guy here: I think we are still lacking clarity on a lot of the details that would be nice to understand before going down this road (or any roads).

I opened a separate issue that vaguely explains what my problem is in #2370

In short: I'd like us to write down everything we know (goals, non-goals, scope, etc) so we can make good decisions.

I'm already working on a document for the middleware thing. Do you think @nabokihms you could come up with something for this one?

Basically, I'd like to understand:
- What functionality do we want to "externalize"?
- How does the process look like?
- Can we make it abstract enough so that the implementation doesn't matter? If not, what are the requirements towards the implementations?
- What are the attack vectors that we might introduce into Dex this way?

It's pretty late here, so we might want to have a discussion at some point instead of me just trying to dump my thoughts in the evening. Hopefully it's somewhat clear what I'd like us to do.

PR #2371, opened 2022-01-14
Allow configuration of returned groups via authproxy connector

seuf:
#### Overview

Added `groupHeader` and `groups` configuration for authproxy connector. 

Example :

```
connectors:
      - type: authproxy
        id: http-proxy
        name: http-proxy
        config:
          userHeader: X-Forwarded-User # default is X-Remote-User
          groupHeader: X-Forwarded-Group # default is X-Remote-Group
          staticGroups:
            - admin
```

This allow users to configure argocd dex to use authproxy connector and specify the users default group and the HTTP header to read to get the group.

### DIFF
--- a/connector/authproxy/authproxy.go
+++ b/connector/authproxy/authproxy.go
@@ -13,9 +13,14 @@ import (
 )
 
 // Config holds the configuration parameters for a connector which returns an
-// identity with the HTTP header X-Remote-User as verified email.
+// identity with the HTTP header X-Remote-User as verified email,
+// X-Remote-Group and configured staticGroups as user's group.
+// Headers retrieved to fetch user's email and group can be configured
+// with userHeader and groupHeader.
 type Config struct {
-	UserHeader string `json:"userHeader"`
+	UserHeader  string   `json:"userHeader"`
+	GroupHeader string   `json:"groupHeader"`
+	Groups      []string `json:"staticGroups"`
 }
 
 // Open returns an authentication strategy which requires no user interaction.
@@ -24,16 +29,22 @@ func (c *Config) Open(id string, logger log.Logger) (connector.Connector, error)
 	if userHeader == "" {
 		userHeader = "X-Remote-User"
 	}
+	groupHeader := c.GroupHeader
+	if groupHeader == "" {
+		groupHeader = "X-Remote-Group"
+	}
 
-	return &callback{userHeader: userHeader, logger: logger, pathSuffix: "/" + id}, nil
+	return &callback{userHeader: userHeader, groupHeader: groupHeader, logger: logger, pathSuffix: "/" + id, groups: c.Groups}, nil
 }
 
 // Callback is a connector which returns an identity with the HTTP header
 // X-Remote-User as verified email.
 type callback struct {
-	userHeader string
-	logger     log.Logger
-	pathSuffix string
+	userHeader  string
+	groupHeader string
+	groups      []string
+	logger      log.Logger
+	pathSuffix  string
 }
 
 // LoginURL returns the URL to redirect the user to login with.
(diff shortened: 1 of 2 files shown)


connector/authproxy/authproxy.go
```diff
@@ -15,7 +15,8 @@ import (
 // Config holds the configuration parameters for a connector which returns an
 // identity with the HTTP header X-Remote-User as verified email.
 type Config struct {
-	UserHeader string `json:"userHeader"`
+	UserHeader string   `json:"userHeader"`
+	Groups     []string `json:"groups"`
```
nabokihms, line 19: From my point of view, it would be better (and more generic) to extract groups from headers (as we do for username).

For example, the oauth2 proxy can return groups via the `X-Forwarded-Groups` header.

We even have a TODO in code to implement groups support this way:
https://github.com/dexidp/dex/blob/716eef83bca805e1607338fe0a271affc2d097b0/connector/authproxy/authproxy.go#L58

WDYT? Will extracting groups from headers cover your use case?

nabokihms (changes requested):
Hello, @seuf. Thank you for your intention to work on this feature.

1. I left a comment. We need to figure out the right approach before moving further.
2. Would you mind adding the note that authproxy now supports groups to the README.md? https://github.com/dexidp/dex/blob/716eef83bca805e1607338fe0a271affc2d097b0/README.md?plain=1#L78

connector/authproxy/authproxy.go
seuf: Maybe we can do both ? My auth proxy (traefik-forward-auth) does not return `X-Forwarded-Group` Header.
So if the header is present we use it else we fallback to the configured group.

seuf: I've updated the PR to allow custom group Header configuration and use both header and configured groups (appended)

seuf: Hello @nabokihms What do you think of my changes ?
default groups are the configured groups (can be empty)
if http header (default `X-Remote-Group` but configurable) is present append it to the configured groups.
I've tested it on my argocd dex deployment, and it works fine.

seuf:
Hi @nabokihms I've updated this PR.

You can now configure both `userHeader` and `groupHeader` (defaults to `X-Remote-User` and `X-Remote-Group`).
You can also configure a list of groups (can be empty)
if http group header is present then it is appended to the configured group list.

I've tested it on my argocd dex deployment, and it works fine.

nabokihms (changes requested):
Looks good! Thank you for your work. There are some comments left. Please look through them when you have time.

And one more thing. I would like you to also make changes to the module documentation to inform users about the feature you have made.
https://github.com/dexidp/website/blob/main/content/docs/connectors/authproxy.md

connector/authproxy/authproxy.go
```diff
@@ -15,7 +15,9 @@ import (
 // Config holds the configuration parameters for a connector which returns an
 // identity with the HTTP header X-Remote-User as verified email.
 type Config struct {
-	UserHeader string `json:"userHeader"`
+	UserHeader  string   `json:"userHeader"`
+	GroupHeader string   `json:"groupHeader"`
+	Groups      []string `json:"groups"`
```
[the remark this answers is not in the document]
seuf: ok, i've renamed this one

connector/authproxy/authproxy.go
```diff
@@ -15,7 +15,9 @@ import (
 // Config holds the configuration parameters for a connector which returns an
 // identity with the HTTP header X-Remote-User as verified email.
```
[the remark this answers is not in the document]
seuf: comment updated with the descirption of all config params.

connector/authproxy/authproxy.go
```diff
@@ -55,11 +63,17 @@ func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connecto
 	if remoteUser == "" {
 		return connector.Identity{}, fmt.Errorf("required HTTP header %s is not set", m.userHeader)
 	}
+	groups := m.groups
+	headerGroup := r.Header.Get(m.groupHeader)
+	if headerGroup != "" {
+		groups = append(groups, headerGroup)
+	}
 	// TODO: add support for X-Remote-Group, see
```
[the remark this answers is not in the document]
seuf: todo removed

seuf:
Hi @nabokihms I've applied the changes you suggested. I'm doing the PR on the website documentation.

seuf:
Hi @nabokihms Do you have time this week to review this again ?

nabokihms (approved):

Issue #1924, opened 2021-01-14
Provide distroless images


sagikazarmark:
#### Is your feature request related to a problem?
Distroless images reduce the attack surface of an application.

#### Describe the solution you'd like to see
Provide distroless images by default. Provide alpine variants with shell for debugging purposes.

#### Additional context
Adding CA certs to distroless images might be a challenge

ankeesler:
have y'all discussed doing this recently? i'd be interested in consuming a distroless dex in my product.

could we do this by providing a `Dockerfile.distroless` that:
1) doesn't support `gomplate` and instead has `ENTRYPOINT ["dex"]`
2) copies CA from a previous `apk add ca-certificates` step in a builder stage

alternatively, if we don't want to give up `gomplate`, could we rewrite `docker-entrypoint.sh` as a go program and inject it into the runner stage as a binary?

thanks for your consideration.

sagikazarmark:
Hey @ankeesler 

To be honest, this isn't our top priority these days, but we'd definitely accept a PR. ;)

As for gomplate: I think it solved a ton of problems for Dex, specifically injecting secrets into various places in the config file.
I don't think we have a better solution at the moment and it would probably weird if the different images worked too differently.

The entrypoint script is quite simple though, so I think it would be trivial to rewrite it in Go.

P.S. Would you consider adding yourself to the ADOPTERS.md file?

ankeesler:
cool, thanks for the response @sagikazarmark 

i opened an initial strawperson implementation here to get the conversation started: https://github.com/dexidp/dex/pull/2378. would love your thoughts if you have some cycles

and i will talk to my coworkers about getting us added as an adopter...

PR #2378, opened 2022-01-19
distroless: Dockerfile works with distroless base image

ankeesler:
#### Overview

This PR hopes to add support for a distroless Dex container image:
```
docker build --build-arg BASEIMAGE=gcr.io/distroless/static:latest -t andrew:distroless .
```

In order to get the Dex container built on distroless, I had to:
1. rewrite docker-entrypoint.sh in go and compile it into the Dex image (to avoid the need for a shell in the Dex image)
1. fetch CA certificates in an earlier build stage and copy them into the Dex image (to avoid the need for a package manager in the Dex image)
1. add a "stager" build stage for creating some directories that should exist in the Dex container and copy them into the Dex image (to avoid the need for `mkdir(1)` in the Dex image)

Closes https://github.com/dexidp/dex/issues/1924.

* Perhaps see initial chatter about this in https://github.com/dexidp/dex/issues/1924 comments.
* Question: how do I setup E2E tests to run in CI with this distroless Dex container image?

```release-note
Dex Dockerfile supports being built with distroless base layer (docker build --build-arg BASEIMAGE=gcr.io/distroless/static:latest ...)
```


Makefile
```diff
@@ -18,7 +18,7 @@ group=$(shell id -g -n)
 
 export GOBIN=$(PWD)/bin
 
-LD_FLAGS="-w -X main.version=$(VERSION)"
+LD_FLAGS="-w -X main.version=$(VERSION) -extldflags \"-static\""
```
ankeesler, line 21: how do y'all feel about always statically compiling C code into Go binaries? i added this since on distroless we don't have some C libraries that Go wants to dynamically load (i think they are related to somewhere around `storage/sql/sqlite.go`)

this looks like it would add XXX to the default alpine-based Dex container image

sagikazarmark:
@ankeesler this is awesome, thanks for taking the time to submit this PR. It's gonna take a bit more time to review it, however I'm inclined to wait with the next release for this to land.

ankeesler:
thanks very much @sagikazarmark for the support and being willing to review and giving us a heads up on timeframe. let me know if you need anything from me in order to review.

Dockerfile
```diff
@@ -1,8 +1,10 @@
+ARG BASEIMAGE=alpine:3.15.0
```
sagikazarmark, line 1: My only bad feeling about this is that we probably can't use dependabot anymore to keep the image up-to-date. But we can't use it for the distroless image anyway, because it's not mentioned in this file anywhere, so there will be manual steps to keep images up-to-date.

Makefile
sagikazarmark: Looks good to me

sagikazarmark (approved):
Thanks @ankeesler , this is great work! Kudos for providing tests.

ankeesler:
thanks very much for reviewing @sagikazarmark - really appreciate you devoting the energy to this. i see you requested review from @nabokihms as well. i'll be on the lookout for that review...

sagikazarmark:
Yeah, @nabokihms did the original implementation of the entrypoint and gomplate, so I wanted to hear his opinion as well before merging.

ankeesler:
thanks very much @nabokihms 

do y'all have an estimate for when this will land in a release?

nabokihms:
1. That is enough to check container could run.
2. According to milestone v2.31.0, there is a couple of PRs we want to have reviewed. After checking them, we will release a new version.

PR #2379, opened 2022-01-19
Add support for group and email headers to authproxy connector

GCHQDeveloper73046:
#### Overview

This commit adds several new options to the authproxy connector to allow additional user
attributes to be extracted.

A new emailHeader config option on the connector allows different values for the UserID
Email fields of the identity to be used, if the option is not specified then Email will
be populated from UserID as before.

User groups may also be extracted from headers in one of two different ways.
In order to extract group information the groupHeader config option must be specified.
If the groupSeparator option is specified and non-empty then the group header is treated
as a delimited list of group names, for example given groupSeparator = "," then the
expected format of header would be

X-Remote-Group: allUsers,admins

Alternatively if groupSeparator is not provided then the header is treated as a repeated
list of group names, for example

X-Remote-Group: allUsers
X-Remote-Group: admins

Note that the connector behaviour remains unchanged in all cases where the new config
options have not been used, ensuring backwards compatibility for existing installations.

```release-note
Add optional support for group and email headers to authproxy connector
```


GCHQDeveloper73046:
I've just spotted #2371 was also raised recently, it's unfortunate that after so long without this feature, two PRs arrive to address it at around the same time. I don't wish to involve myself in the other PR in any way but just wanted to cross-link them for awareness.

I will additionally address in my PR the comment that was raised on the other about updating the readme which I have also neglected to do in the initial commit here.

nabokihms:
@GCHQDeveloper73046, hello and thank you for the contribution. We have merged the #2371 recently. Does this PR cover your use cases?

GCHQDeveloper73046:
> @GCHQDeveloper73046, hello and thank you for the contribution. We have merged the #2371 recently. Does this PR cover your use cases?

Hi, I saw the merge of the other PR  had gone through, I suspect the answer is yes it will cover my use cases but either way this PR as it stands is no longer applicable, so I will close and will raise a new PR with revised changes if necessary. Thanks.

Issue #2417, opened 2022-02-22
Add acr_values support for OIDC


dirien:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

Hi,

Our internal IDP, uses the propery acr_values as a set of arbitrary values that the client and idp agreed upon to communicate the level of authentication that happened. This is to give the client a level of confidence on the qualify of the authentication that took place.

There are no "official" values for them. They should be communicated via the acr_values_supported parameter of OIDC discovery response. Please see [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html) for more on this.

### Proposed Solution

Add new, non-mandatory, config field in OIDC connection.

### Alternatives Considered

-

### Additional Information

I will create the PR for this.

PR #2418, opened 2022-02-22
feat: Add acr_values support for OIDC

dirien:

This PR adds acr_values support for the OIDC connection
Fixes: #2417 

yes, a new non-mandatory field is introduced called `acrValues` in the config of the OIDC connection

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -44,6 +44,11 @@ type Config struct {
 	// InsecureEnableGroups enables groups claims. This is disabled by default until https://github.com/dexidp/dex/issues/1065 is resolved
 	InsecureEnableGroups bool `json:"insecureEnableGroups"`
 
+	// AcrValues (Authentication Context Class Reference Values) that specifies the Authentication Context Class Values
+	// within the Authentication Request that the Authorization Server is being requested to use for
+	// processing requests from this Client, with the values appearing in order of preference.
+	AcrValues []string `json:"acrValues"`
+
 	// GetUserInfo uses the userinfo endpoint to get additional claims for
 	// the token. This is especially useful where upstreams return "thin"
 	// id tokens
@@ -154,6 +159,7 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 		hostedDomains:             c.HostedDomains,
 		insecureSkipEmailVerified: c.InsecureSkipEmailVerified,
 		insecureEnableGroups:      c.InsecureEnableGroups,
+		acrValues:                 c.AcrValues,
 		getUserInfo:               c.GetUserInfo,
 		promptType:                c.PromptType,
 		userIDKey:                 c.UserIDKey,
@@ -180,6 +186,7 @@ type oidcConnector struct {
 	hostedDomains             []string
 	insecureSkipEmailVerified bool
 	insecureEnableGroups      bool
+	acrValues                 []string
 	getUserInfo               bool
 	promptType                string
 	userIDKey                 string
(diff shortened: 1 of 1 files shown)


Kaktor (approved):

connector/oidc/oidc.go
```diff
@@ -44,6 +44,11 @@ type Config struct {
 	// InsecureEnableGroups enables groups claims. This is disabled by default until https://github.com/dexidp/dex/issues/1065 is resolved
 	InsecureEnableGroups bool `json:"insecureEnableGroups"`
 
+	// AcrValues (Authentication Context Class Reference Values) that specifies the Authentication Context Class Values
+	// within the Authentication Request that the Authorization Server is being requested to use for
+	// processing requests from this Client, with the values appearing in order of preference.
+	AcrValues string `json:"acrValues"`
```
nabokihms, line 50:
```suggestion
	AcrValues []string `json:"acrValues"`
```

connector/oidc/oidc.go
```diff
@@ -209,6 +216,10 @@ func (c *oidcConnector) LoginURL(s connector.Scopes, callbackURL, state string)
 		opts = append(opts, oauth2.SetAuthURLParam("hd", preferredDomain))
 	}
 
+	if len(c.acrValues) > 1 {
```
nabokihms, line 219:
```suggestion
	if len(c.acrValues) > 0 {
```

connector/oidc/oidc.go
```diff
@@ -154,6 +159,7 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 		hostedDomains:             c.HostedDomains,
 		insecureSkipEmailVerified: c.InsecureSkipEmailVerified,
 		insecureEnableGroups:      c.InsecureEnableGroups,
+		acrValues:                 c.AcrValues,
```
nabokihms, line 162:
```suggestion
		acrValues:                 strings.Join(c.AcrValues, " "),
```

connector/oidc/oidc.go
nabokihms: Some providers use acr values like `0` or `1`.
See https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens

nabokihms (changes requested):
Hello there! First of all, thank you for the PR. I have a couple of suggestions. Please look through the, when you have time.

1. Because RFC states that 
    > act_values is a space-separated string that specifies the acr values
    
    Let's use the same structure for the  `acrValues` config option for scopes - a slice of strings.
    
2. In the `Open` method of the connector, compose the values of the slice into a single string.

3. What do you think about validating the `acr_values_supported` received from the discovery OIDC enpoint before sending a request or event on connector creation? Does it seem to be a good feature, or is it just a complication without benefits?

connector/oidc/oidc.go
dirien: Very good idea! Fixed

connector/oidc/oidc.go
dirien: Done! 😄

connector/oidc/oidc.go
dirien: Done, the check if the string is empty should really be 0! my bad 😅

dirien:
Hi @nabokihms,

thanks for the feedback! I changed your suggestions in the code.

> What do you think about validating the acr_values_supported received from the discovery OIDC enpoint before sending a request or event on connector creation? Does it seem to be a good feature, or is it just a complication without benefits?

I would create for this, if needed in the future a seperate issue. Our company `well-known/openid-configuration` does not return acr_values_supported. So I can't test it proberly.

Looking for your feedback.

dirien:
I updated the website too, as we have now a list instead a string 

https://github.com/dexidp/website/pull/110

nabokihms (approved):

PR #2430, opened 2022-03-08
Create setting to allow to trust the system root CAs

dhaus67:
#### Overview

This introduces a new function for the OpenShift connector with the ability to inject a `http.Client` for usage. This will enable us to i.e. modify the `x509.CertPool` and support adding system specific certificates as trusted within library usage.

We are currently using dex as a library within our application running on OpenShift as a containerized workload.
When using the connector in combination with the internal OAuth server of OpenShift, they OAuth server may be either 
1. internally exposed, whereas the certificate will be explicitly added under a specific mount point.
2. exposed externally, behind a OpenShift route and a certificate the cluster admin chose.

Since we want to cover both cases ideally when using it as a library, it would be nice to have some way of telling the connector to not only include custom root CAs, but also the system root CAs.
The best way would be to extract the list of trusted CAs from Golang's `x509.CertPool` and use this for configuration, but sadly this is currently not possible.

No.

### DIFF
--- a/connector/openshift/openshift.go
+++ b/connector/openshift/openshift.go
@@ -28,13 +28,14 @@ const (
 
 // Config holds configuration options for OpenShift login
 type Config struct {
-	Issuer       string   `json:"issuer"`
-	ClientID     string   `json:"clientID"`
-	ClientSecret string   `json:"clientSecret"`
-	RedirectURI  string   `json:"redirectURI"`
-	Groups       []string `json:"groups"`
-	InsecureCA   bool     `json:"insecureCA"`
-	RootCA       string   `json:"rootCA"`
+	Issuer               string   `json:"issuer"`
+	ClientID             string   `json:"clientID"`
+	ClientSecret         string   `json:"clientSecret"`
+	RedirectURI          string   `json:"redirectURI"`
+	Groups               []string `json:"groups"`
+	InsecureCA           bool     `json:"insecureCA"`
+	RootCA               string   `json:"rootCA"`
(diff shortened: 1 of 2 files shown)


nabokihms:
Hello, @dhaus67. I finally managed to take a look at this PR. We do not imply that someone uses Dex as a library, but I do not have concerns about rearranging the code to make it usable as a library. It still will be hard to guarantee the compatibility of your changes for future releases, though.

Speaking about this particular solution, I don't think we are willing to expose this option to end-users. If users want to append a certificate to the system root pool, they should follow the way you have mentioned (about adding the mount point).

From my point of view, this is ok to add a method to substitute the whole HTTP transport or client to make this connector more usable as a library. Something like this:
```golang
func (c *Config) OpenWithHttpClient(id string, logger log.Logger, httpClient *http.Client) (conn connector.Connector, err error) {
  ...
}

func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, err error) {
   httpClient, err := newHTTPClient(c.InsecureCA, c.RootCA); err != nil {
     return nil, fmt.Errorf("failed to create HTTP client: %v", err)
   }
   return c.OpenWithHttpClient(id, logger, httpClient)  
}
```

nabokihms:
@sagikazarmark  /cc

dhaus67:
Thanks for the feedback @nabokihms ! I've changed it to not expose it directly as a setting for the end-user, but rather as you mentioned by providing the ability to inject the `http.Client`. Feel free to take a look once more.

dhaus67:
@nabokihms any update on this one?

PR #2445, opened 2022-03-21
Added s390x and power support for makefile

jdao-rh:
#### Overview

Added support for s390x and ppc64le for `make deps` in makefile


sagikazarmark:
Thanks for the contribution @jdao-rh !

May I ask what led you to make this change? Do you actually plan to develop Dex on these platforms? Because they aren't necessary for "just" building Dex for them.

jdao-rh:
Thanks for the quick feedback, I was under the impression that its was needed for building and running the tests for Dex. But after looking into it, the changes are probably not needed. Also, there are no immediate plans for developing Dex for these architectures, besides building the binary and image.

krishvoor:
Hi @sagikazarmark,
Thanks for your prompt response, Dex is leveraged for authentication in [ArgoCD](https://argoproj.github.io/argo-workflows/argo-server-sso-argocd/). 

As an ongoing effort to enable "dexidp/dex", we "just" do not want to build, we also have attempted the tests, and this PR allows us to pass the tests as well. Eventually, would like to propose CI that would allow us to contribute/maintain concurrency in long run. That said, this is the first PR via which we are reaching out to "dexidp" as a community and enabling them for ppc64le/s390x architectures.

Issue #2457, opened 2022-03-30
Dex can not authenticate against USGovernment Azure via OIDC.


jlane67:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.30.2

### Storage Type

Kubernetes

### Installation Type

Other (specify below)

### Expected Behavior

I installed Argo CD with Argo's bundled Dex configured to speak to an Azure AD domain for a US Government Cloud customer.  I got to the Azure SSO page and expected to be redirected back into Argo CD.

### Actual Behavior

I was redirected back to the login page with a 400 error.  The error in the pod logs indicates that Dex is not using the correct method of connecting to Azure AD.

### Additional Information

MS REST NodeAuth issue: https://github.com/Azure/ms-rest-nodeauth/issues/110
Argo CD issue I opened about this bug: https://github.com/argoproj/argo-cd/issues/8946

### Configuration

```yaml
Argo CD handles the configuration of Dex.  The only values I specified for Dex in Argo CD's configuration are private ones: client ID, secret, etc.
```

### Logs

```shell
From the pod log:

`level=error msg="Failed to authenticate: microsoft: failed to get token: oauth2: cannot fetch token: 400 Bad Request\nResponse: {\"error\":\"invalid_request\",\"error_description\":\"AADSTS900432: Confidential Client is not supported in Cross Cloud request.\\r\\n`
```

jlane67:
Allowing Dex to specify the Azure AD endpoint appears to fix this issue.

jlane67:
[These two lines](https://github.com/dexidp/dex/blob/master/connector/microsoft/microsoft.go#L65-L66) should be configurable from JSON as they are not used for non-standard accounts.  The endpoint is microsoftonline.us rather than microsoftonline.com, so making this a constant is wrong.

PR #2459, opened 2022-04-01
enable custom Azure endpoints for Government Cloud Services customers

jlane67:
#### Overview

Stops hard-coding Azure endpoints as they are not universally accurate.

Resolves https://github.com/dexidp/dex/issues/2457.  Pure OIDC mode is unable to retrieve group membership data from Azure AD, meaning RBAC is impossible.

yes

```release-note
This change makes the Azure endpoints configurable.  For most users, the default 
https://login.microsoftonline.com/ and https://graph.microsoft.com/ values 
will remain the correct choice, but this allows users of Azure in special 
regions, such as Government Cloud Services for export-controlled workloads, 
to use the Microsoft connector to get group membership.
```


sagikazarmark:
@jlane67 thanks for the contribution!

On the surface, this seems to be a breaking change as the default values for those URLs are empty strings.

Do you know if the underlying library falls back to the right default values in this case? Can you also point to the code in the SDK?

Last, but not least: can you please update the documentation? https://github.com/dexidp/website/blob/main/content/docs/connectors/microsoft.md

jlane67:
Hi, thank you for pointing that out.  I have added another commit adding a fallback to the default values.  I'm not super experienced at Go, but this is just something that came up in the course of deploying Dex.  If the logical flow looks good I can add a PR to fix the documentation.

jlane67:
I have confirmed in a deployment on our infrastructure that the changes in this PR allow full authentication with Azure Government Cloud Services when using the .us endpoints.

sleterrier:
@sagikazarmark: anything I could do to help this PR get merged?

dsetlock:
@sagikazarmark wanting to pick this thread back up. Trying to deploy Dex into Azure Government I hit this same blocker and came here to investigate but it seems a PR has been make the Azure endpoints more flexible for over a year.

sagikazarmark:
I'm looking at the code and....I'm a little bit confused. I don't think this code even compiles.

Issue #1848, opened 2020-10-28
Switch to Kubernetes client-go library


nabokihms:
Dex can earn some of the benefits from it:
* There a lot of already features in client go like canceling requests, retrying requests, static typing errors, API QPS settings, etc. 
* Many people already familiar with go client, switching to client-go will make the contribution process easier for them
* Kubernetes community uses client-go a lot, which means this library is well tested

PR #2479, opened 2022-04-15
fix: prevent server-side request forgery using Kubernetes storage

nabokihms:

#### Overview
Validate kubernets object names on the Dex side

For some endpoints, we directly get data out of a user request and build a URL for the request to Kubernetes API out of it.
The problem does not look urgent, because Dex always builds the prefix of the URL on his side, yet we need to protect API from malicious or invalid requests.

https://cwe.mitre.org/data/definitions/918.html

```release-note
fix: prevent server-side request forgery using Kubernetes storage
```

### DIFF
--- a/storage/kubernetes/client.go
+++ b/storage/kubernetes/client.go
@@ -18,6 +18,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


storage/kubernetes/client.go
```diff
@@ -83,14 +84,24 @@ func offlineTokenName(userID string, connID string, h func() hash.Hash) string {
...
 	apiVersion, namespace, resource, name string, params url.Values,
-) string {
+) (string, error) {
 	basePath := "apis/"
 	if apiVersion == "v1" {
 		basePath = "api/"
 	}
 
+	if name != "" && (len(name) > kubeResourceMaxLen || !kubeResourceNameRegex.MatchString(name)) {
+		return "", fmt.Errorf(
```
sagikazarmark, line 100: This would result in unsanitized user input being logged. (See GitHub code scanning results below)

Also, just an idea: `s/kubernetes resource name is not valid/invalid kubernetes resource name`

storage/kubernetes/client.go
```diff
@@ -83,14 +84,26 @@ func offlineTokenName(userID string, connID string, h func() hash.Hash) string {
...
+) (string, error) {
 	basePath := "apis/"
 	if apiVersion == "v1" {
 		basePath = "api/"
 	}
 
+	if name != "" && (len(name) > kubeResourceMaxLen || !kubeResourceNameRegex.MatchString(name)) {
+		// The actual name can be found in auth request or auth code objects and equals to the state value
+		return "", fmt.Errorf(
+			"invalid kubernetes resource name: must match the pattern %s and be no longer than %d charactes",
```
sagikazarmark, line 102: I don't really want to nitpick, but Kubernetes has a pretty good error message for invalid names. Can't we just use that?

Maybe even use the same library to validate the resource name?

Anyway, we can always change the error later.

sagikazarmark (approved):

storage/kubernetes/client.go
nabokihms: It seems like related to
https://github.com/dexidp/dex/issues/1848

sagikazarmark: Let's look into that if we can

Issue #778, opened 2017-01-19
Monitoring endpoints for debugging


ericchiang:
Add a Prometheus monitoring endpoint[0] or a pprof endpoint[1].

[0] https://github.com/prometheus/client_golang
[1] https://golang.org/pkg/net/http/pprof/

ericchiang:
cc @rithujohn191 we've also had a request for adding additional prometheus metrics for connector availability etc.

ericchiang:
For example, if the LDAP connector has connectivity.

brancz:
Let us know if you want us to review ;)

fabxc:
I'd recommend adding the debug pprof endpoint – we have it in Prometheus and it's pretty helpful. In general, client_go also comes with a default go collector that exposes all runtime internals (i.e. pprof metrics) we can gather. That's extremely helpful as well.

byxorna:
It would be neat if this was prioritized for an upcoming release; having insight into the health and state of dex is critical when deployed into production environments. Not to mention, exposing a /metrics endpoint is part of being a good container-native citizen :)

brancz:
@byxorna working on it: #1155

sagikazarmark:
We have both metrics and pprof exposed on the telemetry endpoint.

PR #2482, opened 2022-04-20
feat: enable profiling endpoints

nabokihms:

#### Overview

Add profiling endpoints for the Dex telemetry server.

Related to #778

```release-note
feat: enable profiling endpoints
```

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -151,6 +151,8 @@ type Web struct {
 // Telemetry is the config format for telemetry including the HTTP server config.
 type Telemetry struct {
 	HTTP string `json:"http"`
+	// EnableProfiling makes profiling endpoints available via web interface host:port/debug/pprof/
+	EnableProfiling bool `json:"enableProfiling"`
 }
 
 // GRPC is the config for the gRPC API.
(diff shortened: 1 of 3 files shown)


cmd/dex/config.go
```diff
@@ -151,6 +151,8 @@ type Web struct {
 // Telemetry is the config format for telemetry including the HTTP server config.
 type Telemetry struct {
 	HTTP string `json:"http"`
+	// EnableProfiling makes profiling endpoints available via web interface host:port/debug/pprof/
+	EnableProfiling bool `json:"enableProfiling"`
```
sagikazarmark, line 155: Honestly, I'd just add them, not sure if we need a switch for it, but I'm not objecting to it either.

sagikazarmark (approved):

cmd/dex/config.go
nabokihms: It was a security concern because profiling adds overhead, and frauds can use these endpoints for dos attacks.

sagikazarmark: Ideally, the telemetry endpoint shouldn't be public, but I understand that some people publish it, so fine by me.

PR #2483, opened 2022-04-21
Add numeric user ID support for oauth connector

tsl0922:
Signed-off-by: Shuanglei Tao <tsl0922@gmail.com>

Add numeric user ID support for oauth connector

```release-note
Add numeric user ID support for oauth connector
```

### DIFF
--- a/connector/oauth/oauth.go
+++ b/connector/oauth/oauth.go
@@ -209,12 +209,18 @@ func (c *oauthConnector) HandleCallback(s connector.Scopes, r *http.Request) (id
 		return identity, fmt.Errorf("OAuth Connector: failed to parse userinfo: %v", err)
 	}
 
-	userID, found := userInfoResult[c.userIDKey].(string)
+	userID, found := userInfoResult[c.userIDKey]
 	if !found {
 		return identity, fmt.Errorf("OAuth Connector: not found %v claim", c.userIDKey)
 	}
 
-	identity.UserID = userID
+	switch userID.(type) {
+	case float64, int64, string:
+		identity.UserID = fmt.Sprintf("%v", userID)
+	default:
+		return identity, fmt.Errorf("OAuth Connector: %v claim should be string or number, got %T", c.userIDKey, userID)
+	}
+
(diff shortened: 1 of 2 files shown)


connector/oauth/oauth.go
```diff
@@ -209,12 +209,12 @@ func (c *oauthConnector) HandleCallback(s connector.Scopes, r *http.Request) (id
...
 	}
 
-	userID, found := userInfoResult[c.userIDKey].(string)
+	userID, found := userInfoResult[c.userIDKey]
 	if !found {
 		return identity, fmt.Errorf("OAuth Connector: not found %v claim", c.userIDKey)
 	}
 
-	identity.UserID = userID
+	identity.UserID = fmt.Sprintf("%v", userID)
```
nabokihms, line 217: It seems ok for id to be a number or string, yet this piece of the code will also convert arrays and maps to strings.

I suggest using a custom logic to convert claims to strings.

```go
switch userID.(type) {
case float64, int64, string:
    identity.UserID = fmt.Sprintf("%v", userID)
default:
     return identity, fmt.Errorf("OAuth Connector: %v claim should be string or number, got %T", c.userIDKey, userID)
}
```

And one more thing. Maybe we can also apply the same logic to username and preferred username claims. It is ok to have a username like, e.g., 644587, but it is an open question.

nabokihms (changes requested):
Good catch! Thank you for opening this PR. I have only one thing to discuss before we can merge this.

connector/oauth/oauth.go
tsl0922: Done.

nabokihms:
@tsl0922, the linter throws an error. Could you please check it so we can merge this PR?

nabokihms:
@tsl0922, the linter throws an error. Could you please check it so we can merge this PR?

nabokihms (approved):
lgtm, thank you!

nabokihms:
Docker build is not working, but it looks like a problem on our side.

PR #2505, opened 2022-05-06
OIDC connector: Support cases where there is no id_token when using a refresh_token grant

### LINKED ISSUE #2498: Dex OIDC connector requires id_token in response body from IDPs, when using a grant_type of refresh_token
### Preflight Checklist

### Version

2.31.0

### Storage Type

Kubernetes

### Installation Type

Official container image

### Expected Behavior

I would expect that dex be compatible with IDPs that do not issue an id_token again when using a grant_type of refresh_token, or that the reasons why this can't be supported are documented. 

### Actual Behavior

Dex fails to use the IDP refresh token to get a new access token. 

### Steps To Reproduce

_No response_

### Additional Information

As per https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse an id_token may not be contained in a reply on the token endpoint, when using a grant type of refresh_token. 

id_token seems to be required in the current code https://github.com/dexidp/dex/blob/master/connector/oidc/oidc.go#L278

```
Upon successful validation of the Refresh Token, the response body is the Token Response of [Section 3.1.3.3](https://openid.net/specs/openid-connect-core-1_0.html#TokenResponse) except that it might not contain an id_token.
```

The body of the response from my IDP looks like this:

```
{
    "access_token": "redacted_access",
    "expires_in": 3599,
    "refresh_token": "redacted_refresh",
    "token_type": "Bearer"
}
```

### Configuration

_No response_

### Logs

```shell
time="2022-04-28T21:37:31Z" level=error msg="failed to refresh identity: oidc: no id_token in token response"
```

Blorpy:

#### Overview

Allows for dex to function with IDPs that do not return an ID token in the reply when using a refresh_token grant. 

Although this seems to be rare, the OIDC spec does not require that an id_token be present in the reply when using a refresh_token grant. Additional info in #2498. 

This is a toggle in PingFed, so there is at least one common IDP out there where there won't always be a refresh token. 

Closes #2498 

Note that if a user has an IDP where ID tokens aren't present when using a refresh_token grant, AND also chooses not to enable the userinfo endpoint, this will choke. The claims have to come from somewhere, so I would say this really is just an unsupported configuration should someone choose to try this. The OIDC spec says that userinfo must always contain the sub additionally. 

I tried to keep behaviour as close as possible to current. Although its extra code, and probably not needed, I chose to prefer the sub from the ID token if its present, over assigning the one from the userinfo. 

```release-note

NONE

```

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -226,7 +226,6 @@ func (c *oidcConnector) LoginURL(s connector.Scopes, callbackURL, state string) 
 	}
 	return c.oauth2Config.AuthCodeURL(state, opts...), nil
 }
-
 type oauth2Error struct {
 	error            string
 	errorDescription string
@@ -239,6 +238,11 @@ func (e *oauth2Error) Error() string {
 	return e.error + ": " + e.errorDescription
 }
 
+const (
+	createCaller = "create"
+	refreshCaller = "refresh"
+)
+
 func (c *oidcConnector) HandleCallback(s connector.Scopes, r *http.Request) (identity connector.Identity, err error) {
 	q := r.URL.Query()
 	if errType := q.Get("error"); errType != "" {
@@ -248,8 +252,7 @@ func (c *oidcConnector) HandleCallback(s connector.Scopes, r *http.Request) (ide
 	if err != nil {
 		return identity, fmt.Errorf("oidc: failed to get token: %v", err)
 	}
-
-	return c.createIdentity(r.Context(), identity, token)
+	return c.createIdentity(r.Context(), identity, token, createCaller)
 }
 
 // Refresh is used to refresh a session with the refresh token provided by the IdP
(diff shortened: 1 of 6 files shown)


connector/oidc/oidc.go
```diff
@@ -239,6 +239,11 @@ func (e *oauth2Error) Error() string {
 	return e.error + ": " + e.errorDescription
 }
 
+const (
+	createCaller = "create"
```
nabokihms, line 243: NIT: It looks better for me to use more strict internal constant markers.

```go
type caller uint

const (
    createCaller caller = iota
    refreshCaller
)
```

nabokihms (changes requested):
Thank you, @Blorpy. Overall, everything looks good. I left a couple of suggestions to you, and adding tests is also required for these changes.

Blorpy:
Thanks for the excellent feedback! I have implemented the requested changes, but will leave this in draft until I have the tests written.

Blorpy:
I was surprised to see there were no refresh tests already for this connector, so I had to do a bit of plumbing here. I chose to break out my tests to a new function, primarily because the existing function was for the initial identity creation, and I didn't want to mess with it too much. 

I have to deal with test cases where there will be no ID tokens, and because the functions call setupServer once which will dictate the ID token behaviour for the lifetime of that test case, I figured it was just easier to write another small test function.

Blorpy:
Sorry for the messed up push here, will fix

PR #2511, opened 2022-05-06
Remove google specific hd / hosted domain claim config from oidc connector

Blorpy:

#### Overview

At some point in the evolution of dex it seems like some Google specific stuff got left in the OIDC connector, when there is now a Google specific one. 

I would like to gather feedback on this PR whether it makes sense to remove this, given that there could be users out there who are utilizing the OIDC connector but using Google as their IDP, instead of using the Google connector. I don't want to break people, but I also think this stuff is pretty 

Originally added in https://github.com/dexidp/dex/pull/974 which is before the Google connector was split out here https://github.com/dexidp/dex/pull/1185. I believe that this is just a miss since then, I am unaware of any other IDPs (but of course I am not aware of everyone's configuration) that use this claim. 

This is currently also referenced in the docs here https://dexidp.io/docs/connectors/oidc/, I will open a separate PR to update the docs should it be agreed we want to proceed with removing this code. 

```release-note
The OIDC connector will no longer look for, or respect the hosted domain "hd" claim. User's who are using Google as their IDP should use the Google connector. 
```

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -34,10 +34,6 @@ type Config struct {
 
 	Scopes []string `json:"scopes"` // defaults to "profile" and "email"
 
-	// Optional list of whitelisted domains when using Google
-	// If this field is nonempty, only users from a listed domain will be allowed to log in
-	HostedDomains []string `json:"hostedDomains"`
-
 	// Override the value of email_verified to true in the returned claims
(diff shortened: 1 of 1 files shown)


nabokihms (approved):
Thank you for your contribution! Looks good to me.

nabokihms:
We should not forget to mention this in the upper block of the release message.

PR #2512, opened 2022-05-09
Introduce CheckEndpointConnector, implement for OpenShift connector.

dhaus67:
#### Overview

Add the ability to check the endpoints used by the connectors. The goal is to give clients the possibility to check whether a connector's endpoints are reachable.

The ultimate goal behind is to avoid potential misconfiguration when creating a new connector. Right now, it wouldn't be possible to detect this at the configuration point, but rather only when _using_ the connector to login if you i.e. use dex within your application.

Ideally, we'd like to check this during creation. 
Since this would be a major, undesirable change to introduce this during the `Open` method, I've opted to add a new interface which connectors may implement, if they want to offer the functionality to provide such a method.

Alternative to the interface choice, we could also decide to make this method exclusive to the openshift connector, similar to the `OpenWithHTTPClient`, to only expose it for library usage.

```release-note
Added CheckEndpointConnector, provide the ability to test
```


dhaus67:
This might be interesting to you @nabokihms , as it is somewhat similar to #2430 .

nabokihms:
I've looked through this PR, and, in my opinion, it is not right to add more public interfaces only to use Dex as a library. The best for me is to add **optional** endpoints health check to the `Open` method. It just makes more sense.

@sagikazarmark, could you also please have a look at this code?

dhaus67:
@nabokihms thanks for the feedback.
I can see why we don't want to expose a public interface for that as well. I've gone the way you mentioned with regards to the _optional_ endpoints health check via options that can be passed to `OpenWithOptions`, allowing for customization within library usage.

Let me know what you think about it.

nabokihms:
Thank you for the quick fix, @dhaus67. I'm still waiting for the second opinion here. We need to figure out what to do with this (and probably with further) Openshift Connector improvements.

dhaus67:
@nabokihms @sagikazarmark any update on your end regarding this?

PR #2522, opened 2022-05-20
OIDC connector: Support cases where there is no id_token when using a refresh_token grant

Blorpy:

#### Overview

I broke https://github.com/dexidp/dex/pull/2505 with a bad rebase, this is a PR off a fixed branch, sorry for the confusion @nabokihms 

Allows for dex to function with IDPs that do not return an ID token in the reply when using a refresh_token grant.

Although this seems to be rare, the OIDC spec does not require that an id_token be present in the reply when using a refresh_token grant. Additional info in https://github.com/dexidp/dex/issues/2498.

This is a toggle in PingFed, so there is at least one common IDP out there where there won't always be a refresh token.

Closes https://github.com/dexidp/dex/issues/2498

Note that if a user has an IDP where ID tokens aren't present when using a refresh_token grant, AND also chooses not to enable the userinfo endpoint, this will choke. The claims have to come from somewhere, so I would say this really is just an unsupported configuration should someone choose to try this. The OIDC spec says that userinfo must always contain the sub additionally.

I tried to keep behaviour as close as possible to current. Although its extra code, and probably not needed, I chose to prefer the sub from the ID token if its present, over assigning the one from the userinfo.

```release-note
OIDC connector: Support cases where there is no id_token when using a refresh_token grant
```

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -296,7 +296,7 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 	const subjectClaimKey = "sub"
 	subject, found := claims[subjectClaimKey].(string)
 	if !found {
-	    return identity, fmt.Errorf("missing \"%s\" claim", subjectClaimKey)
+		return identity, fmt.Errorf("missing \"%s\" claim", subjectClaimKey)
 	}
 
 	userNameKey := "name"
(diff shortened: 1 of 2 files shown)


nabokihms (approved):
Looks good to me! Thank you for the contribution, @Blorpy. Great work.

PR #2524, opened 2022-05-20
Limit the amount of objects we attempt to GC on each cycle

kellyma2:
If something causes the number k8s resources to increase beyond a
certain threshold, garbage collection can fail because the query to
retrieve those resources will time out, resulting in a perpetual cycle
of being unable to garbage collect resources.

In lieu of trying to get *every* object each cycle, we can limit the
number of resources retrieved per GC cycle to some reasonable number.

### DIFF
--- a/storage/kubernetes/client.go
+++ b/storage/kubernetes/client.go
@@ -96,18 +96,18 @@ func (cli *client) urlForWithParams(
 	} else {
 		p = path.Join(basePath, apiVersion, resource, name)
 	}
-	if strings.HasSuffix(cli.baseURL, "/") {
-		return cli.baseURL + p
-	}
-
-	r := cli.baseURL + "/" + p
 
 	encodedParams := params.Encode()
+	paramsSuffix := ""
 	if len(encodedParams) > 0 {
-		return r + "?" + encodedParams
+		paramsSuffix = "?" + encodedParams
 	}
 
-	return r
+	if strings.HasSuffix(cli.baseURL, "/") {
+		return cli.baseURL + p + paramsSuffix
+	}
+
+	return cli.baseURL + "/" + p + paramsSuffix
 }
 
 func (cli *client) urlFor(apiVersion, namespace, resource, name string) string {
@@ -201,7 +201,7 @@ func (cli *client) listN(resource string, v interface{}, n int) error {
 }
 
 func (cli *client) list(resource string, v interface{}) error {
-	return cli.listN(resource, v, -1)
+	return cli.get(resource, "", v)
 }
 
 func (cli *client) post(resource string, v interface{}) error {
--- a/storage/kubernetes/client.go
+++ b/storage/kubernetes/client.go
@@ -15,6 +15,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"path"
 	"strconv"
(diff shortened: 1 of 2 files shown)


nabokihms:
@kellyma2 thank you for opening this PR. It totally makes sense. I will try to take a look on it after the weekend.

kellyma2:
@nabokihms any comments?

storage/kubernetes/storage.go
```diff
@@ -40,6 +40,10 @@ const (
 	resourceDeviceToken     = "devicetokens"
 )
 
+const (
+	gcResultLimit = 10000
```
nabokihms, line 44: Let's make it 500 as in Kubernetes client-go 
https://github.com/kubernetes/client-go/blob/8344b1c856d6484a78135387551c2c88a3f65a1e/tools/pager/pager.go#L31

storage/kubernetes/client.go
```diff
@@ -97,7 +99,19 @@ func (cli *client) urlFor(apiVersion, namespace, resource, name string) string {
 	if strings.HasSuffix(cli.baseURL, "/") {
 		return cli.baseURL + p
 	}
-	return cli.baseURL + "/" + p
+
```
nabokihms, line 102: If URL has a suffix /, then limits will not be applied.
https://github.com/dexidp/dex/blob/8ffb53869ec552605fb1af6816bb3936e0963aec/storage/kubernetes/client.go#L99-L101

storage/kubernetes/client.go
```diff
@@ -176,8 +189,19 @@ func (cli *client) getResource(apiVersion, namespace, resource, name string, v i
...
+func (cli *client) listN(resource string, v interface{}, n int) error {
+	params := url.Values{}
+	params.Add("limit", fmt.Sprintf("%d", n))
+	u := cli.urlForWithParams(cli.apiVersion, cli.namespace, resource, "", params)
+	return cli.getURL(u, v)
+}
+
 func (cli *client) list(resource string, v interface{}) error {
-	return cli.get(resource, "", v)
+	return cli.listN(resource, v, -1)
```
nabokihms, line 204: Personally, I wouldn't say I like having `limit=-1` in all list requests besides GC.

The one option to solve the problem is to avoid `urlForWithParams` for the plain list method.
```go
func (cli *client) list(resource string, v interface{}) error {
	u := cli.urlFor(cli.apiVersion, cli.namespace, resource, "")
	return cli.getURL(u, v)
}
```

nabokihms (changes requested):
@kellyma2, sorry for the delay. My vacation came all of a sudden 😅 

Thank you for the great work. I have a couple of suggestions, but overall it looks excellent.

kellyma2:
Thanks for the feedback @nabokihms . I'll check it out this afternoon!

storage/kubernetes/storage.go
kellyma2: Sure.

storage/kubernetes/client.go
kellyma2: fixed

storage/kubernetes/client.go
kellyma2: Thought about this a bit more.  I don't think this necessarily needs to change in any form - was just trying to minimize recreating large blocks of code, but I think the other adjustments cover this anyway.

kellyma2:
@nabokihms Updated the PR

kellyma2:
> LGTM! @kellyma2, could you please sign comments to pass the DCO check so we can merge this?

@nabokihms Done!

PR #2122, opened 2021-05-15
Google: Add functionality to impersonate GSuite Admin identity without JSON

NaurisSadovskis:
#### Overview

Hi folks, adding an option to impersonate GSuite Admin identity without service account key JSON. This functionality has been recently added as part of [google-api-go-client v0.46.0](https://github.com/googleapis/google-api-go-client/releases/tag/v0.46.0)

#### What this PR does / why we need it
This feature is primarily aimed at folks running within Google Cloud Platform (GCE/GKE):
* Removes the need for (long-lived) user-managed service account keys when configuring Dex

Addresses: #1756 

#### Special notes for your reviewer
I have not tested this with an actual GSuite tenant, as I don't have one, but can test it sometime next week. 

#### Does this PR introduce a user-facing change?
```release-note
* Add ability to impersonate GSuite Admin identity without JSON when running in Google Cloud Platform
```

PR #2530, opened 2022-05-23
Implement Application Default Credentials for the google connector

### LINKED ISSUE #1756: Support workload identity in Google connector
Workload identity accessing GoogleAPIs without having a static service account token. GKE metadata server available in each pod is used to fetch tokens.

Workload identity tokens work as a replacement for GSA token, they are not accepted by GSuite Admin API, because Admin API requires impersonalization for GSuite admin. Impersonalization IS possible with workload identity token, hence this proposal.

Example use of GSuite admin impersonalization can be found here: https://github.com/salrashid123/oauth2#impersonated-credentials-with-domain-wide-delegation

So proposal is following:
 - if google connector config specifies `adminEmail`, but does not specifies `serviceAccountFilePath` Google connector attempts to use workload identity token to impersonate GSuite Admin and create Admin API client instance

ichbinfrog:
#### Overview

Implements [workload identity](https://cloud.google.com/kubernetes-engine/docs/concepts/workload-identity) to the google connector. 

Closes #1756 by implementing workload identity for the google connector. 

This PR has been tested by:
- Building the dex image
- Injecting the image into the Argocd helm chart with Google Groups integration (see related [doc](https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/google/#openid-connect-plus-google-groups-using-dex))
- Testing if Groups are visible within Argocd

I also wanted to add unit test but it required some changes to the `Open` interface to add  `option.ClientOptions...` or `interface{}...`.

Another similar [PR](https://github.com/dexidp/dex/pull/2122) exists but testing it yields an `404: Domain not found` error.

```release-note
Add application default credentials for the google workspace connector
```


nabokihms:
@ichbinfrog sorry for the delay with this. The idea looks great! I only wonder about testing, adding a test is still desired. Besides it, lgtm.

connector/google/google.go
```diff
@@ -279,37 +279,36 @@ func (c *googleConnector) getGroups(email string, fetchTransitiveGroupMembership
...
-
-	srv, err := admin.NewService(ctx, option.WithHTTPClient(client))
+	if serviceAccountFilePath == "" {
+		credential, err := google.FindDefaultCredentials(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch application default credentials: %w", err)
+		}
+		jsonCredentials = credential.JSON
+	} else {
+		jsonCredentials, err = os.ReadFile(serviceAccountFilePath)
```
nabokihms, line 301: Maybe we should log a warning, that default credentials file is used.

nabokihms:
Moving this one to the v2.34 milestone.

jinnjwu:
our dex is hosting in gke, we need this feature too

ichbinfrog:
Thanks for the review @nabokihms and sorry for the delay, I've added the warning log as well as some unit tests for the logic. 

Regarding the unit test implementation, I've given it a try. It did however require that the `Open` interface become to allow creating the Google Directory Service pointing to a local endpoint:

```golang
Open(id string, logger log.Logger, opts ...interface{}) (connector.Connector, error)
```

In the long term, this change would allow us to add further unit testing of the google connector implementation by mocking the API similarly to what's currently in the oauth connector.

connector/google/google.go
```diff
@@ -279,37 +289,37 @@ func (c *googleConnector) getGroups(email string, fetchTransitiveGroupMembership
...
-	config.Subject = email
+	var jsonCredentials []byte
+	var err error
 
 	ctx := context.Background()
-	client := config.Client(ctx)
-
-	srv, err := admin.NewService(ctx, option.WithHTTPClient(client))
+	if serviceAccountFilePath == "" {
+		logger.Warn("the application default credential is used since the service acocunt file path is not used")
```
shuker85, line 305: small typo on `account`

connector/google/google_test.go
```diff
@@ -0,0 +1,146 @@
...
+	// 	json.NewEncoder(w).Encode(&admin.Groups{
+	// 		Groups: []*admin.Group{},
+	// 	})
+	// })
+	return httptest.NewServer(mux)
+}
+
+func newConnector(config *Config, serverURL string) (*googleConnector, error) {
+	log := logrus.New()
+	conn, err := config.Open("id", log, option.WithEndpoint(serverURL))
```
nabokihms, line 30: If only endpoint changes are required for now, I'd suggest making it through the private connector property avoiding global changes.

nabokihms (changes requested):
Looks great! I have only one convern left. Despite it, everything is ok.

connector/google/google_test.go
ichbinfrog: Agreed, in hindsight, the behavior to be tested in this PR does not even require that option.

ichbinfrog:
There's no problem regarding the delay. My PR updates are also extremely delayed 😅

nabokihms (approved):
@ichbinfrog thanks! LGTM.

PR #2533, opened 2022-05-24
fix: add notification about groups access to the Grant Access page

nabokihms:

#### Overview

Show a warning that the application requests user groups access

Groups list can contain sensitive data which should not be exposed for some applications

```release-note
Notify about requesting group access.
```


sagikazarmark (approved):
While claims can be anything, group information is not standard OIDC and Dex only supports groups if the upstream provider does too (which is not always the case).

Not saying we shouldn't add it, but it's worth noting it's not a standard claim. Maybe just add a comment explaining we added it for convenience?

Issue #2537, opened 2022-05-26
Device Code Flow does not return `application/json` in `Content-Type` header


ShivanshVij:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.31.2

### Storage Type

etcd

### Installation Type

Official container image, Other (specify below)

### Expected Behavior

When using Dex with [Github's official Device code flow library for Go](https://github.com/cli/oauth), the flow does not work an instead falls back to using the `Web Auth` flow. 

It **should** work just fine as long as Dex implements the Device Code Flow Spec (https://datatracker.ietf.org/doc/html/rfc8628#section-3.3 and https://datatracker.ietf.org/doc/html/rfc6749#section-5.1), but instead the library returns an error saying that it cannot retrieve the `verification URL`. 

### Actual Behavior

After digging into the code on both sides, I've confirmed that when  the `device code flow` authorization grant that was implemented during [this PR](https://github.com/dexidp/dex/pull/1706), the [handleDeviceCode](https://github.com/dexidp/dex/blob/v2.31.2/server/deviceflowhandlers.go#L142) function did not append a `content-type: application/json` header to the [JSON-encoded response](https://github.com/dexidp/dex/blob/v2.31.2/server/deviceflowhandlers.go#L142). 

Because of this, the response header is just `text/plain` as can be seen in the attached screenshot. 

![Screen Shot 2022-05-26 at 3 26 59 PM](https://user-images.githubusercontent.com/27162109/170562439-a3d7ceda-ee59-4bd4-9843-07b40dc86eba.png)

### Steps To Reproduce

1. Install and set up Dex with the Github Provider and start it listening on `localhost:8080`
2. Set up a simple "test" program using Github's OAuth library:
```go
       host := &oauth.Host{
		DeviceCodeURL: "http://localhost:8080/device/code",
		AuthorizeURL:  "http://localhost:8080/auth",
		TokenURL:      "http://localhost:8080/token",
	}
	flow := &oauth.Flow{
		Host:         host,
		ClientID:     "000000",
[...]

PR #1706, opened 2020-05-11
Implementing the OAuth2 Device Authorization Grant

justin-slowik:
Fixes #911

This PR adds functionality for the [OAuth 2 Device Flow](https://oauth.net/2/device-flow/)

I initially wrote this a few months back for my company's private fork, but have rebased it and would like to submit if possible.

Basic usage:
* Device sends a POST to the /device/code endpoint with the client id/secret, and scopes.  Dex returns a device code, user code, expire time, query interval, verification uri and "complete" verification uri (pre-populates the verification page with the user code)
`curl https://127.0.0.1/dex/device/code -d client_id=platform_device -d scope="openid profile email offline_access"
`
> {
   "device_code": "powkm5zcpz7njeoyh63fminzxrcb5ystrjocjx5ch7s47ezmpli",
   "user_code": "FQXZ-HRCG",
   "verification_uri": "http://127.0.0.1/dex/device",
   "verification_uri_complete": "http://127.0.0.1/dex/device?user_code=FQXZ-HRCG",
   "expires_in": 300,
   "interval": 5
}

* The device periodically (at the rate defined by "interval") queries the /device/token endpoint, including the device_code in the POST body.  If the end user has not gone to the device_verification_uri and logged in yet, the error "authorization_pending" will be returned.  If the /device/token endpoint is queried with the device code faster than a rate defined by "interval", a "slow_down" message is retuned, and the wait interval is increased by an additional 5 seconds (from 5 seconds, to 10 seconds, to 15, etc)

* As the device queries the /device/token endpoint, the user can go to the verification URI, enter their verification code, then log in using the standard authorization code flow.  Upon successful login, the user will be alerted to return to their device.

* After the user has logged in, when the advice hits the /device/token endpoint, the access_token will be returned.

Just a note:   rate limiting is only included for the /device/token endpoint.  For our use case, all other rate limiting is currently handled at the API gateway.  

I do not have a large amount of time to do additional work on this branch due to other commitments (other than possibly fixing showstopping issues), so people are welcome to take over.

PR #2538, opened 2022-05-26
Fixes https://github.com/dexidp/dex/issues/2537

ShivanshVij:
Signed-off-by: Shivansh Vij <shivanshvij@outlook.com>

Closes https://github.com/dexidp/dex/issues/2537

This is a bug that was introduced in [this PR](https://github.com/dexidp/dex/pull/1706), and is where the `handleDeviceCode` handler does not set the `content-type` header in the response to `application/json` - because of this, the flow does not conform to the RFC (https://datatracker.ietf.org/doc/html/rfc8628#section-3.3 and https://datatracker.ietf.org/doc/html/rfc6749#section-5.1). 

I've also added a test case to check for this exact bug. 

```release-note
NONE
```

### DIFF
--- a/server/deviceflowhandlers.go
+++ b/server/deviceflowhandlers.go
@@ -141,6 +141,10 @@ func (s *Server) handleDeviceCode(w http.ResponseWriter, r *http.Request) {
 		// https://tools.ietf.org/html/rfc8628#section-3.2
 		w.Header().Set("Cache-Control", "no-store")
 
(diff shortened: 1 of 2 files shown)


nabokihms:
Because our implementation is not compliant with the spec, I think we can accept the PR without worrying about being a breaking change. 

@ShivanshVij, good catch! Thank you for your contribution.

PR #2540, opened 2022-05-30
Fixes Public Clients with Device Code Authorization Flow #2539

### LINKED ISSUE #2539: Device Code Flow does not respect Public Clients
### Preflight Checklist

### Version

2.31.2

### Storage Type

etcd

### Installation Type

Official container image

### Expected Behavior

When performing a `Device Code` Authorization flow (as defined in the spec: https://datatracker.ietf.org/doc/html/rfc8628#section-3.1), the flow should complete correctly and return an authorization response.

### Actual Behavior

Dex returns an `Invalid Client` error at the last stage of the authorization flow if the original request did not include a `Client Secret` because of this line: https://github.com/dexidp/dex/blob/master/server/deviceflowhandlers.go#L287 . 

### Steps To Reproduce

1. Set up Dex with any OAuth 2.0 provider
2. Perform a Device Code Flow request using any spec-conformant library (like https://github.com/cli/oauth) 

### Additional Information

The official [RFC](https://datatracker.ietf.org/doc/html/rfc8628#section-3.1) for this flow specifies that the `client_secret` is not a required field (or even an optional one) for this authorization flow. 

My suggestion is that we forgo checking the client secret for clients that are `Public`. This works because spec-conforming libraries won't send it anyways because they are meant to be used with `Public` clients. 

### Configuration

_No response_

### Logs

_No response_

ShivanshVij:
#### Overview

This PR makes it so that we check if a client is marked as `Public` in the Device Code Authorization flow. If it is, we skip checking that the client secret in the Device Code Authorization flow is the same as the client's actual secret. 

Closes #2539 

When performing a Device Code Authorization flow (as defined in the spec: https://datatracker.ietf.org/doc/html/rfc8628#section-3.1), the flow should complete correctly and return an authorization response.

Instead, Dex returns an `Invalid Client` error at the last stage of the authorization flow if the original request did not include a `Client Secret`. 

The official [RFC](https://datatracker.ietf.org/doc/html/rfc8628#section-3.1) for this flow specifies that the client_secret is not a required field (or even an optional one) for this authorization flow. 

The reason we shouldn't just remove the `Client Secret` verification entirely, is because the spec also defines that a client may authenticate the way it normally does (ie. with a `client secret`) if it wishes to. To that end, this PR allows the client secret to be a requirement for private clients, but for public clients (that don't authenticate using the client secret) it conforms to the spec. 

Existing implementations should continue to work as expected with this change as long as they were conforming to the  official [RFC](https://datatracker.ietf.org/doc/html/rfc8628#section-3.1).

```release-note
NONE
```


nabokihms:
@ShivanshVij, thank you for submitting the PR! I swear that I saw public clients working with the device flow with my eyes 😅 The code looks good to me, yet I would like check the problem first before merging this PR.

ShivanshVij:
I'm not sure how other libraries do it, but the official [github oauth](https://github.com/cli/oauth) library that they use in the github cli definitely does not send the `Client Secret`: https://github.com/cli/oauth/blob/main/device/device_flow.go#L59

nabokihms:
@ShivanshVij I checked this with a current master branch, and it looks like everything is fine with public clients.

1. The following static client is defined in the dex config file:
```yaml
staticClients:
- id: example-device-client
  redirectURIs:
  - /device/callback
  name: 'Static Client for Device Flow'
  public: true
 ```

2. Make a request to the device code endpoint to get the code. 
```bash
curl --request POST \
  --url http://127.0.0.1:5556/dex/device/code \
  --header 'Accept: application/json' \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data-urlencode 'client_id=example-device-client' \
  --data-urlencode 'scope=openid profile offline_access' 
```
```json  
{
   "device_code": "hub4vrwuuhqtyldl4n4fi47y7hhwvgt6bxwibrxulnf6ewjasye",
   "user_code": "WDFN-RBFC",
   "verification_uri": "http://127.0.0.1:5556/dex/device",
   "verification_uri_complete": "http://127.0.0.1:5556/dex/device?user_code=WDFN-RBFC",
   "expires_in": 300,
   "interval": 5
}
```

3. Go to the `verification_uri_complete` and submit the token request.

4. Get the token.
```bash
curl --request POST \
  --url http://127.0.0.1:5556/dex/token \
  --header 'Accept: application/json' \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data-urlencode 'client_id=example-device-client' \
  --data-urlencode 'scope=openid profile offline_access' \
  --data-urlencode 'grant_type=urn:ietf:params:oauth:grant-type:device_code' \
  --data-urlencode 'device_code=hub4vrwuuhqtyldl4n4fi47y7hhwvgt6bxwibrxulnf6ewjasye'
```
```json
{
  "access_token": "eyJhbGciOiJSUzI1NiIsImtpZCI6IjA1MTNhMjAwOWFlZTFmMWVkZTY4YmIzYzllYmEzYTQwNGQxZjFhYmIifQ.eyJpc3MiOiJodHRwOi8vMTI3LjAuMC4xOjU1NTYvZGV4Iiwic3ViIjoiQ2cwd0xUTTROUzB5T0RBNE9TMHdFZ1J0YjJOciIsImF1ZCI6ImV4YW1wbGUtZGV2aWNlLWNsaWVudCIsImV4cCI6MTY1Nzk0MjYyNCwiaWF0IjoxNjU3ODU2MjI0LCJhdF9oYXNoIjoibVVDbGotNmdrcnVGLXRORkFGNGtOQSIsIm5hbWUiOiJLaWxnb3JlIFRyb3V0In0.RsXVuWW4u0UA9rXZCT0F01ew7P7GpgBWpMZJmYU7ei6l5d-FhUA4_IUgHbtS-j9MC0FZV7aTgHAJQR5-c9M5yD5xopVTjqmPz2AnZh4kSjrU-_KFsy7fCvaE9lxPgtE1beOYkbMXG5KUVFbEwcqVbTswyRwyb3zmdRWBNnq7NjyN1AK_MUkinQsdoHJZu8Tk8XL1eumzWm1ecCSdmYtQdkFgkmNHWSsa8z7TWetPzJq2NfVVynfb7v74E1q-g-KP_68_o3HjWh0lOeSA1xAscgXUxansNWAMHPha9gGnKfZCPS12We0OvWHCo13KjhelPWKeIOV10pJHVurw-xMt9A",
  "token_type": "bearer",
  "expires_in": 86399,
  "refresh_token": "Chl5NGpnc3M2aDZ6aWVldmtwc3Zzd2RkM2ZvEhlwYzJoNTc2aTJ0ZW9id3YyYWhsazNhZHRy",
  "id_token": "eyJhbGciOiJSUzI1NiIsImtpZCI6IjA1MTNhMjAwOWFlZTFmMWVkZTY4YmIzYzllYmEzYTQwNGQxZjFhYmIifQ.eyJpc3MiOiJodHRwOi8vMTI3LjAuMC4xOjU1NTYvZGV4Iiwic3ViIjoiQ2cwd0xUTTROUzB5T0RBNE9TMHdFZ1J0YjJOciIsImF1ZCI6ImV4YW1wbGUtZGV2aWNlLWNsaWVudCIsImV4cCI6MTY1Nzk0MjYyNCwiaWF0IjoxNjU3ODU2MjI0LCJhdF9oYXNoIjoiMjR2TE96RFdfeEJqS1p1N3pXWld0dyIsImNfaGFzaCI6InI5N2plandsQTVWR2w1VTJLbkpZWUEiLCJuYW1lIjoiS2lsZ29yZSBUcm91dCJ9.Q7GorMxrZsey3sJP7UePgZYZo65l2Bdx7U4gbxDrAseSlFjuwTMAv5pNT6nAz32Y6-HR7aVuNEQBA0QWoSe-jwdQmaVeHAYz_mcgXEpjzSGdonAcae53p-sp1tb_CfBnaQd9LuulhqQF9N0siiTT1UsuhBG_9OFCt6t9pBaHryiNb_OlzT58un08lPeCzJPNEZmQWHJmGV_AkJ9DL8mMvEX9Pm1_ebb33OGdOc0wduQqJ91w0zlEU5lM7JwTusUGINcivtnQrDbdMZgX0xOOq_S5VuHdcuEXCP9lQhPRLhWmjW-KVUJBvGOgJvcgeYiLfRxRcfQvXHDsIx-AuyNQGw"
}
```

Maybe I've missed something?

ShivanshVij:
I went through your example and something doesn't make sense. 

When you make the request to `http://127.0.0.1:5556/dex/device/code`, the following handler runs: https://github.com/dexidp/dex/blob/master/server/deviceflowhandlers.go#L60

In this handler, we get the Client Secret (https://github.com/dexidp/dex/blob/master/server/deviceflowhandlers.go#L74) and save that into the device request (https://github.com/dexidp/dex/blob/master/server/deviceflowhandlers.go#L93) which gets saved in the storage (https://github.com/dexidp/dex/blob/master/server/deviceflowhandlers.go#L98)

Next, we go through the OAuth flow in the browser and the OAuth provider does a callback to the dex callback endpoint (`/device/callback`) which calls the following handler: https://github.com/dexidp/dex/blob/master/server/deviceflowhandlers.go#L98

This handler then grabs the device request that we saved into storage (https://github.com/dexidp/dex/blob/master/server/deviceflowhandlers.go#L98) and retrieves the client from the storage (https://github.com/dexidp/dex/blob/master/server/deviceflowhandlers.go#L285) and then checks the client secret that was in the original device request against the stored client's secret (https://github.com/dexidp/dex/blob/master/server/deviceflowhandlers.go#L295).

If they don't match, it returns an error. 

I think the difference here is that when you register the static client, you're not specifying a `secret` for the client (and you don't have to if it's a public client: https://github.com/dexidp/dex/blob/master/cmd/dex/serve.go#L204). 

So the check is still happening, but we're just comparing two empty strings together in https://github.com/dexidp/dex/blob/master/server/deviceflowhandlers.go#L295, as opposed to skipping the check entirely. 

So I guess the question is: Are public clients expected to have an empty secret (because right now they do)?

nabokihms:
Yeah, that's it. I always thought there should be no client secret for public clients (probably, this assumption comes through the whole Dex codebase, not only device flow related code).

There is a validation that the secret field is mandatory for confidential clients, but for public clients, it is optional
https://github.com/dexidp/dex/blob/6da5187b474a967d4b096c5a137de048f00f6bbc/cmd/dex/serve.go#L204-L206

One approach to resolve the issue is to ignore secrets for public clients, but it will be a breaking change. I think we can just add a notice to the doc about public clients' nuances.

ShivanshVij:
Yep, this makes perfect sense to me. I'll go ahead and close this PR and I'll update the accompanying issue to be about updating the docs to make it clear that public clients are not expected to have a secret and their secret will be ignored.

PR #2562, opened 2022-06-21
feat: oidc accept groups as string

susguzman:
#### Overview

The current OIDC connector only support the groups claim as an array, but if the response is a string, because the user only has one group, the groups cannot be mapped. I was configuring Dex with an OIDC provider that response a string when the user has only one group, so it's needed a way to handle this scenarios. 

- Allows the group claim be a string and an array.
- It's need because there are some providers that response with a string when the user only have one group. 

NONE

```release-note
OIDC Connector: Support for groups as a string.
```


nabokihms:
My bad, accidently linked unrelated PR.

nabokihms:
Closed by https://github.com/dexidp/dex/pull/2639

PR #2575, opened 2022-06-29
add PKCE support to device code flow

bobcallaway:
This adds support for the PKCE `code_challenge` and `code_verifier` values defined in RFC7636 when using the device code flow. This is especially useful when a client secret can not be used for SPA. 

### DIFF
--- a/server/deviceflowhandlers.go
+++ b/server/deviceflowhandlers.go
@@ -71,6 +71,12 @@ func (s *Server) handleDeviceCode(w http.ResponseWriter, r *http.Request) {
 		clientID := r.Form.Get("client_id")
 		clientSecret := r.Form.Get("client_secret")
 		scopes := strings.Fields(r.Form.Get("scope"))
+		codeChallenge := r.Form.Get("code_challenge")
+		codeChallengeMethod := r.Form.Get("code_challenge_method")
+
+		if codeChallengeMethod == "" {
+			codeChallengeMethod = codeChallengeMethodPlain
+		}
 
 		s.logger.Infof("Received device request for client %v with scopes %v", clientID, scopes)
 
@@ -106,6 +112,10 @@ func (s *Server) handleDeviceCode(w http.ResponseWriter, r *http.Request) {
 			Expiry:              expireTime,
 			LastRequestTime:     s.now(),
 			PollIntervalSeconds: 0,
+			PKCE: storage.PKCE{
+				CodeChallenge:       codeChallenge,
+				CodeChallengeMethod: codeChallengeMethod,
+			},
 		}
 
 		if err := s.storage.CreateDeviceToken(deviceToken); err != nil {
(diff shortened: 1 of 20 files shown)


storage/storage.go
```diff
@@ -427,4 +427,5 @@ type DeviceToken struct {
 	Expiry              time.Time
 	LastRequestTime     time.Time
 	PollIntervalSeconds int
+	PKCE                PKCE
```
nabokihms, line 430: It seems that changes to the database schema are required.  Otherwise, PKCE settings will not persist. 
It would also be nice to add [a conformance test](https://github.com/dexidp/dex/blob/f49e7bc218ebffdc5bfb800c53c06d273958eef2/storage/conformance/conformance.go#L991-L1032) for the storage.

Pieces of a code that should be changed (and the corresponding Get and Update functions too):
https://github.com/dexidp/dex/blob/f49e7bc218ebffdc5bfb800c53c06d273958eef2/storage/ent/client/devicetoken.go#L11-L25

https://github.com/dexidp/dex/blob/f49e7bc218ebffdc5bfb800c53c06d273958eef2/storage/kubernetes/types.go#L814-L831

https://github.com/dexidp/dex/blob/f49e7bc218ebffdc5bfb800c53c06d273958eef2/storage/sql/crud.go#L927-L944

https://github.com/dexidp/dex/blob/f49e7bc218ebffdc5bfb800c53c06d273958eef2/storage/etcd/types.go#L286-L295

nabokihms (changes requested):
@bobcallaway, thank you for opening this PR. It looks excellent. I have only one concern regarding storage schema changes. Please, look through them when you have time.

nabokihms:
@bobcallaway Thanks, the code seems ok to me. I will do some manual tests in the cluster and merge it if everything is ok.

server/deviceflowhandlers.go
```diff
@@ -71,6 +71,12 @@ func (s *Server) handleDeviceCode(w http.ResponseWriter, r *http.Request) {
 		clientID := r.Form.Get("client_id")
 		clientSecret := r.Form.Get("client_secret")
 		scopes := strings.Fields(r.Form.Get("scope"))
+		codeChallenge := r.Form.Get("code_challenge")
+		codeChallengeMethod := r.Form.Get("code_challenge_method")
+
+		if codeChallengeMethod == "" {
```
nabokihms, line 77: I've found a problem here. The cause is that we do not validate `code_challenge_method` on parsing a device auth request.

How to reproduce:
1. Request the `/device/code` endpoint.There will be no error in the response.
```sh
 curl --request POST \
  --url http://127.0.0.1:5556/dex/device/code \
  --header 'Accept: application/json' \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data-urlencode 'client_id=example-device-client' \
  --data-urlencode 'scope=openid profile offline_access' \
  --data-urlencode 'code_challenge=test' \                                      
  --data-urlencode 'code_challenge_method=ivalid-method' 
```
2. Authenticate.
3. Get the token.
```sh
curl --request POST \                              
  --url http://127.0.0.1:5556/dex/token \
  --header 'Accept: application/json' \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data-urlencode 'client_id=example-device-client' \
  --data-urlencode 'scope=openid profile offline_access' \
  --data-urlencode 'grant_type=urn:ietf:params:oauth:grant-type:device_code' \
  --data-urlencode 'device_code=xxxxxxxxxxxxxxxxxxx' \
  --data-urlencode 'code_verifier=test'
```
4. There will be a `500 Internal Server Error` in response.
```
< HTTP/1.1 500 Internal Server Error
< Content-Length: 24
< Content-Type: application/json
< Date: Tue, 26 Jul 2022 14:05:04 GMT
< 
* Connection #0 to host 127.0.0.1 left intact
{"error":"server_error"}%  
```

I think we must validate the `code_challenge_method` values is either `plain`, `S256`, or empty. Corresponding validation for the authcode grant can be found here:
https://github.com/dexidp/dex/blob/50dc2f55187a0e72d73809485725e45eba42b4eb/server/oauth2.go#L475-L478

nabokihms (changes requested):
@bobcallaway sorry for the delay. I tested this PR manually, and everything's fine besides one minor issue. Please, take a look at it when you have time.

server/deviceflowhandlers.go
bobcallaway: It is validated on the token exchange side, but I added a check on the device code endpoint to fail earlier if the value is invalid.

nabokihms (approved):

nabokihms:
@bobcallaway thanks for this. LGTM.

PR #2582, opened 2022-07-06
add config to explicitly set scopes for microsoft connector

bobcallaway:
#### Overview

Adds a new configuration option for the Microsoft connector to explicitly declare the scopes to be requested in tokens

[Users have noted](https://github.com/sigstore/fulcio/issues/157) that requesting a default scope of `user.read` (aka `profile`) vs `["openid", "email"]` may be too broad for a simple ID token. This PR allows for the Microsoft connector configuration to explicitly set the minimum scope to something other than `user.read` (per Microsoft [documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent)), which is available in other similar connectors.

New config option for connector

### DIFF
--- a/connector/microsoft/microsoft.go
+++ b/connector/microsoft/microsoft.go
@@ -58,6 +58,8 @@ type Config struct {
 	// For valid values, see https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code.
 	PromptType string `json:"promptType"`
 	DomainHint string `json:"domainHint"`
+
+	Scopes []string `json:"scopes"` // defaults to scopeUser (user.read)
 }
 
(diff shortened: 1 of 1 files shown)


nabokihms:
@bobcallaway hello! I'd like to take this one to 2.33. Could you please rebase the branch onto master and, if it is possible, open a PR to also fix [the doc](https://github.com/dexidp/website)?

nabokihms (approved):

brokenjacobs:
This still manually takes the directory.read_all scope to the end of the scope list. If this could be dropped when scopes are specified, it would allow using the default scope:
```https://graph.microsoft.com/.default ```
Right now this can't be used at the same time as user.read.all.

Basically... if the user passes scopes, just use them. Don't keep adding the extra one onto the end.

PR #2583, opened 2022-07-07
[Connector/SAML] Support metadata xml discovery

kenchan0130:
#### Overview

Some values are now automatically retrieved from IdP metadata URLs.

In SAML connector, the implementation of the metadata discovery function remained a TODO in source code.

A new config option for SAML connector.
This is not a breaking change.

In https://dexidp.io/docs/connectors/saml/#configuration, `idpMetadataURL` attribute must be optionally added to the config value.


kenchan0130:
@nabokihms 

Could you check and review this pull request?

PR #2586, opened 2022-07-12
Add domainHint parameter to Microsoft Connector

### LINKED ISSUE #2585: Microsoft Connector adds extra interactive step in login for tenants with known domains
### Preflight Checklist

### Problem Description

When a user with multiple Microsoft sessions uses the Microsoft connector in Dex, the `/authorize` page displays an interactive UI rather than a seamless redirect.  The page with title "Pick an account" allows the user to select from a list of accounts known to Microsoft.

PickAnAccount-censored

This account list contains a mix of tenant and personal accounts, including accounts without current sessions.
Upon selecting an account, the authentication proceeds normally.

In cases where the specific `tenant` parameter is configured for the Microsoft connector, the administrator is intentionally restricting access to users in the specified tenant.  Should a user select an account outside of the tenant on the "Select an account" page, the request will be rejected within Microsoft's OAuth flow.

**For these types of tenant deployments, Dex should be configurable so that the user may login using an existing session at Microsoft without seeing this prompt.**

### Proposed Solution

Microsoft offers several "hinting" strings for an application to provide at the Authorization code redirect as query parameters.  The simplest is the `domain_hint` parameter, which allows the application to provide the user email's domain name.

[Microsoft's "Request an authorization code" documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code) excerpt:

> `domain_hint` | optional | If included, the app skips the email-based discovery process that user goes through on the sign-in page, leading to a slightly more streamlined user experience. For example, sending them to their federated identity provider. Apps can use this parameter during reauthentication, by extracting the tid from a previous sign-in.
> -- | -- | --

**A parameter could be added in the Microsoft Connector configuration as `domainHint`.  Administrators who want to streamline this behavior can set this configuration string to their domain and Dex will be pass the string to Microsoft as `domain_hint`.**

No change to behavior for Dex deployments without this parameter.

### Alternatives Considered

Microsoft offers several options for hinting at the correct account, [documented here](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code)

If the user account is known to Dex before the redirect, Dex could supply the `login_hint` dynamically based on previous logins.  However, this adds significant implementation complexity and would likely require an opt-in configuration flag to prevent changes to existing behavior for the connector.

I think that the `domain_hint` approach still provides value, even if the `login_hint` approach is also added in the future, because it will work for logins without an existing session.

### Additional Information

I've tested the proposed solution in a fork and I am happy with the results.
I hope to PR the changes into Dex if the solution is acceptable to maintainers!

josephtknight:
#### Overview

This PR implements the solution proposed in https://github.com/dexidp/dex/issues/2585 to accelerate users logins using the Microsoft Connector when multiple sessions are present.

Closes #2585

Adds optional `domainHint` string configuration parameter to the Microsoft Connector
When `domainHint` is configured, the Microsoft Connector will encode the string as `domain_hint` in the Authorization Code URL.

Added 2 new tests to cover the `TestLoginURL` function in the Microsoft Connector.

```release-note
Adds optional `domainHint` configuration parameter to the Microsoft connector to streamline logins for single-tenant deployments.
```

### DIFF
--- a/connector/microsoft/microsoft.go
+++ b/connector/microsoft/microsoft.go
@@ -57,6 +57,7 @@ type Config struct {
 	// PromptType is used for the prompt query parameter.
 	// For valid values, see https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code.
 	PromptType string `json:"promptType"`
+	DomainHint string `json:"domainHint"`
 }
 
 // Open returns a strategy for logging in through Microsoft.
@@ -75,6 +76,7 @@ func (c *Config) Open(id string, logger log.Logger) (connector.Connector, error)
 		logger:               logger,
 		emailToLowercase:     c.EmailToLowercase,
 		promptType:           c.PromptType,
+		domainHint:			  c.DomainHint,
 	}
 	// By default allow logins from both personal and business/school
 	// accounts.
@@ -119,6 +121,7 @@ type microsoftConnector struct {
 	logger               log.Logger
 	emailToLowercase     bool
 	promptType           string
+	domainHint			 string
 }
 
 func (c *microsoftConnector) isOrgTenant() bool {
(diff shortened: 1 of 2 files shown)


josephtknight:
Thanks for feedback @nabokihms 
I've made the requested addition to the docs website.

CI also pointed out some lint issues in my first submission, I pushed a commit that should resolve them.
Please re-kick the workflow so I can make sure all the checks are clear.

Let me know if any other changes are needed!

nabokihms (approved):
@josephtknight looks good to me! Thank you a lot for contributing.

PR #2601, opened 2022-07-21
A connector for SMTP authentication

jeffallen:
#### Overview

This is a new connector that attempts to do an authentication against an SMTP server.

If your company subscribes to an e-mail service provider who does not support LDAP, another way to let "the email password" be your single sign-on is to use `dex` to talk SMTP over TLS to the service provider's SMTP server.

Yes, this is a new connector. If you guys would like to go forward with it, I will propose an update to the docs in the website repo as well.

```release-note
A new connector is available to authenticate against SMTP servers.
```


jeffallen:
Ping @sagikazarmark?

sagikazarmark:
Honestly, I find this weird. 😄 

On one hand, we are generally against adding more connectors to the core. We can't really maintain them all and it adds to the complexity.

(We really do have to prioritize #1907)

On the other hand, the amount of code here appears to be minimal.

I'm going to let other maintainers chime in on this one, but I believe we can let this one in with the following conditions:
- It's going to marked as experimental/community maintained and stays that way until we see it fit or we can remove it to an external plugin
- There will be no backwards compatibility support on it
- Lastly, we would refer to you whenever someone reports an issue about this particular connector

These are similar terms for what we agreed to in case of the CloudFoundry connector.

Are these acceptable for you @jeffallen ?

jeffallen:
Yes, I'm willing to support it until it is removed in favor of a plugin system. And then I will maintain a plugin instead, but in my company's repo.

I will update the PR, and include some docs, not sure when, maybe tomorrow depending on other jobs.

emanuelegissi:
I found this connector by chance.
It is indeed a great idea.
Do you intend to merge this connector?
Or should I search for another solution?
Thank you very much indeed, and happy festivities

jeffallen:
Well, it's working for me. But I may soon stop working at my current job, and would probably not continue working on this. So if you start using this, be aware you will be self-supporting.

emanuelegissi:
Hi Jeff, thank you very much for your answer.
Our SMTP server is on port 465, but supports only SSL, not TSL.

I'm not a go programmer, but it seems to me that this line of your connector uses TSL every time the port is 465. Am I right?

```
if p == "" || p == "465" {
		conn, err = tls.Dial("tcp", sc.cfg.Host, nil)
		if err != nil {
			return
		}
	} else {
		conn, err = net.Dial("tcp", sc.cfg.Host)
		if err != nil {
			return
		}
	}
```

How could I configure SSL instead of TSL on port 465?

Thank you again and happy festivities, if applicable.
Emanuele

PS maybe a good excuse to start learning go in 2024...

emanuelegissi:
Disclaimer: this is the first time in my life that I touch some golang code...

I studied go a little more, and adapted the following code:
https://gist.github.com/chrisgillis/10888032

With this code I can authenticate to my SMTP server and send an email.
So, it seems to me that your code should work in my case as well.

Next step, forking dex, merging your code and testing it. 

Again thank you in advance, and happy new year!

jeffallen:
I am closing this PR for now, because I'm unlikely to work further on it, and if @emanuelegissi picks it up, it's better if it's all his.

@emanuelegissi, for the purposes of the Apache license, you have my permission to use this PR as a base for your contribution. If you choose to give my employer Pie Aéronefs credit as well, that would be nice.

emanuelegissi:
Thank you very much Jeff.

If I will be able to become a contributor, I am sure going to acknowledge
Pie Aéronefs contribution in addition to your name.

In particular, I wish your employer is going to find the resources to keep
on the good work on electric planes, with definitely need that to save the
planet from climate doom.

Il mer 3 gen 2024, 14:02 Jeff Allen ***@***.***> ha scritto:

PR #2631, opened 2022-08-15
Add general support for adding additional auth params to OIDC connector

roddyherries:
#### Overview

Some IdPs support login features that are accessible via additional parameters on an authorization request. The current OIDC connector implementation provides no mechanism to configure additional authorization request parameters and hence the IdP specific features are out of reach. This PR adds support to for such additional params to the OIDC connector.

- Adds a new "additionalAuthRequestParams" map to the OIDC Config struct
- Populates the map from config at connector "open"
- Adds any additional auth params to the auth LoginURL
- Prevents mis-configuration attempts to override OIDC specified auth params.

Yes

```release-note
Support additional parameters on the OIDC auth request
```


connector/oidc/oidc_test.go
```diff
@@ -412,6 +415,178 @@ func TestRefresh(t *testing.T) {
...
+	cfg := Config{
+		ClientID:    "client",
+		RedirectURI: "callback",
+		AdditionalAuthRequestParams: map[string]string{
+			"acr_values": "not-so-fast",
+		},
+	}
+	_, err := testLoginURL(t, cfg, "1234")
+	assert.EqualErrorf(t, err, "parameter 'acr_values' is already managed by this connector", "")
+}
```
sagikazarmark, line 583: Would you mind merging these into a single test function using table-driven tests?

connector/oidc/oidc.go
```diff
@@ -211,6 +229,15 @@ func (c *oidcConnector) LoginURL(s connector.Scopes, callbackURL, state string)
 	if s.OfflineAccess {
 		opts = append(opts, oauth2.AccessTypeOffline, oauth2.SetAuthURLParam("prompt", c.promptType))
 	}
+
+	if len(c.additionalAuthRequestParams) > 0 {
+		for k, v := range c.additionalAuthRequestParams {
+			if contains(managedAuthParams, k) {
```
sagikazarmark, line 235: `golang.org/x/exp` is already a dependency of this project. You could use [golang.org/x/exp/slices.Contains](https://pkg.go.dev/golang.org/x/exp/slices#Contains) instead.

connector/oidc/oidc.go
```diff
@@ -87,6 +91,18 @@ type connectorData struct {
 	RefreshToken []byte
 }
 
+var managedAuthParams = []string{
```
sagikazarmark, line 94: I'm a little concerned that this list is the only thing that protects parameters from being overwritten down below. Any chance we could make it so that parameters don't get overwritten? (Not sure it's easy or feasible at all, but it's worth a check and adding a comment here if it's not)

connector/oidc/oidc.go
```diff
@@ -211,6 +229,15 @@ func (c *oidcConnector) LoginURL(s connector.Scopes, callbackURL, state string)
 	if s.OfflineAccess {
 		opts = append(opts, oauth2.AccessTypeOffline, oauth2.SetAuthURLParam("prompt", c.promptType))
 	}
+
+	if len(c.additionalAuthRequestParams) > 0 {
```
sagikazarmark, line 233: This if is noop: if the list is empty, the loop won't go through it. Am I missing something?

connector/oidc/oidc.go
sagikazarmark: Also, I wonder if this check would be better of in the connector open function as a validation.

connector/oidc/oidc.go
```diff
@@ -72,6 +72,10 @@ type Config struct {
 		// Configurable key which contains the groups claims
 		GroupsKey string `json:"groups"` // defaults to "groups"
 	} `json:"claimMapping"`
+
+	// Add additional authorization request parameters to acceess IdP specific features.
```
sagikazarmark, line 76:
```suggestion
	// Add additional authorization request parameters to access IdP specific features.
```

connector/oidc/oidc.go
```diff
@@ -72,6 +72,10 @@ type Config struct {
...
+	// Take care not to override standard OICD authorization requests parameters.
```
sagikazarmark, line 77: I'd probably say that doing so would result in an invalid config.

sagikazarmark (changes requested):
Thank you for your PR @roddyherries!

Would you mind taking a look at the comments (and rebase the PR)?

PR #2643, opened 2022-08-29
feat(connector/authproxy): support multiple groups

mclavel:
#### Overview

Added support to use multiple values for group header on authproxy connector.

If you use multiples values using comma-separated, the connector can't understand there are different groups.

Definition to use comma-separated list [RFC2616 section 4.2](https://www.rfc-editor.org/rfc/rfc2616#section-4.2)

```
NONE
```

### DIFF
--- a/connector/authproxy/authproxy.go
+++ b/connector/authproxy/authproxy.go
@@ -70,8 +70,11 @@ func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connecto
 	groups := m.groups
 	headerGroup := r.Header.Get(m.groupHeader)
 	if headerGroup != "" {
-		trimHeaderGroup := strings.Replace(headerGroup, " ", "", -1)
-		groups = append(strings.Split(trimHeaderGroup, ","), groups...)
+		splitheaderGroup := strings.Split(headerGroup, ",")
+		for i, v := range splitheaderGroup {
+			splitheaderGroup[i] = strings.TrimSpace(v)
+		}
+		groups = append(splitheaderGroup, groups...)
 	}
 	return connector.Identity{
 		UserID:        remoteUser, // TODO: figure out if this is a bad ID value.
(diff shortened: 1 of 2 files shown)


connector/authproxy/authproxy.go
```diff
@@ -69,7 +70,7 @@ func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connecto
 	groups := m.groups
 	headerGroup := r.Header.Get(m.groupHeader)
 	if headerGroup != "" {
-		groups = append(groups, headerGroup)
+		groups = append(strings.Split(headerGroup, ","), groups...)
```
nabokihms, line 73:
```suggestion
		groups = append(strings.Split(headerGroup, ","), groups...)
```
I assume that trimming is also required, because `X-Groups: group-1, group2` is also a valid value.

nabokihms (changes requested):
I left a small comment. Please take a look when you have time. I'd also like to see test, but sadly the authproxy code is not covered... I think we can merge this PR without tests, but it would be awesome to add any.

connector/authproxy/authproxy.go
mclavel: Cool 🙌, I cleaned the headerGroup to support spaces

mclavel: I added some tests using this [PR](https://github.com/dexidp/dex/pull/2379/files) as a guide

connector/authproxy/authproxy.go
```diff
@@ -69,7 +70,8 @@ func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connecto
 	groups := m.groups
 	headerGroup := r.Header.Get(m.groupHeader)
 	if headerGroup != "" {
-		groups = append(groups, headerGroup)
+		trimHeaderGroup := strings.Replace(headerGroup, " ", "", -1)
```
nabokihms, line 73: I wonder whether a group name can contain whitespaces? Like
```
group 1, group 2
```

mclavel: I updated the connector to support that cases 😄

nabokihms (approved):
@mclavel thank for your contribution!

mclavel:
@nabokihms 🙌 , can you add this change to milestone v2.34?

Issue #2644, opened 2022-08-30
Crowd connector does not return display name


ihulsbus:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.33.0

### Storage Type

SQLite

### Installation Type

Official container image

### Expected Behavior

Expected to get the user's display name after logging in via the Crowd connector in the `name` claim. 

### Actual Behavior

Crowd connector returns the username instead of the display name in the `name` claim, e.g.: johndoe instead of John Doe 

### Steps To Reproduce

1. Use official container
2. Use latest Crowd release (5.0.2 at the time of writing)
3. Create crowd connection and static client
4. Login
5. check the `/userinfo` endpoint. 
6. user claim shows username in the `name` claim instead of the display name

### Additional Information

Got:
```
{
    "iss": "https://localhost:5556",
    "sub": "<hash>",
    "aud": "testclient",
    "exp": 1661942538,
    "iat": 1661856138,
    "at_hash": "<hash>",
    "email": "johndoe@example.com",
    "email_verified": true,
    "groups": [
        "group"
    ],
    "name": "johndoe",
    "preferred_username": "johndoe"
}
```
Expected:
```
{
    "iss": "https://localhost:5556",
    "sub": "<hash>",
    "aud": "testclient",
    "exp": 1661942538,
    "iat": 1661856138,
    "at_hash": "<hash>",
    "email": "johndoe@example.com",
    "email_verified": true,
    "groups": [
        "group"
    ],
    "name": "John Doe",
    "preferred_username": "johndoe"
}
```

### Configuration

```yaml
issuer: https://localhost:5556
storage:
  type: sqlite3
  config:
    file: dex.db
web:
  http: 0.0.0.0:5556
telemetry:
  http: 0.0.0.0:5558
expiry:
  signingKeys: "6h"
  idTokens: "24h"
oauth2:
    skipApprovalScreen: true
    alwaysShowLoginScreen: false
staticClients:
- id: testclient
  redirectURIs:
  - 'urn:ietf:wg:oauth:2.0:oob'
[...]

PR #2661, opened 2022-09-14
Use `display-name` for Crowd user names

xeals:
#### Overview

Modifies the Atlassian Crowd connector to set the `name` claim to the user's full name and adjusts supporting code in response.

This is a breaking change.

Some feedback around how to handle refreshes would be appreciated, as unlike other connectors Crowd doesn't assume that the preferred username is set (which is now the only place the login username is persisted in the token)

#2644 notes that the Crowd connector does not provide the full name of the user, though Crowd itself does provide it. This is unlike most other connectors and the OAuth2 specification and is surprising to consumers.

Authentication to the group checking endpoints now uses the user ID provided by the authentication flow instead of the `identity.Username`, since the attribute is now the full name and not the same ID that was provided initially.

```release-note
The Atlassian Crowd connector now returns the user's full display name instead of their username in the `name` claim. Clients that require the username should set the `preferredUsernameField` of the connector configuration to `name`.
```


nabokihms:
@xeals, thank you for opening this PR, and sorry for the long delay.

These are breaking changes because they change the username field. If users have authorization rules configured to give access to particular users, it will be problematic for them to rewrite them.

The best way to handle this is to add an option to change the username field exported from Crowd API. However, I think the name field is better because it is a slug that cannot be changed in Crowd in case of authentication. The display name is customizable in Crowd and is a better choice for the preferred username IMO.

Could you elaborate more on why the preferred username is unsuitable for your case?

xeals:
It's been some time since I wrote this so my reasoning is a bit pieced together:

1. Dex replies to an identity claim mapping the `connector.Identity.Username` field to `name`. This is [specified](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) as the user's full display name. That this is referred to and treated as a "username" in Dex's codebase is somewhat unfortunate and misleading.
https://github.com/dexidp/dex/blob/a17bc05ee6d896b98b8c4e25c2af0b3bb507ce6e/server/oauth2.go#L359-L361
2. The preferred username field was considered, but the Crowd attribute in question is not a username, nor is it intended as one.
3. I likely didn't see any way of mapping the partial name fields (`given_name`, `family_name`, etc.) at the time of writing, and don't see a way to now either.

Specifically addressing `preferred_username`:
- Following specification, this field should not be relied upon to be unique, and Crowd reasonably doesn't enforce that two people can't have the same full name.
- Peoples' names typically do not follow rules that application developers might typically assign usernames, such as not containing whitespace, punctuation, or non-Latin characters.
- While the format of the display name is customisable in Crowd, the sane default is `first_name last_name`. I imagine this is the behaviour most consumers will be using, or a similar format as befitting their locale. Requiring to change this to a semi-unique username format in order to migrate to a Dex frontend seems needlessly difficult.

In light of the existing codebase and concerns, I'd be happy to compromise with a configuration flag for the connector, rather than an option as a preferred username field.

Issue #2670, opened 2022-09-19
Google without groups seems broken after #2530


jonkerj:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.34.0

### Storage Type

Kubernetes

### Installation Type

Official Helm chart

### Expected Behavior

I've upgraded dex from v2.33.0 to v2.34.0, and expected the Google connector to keep working.

### Actual Behavior

Dex was crashlooping, complaining about `adminEmail` not being set. After setting `adminEmail`, there was something wrong with default application credentials. 

### Steps To Reproduce

1. Install dex 2.34.0, configure Google according to [the docs](https://dexidp.io/docs/connectors/google/), without `groups`, `serviceAccountFilePath` or `adminEmail`

### Additional Information

It is not really clear to me if this is an actual bug introduced by #2530, or a lack of documentation (which could have been part of #2530). The documentation suggest group/svc account is optional, but these errors suggest otherwise.

In other words: is using Google as a provider without group/svc account still supported?

Downgrading to v2.33.0 still works, so I am guessing it might be a bug.

### Configuration

```yaml
connectors:
- adminEmail: re@dact.ed
  config:
    clientID: $GOOGLE_CLIENT_ID
    clientSecret: $GOOGLE_CLIENT_SECRET
    redirectURI: https://re.dact.ed/callback
  hostedDomains:
  - kippendief.biz
  id: google
  name: Google
  type: google
issuer: https://dex.red.act.ed
staticClients:
- idEnv: CLIENT_OAUTH2_PROXY_ID
  name: OAuth2 Proxy
  redirectURIs:
  - https://foo.bar.red.act.ed/oauth2/callback
  secretEnv: CLIENT_OAUTH2_PROXY_SECRET
- idEnv: CLIENT_GRAFANA_ID
  name: Grafana
  redirectURIs:
  - https://grafana.red.act.ed/login/generic_oauth
  secretEnv: CLIENT_GRAFANA_SECRET
storage:
  config:
    inCluster: true
  type: kubernetes
web:
  http: 0.0.0.0:5556
```

### Logs

[...]

PR #2679, opened 2022-09-22
fix for issue 2670; check for no serviceAccountFilePath and no email

bobcallaway:
#### Overview

This (hopefully) reverts a bug introduced in #2530 which does not handle the case where application default credentials nor a service account file (containing credentials) are needed to operate a Google connector.

Fixes: #2670 

NONE

```release-note
fixes regression for Google connector when used without groups
```

@ichbinfrog

### DIFF
--- a/connector/google/google.go
+++ b/connector/google/google.go
@@ -71,10 +71,13 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 		scopes = append(scopes, "profile", "email")
 	}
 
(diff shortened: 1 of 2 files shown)


nabokihms:
@bobcallaway Thanks! I think we can make a patch release with this fix, but there is a test that is failing.

nabokihms:
LGTM from my point of view!
@ichbinfrog , do you have to add something (as the original author of the feature)?

ichbinfrog:
The PR LGTM @nabokihms

sagikazarmark (approved):
LGTM

@nabokihms did you have anything else to add?

PR #2680, opened 2022-09-23
Fix Google connector ADC implementation within GKE environments

### LINKED ISSUE #2676: Google connector with ADC: unexpected end of JSON input
### Preflight Checklist

### Version

ghcr.io/dexidp/dex:v2.34.0

### Storage Type

Kubernetes

### Installation Type

Other (specify below)

### Expected Behavior

google connector successfully use Application Default Credential via Compute Engine metadata server

### Actual Behavior

failed to initialize server: server: Failed to open connector google: failed to open connector: failed to create connector google: could not create directory service: unable to parse credentials to config: unexpected end of JSON input

### Steps To Reproduce

1. deploy dex on GCE / GKE. Use google connector without specifying `serviceAccountFilePath`
2. check the logs

### Additional Information

I installed dex via ArgoCD's `install.yaml`, and replaced its version from v2.30.2 to v2.34.0.
My setup worked OK before, with a key file in serviceAccountFilePath.

I tried running the following code on the same GKE pod by `kubectl exec`, and it went all good.
```
func main() {
        cred, err := google.FindDefaultCredentials(context.Background())
        if err != nil {
                fmt.Errorf("FindDefaultCredentials: %w", err)
        }
        fmt.Printf("Json: %v\n", cred.JSON)
        cfg, err := google.JWTConfigFromJSON(cred.JSON, admin.AdminDirectoryGroupReadonlyScope)
        if err != nil {
                fmt.Errorf("JWTConfigFromJSON: %w\n", err)
        }
        fmt.Printf("cfg: %v\n", cfg)
}
```
The above code works OK with following output, which is something I expected.
```
cred: &{my-project-name 0xc0000aed00 []}
Json:
cfg: <nil>
```

The code does basically the same as dex v2.34.0, so I am not sure why it doesn't work in dex.

cc @ichbinfrog

### Configuration

```yaml
connectors:
    - type: google
      id: google
      name: Google
      config:
        issuer: https://accounts.google.com
        clientID: $GOOGLE_CLIENT_ID
        clientSecret: $GOOGLE_CLIENT_SECRET
        redirectURI: https://my.domain/api/dex/callback
        hostedDomains: ['my.domain']
        serviceAccountFilePath: /tmp/keyfile.json
        adminEmail: test@my.domain
```

### Logs

```shell
Dex Version: v2.34.0-dirty, Go Version: go1.19.1, Go OS/ARCH: linux amd64
config issuer: https://***
config connector: google
config skipping approval screen
config refresh tokens rotation enabled: true
the application default credential is used since the service account file path is not used
failed to initialize server: server: Failed to open connector google: failed to open connector: failed to create connector google: could not create directory service: unable to parse credentials to config: unexpected end of JSON input
```

ichbinfrog:
#### Overview

Adapt the google connector for using application default credentials in GKE.

Closes #2676

Upon further testing in a GKE cluster, this [line](https://github.com/dexidp/dex/blob/master/connector/google/google.go#L300) seems to return an empty JSON credential and therefore making ADC login fail. The default behavior for the [google.FindDefaultCredentials](https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentials) function is the following:

> On Google Compute Engine, Google App Engine standard second generation runtimes (>= Go 1.11), and Google App Engine flexible environment, it fetches credentials from the metadata server.

The `credential.JSON` is empty in GKE environments, thus the error. In order to fetch the credentials (token) from the metadata server, the only method I found which worked did not return a `PERMISSION_DENIED` is to use the flow defined in the [impersonate package](https://pkg.go.dev/google.golang.org/api/impersonate):

- The base Application Default Credentials is used (SA1)
- That service account (SA1) is then used to impersonate the service account designated by the targetPrincipal field (SA2)
- SA2's token source is then used to impersonate the Google Workspace super user designated by adminEmail

TBH, I'm not sure if this is the canonical way of doing it and I've raised a question here https://github.com/googleapis/google-api-go-client/issues/1698.

```release-note

```

### DIFF
--- a/connector/google/google.go
+++ b/connector/google/google.go
@@ -13,6 +13,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	admin "google.golang.org/api/admin/directory/v1"
+	"google.golang.org/api/impersonate"
 	"google.golang.org/api/option"
 
 	"github.com/dexidp/dex/connector"
@@ -45,6 +46,11 @@ type Config struct {
 	// check groups with the admin directory api
 	ServiceAccountFilePath string `json:"serviceAccountFilePath"`
 
+	// Required if ServiceAccountFilePath empty
+	// If nonempty, the base google service account which will be used to impersonate
+	// the workspace super user.
+	TargetPrincipal string `json:"targetPrincipal"`
+
 	// Required if ServiceAccountFilePath
 	// The email of a GSuite super user which the service account will impersonate
 	// when listing groups
@@ -71,7 +77,7 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 		scopes = append(scopes, "profile", "email")
 	}
 
-	srv, err := createDirectoryService(c.ServiceAccountFilePath, c.AdminEmail, logger)
+	srv, err := createDirectoryService(c.ServiceAccountFilePath, c.TargetPrincipal, c.AdminEmail, logger)
 	if err != nil {
 		cancel()
 		return nil, fmt.Errorf("could not create directory service: %v", err)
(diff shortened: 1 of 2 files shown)


nabokihms:
@ichbinfrog Before starting the review process, I have a little question. Did it work previously on GKE and GCP without specifying a service account path (without groups)?

ichbinfrog:
@nabokihms during my initial tests it worked on GKE. However, I realized when the issue was submitted that I had a residual json key in the container which was picked up by the [google.FindDefaultCredentials](https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentials) function so authentication still used the JSON credentials instead of workload identity.

Issue #2547, opened 2022-06-01
Refresh with connector only once


nabokihms:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

There is a problem with concurrent token refresh requests in Dex. If there is no lock on the application side, Dex receives many refresh requests simultaneously. Then Dex tries to go to an external provider for each request and refresh user info.

It works ok for providers like LDAP or Atlassian Crowd. Dex sends them credentials. They send user info back. Users can face minor inconsistency in returned data, but overall it works (yet with the number of unnecessary requests). 

The most problematic providers have refresh flow with refresh tokens rotation, e.g., OIDC and Gitlab. Dex stores refresh tokens in the offline session's connector data field for future refreshes. It is impossible to send these providers the same refresh token twice because this token will be immediately invalidated.

Dex should provide consistency for refresh requests to avoid errors. Each refresh request should use data from the previous refresh request.

### Proposed Solution

Dex has `expiry.refreshToken.reuseInteval` option. With this option enabled, Dex does not issue new refresh tokens for each concurrent request. Instead, Dex replies with the previously issued token that helps avoid concurrent request failures.

However, Dex still makes requests to an external provider, which was mentioned in https://github.com/dexidp/dex/issues/2113
***
My proposal - refresh a token with a connector only once during the reuse interval. Most Dex backends, e.g., etcd, Kubernetes provide consistency out of the box. We only have to use it right.

I made a POC for Kubernetes storage. The full patch can be found [here](https://github.com/deckhouse/deckhouse/blob/12673950fc8a89a9598b3b4faff4e09f7dc94dc1/modules/150-user-authn/images/dex/patches/refresh-once.patch). 

Key points of the patch:
1. Call of the refresh token with the connector method was moved to the updater function, which means Dex will call refresh between `get` and `update` operations.
```go
refreshTokenUpdater := func(old storage.RefreshToken) (storage.RefreshToken, error) {
        ...
	// Dex will call the connector's Refresh method only once if request is not in reuse interval.
	newIdent, rerr := s.refreshWithConnector(ctx, rCtx, ident)
	if rerr != nil {
		return old, rerr
	}
	...
}
```
[...]

PR #2692, opened 2022-10-02
fix: refresh token only once for all concurrent requests

nabokihms:

#### Overview

This PR moves the external provider refresh call to the transaction. If the underlying Dex storage has a proper locking mechanism, in combination with the `reuseInterval` option for all concurrent requests, Dex will call refresh for an external provider only once.

Closes https://github.com/dexidp/dex/issues/2547

```release-note
Refresh the id token only once for all concurrent requests. Fixes errors on token refreshing for Gitlab/OIDC providers.
```

### DIFF
--- a/server/refreshhandlers.go
+++ b/server/refreshhandlers.go
@@ -28,6 +28,10 @@ type refreshError struct {
 	desc string
 }
 
+func (r *refreshError) Error() string {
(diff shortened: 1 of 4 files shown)


sagikazarmark (approved):
LGTM, thanks @nabokihms !

Issue #2728, opened 2022-11-02
Improve connectorData flow


nabokihms:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

There are four places for connector data to be persisted:

1. After successful login Dex saves `connectorData` to the AuthRequest object.
https://github.com/dexidp/dex/blob/6172b492a9ba7462871ef39075c1116a0b004026/server/handlers.go#L485-L490

2. And the same data is saved to the OfflineSession object.
https://github.com/dexidp/dex/blob/6172b492a9ba7462871ef39075c1116a0b004026/server/handlers.go#L524-L529
https://github.com/dexidp/dex/blob/6172b492a9ba7462871ef39075c1116a0b004026/server/handlers.go#L542-L550

3. After login, `connectorData` is moved from the AuthRequest object to the AuthCode object.
https://github.com/dexidp/dex/blob/6172b492a9ba7462871ef39075c1116a0b004026/server/handlers.go#L654-L664 

4. Then `connectorData` is saved from the AuthCode object to the RefreshToken object.
https://github.com/dexidp/dex/blob/6172b492a9ba7462871ef39075c1116a0b004026/server/handlers.go#L938-L949

Bonus: On the first refresh token update connectorData is deleted from the RefreshToken object.

### Proposed Solution

Connector data is a fundamental thing for connectors that refresh tokens. Connector data can contain refresh tokens of other providers or access tokens. Providers like Gitlab or some OIDC providers can punish users for using the stale connector data, leading to the loss of the refresh token issued by Dex.

It seems that connector data should only be saved to the offline session after login finalizing and nowhere else. It is possible to skip stages 1, 3, and 4.

### Additional Information

Maybe it is also good to consider having connector data per refresh token because it makes more sense to map provider refresh tokens and dex refresh tokens like 1:1.

PR #2729, opened 2022-11-02
fix: Do not use connector data from the refresh token field

nabokihms:

There is a possibility that connector data will be saved to refresh token, which can lead to refresh tokens invalidation.

The flow of connectorData is too complicated. I opened an issue to think about it in the future.
See https://github.com/dexidp/dex/issues/2728

```release-note
Do not use connector data from the refresh token field.
```


sagikazarmark (approved):

sagikazarmark:
Are there any real user facing implications of this change? Can't really see any.

nabokihms:
In some rare cases, e.g., if the refresh token rotation feature is enabled in Okta and a client application does not support locking, users can face odd behavior, but it is definitely a minor fix.

PR #2740, opened 2022-11-15
Add preferredEmailDomain config option for GitHub connector

nobuyo:
#### Overview

This PR is a revamped version of https://github.com/dexidp/dex/pull/1298.
It allows identifying users by verified company email without marking it as primary.

This PR adds new `preferredEmailDomain` config option for GitHub connector.
If set, an email address with the specified domain takes precedence over the primary email.

When a user is logged into a service that uses Dex, such as ArgoCD, via GitHub, the user is identified by the account's primary email address.

However, in a company, we have the following needs:
* Identify employees by their company domain email address
* Allow employees to use their own(not-for-only-work) GitHub account

To satisfy these, we need an option that can return a specific domain email address as the user's email address.

N/A

```release-note
Add new `preferredEmailDomain` config option for GitHub connector.
```

### DIFF
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -553,18 +553,11 @@ type userEmail struct {
 // which inserts a bearer token as part of the request.
 func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (string, error) {
 	var (
-		preferredEmailDomainRegexp *regexp.Regexp
-		primaryEmail               userEmail
-		preferredEmails            []userEmail
+		primaryEmail    userEmail
+		preferredEmails []userEmail
 	)
 
 	apiURL := c.apiURL + "/user/emails"
-
-	if c.preferredEmailDomain != "" {
-		globExtracted := strings.ReplaceAll(c.preferredEmailDomain, "*", "[a-zA-Z0-9][A-Za-z0-9-][a-zA-Z0-9-]+")
-		patten := fmt.Sprintf("^%s$", globExtracted)
-		preferredEmailDomainRegexp = regexp.MustCompile(patten)
-	}
 
 	for {
 		// https://developer.github.com/v3/users/emails/#list-email-addresses-for-a-user
@@ -595,12 +588,12 @@ func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (s
 				primaryEmail = email
 			}
 
-			if preferredEmailDomainRegexp != nil {
+			if c.preferredEmailDomain != "" {
 				_, domainPart, ok := strings.Cut(email.Email, "@")
 				if !ok {
 					return "", errors.New("github: invalid format email is detected")
 				}
-				if email.Verified && preferredEmailDomainRegexp.MatchString(domainPart) {
+				if email.Verified && c.isPreferredEmailDomain(domainPart) {
 					preferredEmails = append(preferredEmails, email)
 				}
 			}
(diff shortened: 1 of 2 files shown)


nobuyo:
@sagikazarmark @nabokihms would it be ok if i ask you to review about this?

connector/github/github.go
```diff
@@ -575,7 +580,21 @@ func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (s
...
+
+			if c.preferredEmailDomain != "" {
+				components := strings.Split(email.Email, "@")
+				if email.Verified && components[1] == c.preferredEmailDomain {
+					return email.Email, nil
+				}
+
+				// fallback to primary email if there is no preferred-domain email
+				if idx == len(emails)-1 {
+					return primaryEmail.Email, nil
```
nabokihms, line 594: If both primary email and email with a preferred domain were missing, an error should be returned.

connector/github/github.go
```diff
@@ -575,7 +580,21 @@ func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (s
 			}
 
 			if email.Verified && email.Primary {
-				return email.Email, nil
+				primaryEmail = email
+			}
+
+			if c.preferredEmailDomain != "" {
+				components := strings.Split(email.Email, "@")
+				if email.Verified && components[1] == c.preferredEmailDomain {
```
nabokihms, line 588: NIT: WDYT about making preferredEmailDomain a glob so that users can provide a pattern (not the exact domain)?

connector/github/github.go
```diff
@@ -575,7 +580,21 @@ func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (s
 			}
 
 			if email.Verified && email.Primary {
-				return email.Email, nil
+				primaryEmail = email
+			}
+
+			if c.preferredEmailDomain != "" {
+				components := strings.Split(email.Email, "@")
```
nabokihms, line 587: NIT: `strings.Cut` is better for this case.

connector/github/github.go
```diff
@@ -575,7 +580,21 @@ func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (s
...
+				if email.Verified && components[1] == c.preferredEmailDomain {
+					return email.Email, nil
+				}
+
+				// fallback to primary email if there is no preferred-domain email
+				if idx == len(emails)-1 {
```
nabokihms, line 593: I'd like to suggest splitting logic.

1. Iterate through all domains and find the primary and preferred domain emails. Just find.
2. Decide what should be returned: primary email, email with preferred domain, or an error.

nabokihms (changes requested):
Hello, @nobuyo. Sorry for the long delay. I left some comments. Please take a look when you have time.

connector/github/github.go
```diff
@@ -575,7 +580,21 @@ func (c *githubConnector) userEmail(ctx context.Context, client *http.Client) (s
 			}
 
 			if email.Verified && email.Primary {
-				return email.Email, nil
+				primaryEmail = email
+			}
+
+			if c.preferredEmailDomain != "" {
```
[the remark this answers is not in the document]
nobuyo: Although not explicitly documented, email addresses seem to be returned from the API in the following order:
  1. Primary email
  2. Email with github domain (e.g. `nobuyo@users.noreply.github.com`)
  3. Other email addresses (in order of oldest added to account)

At least the order did not change every API call as far as I tried.

I thought about sorting the emails alphabetically, but that would make the user uncontrollable about email priority at all if user has multiple preferred-domain emails. If we let GitHub decide the order, perhaps user can control it by delete and re-registering the emails to the account. That may not be reasonable, but...

So I think we can respect the order by the API response, for now.

nobuyo:
Thank you for your review! I've updated this PR.

connector/github/github.go
```diff
@@ -548,7 +552,20 @@ type userEmail struct {
...
+	var (
+		preferredEmailDomainRegexp *regexp.Regexp
+		primaryEmail               userEmail
+		preferredEmails            []userEmail
+	)
+
 	apiURL := c.apiURL + "/user/emails"
+
+	if c.preferredEmailDomain != "" {
+		globExtracted := strings.ReplaceAll(c.preferredEmailDomain, "*", "[a-zA-Z0-9][A-Za-z0-9-][a-zA-Z0-9-]+")
```
nabokihms, line 564: It seems that the preferred email domain is now a proper regex with a side effect.
Users can provide something like `(.+).my[a-A+].(google|yahoo).com`, and it will be valid.

I'd like to suggest avoiding regex here and using a plain method instead with the following logic (see example here: https://go.dev/play/p/S1X_aGIezCB). With this approach, there are fewer pitfalls, and the method seems more controllable.

nabokihms (changes requested):
Thank you. Overall, everything looks good to me. I have only one suggestion regarding the regex part. Please take a look when you have time.

nabokihms (approved):
LGTM! Thank you for your work, @nobuyo!

PR #2771, opened 2023-01-04
feat: support redis storage

houxul:


storage/redis/redis.go
```diff
@@ -0,0 +1,545 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), defaultStorageTimeout)
+	defer cancel()
+
+	kvs, err := c.getKvs(ctx, connectorPrefix)
+	if err != nil {
+		return nil, err
+	}
+	for _, val := range kvs {
+		var connector storage.Connector
+		err = json.Unmarshal([]byte(val), &connector)
```
AmitKatyal-Sophos, line 373: storage.Connector has config json field of type []byte. If we have config json object stored in the redis, then json unmarshaling returns an error.

"{\"type\":\"ldap\",\"name\":\"OpenLDAP\",\"id\":\"ldap\",\"config\":{\"host\":\"localhost:389\",\"insecureNoSSL\":true,\"bindDN\":\"cn=admin,dc=example,dc=org\",\"bindPW\":\"admin\",\"usernamePrompt\":\"EmailAddress\",\"userSearch\":{\"baseDN\":\"ou=People,dc=example,dc=org\",\"filter\":\"(objectClass=person)\",\"username\":\"mail\",\"idAttr\":\"DN\",\"emailAttr\":\"mail\",\"nameAttr\":\"cn\"},\"groupSearch\":{\"baseDN\":\"ou=Groups,dc=example,dc=org\",\"filter\":\"(objectClass=groupOfNames)\",\"userMatchers\":[{\"userAttr\":\"DN\",\"groupAttr\":\"member\"}],\"nameAttr\":\"cn\"}}}"

Issue is fixed after declaring config of type json.RawMessage instead of []byte.

storage/redis/redis.go
```diff
@@ -0,0 +1,545 @@
...
+type client struct {
+	db     redisv8.UniversalClient
+	logger log.Logger
+}
+
+func (c *client) Close() error {
+	return c.db.Close()
+}
+
+func (c *client) GarbageCollect(now time.Time) (result storage.GCResult, err error) {
```
AmitKatyal-Sophos, line 44: We can set the key ttl while creating the key to avoid garbage collection.

PR #2776, opened 2023-01-06
Support multiple LDAP hosts

onetwopunch:
#### Overview

Often there is a need, as there is in my company, to iterate through multiple LDAP hosts in the event that one has failed. DNS is one possible solution, but there are cases where that may not be possible, for example with Kerberos. 

This change allows users to specify a comma-separated list of LDAP hosts to iterate through in case one is unreachable. It will function the same as always with one host, but will iterate through if multiple are given and the prior hosts fail.

Happy to add tests if necessary, but for such a simple change, I'm not sure it is. 

```release-note
LDAP connector now supports multiple hosts in a comma-separated list. No changes required.
```


onetwopunch:
@ericchiang I see you worked on the LDAP bits pretty heavily. Any thoughts on this?

ericchiang:
I haven't worked on dex in years. You'll have to find an actual maintainer to review :(

onetwopunch:
Thanks for letting me know and good luck on your current endeavors 😄 

@sagikazarmark or @nabokihms any thoughts on this PR?

dlipovetsky:
(I'm not a dex maintainer, so please consider this my humble, unsolicited opinion)

The change to the configuration schema is backward compatible. On the one hand, it may be confusing that the `Host` field supports multiple hosts, from which the first is chosen when the connector is initialized, but, on the other hand, the change is concise, and the common use case (one host) works as it did before.

The change to connector initialization, however, is not backward compatible. Before this change, the connector made no calls to the LDAP host during initialization. With this change, it does, and at least one LDAP host must respond for initialization to succeed.

onetwopunch:
@dlipovetsky thanks for the feedback! 

> The change to connector initialization, however, is not backward compatible

This actually isn't quite . While the connection is checked at startup time, if all of them error out, the last host in the list will be returned regardless of if the connection succeeded, I did this specifically for backwards compatibility.

However, thanks for commenting on this as I realize now that I'm only calling this connection once during startup and not during invocation, which kind of defeats the goal haha. 

Not that any maintainers seem interested in this change anyway after 2 weeks of silence. I'll probably just end up forking dex for this and other changes that no one seems interested in addressing 🤷 But thanks again for the feedback!

dlipovetsky:
> While the connection is checked at startup time, if all of them error out, the last host in the list will be returned regardless of if the connection succeeded, I did this specifically for backwards compatibility.

My mistake! :facepalm: I see now that you're [breaking out of the loop](https://github.com/dexidp/dex/pull/2776/files#diff-b877a4fe9a6210f40c1a5fbcf4c6dd366856b31ec0661975dcf60915cf3985d1R657-R661) as soon as you get a successful connection.

dlipovetsky:
> However, thanks for commenting on this as I realize now that I'm only calling this connection once during startup and not during invocation, which kind of defeats the goal haha.

A silver lining! :wink: 

Could you accomplish your goal in https://github.com/dexidp/dex/blob/70e6cc2205c30788e3d1fe28cc66969266c59336/connector/ldap/ldap.go#L310-L348?

onetwopunch:
Yes I believe I can...I actually started there, but then realized I was making some breaking changes so I switched it but totally missed that that would only do the check at startup 😂 Thanks be to Code Review

PR #2781, opened 2023-01-10
fix: propagate http client to userInfo requests for OIDC connector

nabokihms:

#### Overview

Fix for an error (it occurs when the `getUserInfo` is set to true and a custom ca is used)

image

A fix for https://github.com/dexidp/dex/pull/1632

This is a fix for the feature that is not released.
```release-note
NONE
```


nabokihms:
I tested it manually, and it works as expected now. I'm going to merge this PR. It is better to have a working version in the main than to live with the bug (and it blocks release, ofc).

PR #2789, opened 2023-01-13
feat: Initialize connectors lazily (not on server start)

### LINKED ISSUE #1723: OIDC discovery blocks other connectors
It seems like there is no way to disable OpenID Connect Discovery (see also #1228) in the OIDC connector. This breaks use cases when the remote providers may be unavailable during startup and in turn it also makes other connectors non-functional (because the server won't start).

Are there any plans to either make remote OIDC discovery optional or let the server start with broken/failed/not ready connectors? That would make it possible to use DEX in environments where the local login must be provided even if external connections fail.

dlipovetsky:
#### Overview

Initialize connectors lazily (not on server start).

Dex initializes all connectors on server start. Connectors may fail to initialize, e.g. if they require a remote service that is unavailable, or if they are misconfigured. If any connector fails to initialize, dex exits. This PR adds an option to skip connector initialization during server start. Dex will initialize each connectors as it needs them to serve requests.

The current behavior (exiting) remains the default; the new behavior is enabled by a field in the configuration file:

```yaml
lazyInitConnectors: true
```

Closes #1723 

```release-note
Add a configuration option to start the server without initializing connectors.
```


cmd/dex/config.go
```diff
@@ -49,6 +49,13 @@ type Config struct {
 	// querying the storage. Cannot be specified without enabling a passwords
 	// database.
 	StaticPasswords []password `json:"staticPasswords"`
+
+	// If enabled, the server does not initialize connectors when it starts, but
+	// initializes connectors when it handles requests. This allows the server to
+	// start even if some connectors cannot be initialized, for example, because
+	// because it is misconfigured, or because it initialization requires a
```
msdolbey, line 56: sorry - tiny nit

```suggestion
	// because it is misconfigured, or because the initialization requires a
```

msdolbey (approved):
👍  ❤️

cmd/dex/config.go
dlipovetsky: Fixed, amended, and force-pushed

dlipovetsky:
@nabokihms Could you please allow the CI workflows to run? Thanks!

dlipovetsky:
The PR is backward compatible, addresses a long-standing issue, comes with a unit test, and CI is passing (ne job fails because the PR needs some labels, which I don't have the permission to create) 

@nabokihms and/or @sagikazarmark, please let me know if there's anything else I can do to help it along, or if you disagree with the approach I took. Thank you! :pray:

nabokihms:
Sorry, slacked a little because of KubeCon. I will take a look till the end of the week and give my feedback.

nabokihms:
@dlipovetsky Sorry for the long delay. This feature overlaps with what we in Palak also need, so I'd be happy to work with you on this feature.

I had a chance to look through the code and even test the PR, and there are some questions I'd like to discuss upfront.

1. A connector is available from the start if it is a lazily initialized connector. Thus, users can see the connector on the list of available connectors and click on it, attempting to authenticate with it (despite Dex knowing that the connector is not ready).

    This does not seem like a valid UX to me. Connectors should not be pickable until initialized.

2. Each failed attempt is a request to the external provider. Dex will try to make a request to the provider without exponential backoff or any other protection mechanisms in case of unavailability.

3. If all connectors are lazily initialized, it is possible to end up with a working Dex without working connectors, which is misleading. Dex may be working but not ready to serve to auth requests.

What do you think about the following proposal? This is more complicated, but I believe provides better UX for users.
* Initialize connectors in the background in a loop if the first initialization attempt fails ([like Kubernetes does](https://github.com/kubernetes/kubernetes/blob/2cc529282a37e5a8aa6479ce71f8660645b18f52/staging/src/k8s.io/apiserver/plugin/pkg/authenticator/token/oidc/oidc.go#L132-L185)).
* Do not show the connector in the list of connectors until initialized.
* Fail the readiness probe if there are no initialized connectors.

dlipovetsky:
> @dlipovetsky Sorry for the long delay. This feature overlaps with what we in Palak also need, so I'd be happy to work with you on this feature.

Thanks for the review!

> 1. A connector is available from the start if it is a lazily initialized connector. Thus, users can see the connector on the list of available connectors and click on it, attempting to authenticate with it (despite Dex knowing that the connector is not ready).
>    This does not seem like a valid UX to me. Connectors should not be pickable until initialized.

Good point. I do not use the dex UI directly, so I did not see (or consider) this.

> 2. Each failed attempt is a request to the external provider. Dex will try to make a request to the provider without exponential backoff or any other protection mechanisms in case of unavailability.

Good point. A couple of questions: Does "attempt" mean a login attempt? When you say "Dex will try to make a request to the provider," are you referring to this code path: https://github.com/dexidp/dex/blob/1890bf37b67b6161da42804e8ead332132fdbfcb/server/server.go#L639

> 3. If all connectors are lazily initialized, it is possible to end up with a working Dex without working connectors, which is misleading. Dex may be working but not ready to serve to auth requests.

Good point.

> What do you think about the following proposal? This is more complicated, but I believe provides better UX for users.

I agree, your proposal does provide better UX. Being unfamiliar with the dex code, I made the smallest possible change that allowed dex to start in the presence of a misconfigured connector :sweat_smile: 

Let me know if you'd like to me to work on this, or if you'd like to do it yourself. Also, let me know if you'd like the work to go in a new PR.

nabokihms:
@dlipovetsky I would be happy if you had time to work on this feature. It is ok to work on this improvement in the scope of this PR. Opening a new PR is not necessary.

nabokihms:
Fixed by https://github.com/dexidp/dex/pull/4159

PR #2804, opened 2023-01-28
Google: Fetch groups of parent grous in parallel

torfjor:
#### Overview

Fetching transitive groups is done in a single goroutine and is very slow. One of our users has 204 direct group memberships and a login through the google connector takes ~70s. By iterating through all the direct group memberships for transitive memberships in parallel this is reduced to ~4s.

No.


connector/google/google.go
```diff
@@ -227,8 +235,11 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
 
 	var groups []string
 	if s.Groups && c.adminSrv != nil {
-		checkedGroups := make(map[string]struct{})
-		groups, err = c.getGroups(claims.Email, c.fetchTransitiveGroupMembership, checkedGroups)
+		if c.fetchTransitiveGroupMembership {
+			groups, err = c.getAllGroups(ctx, claims.Email)
+		} else {
+			groups, err = c.getGroups(ctx, claims.Email, &sync.Map{})
```
nabokihms, line 241: sync.Map is only suitable for special cases. I'd like to suggest avoiding it here.

connector/google/google.go
```diff
@@ -252,34 +263,95 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
 	return identity, nil
 }
 
+func (c *googleConnector) getAllGroups(ctx context.Context, userKey string) ([]string, error) {
+	directMemberships, err := c.adminSrv.Groups.List().
+		UserKey(userKey).
+		Domain(c.domain).
+		Context(ctx).
```
nabokihms, line 270: The pagination token is missing here. Is it ok? Previously it was always specified.

connector/google/google.go
```diff
@@ -252,34 +263,95 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
...
+	if err != nil {
+		return nil, err
+	}
+
+	var groups []string
+	var groupsCh = make(chan string)
+	checkedGroups := sync.Map{}
+	g, cctx := errgroup.WithContext(ctx)
+
+	for _, group := range directMemberships.Groups {
```
nabokihms, line 281: Is there a limit for concurrent requests to Google API? Can a concurrent request be rejected because of too many requests?

connector/google/google.go
```diff
@@ -252,34 +263,95 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
...
+		email := group.Email
+		g.Go(func() error {
+			transitiveGroups, err := c.getGroups(cctx, email, &checkedGroups)
```
nabokihms, line 284: The `getGroups` method returns a slice, and then this slice is sent to the channel. Can this be optimized to use a channel to return groups for the `getGroups` method also?

nabokihms (changes requested):
Hello @torfjor. Thank you for opening the PR. I left some notes about the code. Please take a look when you have time.

connector/google/google.go
torfjor: Acoording to the [docs](https://developers.google.com/admin-sdk/directory/v1/limits) the quota is 3000 requests/100s and can be raised in the developer console. It doesn't look like the generated googleapis clients handle 403/userRateLimitExceeded, quotaExceeded or 429/rateLimitExceeded out of the box. We could potentially provide a hook for the user to provide a limit on in-flight requests through the `errgroup.Group`'s `SetLimit()` method.

connector/google/google.go
torfjor: It was actually not handled at all in the original either. Pushed a change which fetches all pages.

connector/google/google.go
torfjor: Fixed

connector/google/google.go
torfjor: I find the readability better if we avoid passing channels around and instead only use them at synchronization points. If you strongly disagree, I can change it.

Issue #2657, opened 2022-09-09
Get OIDC token issued by Dex using a token issued by one of the connectors.


shankerwangmiao:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

In some scenarios, there can only be one OIDC provider, e.g. kube-apiserver. Dex can serve this role well by providing various connectors. However, when it comes to a program which has been issued an OIDC token from another issuer, it cannot be used to exchange for a Dex issued token programmingly.

### Proposed Solution

OIDC connectors can be configured as ``passive", which will be hidden to the interactive users.

An api interface can be added to Dex, say `/token`. When a `POST` request like below is received, the token from the Authorization header will be extracted and verified against each configured passive OIDC connector. If one of them get passed, the client will be issued an OIDC token issued by Dex.

```
POST /token HTTP/1.1
Host: server.example.com
Authorization: Basic <OIDC_Token_issued_by_a_connector_goes_here>
Content-Type: application/x-www-form-urlencoded

grant_type=client_credentials&client_id=<foo_client>
```

### Additional Information

The use case is, I would like a Kubernetes cluster can be accessed by both interactive normal users and programmingly CI jobs. CI jobs running on gitlab runners are issued an OIDC token by the gitlab instance via an environment variable. If this is implemented, then the CI job can get a token issued by Dex using this OIDC token, which can then be used to access the kubernetes cluster.

iamnoah:
[RFC8693](https://www.rfc-editor.org/rfc/rfc8693.html#name-request) defines a token-exchange grant_type that accepts a token for authentication. Dex could accept this grant type in on the token endpoint for an OIDC connector, validate the token via JWKS, and issue a new token using the claimMapping.

blairdrummond:
I would love to see this implemented. My use case is I use ArgoCD with Dex on Azure with AzureAD. Users are using AAD no problem, but I want my pipelines to be able to access ArgoCD's API with a JWT (avoiding local accounts). The pipeline easily gets a sts.windows.net JWT from the machine identity/service principal in the pipeline, so if I could exchange that AAD token for a Dex token then machine access would be very smooth.

blairdrummond:
[...]

Issue #1668, opened 2020-03-11
Question: non-web based clients?


rezagh2:
is it possible to have a dex client that is not web based? i.e.b2b scenarios where an applications wants to get a ID token via dex which is connected to LDAP?

devfd:
In the case you are referring to a native application it depends if you have access to a web browser. If you do, your application starts a local web server and then initiates an authorization flow with a redirect_url to the local web server. You can then exchange the code you receive for an access token.

rezagh2:
Thanks; so if the application has no access to a web browser (a backend code making an api call to another backend service) the answer i assume would be no?

candlerb:
Sounds like you want the password grant flow: open ticket #926 

It may be better to "prime" the application with an initial id-token and refresh-token (obtained via a one-time web exchange), and thereafter it can use the refresh-token without any further web interaction.

squat:
Support for the password grant flow was added in https://github.com/dexidp/dex/pull/1621.

rezagh2:
@candlerb I am after client credentials b2b flow.

Issue #1484, opened 2019-07-12
Token exchange for external tokens


onatm:
Is there any plans to add a token exchange for the access tokens of connectors?

[`keycloak`](https://github.com/keycloak/keycloak) has this functionality. If there is a plan to add it then I'd be happy to go on and create a PR for it.

You can find the documentation of their external token exchange [here](https://github.com/keycloak/keycloak-documentation/blob/master/securing_apps/topics/token-exchange/token-exchange.adoc#internal-token-to-external-token-exchange).

srenatus:
Thanks for providing links! There's currently no _plans_ or a roadmap or anything here -- Just a few people spending some spare time on keeping this from rotting away 😉  -- So, unless there's someone implementing it, that is. Would you want to be that person?

If so, let's sketch out how this is going to look like before diving into the code, please. It seems like a bigger endeavour than what you'd commonly _"just send a PR for"_ 😃

onatm:
I definitely want to be that person 😄 

I'd be happy to share ideas and discuss the best possible way to implement it. Where is the best place to continue the conversation? Here or gitter or discord?

srenatus:
💯-- great.

I'm afraid I am still lacking a complete idea how this is going to look like, and what you'd achieve with that. My current best guess (after merely skimming the linked docs): you're exchanging a connector-id-token for a dex-id-token? So, this would only apply to using the OIDC connector? Or what's the intended scope here? 😃

onatm:
Your guess is half right 😄 It is for exchanging a dex `id_token` with a connector `access_token`. I've already implemented a super naive version of it by only having `github` connector in mind.

You can check the implementation here: https://github.com/onatm/dex/tree/external/token-exchange

srenatus:
Right now, it's quite nice that dex works like a funnel with many external identity providers on the top side, and a single OIDC IdP for anything _below dex_... This would break this abstraction, wouldn't it? 🤔

onatm:
I think this would not break that abstraction (_and it shouldn't_). `ConnectorData` is already stored in the backing data store and it would be a shame not to use that data 😃.

The implementation I've done accepts an `idToken` with some required parameters (based on `keycloak` specification) and returns the `ConnectorData` to the caller. dex still on the top side and now behaves like a bridge between the client and underlying identity provider.

onatm:
I forgot to add: to be able to use this _token exchange_, the `idToken` should have `federated_claims` in it.

[...]

Issue #2143, opened 2021-05-26
feat: nested map support in claimMapping 


lukasmrtvy:
#### Is your feature request related to a problem?

Its not possible to reference nested claim in claimMapping ( Keycloak exports roles on realm_access.roles path for example )

#### Describe the solution you'd like to see

```
claimMapping:
  groups: realm_access.roles
```

#### Describe alternatives you've considered

Remapping claim. ( Keycloak can )

#### Additional context

None

michaelajr:
We really need this as well. Any thoughts on getting this implemented?

PR #2806, opened 2023-01-29
RFC 8693 OAuth 2.0 Token Exchange

seankhliao:
#### Overview

Implement [RFC 8693](https://www.rfc-editor.org/rfc/rfc8693.html) which specifies exchanging third party tokens for native ones.

There are times when we wish to grant machines access to a protected endpoint without using long-lived, static API keys or user/password combinations.
Many modern execution environments can expose short-lived OIDC tokens for the duration of the execution, examples include:

- [GCP Service Identity](https://cloud.google.com/run/docs/securing/service-identity)
- [AWS Execution Role](https://docs.aws.amazon.com/lambda/latest/dg/lambda-intro-execution-role.html)
- [Github Actions OIDC](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect)
- [CircleCI OIDC](https://circleci.com/docs/openid-connect-tokens/)
- [Kubernetes Service Accounts](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/)

As these are machine users, handling a redirect is not very feasible.
In these cases, clients will have independently authenticated to the upstream IDP (sometimes implicitly by virtue of the IDP integrating with the execution environment) and obtained an ID token.
This PR allows using this prior authentication to be used in obtaining a dex issued token for use with downstream clients.

As a concrete example:
We're running [ArgoCD](https://argo-cd.readthedocs.io/en/stable/getting_started/) with dex as the identity provider, connected to Okta for humans.
We want to access the ArgoCD API from our CI environments without keeping around a static key which [may be compromised](https://circleci.com/blog/january-4-2023-security-alert/).

Related issues:

- #1484
- #1668
- #2657

The RFC: 

- Isn't very clear on what are the accepted subject/requested token types. I'm assuming it's ID tokens for subject, and ID/Access for requested.
- `actor_token` and `actor_token_type` are "MUST ... if the actor token is present, also perform the appropriate validation procedures for its indicated token type", but it's unclear what to authenticate them in this context (an existing user?)
- specifies an optional `resource` field which I don't see any use for
- specifies an `audience` field which I'm using for connector ID, based on how [GCP Workload Identity Pools use them](https://cloud.google.com/iam/docs/workload-identity-federation-with-other-providers#prepare)

It was also unclear if the correct arguments are passed to `newIDToken` / `newAccessToken`

```release-note
oidc connections now support the grant type "urn:ietf:params:oauth:grant-type:token-exchange"
```

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -129,6 +129,10 @@ func (p *password) UnmarshalJSON(b []byte) error {
 
 // OAuth2 describes enabled OAuth2 extensions.
 type OAuth2 struct {
+	// list of allowed grant types,
+	// defaults to all supported types
+	GrantTypes []string `json:"grantTypes"`
+
 	ResponseTypes []string `json:"responseTypes"`
 	// If specified, do not prompt the user to approve client authorization. The
 	// act of logging in implies authorization.
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -246,6 +246,7 @@ type caller uint
 const (
 	createCaller caller = iota
 	refreshCaller
+	exchangeCaller
 )
 
 func (c *oidcConnector) HandleCallback(s connector.Scopes, r *http.Request) (identity connector.Identity, err error) {
@@ -284,16 +285,32 @@ func (c *oidcConnector) Refresh(ctx context.Context, s connector.Scopes, identit
 	return c.createIdentity(ctx, identity, token, refreshCaller)
 }
 
+func (c *oidcConnector) TokenIdentity(ctx context.Context, subjectTokenType, subjectToken string) (connector.Identity, error) {
+	var identity connector.Identity
+	token := &oauth2.Token{
+		AccessToken: subjectToken,
+	}
+	return c.createIdentity(ctx, identity, token, exchangeCaller)
+}
+
 func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.Identity, token *oauth2.Token, caller caller) (connector.Identity, error) {
 	var claims map[string]interface{}
 
-	rawIDToken, ok := token.Extra("id_token").(string)
-	if ok {
+	if rawIDToken, ok := token.Extra("id_token").(string); ok {
 		idToken, err := c.verifier.Verify(ctx, rawIDToken)
 		if err != nil {
 			return identity, fmt.Errorf("oidc: failed to verify ID Token: %v", err)
 		}
 
+		if err := idToken.Claims(&claims); err != nil {
+			return identity, fmt.Errorf("oidc: failed to decode claims: %v", err)
+		}
+	} else if caller == exchangeCaller {
+		// AccessToken here could be either an id token or an access token
+		idToken, err := c.provider.Verifier(&oidc.Config{SkipClientIDCheck: true}).Verify(ctx, token.AccessToken)
+		if err != nil {
+			return identity, fmt.Errorf("oidc: failed to verify token: %v", err)
+		}
 		if err := idToken.Claims(&claims); err != nil {
 			return identity, fmt.Errorf("oidc: failed to decode claims: %v", err)
 		}
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -845,7 +845,7 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
 	case grantTypePassword:
 		s.withClientFromStorage(w, r, s.handlePasswordGrant)
 	case grantTypeTokenExchange:
-		s.handleTokenExchange(w, r)
+		s.withClientFromStorage(w, r, s.handleTokenExchange)
 	default:
 		s.tokenErrHelper(w, errUnsupportedGrantType, "", http.StatusBadRequest)
 	}
@@ -1326,7 +1326,7 @@ func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, cli
 	s.writeAccessToken(w, resp)
 }
 
-func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request) {
+func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request, client storage.Client) {
 	ctx := r.Context()
 
 	if err := r.ParseForm(); err != nil {
@@ -1336,15 +1336,14 @@ func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request) {
 	}
 	q := r.Form
 
-	resource := q.Get("resource")                       // OPTIONAL, use for issued token audience
-	scopes := strings.Fields(q.Get("scope"))            // OPTIONAL, map to issed token scope
+	scopes := strings.Fields(q.Get("scope"))            // OPTIONAL, map to issued token scope
 	requestedTokenType := q.Get("requested_token_type") // OPTIONAL, default to access token
 	if requestedTokenType == "" {
 		requestedTokenType = tokenTypeAccess
 	}
-	connID := q.Get("audience")                     // REQUIRED (RFC 8693 optional), use for connector ID
 	subjectToken := q.Get("subject_token")          // REQUIRED
 	subjectTokenType := q.Get("subject_token_type") // REQUIRED
+	connID := q.Get("connector_id")                 // REQUIRED, not in RFC
 
 	switch subjectTokenType {
 	case tokenTypeID, tokenTypeAccess: // ok, continue
@@ -1392,9 +1391,9 @@ func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request) {
 	var expiry time.Time
 	switch requestedTokenType {
 	case tokenTypeID:
-		resp.AccessToken, expiry, err = s.newIDToken(resource, claims, scopes, "", "", "", connID)
+		resp.AccessToken, expiry, err = s.newIDToken(client.ID, claims, scopes, "", "", "", connID)
 	case tokenTypeAccess:
-		resp.AccessToken, expiry, err = s.newAccessToken(resource, claims, scopes, "", connID)
+		resp.AccessToken, expiry, err = s.newAccessToken(client.ID, claims, scopes, "", connID)
 	default:
 		s.tokenErrHelper(w, errRequestNotSupported, "Invalid requested_token_type.", http.StatusBadRequest)
 		return
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -1329,8 +1329,6 @@ func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, cli
 func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
-	// TODO: check global allowed grant types?
-
 	if err := r.ParseForm(); err != nil {
 		s.logger.Errorf("could not parse request body: %v", err)
 		s.tokenErrHelper(w, errInvalidRequest, "", http.StatusBadRequest)
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -710,7 +710,7 @@ func (s *Server) sendCodeResponse(w http.ResponseWriter, r *http.Request, authRe
 			implicitOrHybrid = true
 			var err error
 
-			accessToken, err = s.newAccessToken(authReq.ClientID, authReq.Claims, authReq.Scopes, authReq.Nonce, authReq.ConnectorID)
+			accessToken, _, err = s.newAccessToken(authReq.ClientID, authReq.Claims, authReq.Scopes, authReq.Nonce, authReq.ConnectorID)
 			if err != nil {
 				s.logger.Errorf("failed to create new access token: %v", err)
 				s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
@@ -830,6 +830,11 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
 	}
 
 	grantType := r.PostFormValue("grant_type")
+	if !contains(s.supportedGrantTypes, grantType) {
+		s.logger.Errorf("unsupported grant type: %v", grantType)
+		s.tokenErrHelper(w, errUnsupportedGrantType, "", http.StatusBadRequest)
+		return
+	}
 	switch grantType {
 	case grantTypeDeviceCode:
 		s.handleDeviceToken(w, r)
@@ -839,6 +844,8 @@ func (s *Server) handleToken(w http.ResponseWriter, r *http.Request) {
 		s.withClientFromStorage(w, r, s.handleRefreshToken)
 	case grantTypePassword:
 		s.withClientFromStorage(w, r, s.handlePasswordGrant)
+	case grantTypeTokenExchange:
+		s.handleTokenExchange(w, r)
 	default:
 		s.tokenErrHelper(w, errUnsupportedGrantType, "", http.StatusBadRequest)
 	}
@@ -917,7 +924,7 @@ func (s *Server) handleAuthCode(w http.ResponseWriter, r *http.Request, client s
 }
 
 func (s *Server) exchangeAuthCode(w http.ResponseWriter, authCode storage.AuthCode, client storage.Client) (*accessTokenResponse, error) {
-	accessToken, err := s.newAccessToken(client.ID, authCode.Claims, authCode.Scopes, authCode.Nonce, authCode.ConnectorID)
+	accessToken, _, err := s.newAccessToken(client.ID, authCode.Claims, authCode.Scopes, authCode.Nonce, authCode.ConnectorID)
 	if err != nil {
 		s.logger.Errorf("failed to create new access token: %v", err)
 		s.tokenErrHelper(w, errServerError, "", http.StatusInternalServerError)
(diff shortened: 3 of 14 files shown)


seankhliao:
I think this is good enough for a first round of reviews.
Not quite sure about what to do with the CodeQL failures on logging user input

baz-bakerhughes:
Truly would like this merged into main so dex can be our complete Authentication proxy solution.

sagikazarmark:
I can't promise it's gonna happen before KubeCon, but we will get to this.

gnunn1:
Just as a data point if it helps, I tested this PR with the Openshift connector after adding a very hacky implementation of the TokenIdentity method to the connector and it worked fine.

blairdrummond:
Are there any docs/examples of this flow for Github Actions kicking around? I would be very interested in trying it out

nabokihms:
@seankhliao, sorry for the long delay. I have no excuses (besides KubeCon 😄 ). The code looks good to me.
I want to test the feature manually and give my green light after that.

BTW could you please rebase the PR to resolve conflicts?

seankhliao:
@nabokihms rebased, no need for excuses :)

@blairdrummond https://github.com/seankhliao/testrepo0191
dex config is in config.yaml
curl commands in the workflow

blairdrummond:
@aidansteele, just incase you're interested, I recently read [aws-iam-oidc-idps-need-more-controls](https://awsteele.com/blog/2021/10/12/aws-iam-oidc-idps-need-more-controls.html), and it seems like this might help resolve the things you have implemented via [jwtex](https://github.com/aidansteele/jwtex). I think the only thing missing is [nested json claim mapping](https://github.com/dexidp/dex/issues/2143) which is still a TODO.

gnunn1:
I have the OpenShift connector updated to implement token exchange for this PR, should I wait for this to be merged before submitting a PR since it depends on this PR?

dhasmac-bh:
Is there any info on how we can test with oidc type connector like auth0.  I want to test if the access token through client-credentials grant be exchanged for DEX approved token through token-exchange.

dhasmac-bh:
> Is there any info on how we can test with oidc type connector like auth0. I want to test if the access token through client-credentials grant be exchanged for DEX approved token through token-exchange.

It worked.

baz-bakerhughes:
OK.  We have been running this build with this PR for about 3 weeks in our test environment. I am curious if there is any concern that this will be rejected.  If it eventually would be accepted, we can promote the version with the PR and then swap once its in.  But if there is a concern that having dexidp support this functionality (even if it was re coded in a different way) but there is a philosophical reason against this that would change our plans. Any thoughts?

baz-bakerhughes:
@nabokihms Is there any assistance needed to get this merged into main.  We have been using the branch build in our environment for a while now and tested all the functional options on tokens we are using. That is @dhasmac-bh has and it all works great.

- auth code flow 
- token exchange

nabokihms:
@baz-bakerhughes, for now, no assistance is needed. RFC was accepted, so it is only a matter of testing. Because of the PTO season, we a throttling with this PR a bit, but no worries. We will merge it eventually.

cmd/dex/config.go
```diff
@@ -129,6 +129,10 @@ func (p *password) UnmarshalJSON(b []byte) error {
 
 // OAuth2 describes enabled OAuth2 extensions.
 type OAuth2 struct {
+	// list of allowed grant types,
+	// defaults to all supported types
+	GrantTypes []string `json:"grantTypes"`
```
nabokihms, line 134: There is an exciting nug when the grantTypes contains the `password` grant.

The discovery endpoint will return 
```
"grant_types_supported": [
  "password",
  "password"
],
```

For unknown grant type, it will return this claim. For example, for `test` grantTypes it will be 
```
"grant_types_supported": [
  "test"
],
```
which is not valid.

I think grant types should be applied as the final filter here only to find the intersection between known enabled grants and allowed grants from the config.
https://github.com/dexidp/dex/blob/47a0e067a174119224b49aba1549eb8af1550112/server/server.go#L216-L237

nabokihms (changes requested):

server/handlers.go
```diff
@@ -1319,21 +1326,121 @@ func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, cli
...
+		Groups:            identity.Groups,
+	}
+	resp := accessTokenResponse{
+		IssuedTokenType: requestedTokenType,
+		TokenType:       "bearer",
+	}
+	var expiry time.Time
+	switch requestedTokenType {
+	case tokenTypeID:
+		resp.AccessToken, expiry, err = s.newIDToken(resource, claims, scopes, "", "", "", connID)
```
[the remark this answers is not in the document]
seankhliao: My reading of the RFC: https://datatracker.ietf.org/doc/html/rfc8693#section-2.2.1
says of the access_token field:

> access_token
>   REQUIRED. The security token issued by the authorization server in response to the token exchange request. The access_token parameter from [Section 5.1](https://www.rfc-editor.org/rfc/rfc6749#section-5.1) of [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)] is used here to carry the requested token, which allows this token exchange protocol to use the existing OAuth 2.0 request and response constructs defined for the token endpoint. The identifier access_token is used for historical reasons and the issued token need not be an OAuth access token.

The id_token field seems to only be used in the response for eachanging an auth code in a normal oauth2 flow

server/handlers.go
```diff
@@ -1319,21 +1326,121 @@ func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, cli
...
+	conn, err := s.getConnector(connID)
+	if err != nil {
+		s.logger.Errorf("Failed to get connector: %v", err)
+		s.renderError(r, w, http.StatusBadRequest, "Requested resource does not exist")
+		return
+	}
+	teConn, ok := conn.Connector.(connector.TokenIdentityConnector)
+	if !ok {
+		s.logger.Errorf("Connector doesn't implement token exchange: %v", connID)
+		s.renderError(r, w, http.StatusBadRequest, "Requested resource does not exist")
```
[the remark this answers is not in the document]
seankhliao: done, same for renderError a few lines above

connector/oidc/oidc.go
```diff
@@ -284,12 +285,29 @@ func (c *oidcConnector) Refresh(ctx context.Context, s connector.Scopes, identit
 	return c.createIdentity(ctx, identity, token, refreshCaller)
 }
 
+func (c *oidcConnector) TokenIdentity(ctx context.Context, subjectTokenType, subjectToken string) (connector.Identity, error) {
+	var identity connector.Identity
+	token := &oauth2.Token{
+		AccessToken: subjectToken,
+	}
+	if subjectTokenType == "urn:ietf:params:oauth:token-type:access_token" {
```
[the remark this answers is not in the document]
seankhliao: They're currently consts in server/ , but server imports the connectors, so exporting them there would be an import cycle.
Is there a more appropriate place to export them?

server/handlers.go
```diff
@@ -1319,21 +1326,121 @@ func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, cli
...
+		return
+	}
+	identity, err := teConn.TokenIdentity(ctx, subjectTokenType, subjectToken)
+	if err != nil {
+		s.logger.Errorf("Failed to verify subject token: %v", err)
```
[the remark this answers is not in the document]
seankhliao: I've gone with all lowercase

server/handlers.go
```diff
@@ -1319,21 +1326,121 @@ func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, cli
...
+	// TODO: check global allowed grant types?
+
+	if err := r.ParseForm(); err != nil {
+		s.logger.Errorf("Could not parse request body: %v", err)
+		s.tokenErrHelper(w, errInvalidRequest, "", http.StatusBadRequest)
+		return
+	}
+	q := r.Form
+
+	connID := q.Get("audience")                         // OPTIONAL, use for connector ID
```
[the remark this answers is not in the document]
seankhliao: I've moved this down and clarified that it is required for dex, but optional in the RFC.
(the comments we left over from working out which fields we'd need/use)

connector/oidc/oidc.go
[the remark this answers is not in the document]
seankhliao: I added this so that createIdentity called below would attempt to verify the token and get claims.

I think I need to restructure this part a bit: given an exchange caller, we must always verify the token first.
I've made the exchange verifier block a separate condition in createIdentity

connector/oidc/oidc.go
```diff
@@ -284,12 +285,29 @@ func (c *oidcConnector) Refresh(ctx context.Context, s connector.Scopes, identit
...
+		token = token.WithExtra(map[string]any{
+			"id_token": subjectToken,
+		})
+	}
+	return c.createIdentity(ctx, identity, token, exchangeCaller)
```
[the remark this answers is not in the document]
seankhliao: I'm not sure how to get that error.
Dex config says this, which makes me think it should be valid to set getUserInfo

>     # When enabled, the OpenID Connector will query the UserInfo endpoint for additional claims. UserInfo claims
>     # take priority over claims returned by the IDToken. This option should be used when the IDToken doesn't contain
>     # all the claims requested.

connector/oidc/oidc.go
```diff
@@ -284,12 +285,29 @@ func (c *oidcConnector) Refresh(ctx context.Context, s connector.Scopes, identit
...
 	var claims map[string]interface{}
 
 	rawIDToken, ok := token.Extra("id_token").(string)
 	if ok {
-		idToken, err := c.verifier.Verify(ctx, rawIDToken)
+		verifier := c.verifier
+		if caller == exchangeCaller {
+			verifier = c.provider.Verifier(&oidc.Config{SkipClientIDCheck: true})
+		}
+		idToken, err := verifier.Verify(ctx, rawIDToken)
```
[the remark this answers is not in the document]
seankhliao: I think this was a bug from me only testing with access tokens, testing with my new code, I get 

> time="2023-06-10T11:09:22Z" level=error msg="failed to verify subject token: oidc: error loading userinfo: oidc: user info endpoint is not supported by this provider"

Which I think just a restriction on the provider I used rather than a fundamental limitation of access -> id

cmd/dex/config.go
seankhliao: I've implemented this

connector/oidc/oidc.go
```diff
@@ -284,12 +285,29 @@ func (c *oidcConnector) Refresh(ctx context.Context, s connector.Scopes, identit
 	return c.createIdentity(ctx, identity, token, refreshCaller)
 }
 
+func (c *oidcConnector) TokenIdentity(ctx context.Context, subjectTokenType, subjectToken string) (connector.Identity, error) {
```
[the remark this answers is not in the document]
seankhliao: I've added (a very basic) one

seankhliao:
I've reworked the TokenIdentiy part a bit so it should properly work with both id and access tokens.
I think getUserInfo should work now, except I don't have a real backend to test against, the providers I'm using don't support that endpoint at all

nabokihms:
@seankhliao I tested a dex - dex communication (configuring the second dex instance as an oidc provider for the first) 😄.

seankhliao:
Ok, I think dex - dex works too with my updated code

server/handlers.go
```diff
@@ -1319,21 +1326,112 @@ func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, cli
...
+	// TODO: check global allowed grant types?
+
+	if err := r.ParseForm(); err != nil {
+		s.logger.Errorf("could not parse request body: %v", err)
+		s.tokenErrHelper(w, errInvalidRequest, "", http.StatusBadRequest)
+		return
+	}
+	q := r.Form
+
+	resource := q.Get("resource")                       // OPTIONAL, use for issued token audience
```
nabokihms, line 1341: I started thinking that adding this parameter was not at all right.
In Dex, the audience claim always points to a client (or clients). There may be private or public clients.

For private clients, credentials are always required. With the `resource` parameter now, users can issue tokens for any client they want without credentials (even for not existing ones, which seems to be a hole.

Before merging, we need to resolve this. One way is to use the client authentication, as for some other grants (with a decorator), and abstain from the resource parameter. https://github.com/dexidp/dex/blob/2e4d724935fb121aff49a0bbf35325eafc79fa57/server/handlers.go#L841

The other option is to reject any requests with the resource parameter equals to one of the existing client ids (which still leaves space for security holes but is less problematic).

connector/oidc/oidc.go
```diff
@@ -284,16 +285,32 @@ func (c *oidcConnector) Refresh(ctx context.Context, s connector.Scopes, identit
...
 		if err != nil {
 			return identity, fmt.Errorf("oidc: failed to verify ID Token: %v", err)
 		}
 
+		if err := idToken.Claims(&claims); err != nil {
+			return identity, fmt.Errorf("oidc: failed to decode claims: %v", err)
+		}
+	} else if caller == exchangeCaller {
+		// AccessToken here could be either an id token or an access token
+		idToken, err := c.provider.Verifier(&oidc.Config{SkipClientIDCheck: true}).Verify(ctx, token.AccessToken)
```
nabokihms, line 310: Now it works even if I use an access subject token with the id token subject token type 😅 
It may be my provider-specific problem.

server/handlers.go
```diff
@@ -1319,21 +1326,112 @@ func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, cli
 	s.writeAccessToken(w, resp)
 }
 
+func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	// TODO: check global allowed grant types?
```
nabokihms, line 1332: It seems not valid anymore because there is a check already. Let's just remove the TODO.

server/handlers.go
nabokihms: It is also strange because the audience claim is required, but the resource parameter is optional. Thus, by default, each token has an empty aud claim (because the aud claim is required by spec).

I checked some popular providers:
1. Okta uses the audience parameter value for the audience claim value https://developer.okta.com/docs/guides/set-up-token-exchange/main/#token-exchange-request-from-service-app-to-api
2. Keycloak uses the audience claim as the client_id, but they also have semi-required client id and client secret parameters (as suggested per my top comment) https://www.keycloak.org/docs/latest/securing_apps/#_token-exchange

We probably need to use our usual `connector` parameter to specify the connector instead of the resource. What do you think?

nabokihms (changes requested):

connector/oidc/oidc.go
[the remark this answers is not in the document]
nabokihms: I cannot see the error anymore, either. Guess it is resolved.

server/handlers.go
seankhliao: done

connector/oidc/oidc.go
seankhliao: maybe, at least the CI providers I'm working with appear to use ID and access tokens interchangeably. I don't think it really matters since they all get put into `authorization: Bearer $token` anyway?

server/handlers.go
seankhliao: I hadn't considered that dex required the issued token audience to be that of a known client (I've been working with sigstore/cosign in gcp recently, sigstore requires a specific audience and gcp doesn't require registering clients).

If I understand correctly, with dex's support cross client trust, we need 3 pieces of information:
- current client id
  - we can, and should require client auth via the decorator like with other handlers
- target client id (optional)
  - use `audience` in the incoming request, defaults to current client id
- connector id
  - use `connector_id` like the `/auth` endpoint currently does
  - I thought about using `resource` for this, but rereading the RFC it seems `resource` is closer to `audience`

Thoughts on this approach?

seankhliao: actually, no dex uses a scope for cross client trust (target client id).
Then I think we should just ignore `resource` and `audience` in the incoming request then

seankhliao: I've updated the PR to match the approach I laid out above.

connector/oidc/oidc.go
nabokihms: Ok, this is just an odd thing I noticed. Don't thing this is a problem now.

nabokihms (approved):
@seankhliao awesome! Thank you for your work. It was a long journey!

P.S. It would be awesome if you could add a doc about the Token Exchange feature to [the Dex documentation](https://github.com/dexidp/website).

sagikazarmark:
Thank you @seankhliao for making this happen!

seankhliao:
I've created https://github.com/dexidp/website/pull/140 for website documentation
and https://github.com/dexidp/dex/pull/3027 for adding grantTypes to the config example (linked from the website).

nabokihms:
There is no release date for the upcoming releases, but it will probably be August something.

sl1pm4t:
Thanks for this feature, we are using it already to enable ArgoCD to authenticate with managed clusters across multiple clouds. 
To facilitate this, and in case it's helpful for anyone else, I wrote this Kubernetes Credential helper:
https://github.com/sl1pm4t/tokenxchange

nabokihms:
@sl1pm4t great, thank you for sharing 🙌

PR #2812, opened 2023-02-03
DEP for RFC 8693 OAuth 2 Token Exchange

seankhliao:
#### Overview

DEP for RFC 8693 OAuth 2.0 Token Exchange / grant_type urn:ietf:params:oauth:grant-type:token-exchange

Discussion / design requested before implementation

```release-note
NONE
```

### DIFF
--- a/docs/enhancements/token-exchange-2023-02-03-#2812.md
+++ b/docs/enhancements/token-exchange-2023-02-03-#2812.md
@@ -1,0 +1,175 @@
+# Dex Enhancement Proposal (DEP) 2812 - 2023-02-03 - Token Exchange
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals/Pain](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+    - [User Experience](#user-experience)
+    - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+    - [Risks and Mitigations](#risks-and-mitigations)
+    - [Alternatives](#alternatives)
+- [Future Improvements](#future-improvements)
+
+## Summary
+
+[RFC 8693] specifies a new OAuth2 `grant_type` of `urn:ietf:params:oauth:grant-type:token-exchange`.
+Using this grant type, when clients start an authentication flow with Dex,
+in lieu of being redirected to their upstream IDP for authentication on demand,
+clients can present an independently obtained, valid token from their IDP to Dex.
+This is primarily useful in fully automated environments with job/machine identities,
+where there is no human in the loop to handle browser-based login flows.
+This DEP proposes to implement the new grant type for Dex.
+
+[RFC 8693]: https://www.rfc-editor.org/rfc/rfc8693.html
+
+## Context
+
+- [#1668 Question: non-web based clients?]
+  was closed with no real resolution
+- [#1484 Token exchange for external tokens]
+  mentions that Keycloak has a similar capability
+- [#2657 Get OIDC token issued by Dex using a token issued by one of the connectors] 
+  is similar to the previous issue, but this time links to the new (January 2020) [RFC 8693].
+
+I believe the context for all of these are similar:
+a downstream project using Dex as its only IDP wants to grant access to programmatic clients
+without issuing long lived API tokens.
+
+Examples of downstream issues:
+
+- [argoproj/argo-cd#11632 Argocd SSO login via Azure AD Auth using OIDC not work for cli sso login]
+
+Other related Dex issues:
+
+- [#2450 Non-OIDC JWT Connector] is a functionally similar request, but expanded to arbitrary JWTs
+- [#1225 GitHub Non-Web application flow support] also asks for an exchange, but for an opaque Github PAT
+
+More broadly, this fits into recent movements to issue machine identities:
+
+- [GCP Service Identity](https://cloud.google.com/run/docs/securing/service-identity)
+- [AWS Execution Role](https://docs.aws.amazon.com/lambda/latest/dg/lambda-intro-execution-role.html)
+- [Github Actions OIDC](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect)
+- [CircleCI OIDC](https://circleci.com/docs/openid-connect-tokens/)
+- [Kubernetes Service Accounts](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/)
+- [SPIFFE](https://spiffe.io/)
+
+and granting access to resources based on trusting federated identities:
+
+- [GCP Workload Identity Federation](https://cloud.google.com/iam/docs/workload-identity-federation)
+- [AWS STS AssumeRoleWithWebIdentity](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html)
+
+[#1484 Token exchange for external tokens]: https://github.com/dexidp/dex/issues/1484
+[#1668 Question: non-web based clients?]: https://github.com/dexidp/dex/issues/1668
+[#2657 Get OIDC token issued by Dex using a token issued by one of the connectors]: https://github.com/dexidp/dex/issues/2657
+[argoproj/argo-cd#11632 Argocd SSO login via Azure AD Auth using OIDC not work for cli sso login]: https://github.com/argoproj/argo-cd/issues/11632
+[#2450 Non-OIDC JWT Connector]: https://github.com/dexidp/dex/issues/2450
+[#1225 GitHub Non-Web application flow support]: https://github.com/dexidp/dex/issues/1225
+
+An initial attempt is at [#2806](https://github.com/dexidp/dex/pull/2806)
+
+## Motivation
+
+### Goals/Pain
+
+The goal is to allow programmatic access to Dex-protected resources 
+without the use of static/long-lived secret tokens (API keys, username/password)
+or web-based redirect flows.
+Such scenarios are common in CI/CD workflows,
+and in general automation of common tasks.
+
+### Non-goals
+
+- Work will be scoped to just the OIDC connector
+- [RFC 8693 Section 2.1.1. Relationship between Resource, Audience, and Scope]
+  details more complex authorization checks based on targeted resources.
+  This is considered out of scope.
+
+[RFC 8693 Section 2.1.1. Relationship between Resource, Audience, and Scope]: https://www.rfc-editor.org/rfc/rfc8693.html#name-relationship-between-resour
+
+## Proposal
+
+### User Experience
+
+Clients can make `POST` requests with `application/x-www-form-urlencoded` 
+parameters as specified by [RFC 8693] to Dex's `/token` endpoint.
+If successful, an access token will be returned,
+allowing direct authentication with Dex.
+No refresh tokens will be issued,
+perform a new exchange (possibly with refreshed upstream tokens) to obtain a new access token.
+
+The request parameters from [RFC 8693 Section 2.1](https://www.rfc-editor.org/rfc/rfc8693.html#name-request):
+
+- `grant_type`: REQUIRED - `urn:ietf:params:oauth:grant-type:token-exchange`
+- `resource`: OPTIONAL - the `audience` in the issued Dex token
+- `audience`: REQUIRED (RFC OPTIONAL) - the connector to verify the provided token against
+- `scope`: OPTIONAL - the `scope` in the issued Dex token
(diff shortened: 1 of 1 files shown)


sagikazarmark:
@seankhliao I really appreciate writing this up. We'll take a look as soon as we can!

docs/enhancements/token-exchange-2023-02-03-#2812.md
```diff
@@ -0,0 +1,171 @@
+# Dex Enhancement Proposal (DEP) 2812 - 2023-02-03 - Token Exchnage
```
thinkverse, line 1: Little typo fix. 👍

```suggestion
# Dex Enhancement Proposal (DEP) 2812 - 2023-02-03 - Token Exchange
```

docs/enhancements/token-exchange-2023-02-03-#2812.md
```diff
@@ -0,0 +1,171 @@
...
+Examples of downstream issues:
+
+- [argoproj/argo-cd#11632 Argocd SSO login via Azure AD Auth using OIDC not work for cli sso login]
+
+Other related Dex issues:
+
+- [#2450 Non-OIDC JWT Connector] is a functionally similar request, but expanded to arbitrary JWTs
+- [#1225 GitHub Non-Web application flow support] also asks for an exchange, but for an opaque Github PAT
+
+More broadly, this fits into recent movements to issue machines identies:
```
thinkverse, line 50: Another small typo. 🙂

```suggestion
More broadly, this fits into recent movements to issue machine identities:
```

docs/enhancements/token-exchange-2023-02-03-#2812.md
```diff
@@ -0,0 +1,171 @@
...
+- [Proposal](#proposal)
+    - [User Experience](#user-experience)
+    - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+    - [Risks and Mitigations](#risks-and-mitigations)
+    - [Alternatives](#alternatives)
+- [Future Improvements](#future-improvements)
+
+## Summary
+
+[RFC 8693] specifies a new OAuth2 `grant_type` of `urn:ietf:params:oauth:grant-type:token-exchange`.
```
nabokihms, line 18: Note: do not forget to add it to the list of supported grants of the discovery endpoint. 
(just to not forget)

docs/enhancements/token-exchange-2023-02-03-#2812.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+### Implementation Details/Notes/Constraints
+
+- Connectors expose a new interface `TokenIdentity` that will verify the given token and return the associated identity.
+  A Dex access/id token is then minted for the given identity.
+
+- It's unclear if passing through `scope` is correct
+- Same for `audience` and `resource`. The current plan is based on how [GCP Workload Identity] Pools use the fields.
+- The internal `newAccessToken()` should also include the expiry time?
```
nabokihms, line 143: In token, there are `iat` and `exp` claims. `exp` - `iat` = expires_in, so it is already included.

(Maybe I got the question wrong)

docs/enhancements/token-exchange-2023-02-03-#2812.md
```diff
@@ -0,0 +1,171 @@
...
+  config:
+    issuer: https://oidc.my-platform.example/
+```
+
+### Implementation Details/Notes/Constraints
+
+- Connectors expose a new interface `TokenIdentity` that will verify the given token and return the associated identity.
+  A Dex access/id token is then minted for the given identity.
+
+- It's unclear if passing through `scope` is correct
```
nabokihms, line 141: What is the main concern there?

docs/enhancements/token-exchange-2023-02-03-#2812.md
```diff
@@ -0,0 +1,171 @@
...
+- Same for `audience` and `resource`. The current plan is based on how [GCP Workload Identity] Pools use the fields.
+- The internal `newAccessToken()` should also include the expiry time?
+- The `accessTokenResponse` struct will need new optional fields, as well as making some existing fields optional (`omitempty`)
+- `actor_token` and `actor_token_type` are "MUST ... if the actor token is present,
```
nabokihms, line 145: We can leave these two fields out of the scope of the initial implementation.

docs/enhancements/token-exchange-2023-02-03-#2812.md
```diff
@@ -0,0 +1,171 @@
...
+If successful, an access token will be returned,
+allowing direct authentication with Dex.
+No refresh tokens will be issued,
+perform a new exchange (possibly with refreshed upstream tokens) to obtain a new access token.
+
+The request parameters from [RFC 8693 Section 2.1](https://www.rfc-editor.org/rfc/rfc8693.html#name-request):
+
+- `grant_type`: REQUIRED - `urn:ietf:params:oauth:grant-type:token-exchange`
+- `resource`: OPTIONAL - the `audience` in the issued Dex token
+- `audience`: REQUIRED (RFC OPTIONAL) - the connector to verify the provided token against
```
nabokihms, line 107: Can it be optional if only a single connector is configured?

docs/enhancements/token-exchange-2023-02-03-#2812.md
```diff
@@ -0,0 +1,171 @@
...
+- `scope`: OPTIONAL - the `scope` in the issued Dex token
+- `requested_token_type`: OPTIONAL - one of `urn:ietf:params:oauth:token-type:access_token` or `urn:ietf:params:oauth:token-type:id_token`, defaulting to access token
+- `subject_token`: REQUIRED - the token issued by the upstream IDP
+- `subject_token_type`: REQUIRED - only  `urn:ietf:params:oauth:token-type:id_token` will be accepted
```
nabokihms, line 111: 1. OIDC connector can work with access tokens by requesting the userInfo endpoint. 
2. In the future, the token type should be the `...:access_token` for other connectors like the LDAP connector.

My opinion is that Dex should accept both tokens and pass them to a connector with an additional tag that this token is the access / id token.

docs/enhancements/token-exchange-2023-02-03-#2812.md
```diff
@@ -0,0 +1,171 @@
...
+- type: oidc
+  id: my-platform
+  name: My Platform
+  config:
+    issuer: https://oidc.my-platform.example/
+```
+
+### Implementation Details/Notes/Constraints
+
+- Connectors expose a new interface `TokenIdentity` that will verify the given token and return the associated identity.
```
nabokihms, line 138: This is controversial because it adds more entropy to the connector's interface.

Dex has three connector types (four, to be precise, but let's forget about SAML for now) - Password, Callback, and Refresh. It makes seven possible combinations of connector functions. With an additional interface, the number will be fourteen. 

However, thinking about alternatives, I like the solution of exposing an additional interface. In the future, a JWT connector can be mentioned in this DEP that will implement only the TokenIdentity interface.

docs/enhancements/token-exchange-2023-02-03-#2812.md
```diff
@@ -0,0 +1,171 @@
...
+We limit the potential impact by not issuing refresh tokens, preventing persistent access.
+Combined with short token lifetimes, it should limit the period of time between authentication to upstream IDPs.
+
+### Alternatives
+
+- Continue to use static keys - 
+  this is a secret management nightmare 
+  and quite painful when client storage of keys is [breached](https://circleci.com/blog/january-4-2023-security-alert/)
+
+## Future Improvements
```
nabokihms, line 163: I like that the password credentials grant can be deprecated entirely with this feature. Password connectors can accept encoded credentials (like basic auth header format) as a token and exchange them for Dex tokens.

The real pain is that Dex only allows a single connector to serve password credentials grant requests, but with this brand new rfc users will be able to specify a target connector. As for me, seems like a great additional improvement.

nabokihms (commented):
I have a couple of questions, but overall it looks solid. Thank you, @seankhliao.

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: Probably. it wasn't really clear how I'd do that (iterate over connectors?).
I went with this as it made implementation easier and matched what GCP was doing.
In theory, dex has all the info (at least for oauth2) to not need the connector id, since it can match on the issuer url (from config and from subject id token)

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: Ah right, sort of forgot about the possibility.

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: Yes, I was initially hesitant to expose an additional interface, but felt that this was the cleanest way to expose the logic (I started with inlining the verify call in the handler)

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: I wasn't sure if scope applied to dex or to the downstream resource.

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: The info is encoded in the token itself, but the token response json object has an optional `expires_in` field that could be filled

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: I'll add this as a future improvement,

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: I probably would have forgotten...

docs/enhancements/token-exchange-2023-02-03-#2812.md
nabokihms: Yeah, it must be an error.

docs/enhancements/token-exchange-2023-02-03-#2812.md
nabokihms: I assume this affects the list of claims that dex should return.

docs/enhancements/token-exchange-2023-02-03-#2812.md
nabokihms: I think we will figure it out during the implementation stage. Maybe returning the expiration time is not a bad idea.

docs/enhancements/token-exchange-2023-02-03-#2812.md
nabokihms: We can make this field optional in the future without breaking anything. Let's leave this topic out of the initial implementation.

nabokihms:
I started thinking users may want to disable this grant because it opens a new attack surface. We have this option for password credentials granted, but with the current design, it will not be possible to restrict the possibility of exchanging a token.

The easiest way to solve this issue is to add a global setting that configures allowed grants.

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: I've removed it from the doc

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: Alright, that clears it up, removed.

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: I've added it under future improvements

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: also added to future improvements

docs/enhancements/token-exchange-2023-02-03-#2812.md
seankhliao: added that access_token is also allowed

seankhliao:
I've added a mention of global/connector settings in the UX section.
eg for us, we'll likely want our Okta (human) connector to disallow it,
only allowing it for our CI connector.

nabokihms (approved):

nabokihms:
From my perspective, the proposal is solid enough to be implemented! Good job.
@sagikazarmark, since this will be a significant change, do you want to add something to the design stage?

sagikazarmark (approved):

PR #2831, opened 2023-02-16
chore: Build Dex in parallel

nabokihms:
#### Overview

This is a proof of concept of CI refactoring.
1. Run the build in parallel.
2. Use Go for cross-compilation instead of QEMU.
3. Bump Go version  (1.20) and golangci-lint (1.51).
4. ~Use runners with more CPU.~ Use standard lightweight runners. They are always available and pick up jobs faster.
5. Remove the `Docker` job.

This PR is supposed to help reduce the number of random freezes during cross-compilation and consume fewer resources/time.

Do we need integration tests for final builds to be sure that Dex works as expected? Do we want to execute tests for different architectures?

```release-note
chore: Build Dex in parallel and cross-compile with Go
```


sagikazarmark:
Uh, that's a massive change @nabokihms. Are you sure you branched off from the right tree? (eg. what's ent generated code doing in this PR?)

sagikazarmark:
> Use Go for cross-compilation instead of QEMU.

I recall using CGO somewhere (for DNS?). Is there something we lose by cross compiling?

sagikazarmark:
The problem with this approach is that the container registry does not "merge" images for different platforms, so basically the last one that gets pushed wins.

nabokihms:
1. Yes, there is an issue in master. Fix is here https://github.com/dexidp/dex/pull/2833
2. SQlite I believe
3. Are you sure? We still use build-and-push action for images and still provide architectures.

sagikazarmark:
> SQlite I believe

That is a problem.

> Are you sure? We still use build-and-push action for images and still provide architectures.

I'm fairly confident. I've tried this before and it didn't work.

That's exactly why there is a separate step for creating the final image [here](https://github.com/docker/build-push-action/issues/671#issuecomment-1373055782).

nabokihms:
@sagikazarmark ok, I think I can fix the build and avoid executing the final step in parallel. What do you think about the concept? Are you strongly disagree with the approach?

.github/workflows/artifacts.yaml
```diff
@@ -55,7 +111,7 @@ jobs:
         uses: docker/login-action@v2
         with:
           registry: ghcr.io
-          username: ${{ github.repository_owner }}
+          username: ${{ github.actor }}
```
sagikazarmark, line 114: `repository_owner` makes it easier to fork the project.

sagikazarmark:
I'm not against it per se. I absolutely love the reduced build times. Let's just say I have a few reservations:

- By not having a self-contained build pipeline in Docker, the entire process becomes less portable and more tedious to run builds manually if necessary
- Locking ourselves into GHA makes me feel uncomfortable. It's not like we can't move or that we want to, but I like the idea of running the same pipelines locally.

Maybe we can do some improvements on these fronts. I'll also try an alternative workflow that parallelizes the docker builds as well. It won't be this performant, but we will have at least alternatives to choose from.

Overall, I think this is great work and shows us how much room we have for improvement.

nabokihms:
@sagikazarmark 
I refactored the release-binary logic to cross-compile binaries in a container. It should fix some issues of the previous implementation and resolve some of your concerns.

Could you please take a look and share your opinion?

nabokihms:
Closing if favor of https://github.com/dexidp/dex/pull/2898#event-9126963668

PR #2842, opened 2023-03-01
fix: exit serve with cascaded errors

aslafy-z:
#### Overview

Cascade `runServe` errors to cobra CLI and exit 1.

When running in Kubernetes, we encountered `failed to initialize server: server: Failed to open connector google: failed to open connector: failed to create connector google: failed to get provider: Get "https://accounts.google.com/.well-known/openid-configuration": dial tcp: lookup accounts.google.com: i/o timeout` that did not triggered a restart. As there is no retry mecanism, IMHO, the dex server should exit so kubernetes (systemd/you name it) can restart it. 

I believe this was changed by https://github.com/dexidp/dex/pull/1921

```release-note
fix: exit serve with cascaded errors
```

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -57,7 +57,7 @@ func commandServe() *cobra.Command {
 		Args:    cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
-			cmd.SilenceErrors = true
+			cmd.SilenceErrors = false
 
 			options.config = args[0]
 


sagikazarmark:
Thanks for the contribution @aslafy-z 

I'm afraid this is not the solution though. That flag only controls whether Cobra prints the error or not...which doesn't really matter, because Dex does: https://github.com/dexidp/dex/blob/master/cmd/dex/main.go#L25

If you received that error, then Dex should have exited as the next line is an `os.Exit`.

One thing I can imagine is that the error is not being returned, thus not reaching Cobra at all, although that would be weird to say the least.

In any case, I doubt this flag is the problem.

Ben10k:
Hi. We are encountering the same behaviour. This usually happens, when dex is scheduled on a newly provisioned node, most likely the network stack is not completely initiatated at that point in time, and then `dex` prints out the log line
```
dex-server failed to initialize server: server: Failed to open connector google: failed to open connector: failed to create connector google: failed to get provider: Get "https://accounts.google.com/.well-known/openid-configuration": dial tcp: lookup accounts.google.com: i/o timeout
```
and hangs without exiting.

Ben10k:
Hi. We are encountering the same behaviour. This usually happens, when dex is scheduled on a newly provisioned node, most likely the network stack is not completely initiatated at that point in time, and then `dex` prints out the log line
```
dex-server failed to initialize server: server: Failed to open connector google: failed to open connector: failed to create connector google: failed to get provider: Get "https://accounts.google.com/.well-known/openid-configuration": dial tcp: lookup accounts.google.com: i/o timeout
```
and hangs without exiting.

@aslafy-z were you able to solve this issue?

PR #2844, opened 2023-03-03
feat: Bump dependencies and Makefile refactoring

nabokihms:
#### Overview

* Refactor Makefile
* Go 1.20 for test
* Bump kind
* Bump golangci-lint and fix errors
* Bumo ent 0.11.9

```release-note
feat: Bump dependencies and Makefile refactoring
```


.github/workflows/ci.yaml
```diff
@@ -64,22 +64,26 @@ jobs:
...
       - name: Checkout code
         uses: actions/checkout@v3
 
+      - name: Download tool dependencies
+        run: make deps
+
+      # Ensure that generated files were committed.
+      - name: Verify
+        run: make verify
+
```
sagikazarmark, line 78: Any particular reason to move this up?

connector/openshift/openshift.go
```diff
@@ -85,6 +85,10 @@ func (c *Config) OpenWithHTTPClient(id string, logger log.Logger,
 
 	wellKnownURL := strings.TrimSuffix(c.Issuer, "/") + wellKnownURLPath
 	req, err := http.NewRequest(http.MethodGet, wellKnownURL, nil)
+	if err != nil {
+		cancel()
```
sagikazarmark, line 89: TODO: wouldn't defer cancel make more sense here?

nabokihms: Yeah, maybe. i will change this.

.github/workflows/ci.yaml
nabokihms: I did it to avoid running CI scripts if generated code was not committed. In this scenario, we test not the actual code that should be tested but some intermediate state.  I wanted to avoid it as early as possible. If you think it is worth clarifying, I can leave a comment.

sagikazarmark: Makes sense!

sagikazarmark (approved):
LGTM

PR #2851, opened 2023-03-08
Add support for extra claims to authproxy connector

mkjpryor:
#### Overview

This PR adds support for some additional headers to be propagated into claims in the authproxy connector.

I am using Dex with authproxy to turn an existing IdP, whose code is not under my control, into an OIDC provider. I want to be able to propagate the user ID and email address from the underlying IdP as OIDC claims alongside the username, which is not possible with the current headers and claims supported by the authproxy connector.

This PR adds support for two extra headers that can be used to propagate the user ID and email from the underlying IdP, and sets the claims. When these additional headers are not set, it reduces to the previous mode.

```release-note
Adds support for two additional headers, for user ID and email, to the authproxy connector.
```

### DIFF
--- a/connector/authproxy/authproxy.go
+++ b/connector/authproxy/authproxy.go
@@ -106,7 +106,6 @@ func (m *callback) HandleCallback(s connector.Scopes, r *http.Request) (connecto
 	}
 	return connector.Identity{
 		UserID:            remoteUserId,
-		Username:          remoteUser,
 		PreferredUsername: remoteUser,
 		Email:             remoteUserEmail,
 		EmailVerified:     true,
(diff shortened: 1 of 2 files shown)


nabokihms (changes requested):
The feature looks good to me.
1. The test is required.
2. Linter has its suggestions that should be satisfied 🙂

mkjpryor:
@nabokihms 

I have rebased, addressed the linting issues and added an extra test. Can you re-review please?

nabokihms (approved):

nabokihms:
@mkjpryor sorry for the long delay. Merged!

PR #2857, opened 2023-03-13
Fix verify

sagikazarmark:


sagikazarmark:
Good question. But I suspect it may not entirely be dependabot's fault.

I ran `go mod tidy` with 1.20.2. CI runs with 1.20.1 and it fails in this PR.

nabokihms:
Quick fix https://github.com/dexidp/dex/pull/2858

PR #2047, opened 2021-03-16
Adjust scopes to be in line with the Microsoft Identity Platform v2.

schuhu:
#### Overview

This PR adds the openid scope necessary for the Microsoft Identity Platform v2 as well as changes the user and group scopes to the ones now supported on the Microsoft graph API. 
This PR was created together with Microsoft backend engineers and fixes https://github.com/dexidp/dex/issues/1855

#### What this PR does / why we need it

The current Microsoft connector just works. But if you enabled features like 2fa on your Microsoft Azure AD Application this is not enforced, because the scopes used trigger the Azure App in a different way than intended. 

So, to be able to use 2fa on a Microsoft Azure AD App, you need those changes. 

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #2877, opened 2023-03-27
fix: support Microsoft Identity Platform v2 scopes

### LINKED ISSUE #1855: Add openid scope to request for Microsoft Azure connector
We have a Conditional Access policy set up in Azure with a custom control for Duo MFA but it is not being applied to Dex authentications because the request is missing the openid scope. 

Per Microsoft, the openid scope needs to be added to apply MFA at the application level rather than the resource level (Microsoft Graph). 

Is there a way to add this scope to the request?

scopes currently returned:
user.read
directory.read.all

hudac:
#### Overview

Added Microsoft Identity Platform v2 specific scopes in order to also support conditional access policies, like MFA, TermsOfUse, etc. It does not change the behaviour of current implementation and works with or without conditional access policy configured.

We have this fix in production for five months without any issues.

In order to support enforced policies, like MFA, you need to use the Microsoft Identity Platform v2 with according scopes. The configuration on Dex-side remains the same and works as of today, with or without conditional access, groups and refresh tokens.

See also: https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#the-default-scope

Closes #1855

This PR is based on #2047 and contains the changes to the MS connector only. It has been tested on top of v2.35 and v2.36.

```release-note
NONE
```


hudac:
Hi @nabokihms is there anything I can support with to get this PR merged?

We have this fix on production (on v2.41) since at least 2023 without any issues.

Issue #2875, opened 2023-03-23
Skip redirection approval regression


nabokihms:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.36.0

### Storage Type

Kubernetes

### Installation Type

Binary, Official Helm chart

### Expected Behavior

I'd like to see a consent screen if a provider requested some of my personal data.

### Actual Behavior

No consent screen.

### Steps To Reproduce

Launch our demo application:

```
make build
make example
bin/dex examples/config-dev.yaml
bin/example-app
```

### Additional Information

This is a regression after https://github.com/dexidp/dex/pull/2805 because of the following clause 
https://github.com/dexidp/dex/blob/2bb4896d120efec1adc98f5db178a6d7206c9cc7/server/handlers.go#L526

1. By default consent screen should always be shown
2. If the consent screen is off by settings, it is still possible to show it using the force approval prompt option.

sagikazarmark:
@nabokihms nice catch. Are you planning to provide a patch?

nabokihms:
Yes, I will.

PR #2897, opened 2023-04-13
fix: Do not skip approval screen by default

nabokihms:
#### Overview

Skip redirect to approval only if skip approval is enabled and the approval prompt is not forced.

Fixes https://github.com/dexidp/dex/issues/2875

```release-note
Fix the bug when Dex skips approval screen by default
```


sagikazarmark (approved):
LGTM

sagikazarmark:
Do we want to backport this as a patch release?

PR #2902, opened 2023-04-15
microsoft: get rid of User.Read scope which leads to additional consents

xaurx:
It seems that when a user is already logged in with Microsoft in a browser and dex requires User.Read scope it forces Microsoft to show an additional screen from time to time:

-----------
Permissions Requested
...
This app would like to:
Sign in and read user profile
-----------

If instead a more standart OIDC connector procedure is used with openid,profile,email scopes - it doesn't happen and no microsoft screens are shown.

This patch kills the need for User.Read scope and follows a more standart OIDC connector procedure for Microsoft login. Note, microsoft doesn't fully follow spec, so issuer field in well-known configuration and token is verified manually.

```release-note

```


cardoe:
I'd be a -1 on this as the Microsoft connector is oauth2 and not oidc. It's designed to work with Microsoft's not quite standard v1 endpoints while their v2 endpoints which are OIDC standard can be used via the regular oidc backend. Since this is meant for the non-standard v1 it's using the necessary scopes for that. What you're doing here is making the Microsoft connector behave like the oidc connector. So just use that connector.

Issue #2653, opened 2022-09-07
Google Connector with groups gives invalid_scope


andrejvanderzee:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

v2.33.0

### Storage Type

Kubernetes

### Installation Type

Official container image

### Expected Behavior

When specifying `scopes: ["email", "profile", "groups"]`  I expect groups in the log output. 

### Actual Behavior

Unfortunately I get the following error in my browser:

```
Sign in with Google

Authorization Error
Error 400: invalid_scope
Some requested scopes were invalid. {valid=[openid, https://www.googleapis.com/auth/userinfo.email, https://www.googleapis.com/auth/userinfo.profile], invalid=[groups]}

[Learn more]
```

### Additional Information

The problem is that we have to add `groups` to scope otherwise `s.Groups` will be false and groups are never fetched:
https://github.com/dexidp/dex/blob/v2.33.0/connector/google/google.go#L221

When I patch google.go and set `s.Groups = true` we get the groups, as expected.

### Logs

```shell
Without the patch:

{
  "level": "info",
  "msg": "login successful: connector \"google\", username=\"Test User\", preferred_username=\"\", email=\"test@***.io\", groups=[]",
  "time": "2022-09-07T06:59:07Z"
}
```

With the patch:
```{
  "level": "info",
  "msg": "login successful: connector \"google\", username=\"Test User\", preferred_username=\"\", email=\"test@***.io\", groups=[\"test-group@***.io\"]",
  "time": "2022-09-07T06:59:07Z"
}
```
```

**Comment by andrejvanderzee** [comment]
Well problem seems that Google signin does not accept scope `groups` anymore. This (in my opinion acceptable) patch makes it work for us (so if one specifies `serviceAccountFilePath` and `adminEmail`, then groups are fetched):

```
$ git diff connector/google/google.go
diff --git a/connector/google/google.go b/connector/google/google.go
index 1b515b8e..b6583ec8 100644
--- a/connector/google/google.go
+++ b/connector/google/google.go
[...]

PR #2917, opened 2023-04-26
fix: remove redundant scope check from Google connector

devholic:
#### Overview

Remove redundant scope check from Google connector to fetching groups available.

Since `groups` are not considered a valid scope by Google, we cannot pass to the `scope`. Although @damieva introduced `claimMapping` in https://github.com/dexidp/dex/issues/2653 to address this issue, it does not seem to work with the current implementation of the Google connector.

This commit resolves the issue by removing the unnecessary `group` scope check, as `adminSrv` is only initialized when the user wants to retrieve groups.

None

```release-note
NONE
```

### DIFF
--- a/connector/google/google.go
+++ b/connector/google/google.go
@@ -226,7 +226,7 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
 	}
 
 	var groups []string
-	if s.Groups && c.adminSrv != nil {
+	if c.adminSrv != nil {
 		checkedGroups := make(map[string]struct{})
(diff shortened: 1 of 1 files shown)


devholic:
@sagikazarmark 

No problem, Thanks! :)

devholic:
Hey @sagikazarmark, I was wondering if I could get an update for this PR (No rushes 🙂)

Deams51:
Any chance this PR will make it in? 😄 
Latest is unfortunately still broken, thus forcing us to rely on generating a customer docker container.

PR #2928, opened 2023-05-05
fix: support workload identity federation in google connector

torfjor:
#### Overview

Service account keys with type `external_account` are not supported by `google.JWTConfigFromJSON`. While here, support another usecase where a service account has been granted group admin directly, bypassing the need for domain wide delegation and impersonating a user.

Trying to pass [external account credentials](https://google.aip.dev/auth/4117) fails because `google.JWTConfigFromJSON` expects service account keys. External account credentials are used for Anthos Fleet/Fleet Identity or when authenticating workloads external to GCP.

None. Should be compatible.


torfjor:
Upon further testing I discovered that this broke the usecase where you impersonate a user account with permissions to read groups. Converting to draft while investigating.

torfjor:
Should be fixed as of latest commit.

Tested successfully with all the listed usecases in the comment.

nabokihms:
@torfjor Hello, I've seen this. Google connector fixes are always a bit problematic for us to merge because we do not have integration tests for this connector. The review of the PR may take some time. Sorry for the inconvenience.

PR #2937, opened 2023-05-10
Fail if OIDC config contains hosted domains

nabokihms:
#### Overview

Fail if the hostedDomains option is specified in an OIDC connector config

The option was removed. After the upgrade, users still using OIDC connectors for their Google providers may accidentally open access to more people than before.

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -35,6 +35,13 @@ type Config struct {
 
 	Scopes []string `json:"scopes"` // defaults to "profile" and "email"
 
+	// Previously it was an optional list of whitelisted domains when using Google.
+	// Only users from a listed domain will be allowed to log in.
+	// Now this option does nothing. To work with Google, users should migrate to using the Google connector.
+	//
+	// Deprecated: will be removed in feature releases.
+	HostedDomains []string `json:"hostedDomains"`
+
 	// Certificates for SSL validation
 	RootCAs []string `json:"rootCAs"`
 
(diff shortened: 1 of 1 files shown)


connector/oidc/oidc.go
```diff
@@ -35,6 +35,13 @@ type Config struct {
 
 	Scopes []string `json:"scopes"` // defaults to "profile" and "email"
 
+	// Previously it was an optional list of whitelisted domains when using Google.
+	// Only users from a listed domain will be allowed to log in.
+	// Now this option does nothing. To work with Google, users should migrate to using the Google connector.
+	//
+	// Deprecated: will be removed in feature releases.
```
sagikazarmark, line 42: Nit: s/feature releases/a future release/ ?

connector/oidc/oidc.go
```diff
@@ -112,6 +119,10 @@ func knownBrokenAuthHeaderProvider(issuerURL string) bool {
 // Open returns a connector which can be used to login users through an upstream
 // OpenID Connect provider.
 func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, err error) {
+	if len(c.HostedDomains) > 0 {
+		return nil, fmt.Errorf("hostedDomains option does not work anymore, consider switching to the Google connector")
```
sagikazarmark, line 123:
```suggestion
		return nil, fmt.Errorf("Hosted domains option had been deprecated and removed. Consider switching to the Google connector")
```

sagikazarmark (changes requested):
Retargeted to 2.37.0

connector/oidc/oidc.go
```diff
@@ -35,6 +35,13 @@ type Config struct {
 
 	Scopes []string `json:"scopes"` // defaults to "profile" and "email"
 
+	// Previously it was an optional list of whitelisted domains when using Google.
+	// Only users from a listed domain will be allowed to log in.
+	// Now this option does nothing. To work with Google, users should migrate to using the Google connector.
```
dlipovetsky, line 40:
```suggestion
	// HostedDomains was an optional list of whitelisted domains when using the OIDC connector with Google. Only users from a whitelisted domain were allowed to log in.
	// Support for this option was removed from the OIDC connector in f07a58a7f1514eaf6e634bb773584cefe98e84ed.
	// The Google connector supports this option.
```

connector/oidc/oidc.go
dlipovetsky, line 123:
```suggestion
		return nil, fmt.Errorf("Support for the HostedDomains option was removed from the OIDC connector. Consider using the Google connector, which supports this option.")
```

nabokihms:
Because of the overlapping review comments, I took parts from both suggestions. Thanks @dlipovetsky @sagikazarmark

Issue #2876, opened 2023-03-27
Extend payload


loafoe:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

Dex has very rigid `claims` support. Adding additional claims via a connector requires forking the code and making changes to various internal structures. There are a number of proposals where this discussed, namely https://github.com/dexidp/dex/issues/1635 . The scope of these is fairly large and we are exploring a possible solution that limits itself to just mutating of the claims.

### Proposed Solution

We propose adding a new `PayloadExtender` interface which connectors can choose to implement:

```golang
// PayloadExtender allows connectors to enhance the payload before signing
type PayloadExtender interface {
	ExtendPayload(scopes []string, payload []byte, connectorData []byte) ([]byte, error)
}
```

By implementing this interface connectors get a chance to mutate the `id_token` payload just before it is signed. The `scopes` may be used to perform conditional mutation. The `payload` is passed as a byte array as well as the `connectorData` associated with the authorization request. This allows the connector to pass context via `connectorData`. The resulting mutated structure is returned including an error condition.

An working example of this interface implementation can look like this:

```golang
func (c *hsdpConnector) ExtendPayload(scopes []string, payload []byte, cdata []byte) ([]byte, error) {
	var cd connectorData
	var originalClaims map[string]interface{}

	c.logger.Info("ExtendPayload called")

	if err := json.Unmarshal(cdata, &cd); err != nil {
		return payload, err
	}
	if err := json.Unmarshal(payload, &originalClaims); err != nil {
		return payload, err
	}
	
	// Experimental teams
	var teams []string
	teams = append(teams, cd.Introspect.Organizations.ManagingOrganization)
	originalClaims["teams"] = teams

	extendedPayload, err := json.Marshal(originalClaims)
	if err != nil {
		return payload, err
	}
	return extendedPayload, nil
}
```

This results in a `teams` claim being available in the `id_token`. A working implementation can be found in this [fork](https://github.com/philips-forks/dex). 

### Alternatives Considered

Alternatives are to maintain a fork with these changes.

### Additional Information

[...]

PR #2954, opened 2023-05-17
DEP-2876

loafoe:
#### Overview

DEP-2876 

Based on #2876 we are capturing the proposal details.


pagalba-com:
+1, highly needed functionality in specific cases

PR #2963, opened 2023-05-19
DEP Implementation: ExtendPayload interface support #2876

loafoe:
#### Overview

Implementation of #2876 

Implements DEP-2876

```release-note

```

### DIFF
--- a/connector/connector.go
+++ b/connector/connector.go
@@ -136,3 +136,7 @@ type LogoutCallbackConnector interface {
 	// return nil.
 	HandleLogoutCallback(ctx context.Context, r *http.Request) error
 }
+
+type PayloadExtender interface {
+	ExtendPayload(scopes []string, payload []byte, connectorData []byte) ([]byte, error)
+}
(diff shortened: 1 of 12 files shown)


kromanow94:
👍

connector/connector.go
```diff
@@ -103,3 +103,7 @@ type RefreshConnector interface {
 type TokenIdentityConnector interface {
 	TokenIdentity(ctx context.Context, subjectTokenType, subjectToken string) (Identity, error)
 }
+
+type PayloadExtender interface {
+	ExtendPayload(scopes []string, payload []byte, connectorData []byte) ([]byte, error)
```
maelvls, line 140: Would be great to pass the claims as well as context.Context here, WDYT?

E.g.:

```go

// PayloadExtender allows connectors to enhance the payload before signing. See:
// https://github.com/dexidp/dex/issues/2876.
type PayloadExtender interface {
	ExtendPayload(scopes []string, claims storage.Claims, payload []byte, connectorData []byte) ([]byte, error)
}
```

PR #2972, opened 2023-05-26
Add GetClient grpc service

marcofranssen:

This PR adds a GetClient grpc endpoint which I need to implement a Terraform provider
https://github.com/marcofranssen/terraform-provider-dexidp/pull/7

```release-note

```


server/api.go
```diff
@@ -45,6 +46,30 @@ type dexAPI struct {
...
 
+func (d dexAPI) GetClient(ctx context.Context, req *api.GetClientReq) (*api.GetClientResp, error) {
+	sanitizedID := strings.Trim(req.Id, " \t\n")
+	if sanitizedID == "" {
+		return nil, errors.New("get client: no client ID supplied")
+	}
+
+	c, err := d.s.GetClient(sanitizedID)
+	if err != nil {
+		return nil, errors.New("")
```
sagikazarmark, line 57: This error looks incorrect.

server/api.go
```diff
@@ -45,6 +46,30 @@ type dexAPI struct {
 	version string
 }
 
+func (d dexAPI) GetClient(ctx context.Context, req *api.GetClientReq) (*api.GetClientResp, error) {
+	sanitizedID := strings.Trim(req.Id, " \t\n")
+	if sanitizedID == "" {
+		return nil, errors.New("get client: no client ID supplied")
+	}
```
sagikazarmark, line 53: I wonder why sanitization needs to happen here.

sagikazarmark (changes requested):

server/api.go
marcofranssen: in theory, someone could send a string like "  my-id ", this way we keep stuff clean.

marcofranssen: Ping? What do you want me to do? Can we merge this one? Do you want me to change something?

nabokihms: Hello, @marcofranssen. Let's go without sanitizing ids for now. If a user sends an id, it must be the exact id without trailing spaces.

marcofranssen: ok, that is exactly what the sanitizing does. It removes all trailing and leading whitespacing.

nabokihms (approved):

nabokihms (approved):
Thanks, @marcofranssen!

marcofranssen:
When can we expect a release with this feature? So I can finish this implementation

https://github.com/marcofranssen/terraform-provider-dexidp/pull/7

nabokihms:
October something I believe.

marcofranssen:
Its december now. Any update on when a release is planned?

marcofranssen:
Is it possible to also create a release using the `api/v2.x` tagging so the change also shows up in https://pkg.go.dev/github.com/dexidp/dex/api/v2?

PR #2980, opened 2023-06-01
Fix lint and fix Makefile tasks

marcofranssen:
#### Overview

Fix the lint and fix tasks in Makefile
    
This uses native Make features to automatically install the tool if not already there and then use the `$<` to take the first dependency for invoking the tool
    
On my Macbook it is unable to resolve the PATH export accordingly
    
Once the file exists the dependency will ofcourse not execute

Fixes and makes it easier for contributors running these tasks. 

```release-note

```


sagikazarmark:
Closing as it's not the preferred way to install tool dependencies.

PR #3031, opened 2023-07-04
verify access tokens by checking getuserinfo during a token exchange

seankhliao:
#### Overview

Followup on #2806 for better access token handling.
verify access tokens by checking getuserinfo during a token exchange

The provider.Verifier.Verify endpoint we were using only works with ID tokens. This isn't an issue with systems which use ID tokens as access tokens (e.g. dex), but for systems with opaque access tokens (e.g. Google / GCP), those access tokens could not be verified. Instead, check the access token against the getUserInfo endpoint.

```release-note
NONE
```

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -301,6 +301,7 @@ func (c *oidcConnector) TokenIdentity(ctx context.Context, subjectTokenType, sub
 	var identity connector.Identity
 	token := &oauth2.Token{
 		AccessToken: subjectToken,
+		TokenType:   subjectTokenType,
 	}
 	return c.createIdentity(ctx, identity, token, exchangeCaller)
 }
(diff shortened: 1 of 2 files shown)


connector/oidc/oidc.go
```diff
@@ -318,21 +320,29 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
 	} else if caller != refreshCaller {
 		// ID tokens aren't mandatory in the reply when using a refresh_token grant
 		return identity, errors.New("oidc: no id_token in token response")
 	}
 
-	// We immediately want to run getUserInfo if configured before we validate the claims
-	if c.getUserInfo {
+	// We immediately want to run getUserInfo if configured before we validate the claims.
+	// For token exchanges with access tokens, this is how we verify the token.
+	if c.getUserInfo || checkAccessToken {
```
nabokihms, line 345:
```suggestion
	if c.getUserInfo {
```
The provider may not have the user info endpoint. Not all providers are entirely OIDC compatible. For these providers, `c.getUserInfo` is a kind of protection.

connector/oidc/oidc.go
```diff
@@ -318,21 +320,29 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+			// Verify only works on ID tokens
+			idToken, err := c.provider.Verifier(&oidc.Config{SkipClientIDCheck: true}).Verify(ctx, token.AccessToken)
+			if err != nil {
+				return identity, fmt.Errorf("oidc: failed to verify token: %v", err)
+			}
+			if err := idToken.Claims(&claims); err != nil {
+				return identity, fmt.Errorf("oidc: failed to decode claims: %v", err)
+			}
+		case "urn:ietf:params:oauth:token-type:access_token":
+			checkAccessToken = true
```
nabokihms, line 334:
```suggestion
```

connector/oidc/oidc.go
```diff
@@ -301,12 +301,14 @@ func (c *oidcConnector) TokenIdentity(ctx context.Context, subjectTokenType, sub
...
 	token := &oauth2.Token{
 		AccessToken: subjectToken,
+		TokenType:   subjectTokenType,
 	}
 	return c.createIdentity(ctx, identity, token, exchangeCaller)
 }
 
 func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.Identity, token *oauth2.Token, caller caller) (connector.Identity, error) {
 	var claims map[string]interface{}
+	var checkAccessToken bool
```
nabokihms, line 311:
```suggestion
```

nabokihms (changes requested):
There is a question to discuss (the small one), and I also wonder wether a unit test can cover this improvement 🤔

seankhliao:
Not sure which question you were referring to?
I've switched to a check ensuring getUserInfo is enabled.
I can test that getUserInfo needs to be set, not sure we can easily test actual opaque tokens.

nabokihms:
@seankhliao it was the question about what you have already fixed. Now everything seems fine.

nabokihms (approved):
LGTM

connector/oidc/oidc.go
```diff
@@ -318,20 +319,30 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+			if err != nil {
+				return identity, fmt.Errorf("oidc: failed to verify token: %v", err)
+			}
+			if err := idToken.Claims(&claims); err != nil {
+				return identity, fmt.Errorf("oidc: failed to decode claims: %v", err)
+			}
+		case "urn:ietf:params:oauth:token-type:access_token":
+			if !c.getUserInfo {
+				return identity, fmt.Errorf("oidc: getUserInfo is required for access token exchange")
+			}
```
Moulick, line 335: Maybe I am mistaking something but this breaks using GitHub Actions (https://token.actions.githubusercontent.com) as an OIDC connector. https://token.actions.githubusercontent.com does not have a userInfo endpoint (https://token.actions.githubusercontent.com/.well-known/openid-configuration). If I set `getUserInfo: false` then this check fails but if `true` then there is no URL to actually do the check!

PR #3056, opened 2023-07-27
Composite claims in OIDC connector

Oded-B:
#### Overview
Adds the ability to compose new group claims elements in the OIDC connector,  based on other claims.
So this OIDC connector config:
``` 
     claimConcatenations:
        - claimList:
           - "organization_slug"
           - "pipeline_slug"
           - "build_branch"
          delimiter: "::"
          prefix: "bk"
```
Would result in this element in the group claim which looks like this:: `bk::acme-inc::super-duper-app::main`

After #2806  we are able to authenticate with JWT generated by CI systems.
The claim provided by those systems are usually static and if the upstream OIDC service doesn't support sophisticated query language in its IAM/RBAC mechanism the operator might be unable to find a configuration that provided the needed granularity.

For example, lets say you want a specific Buildkiite pipeline to run a sync action on an ArgoCD instance. 
These are claim provided by Buildkite CI/CD system:
```
{
  "iss": "https://agent.buildkite.com",
  "sub": "organization:acme-inc:pipeline:super-duper-app:ref:refs/heads/main:commit:9f3182061f1e2cca4702c368cbc039b7dc9d4485:step:build",
  "aud": "https://buildkite.com/acme-inc",
  "iat": 1669014898,
  "nbf": 1669014898,
  "exp": 1669015198,
  "organization_slug": "acme-inc",
  "pipeline_slug": "super-duper-app",
  "build_number": 1,
  "build_branch": "main",
  "build_commit": "9f3182061f1e2cca4702c368cbc039b7dc9d4485",
  "step_key": "build",
  "job_id": "0184990a-477b-4fa8-9968-496074483cee",
  "agent_id": "0184990a-4782-42b5-afc1-16715b10b8ff"
}
```
AFAIKT ArgoCD doesn't have any fancy query language in its RBAC config like big cloud providers IAM (regexs are supported to action field only)  
So If you want to allow pipelines runs from a specific repo and branch combination, you would need to combine  `organization_slug`, `pipeline_slug` and  `build_branch` claims ( notice the `sub` claim includes the git commit, making it unusable as reference for RBAC ).

With this PR you'll be able to concatenate/compose multiple claims to a new element in the group claim:

```      connectors:
        - name: Buildkite Agents OIDC
          type: oidc
          id: bk-oidc
          config:
            issuer: https://agent.buildkite.com
            scopes:
              - groups
              - openid
            userNameKey: sub
            claimConcatenations:
              - claimList:
                  - "organization_slug"
                  - "pipeline_slug"
                  - "build_branch"
                delimiter: "::"
                prefix: "bk"
```
This would be the resulting group : `bk::acme-inc::super-duper-app::main` 
And you can refrence it in ArgoCD RBAC like so:
```
      g, bk::acme-inc::super-duper-app::main, role:cicd-sync
```

This is intended as a temporary solution until a more flexible solution is provided via  #1635.
Because the problem is uniq to service-tied JWTs  and only relevant for OIDC I believe a connecter-specific solution is acceptable.

```release-note
Enhancements:
* Support composing new group claims in OIDC connector.

```

### DIFF
--- a/connector/oidc/oidc_test.go
+++ b/connector/oidc/oidc_test.go
@@ -62,7 +62,7 @@ func TestHandleCallback(t *testing.T) {
 		expectPreferredUsername   string
 		expectedEmailField        string
 		token                     map[string]interface{}
-		newGroupFromClaims       []NewGroupFromClaims
+		newGroupFromClaims        []NewGroupFromClaims
 	}{
 		{
 			name:               "simpleCase",


connector/oidc/oidc_test.go
```diff
@@ -288,6 +289,66 @@ func TestHandleCallback(t *testing.T) {
...
 		},
+		{
+			name:               "concatinateClaim",
+			userIDKey:          "", // not configured
+			userNameKey:        "", // not configured
+			expectUserID:       "subvalue",
+			expectUserName:     "namevalue",
+			expectGroups:       []string{"group1", "gh::acme::pipeline-one", "tfe-acme-foobar", "bk-emailvalue"},
+			expectedEmailField: "emailvalue",
+			claimConcatenations: []ClaimConcatenation{
```
sagikazarmark, line 300: Can you please add a comment above each case which behavior it tests so that they don't get accidentally refactored?

sagikazarmark (changes requested):
@Oded-B Thanks for the contribution.


I expect's going to be some time before we get #1635 implemented, so I suggest we take a more structured approach, anticipating that other similar features may be added to connectors.

First of all, I find it a bit misleading that groups are not mentioned anywhere even though the concatenated claims end up as groups.

Secondly, I think it would be nice to add at least one level of config above it so that we can collect similar future additions.

A couple options I can imagine:

```yaml
# ...

claimModifications:
    groups:
        newFromClaims:
```

```yaml
# ...

claimModifications:
    newGroupsFromClaims:
```

I'm not quite sure which one I like better yet. (@nabokihms WDYT?)

Oded-B:
I used the 2nd option (`claimModifications.newGroupsFromClaims`)  for now but I'll change it per @nabokihms opinion.
Also added comments for each test case.

connector/oidc/oidc.go
```diff
@@ -87,6 +87,23 @@ type Config struct {
 		// Configurable key which contains the groups claims
 		GroupsKey string `json:"groups"` // defaults to "groups"
 	} `json:"claimMapping"`
+
+	// ClaimModifications holds all claim modifications options, current has only newGroupsFromClaims
+	ClaimModifications struct {
+		NewGroupsFromClaims []NewGroupsFromClaims `json:"newGroupsFromClaims"`
+	}
```
sagikazarmark, line 94: Might need a json struct tag here as well.

connector/oidc/oidc.go
```diff
@@ -427,6 +446,26 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+			// Non string claim value are ignored, concatenating them doesn't really make any sense
+			if claimValue, ok := claims[clm].(string); ok {
+				// Removing the delimiier string from the concatenated claim to ensure resulting claim structure
+				// is in full control of Dex operator
+				claimCleanedValue := strings.ReplaceAll(claimValue, cc.Delimiter, "")
+				if newElement == "" {
+					newElement = cc.Prefix + cc.Delimiter + claimCleanedValue
+				} else {
+					newElement = newElement + cc.Delimiter + claimCleanedValue
+				}
```
sagikazarmark, line 461: How about collecting segments into a string slice and calling strings.Join after the loop?

connector/oidc/oidc.go
```diff
@@ -427,6 +446,26 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 		}
 	}
 
+	for _, cc := range c.newGroupsFromClaims {
+		newElement := ""
+		for _, clm := range cc.ClaimList {
```
sagikazarmark, line 451: Can we use some more expressive variables than cc, newElement and clm?

connector/oidc/oidc.go
```diff
@@ -427,6 +446,26 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+			// Non string claim value are ignored, concatenating them doesn't really make any sense
+			if claimValue, ok := claims[clm].(string); ok {
+				// Removing the delimiier string from the concatenated claim to ensure resulting claim structure
+				// is in full control of Dex operator
+				claimCleanedValue := strings.ReplaceAll(claimValue, cc.Delimiter, "")
```
sagikazarmark, line 456: I wonder if this behavior should be controlled by another option (escapeDelimiter/cleanDelimiter)

Oded-B: How about `escapeDelimiterCharacter` config key and if its not set the delimiter character will be removed?

sagikazarmark: Hm, how would that work?

I mean, I doubt people want to _add_ anything like an escape character to their group names.

We can always add that functionality later should anyone need it. Personally, I don't see a use case for it at the moment.

Oded-B: So just keep the current behaviour of removing any occurrences of the delimiter string ? or did misunderstood?

sagikazarmark: `clearDelimiter: true` - remove delimiter as it is right now
`clearDelimiter: false (default)` - leave any delimiter untouched

connector/oidc/oidc.go
```diff
@@ -427,6 +450,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 		}
 	}
 
+	for _, newGroupsElementConfig := range c.newGroupsFromClaims {
```
sagikazarmark, line 453:
```suggestion
	for _, config := range c.newGroupsFromClaims {
```

connector/oidc/oidc.go
```diff
@@ -427,6 +450,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+		newGroupsElement := []string{
```
sagikazarmark, line 454:
```suggestion
		newGroupSegments := []string{
```

connector/oidc/oidc.go
```diff
@@ -427,6 +450,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+			newGroupsElementConfig.Prefix,
+		}
+		for _, claimName := range newGroupsElementConfig.ClaimList {
+			// Non string claim value are ignored, concatenating them doesn't really make any sense
+			if claimValue, ok := claims[claimName].(string); ok {
```
sagikazarmark, line 459: Arrow-pattern is a code smell. You could rewrite this as 

```go
if !ok {
    continue
}
```

connector/oidc/oidc.go
```diff
@@ -427,6 +450,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+				if newGroupsElementConfig.ClearDelimiter {
+					// Removing the delimiier string from the concatenated claim to ensure resulting claim structure
+					// is in full control of Dex operator
+					claimCleanedValue := strings.ReplaceAll(claimValue, newGroupsElementConfig.Delimiter, "")
+					newGroupsElement = append(newGroupsElement, claimCleanedValue)
+				} else {
+					newGroupsElement = append(newGroupsElement, claimValue)
+				}
```
sagikazarmark, line 467: Drop the else and set claimValue directly in the if

connector/oidc/oidc.go
```diff
@@ -427,6 +450,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+			}
+		}
```
sagikazarmark, line 469:
```suggestion
		}

```

connector/oidc/oidc_test.go
```diff
@@ -288,6 +289,79 @@ func TestHandleCallback(t *testing.T) {
 				"email_verified": true,
 			},
 		},
+		{
+			name:               "concatinateClaim",
```
sagikazarmark, line 293:
```suggestion
			name:               "newGroupFromClaims",
```

connector/oidc/oidc.go
```diff
@@ -87,6 +87,27 @@ type Config struct {
 		// Configurable key which contains the groups claims
 		GroupsKey string `json:"groups"` // defaults to "groups"
 	} `json:"claimMapping"`
+
+	// ClaimModifications holds all claim modifications options, current has only newGroupsFromClaims
```
sagikazarmark, line 91:
```suggestion
	// ClaimModifications holds all claim modifications options
```

sagikazarmark (changes requested):

connector/oidc/oidc.go
```diff
@@ -87,6 +87,27 @@ type Config struct {
...
+	// ClaimModifications holds all claim modifications options
+	ClaimModifications struct {
+		NewGroupsFromClaims []NewGroupsFromClaims `json:"newGroupsFromClaims"`
+	} `json:"claimModifications"`
+}
+
+// List of groups claim elements to create by concatenating other claims
+type NewGroupsFromClaims struct {
+	// List of claim to join together
+	ClaimList []string `json:"claimList"`
```
sagikazarmark, line 100:
```suggestion
	Claims []string `json:"claims"`
```

Maybe just call this "claims"?

connector/oidc/oidc.go
```diff
@@ -87,6 +87,27 @@ type Config struct {
 		// Configurable key which contains the groups claims
 		GroupsKey string `json:"groups"` // defaults to "groups"
 	} `json:"claimMapping"`
+
+	// ClaimModifications holds all claim modifications options
+	ClaimModifications struct {
+		NewGroupsFromClaims []NewGroupsFromClaims `json:"newGroupsFromClaims"`
```
sagikazarmark, line 93: I'm thinking `NewGroupFromClaims` (singular) would be better. Each instance would add one group.

connector/oidc/oidc.go
```diff
@@ -87,6 +87,27 @@ type Config struct {
...
+	} `json:"claimModifications"`
+}
+
+// List of groups claim elements to create by concatenating other claims
```
sagikazarmark, line 97:
```suggestion
// NewGroupFromClaims creates a new group from a list of claims and appends it to the list of existing groups.
```

connector/oidc/oidc.go
```diff
@@ -87,6 +87,27 @@ type Config struct {
 		// Configurable key which contains the groups claims
 		GroupsKey string `json:"groups"` // defaults to "groups"
 	} `json:"claimMapping"`
+
+	// ClaimModifications holds all claim modifications options
+	ClaimModifications struct {
```
sagikazarmark, line 92: Let's call this `ClaimMutations`

connector/oidc/oidc.go
```diff
@@ -427,6 +450,29 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
 	}
 
+	for _, config := range c.newGroupsFromClaims {
+		newGroupSegments := []string{
+			config.Prefix,
+		}
+		for _, claimName := range config.ClaimList {
+			claimValue, ok := claims[claimName].(string)
+			// Non string claim value are ignored, concatenating them doesn't really make any sense
+			if !ok {
```
sagikazarmark, line 460:
```suggestion
			if !ok { // Non string claim value are ignored, concatenating them doesn't really make any sense
```

connector/oidc/oidc.go
```diff
@@ -427,6 +450,29 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+				continue
+			}
```
sagikazarmark, line 462:
```suggestion
			}

```

connector/oidc/oidc.go
```diff
@@ -427,6 +450,29 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+			if config.ClearDelimiter {
+				// Removing the delimiter string from the concatenated claim to ensure resulting claim structure
+				// is in full control of Dex operator
+				claimValue = strings.ReplaceAll(claimValue, config.Delimiter, "")
+			}
```
sagikazarmark, line 467:
```suggestion
			}

```

sagikazarmark (changes requested):

sagikazarmark (approved):

sagikazarmark:
LGTM, thank you!

Once the build turns green, good to go.

PR #3063, opened 2023-08-04
add regex for oidc group matching

jacksonargo:
#### Overview

Add a group regex matcher to the OIDC connector config.

Sometimes the IDP's can send very long list of groups, not all IDP's support group filters from the IDP side.

I tested this in the unit tests, ~but not sure how to test this in a live environment.~ and in a live environment, and the filter works as expected.

It adds a new config field for OIDC: `groupsRegex`.

```release-note
ADD groups regex config for oidc connector.
```

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


sagikazarmark:
@jacksonargo thanks for the contribution!

Two thoughts came to my mind:

- People generally want to deny authentication with filters like this. I'm not saying it doesn't make sense, but it can be confusing for some, so we may want to structure config/document this feature to avoid any confusion.
- Would it make sense to support capture groups not just to match, but extract parts of group names?

jacksonargo:
> People generally want to deny authentication with filters like this. I'm not saying it doesn't make sense, but it can be confusing for some, so we may want to structure config/document this feature to avoid any confusion.

Ya, makes sense to me. Would `groupsFilter` be a better name?

> Would it make sense to support capture groups not just to match, but extract parts of group names?

Yes probably! Didn't occur to me at the time.

st-sidjoshi:
Your thoughts @sagikazarmark ?

jacksonargo:
Hi @sagikazarmark, I've updated the pr to put the regex filter under the claim mutations config.

How does this look?

sagikazarmark (approved):

PR #3074, opened 2023-08-09
Minimalistic support for group filtering in oidc connector

mdpradeep:
#### Overview

Filter users based on groups in OIDC connector.

Currently it is not possible to filter users based on groups like it is done with the `google` connector. This PR is an attempt to support group filtering in the `oidc` connector. The connector configuration takes an additional configuration as below to specify what the allowed groups are. If the user's groups belong to at least one of the `allowedGroups`, then authentication will be successful. The other change is to indicate that the user does not belong to allowed groups if authentication fails because the user does not belong to any of the allowed groups.

```
{
            "name": "My-oidc",
            "config": {
                "issuer": "<ISSUER>",
                "clientID": "12345678",
                "redirectURI": "<Dex Callback>",
                "clientSecret": "12345678",
                "insecureSkipEmailVerified": true,
                "insecureEnableGroups": true,
                "getUserInfo": true,
                "allowedGroups": ["XYZ", "ABC"],
                "scopes": ....
            },
            "id": "myoidc",
            "type": "oidc"
        }
```

```release-note
NONE
```

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -12,12 +12,12 @@ import (
 	"time"
 
 	"github.com/coreos/go-oidc/v3/oidc"
-	"golang.org/x/exp/slices"
 	"golang.org/x/oauth2"
 
 	"github.com/dexidp/dex/connector"
 	"github.com/dexidp/dex/pkg/httpclient"
 	"github.com/dexidp/dex/pkg/log"
+	groups_pkg "github.com/dexidp/dex/pkg/groups"
 )
 
 // Config holds configuration options for OpenID Connect logins.
@@ -431,22 +431,15 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 		}
 
 		// Validate that the user is part of allowedGroups
-		accessAllowed := false
 		if len(c.allowedGroups) > 0 {
-			for _, group := range c.allowedGroups {
-				groupPresent := slices.Contains(groups, group)
-				if groupPresent {
-					accessAllowed = true
-					break
-				}
-			}
-		} else {
-			// don't check for groups if allowedGroups is not configured
-			accessAllowed = true
-		}
+			groupMatches := groups_pkg.Filter(groups, c.allowedGroups)
 
-		if !accessAllowed {
-			return identity, errors.New("user is not in the allowed group(s)")
+			if len(groupMatches) == 0 {
+				// No group membership matches found, disallowing
+				return identity, fmt.Errorf("user not a member of allowed groups")
+			}
+
+			groups = groupMatches
 		}
 	}
 
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -15,9 +15,9 @@ import (
 	"golang.org/x/oauth2"
 
 	"github.com/dexidp/dex/connector"
+	groups_pkg "github.com/dexidp/dex/pkg/groups"
 	"github.com/dexidp/dex/pkg/httpclient"
 	"github.com/dexidp/dex/pkg/log"
-	groups_pkg "github.com/dexidp/dex/pkg/groups"
 )
 
 // Config holds configuration options for OpenID Connect logins.
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/coreos/go-oidc/v3/oidc"
+	"golang.org/x/exp/slices"
 	"golang.org/x/oauth2"
 
 	"github.com/dexidp/dex/connector"
@@ -50,7 +51,8 @@ type Config struct {
 	InsecureSkipEmailVerified bool `json:"insecureSkipEmailVerified"`
 
 	// InsecureEnableGroups enables groups claims. This is disabled by default until https://github.com/dexidp/dex/issues/1065 is resolved
-	InsecureEnableGroups bool `json:"insecureEnableGroups"`
+	InsecureEnableGroups bool     `json:"insecureEnableGroups"`
+	AllowedGroups        []string `json:"allowedGroups"`
 
 	// AcrValues (Authentication Context Class Reference Values) that specifies the Authentication Context Class Values
 	// within the Authentication Request that the Authorization Server is being requested to use for
@@ -180,6 +182,7 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 		httpClient:                httpClient,
 		insecureSkipEmailVerified: c.InsecureSkipEmailVerified,
 		insecureEnableGroups:      c.InsecureEnableGroups,
+		allowedGroups:             c.AllowedGroups,
 		acrValues:                 c.AcrValues,
 		getUserInfo:               c.GetUserInfo,
 		promptType:                c.PromptType,
@@ -207,6 +210,7 @@ type oidcConnector struct {
 	httpClient                *http.Client
 	insecureSkipEmailVerified bool
 	insecureEnableGroups      bool
+	allowedGroups             []string
 	acrValues                 []string
 	getUserInfo               bool
 	promptType                string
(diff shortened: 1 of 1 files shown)


connector/oidc/oidc.go
```diff
@@ -341,6 +345,7 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 		return identity, errors.New("oidc: no id_token in token response")
 	}
 
+	c.logger.Debugf("Claims::: %q \n", claims)
```
sagikazarmark, line 348: I assume this stayed here by accident?

connector/oidc/oidc.go
```diff
@@ -351,6 +356,7 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 		if err := userInfo.Claims(&claims); err != nil {
 			return identity, fmt.Errorf("oidc: failed to decode userinfo claims: %v", err)
 		}
+		c.logger.Debugf("UserInfo::: %q \n", userInfo)
```
sagikazarmark, line 359: I assume this stayed here by accident?

connector/oidc/oidc.go
```diff
@@ -51,6 +52,7 @@ type Config struct {
 
 	// InsecureEnableGroups enables groups claims. This is disabled by default until https://github.com/dexidp/dex/issues/1065 is resolved
 	InsecureEnableGroups bool `json:"insecureEnableGroups"`
+	AllowedGroups []string `json:"allowedGroups"` // allowedGroups
```
sagikazarmark, line 55:
```suggestion
	AllowedGroups []string `json:"allowedGroups"`
```

connector/oidc/oidc.go
```diff
@@ -425,6 +431,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 				}
 			}
 		}
+		c.logger.Debugf("Groups::: %q \n", groups)
```
sagikazarmark, line 434: I assume this stayed here by accident?

connector/oidc/oidc.go
```diff
@@ -425,6 +431,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+
+		// Validate that the user is part of allowedGroups
+		var accessAllowed = false
+		if len(c.allowedGroups) > 0 {
+			for _,group := range c.allowedGroups {
+				groupPresent := slices.Contains(groups, group)
+				if groupPresent {
+					accessAllowed = true
+					c.logger.Debugf("Found a match - Group::: %q \n", group)
```
sagikazarmark, line 443: We prefer structured logging over custom message formats

connector/oidc/oidc.go
```diff
@@ -425,6 +431,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+					break
+				}
+			}
+		} else {
+			// don't check for groups if allowedGroups is not configured
+			c.logger.Debugf("Group check not performed \n")
```
sagikazarmark, line 449: this looks unnecessary

connector/oidc/oidc.go
```diff
@@ -425,6 +431,28 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
 			}
 		}
+		c.logger.Debugf("Groups::: %q \n", groups)
+
+		// Validate that the user is part of allowedGroups
+		var accessAllowed = false
+		if len(c.allowedGroups) > 0 {
+			for _,group := range c.allowedGroups {
+				groupPresent := slices.Contains(groups, group)
+				if groupPresent {
```
sagikazarmark, line 441:
```suggestion
				if slices.Contains(groups, group) {
```

server/handlers.go
```diff
@@ -467,7 +467,7 @@ func (s *Server) handleConnectorCallback(w http.ResponseWriter, r *http.Request)
 
 	if err != nil {
 		s.logger.Errorf("Failed to authenticate: %v", err)
-		s.renderError(r, w, http.StatusInternalServerError, fmt.Sprintf("Failed to authenticate: %v", err))
+		s.renderError(r, w, http.StatusUnauthorized, fmt.Sprintf("Failed to authenticate: %v", err))
```
sagikazarmark, line 470: This is not a good idea. This is where we catch unexpected errors as well. If there is a need to return 401 based on the error, the error needs to be checked for such conditions.

sagikazarmark (changes requested):

connector/oidc/oidc.go
mdpradeep: Actually this came as a need when we were recently debugging stuff on a production environment where we did not have access to the IdP. It became painful to prove that some of the claims that we were expecting were not being sent by the IdP. But if you think it is unnecessary I will remove it.

connector/oidc/oidc.go
mdpradeep: Similar to  the previous need, it was helpful to see to which groups the user belonged to. Again, I will remove it if you think it is not necessary.

connector/oidc/oidc.go
mdpradeep: This is redundant, will remove it.

connector/oidc/oidc.go
mdpradeep: Sure, could you point me to an example?

connector/oidc/oidc.go
mdpradeep: Will remove it.

server/handlers.go
mdpradeep: Completely agree. I was not intending to add this to  the PR actually. So, are you suggesting we create a custom Error  type with code + error message and then check for `code` to decide whether to return `StatusInternalServerError` or `StatusUnauthorized` or are you saying that we check for the error message string and then decide whether to return `StatusInternalServerError` or `StatusUnauthorized`?

nabokihms: Let's just revert this change. From the PR description, I don't think this is required.

connector/oidc/oidc.go
```diff
@@ -425,6 +431,27 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
...
+			}
+		} else {
+			// don't check for groups if allowedGroups is not configured
+			c.logger.Debugf("Group check not performed \n")
+			accessAllowed = true
+		}
+
+		if !accessAllowed {
+			return identity, errors.New("User is not in the allowed group(s)")
+		}
```
nabokihms, line 454:
```suggestion
		if len(c.allowedGroups) > 0 {
			groupMatches := groups.Filter(ident.Groups, p.allowedGroups)

			if len(groupMatches) == 0 {
				// No group membership matches found, disallowing
				return identity, fmt.Errorf("user not a member of allowed groups")
			}
			
			groups = groupMatches
		}
```

https://github.com/dexidp/dex/blob/master/pkg/groups/groups.go Should be used here to be inline with other connectors.

nabokihms (changes requested):
Debug messages don't seem like a necessary addition. The better approach is, for example, to trace HTTP requests to providers and check returned bodies. Custom debug messages are blocking this PR.

connector/oidc/oidc.go
mdpradeep: Removed this in 345eba905ddb2664855e79da2acf816ad6b97eb5

connector/oidc/oidc.go
mdpradeep: Removed this in 345eba905ddb2664855e79da2acf816ad6b97eb5

connector/oidc/oidc.go
mdpradeep: Removed in 345eba905ddb2664855e79da2acf816ad6b97eb5

connector/oidc/oidc.go
mdpradeep: Handled in 14f867af2fa25cd4a440c7779431223ab55796af

connector/oidc/oidc.go
mdpradeep: Handled in 14f867af2fa25cd4a440c7779431223ab55796af

server/handlers.go
mdpradeep: Done in 345eba905ddb2664855e79da2acf816ad6b97eb5

nabokihms:
@mdpradeep, everything seems fine now according to the code and ready to be merged (when tests are successfully passed).

 I'd like to take this feature to the upcoming v2.38.0 release, so waiting 🙂

connector/oidc/oidc.go
nabokihms: waiting for this one, because it was lost on rebasing

mdpradeep: Fixed in https://github.com/dexidp/dex/pull/3074/commits/b3e5bf544bbc192d6013f4b361c597ad32a4bd4c

nabokihms (approved):
Looks really nice. Thank you, @mdpradeep, for your contribution!

PR #3086, opened 2023-08-25
fix: Do not evaluate skipApproval on approval page

MM53:
#### Overview

Remove evaluation of skipApproval on approval page

The Pull Request https://github.com/dexidp/dex/pull/2897 caused a bug when using `oauth2.skipApprovalScreen: true` together with a client setting `approval_prompt=force` during the request. The tests validated correctly that the user gets redirected to `/approval` in this case, but the handler of the approval page contains some code which immediately returns the auth code if `oauth2.skipApprovalScreen: true` is set and ignores the `approval_prompt` parameter. Since there is a TODO comment to remove this check after some time, I guess it can be removed now to fix that issue.

```release-note
Fix a bug that Dex ignores approval_prompt=force when skipping approval screen is enabled as default behavior
```

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -613,12 +613,6 @@ func (s *Server) handleApproval(w http.ResponseWriter, r *http.Request) {
 
 	switch r.Method {
 	case http.MethodGet:
-		// TODO: `finalizeLogin()` now sends code directly to client without going through this endpoint,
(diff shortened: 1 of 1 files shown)


nabokihms (approved):
@MM53 sorry for the long delay. It does make sense. Thank you for your contribution.

PR #3114, opened 2023-09-21
feat: Add configurable CORS Headers

josiah-lunit:
#### Overview
The CORS enforced endpoints only allow the `Authorization` header by default, and this behavior is not configurable.  For some environments, it is desirable to allow configuration of which headers should be accepted (e.g. when using Swagger UI with a Dex IDP, which for instance inserts a `x-requested-with` header to the CORS preflight check).

This PR allows the `AllowedHeaders` for CORS endpoints to be configurable, so that the host can decide if headers should be rejected or not.  By default, the `Authorization` header will be accepted to not interfere with any existing setups of Dex if no `allowedHeaders` are defined in the `config.yaml` file.

Yes

```release-note
Adds a configuration option `allowedHeaders` in the `web` section of `config.yaml` to change which headers are accepted on CORS requests.
```

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -150,6 +150,7 @@ type Web struct {
 	TLSCert        string   `json:"tlsCert"`
 	TLSKey         string   `json:"tlsKey"`
 	AllowedOrigins []string `json:"allowedOrigins"`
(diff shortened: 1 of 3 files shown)


nabokihms (approved):
Thank you for your contribution. LGTM.

Issue #3118, opened 2023-09-25
Device Authorization Flow broken with etcd storage


MM53:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.37.0

### Storage Type

etcd

### Installation Type

Official container image

### Expected Behavior

Being able to use the oidc Device Authorization Flow with any type of storage. 

### Actual Behavior

When using the Device Authorization Flow with etcd as storage all login attempts resulted in an error page complaining about missing data like clientId, clientSecret, ... although those information were provided while generating the login url.

nabokihms:
Fixed by https://github.com/dexidp/dex/pull/3119

PR #3119, opened 2023-09-25
Fix etcd device requests

MM53:
#### Overview

Add a missing type conversion while reading device requests from etcd storage and add a check for this case to the storage conformance tests.

This PR resolves the bug described in https://github.com/dexidp/dex/issues/3118

```release-note
Add a missing type conversion while reading device requests from etcd storage
```

### DIFF
--- a/storage/conformance/conformance.go
+++ b/storage/conformance/conformance.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/kylelemons/godebug/pretty"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


nabokihms (approved):

nabokihms (approved):
Looks good to me. What a bug.  😅

nabokihms:
@MM53 there is a problem with the test. Would you mind fixing it?

MM53:
There is an issue with postgres showing "UCT" as default instead of "UTC". I set it to "UTC" for the tests.

nabokihms (approved):

Issue #3124, opened 2023-09-26
Offline sessions stored although not requested


MM53:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.37.0

### Storage Type

etcd

### Installation Type

Official container image

### Expected Behavior

When initiating a login without the 'offline_access' scope, no offline sessions should be created in any storage.

### Actual Behavior

The 'offline_access' scope is only taken into account when using the password grant or while exchanging the authCode. During the initial login it is currently ignored. Therefore an offline session will be created for every login. In most cases this isn't an issue because the session will only be persisted but never used again.

In our setup we have a scenario where some users are machine accounts which could try to login from multiple hosts at the same time. This results in failing login attempts because our storage (etcd) couldn't finish persisting one offline session before an update of this offline session will be triggered by another instance of dex.

Since there is no reason for creating an offline session if the user didn't request the scope offline_access, I think this is an unintended behavior.

PR #3125, opened 2023-09-26
Store offline sessions only if they were requested by the user

MM53:
#### Overview

Create offline sessions only if they are supported by the connector and if the user requests it with the scope `offline_access`.
Extend handler tests to verify this behavior.

This PR resolves the bug described in https://github.com/dexidp/dex/issues/3124

```release-note
Store offline sessions only if they are supported and were requested by the user
```

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -539,6 +539,17 @@ func (s *Server) finalizeLogin(identity connector.Identity, authReq storage.Auth
 		return returnURL, false, nil
 	}
 
+	offlineAccessRequested := false
+	for _, scope := range authReq.Scopes {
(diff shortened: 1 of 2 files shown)


server/handlers.go
```diff
@@ -539,6 +539,17 @@ func (s *Server) finalizeLogin(identity connector.Identity, authReq storage.Auth
 		return returnURL, false, nil
 	}
 
+	offlineAccessRequested := false
```
nabokihms, line 542: NIT
```suggestion
	var offlineAccessRequested bool
```

nabokihms (approved):

PR #3162, opened 2023-10-19
Implement client credentials grant

orange-hbenmabrouk:
#### Overview

This pull request:

- Introduces the "[Client Credentials](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4)" flow to the list of supported grant types.
- Adding an optional field to the oidc connector config in order to introduce a PKCE challenge to the code exchange flow (since some identity providers requires it)
- Fixes an issue with the device_authorization flow of Dex (the issue and it's solution are described below)

- If a client wants to use Dex with client credentials grant type for machine to machine use cases, they are not able to. So the addition of the client credentials flow would be valuable.

- If an external IdP is requiring a PKCE challenge when authenticating an **authorization_code**, the only choice available is to provide it with required information (code_challenge & code_challenge_method as URL params), otherwise the request won't be successful. 
=> The addition of the "pkce" parameter to the config of the oidc connector, enables the client to achieve this authentication with the challenge set:
> **code_challenge** & **code_challenge_method** are set to the authentication endpoint of the external issuer
> **code_verifier** set to the token endpoint of the external issuer

- The fix made for the device_authorization flow is that after the user authenticated via the external IdP through Dex, the external IdP redirects the user to a wrong URI whenever Dex issuer configuration has a suffix appended to it.
Example Dex configuration: 
```yaml
...
issuer: https://example.com/dex
...
```
A faulty redirection for this config would be: `https://example.com/device/callback/...`
A correct redirection for this config would be: `https://example.com/dex/device/callback/...`

```release-note
If
  1. you are using the device authorization flow with an oidc connector
  2. want to authenticate to the external IdP and exchange the access_token using a PKCE challenge
you have to add the "pkce" entry to your connector's config (it's the only action required)

Example:
  connectors:
  - id: my-oidc
    name: My Connector
    type: oidc
    config:
      issuer: https://example-external-issuer.com
      redirectURI: https://example-dex-server.com/callback
      clientID: ***
      clientSecret: ""
      pkce:
        enabled: true # false by default
```


p-kimberley:
+1 for this - highly desirable for our use case

p-kimberley:
Hi @nabokihms could you please review this when you get a chance? This is a blocker for our application and we'd appreciate if this PR could be assessed.

PR #3188, opened 2023-11-08
feat: Support PKCE in OIDC connector

vixus0:
#### Overview

This adds a configuration option `enablePKCE` to the OIDC connector to send code challenge and verifier parameters as required for PKCE.

Some OIDC providers require the PKCE flow, so a code challenge must be passed to the authorize endpoint, and a code verifier must be sent on the subsequent authorization token exchange.

Dex currently supports PKCE for downstream clients, but not for upstream OIDC providers.
Dynamic configuration using the OAuth2 discovery metadata isn't possible as the `coreos/go-oidc` library currently doesn't expose `code_challenge_methods_supported` in the `Provider`.

```release-note
* Adds the OIDC connector configuration option `enablePKCE` to send code
  challenge and verifier parameters to upstream OIDC providers as required for
  PKCE.
```

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -74,6 +74,9 @@ type Config struct {
 	// PromptType will be used fot the prompt parameter (when offline_access, by default prompt=consent)
 	PromptType string `json:"promptType"`
 
+	// Enable PKCE code challenge with upstream provider
+	EnablePKCE bool `json:"enablePKCE"`
+
 	// OverrideClaimMapping will be used to override the options defined in claimMappings.
 	// i.e. if there are 'email' and `preferred_email` claims available, by default Dex will always use the `email` claim independent of the ClaimMapping.EmailKey.
 	// This setting allows you to override the default behavior of Dex and enforce the mappings defined in `claimMapping`.
@@ -184,6 +187,11 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 		c.PromptType = "consent"
 	}
 
+	pkceVerifier := ""
+	if c.EnablePKCE {
+		pkceVerifier = oauth2.GenerateVerifier()
+	}
+
 	clientID := c.ClientID
 	return &oidcConnector{
 		provider:    provider,
@@ -209,6 +217,7 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 		promptType:                c.PromptType,
 		userIDKey:                 c.UserIDKey,
 		userNameKey:               c.UserNameKey,
+		pkceVerifier:              pkceVerifier,
 		overrideClaimMapping:      c.OverrideClaimMapping,
 		preferredUsernameKey:      c.ClaimMapping.PreferredUsernameKey,
 		emailKey:                  c.ClaimMapping.EmailKey,
@@ -243,6 +252,7 @@ type oidcConnector struct {
 	emailKey                  string
 	groupsKey                 string
 	newGroupFromClaims        []NewGroupFromClaims
+	pkceVerifier              string
 }
 
 func (c *oidcConnector) Close() error {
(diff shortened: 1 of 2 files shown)


cameronbrunner:
I did a similar implementation here: #3195 with the biggest difference being that mine generates a new verifier for each login session.  As 'oauth2.GenerateVerifier()' just gets a single random string it is stated that a new one should be done for each authorization: https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.15.0:pkce.go;l=22
Any thoughts on my PR?

cameronbrunner:
Sorry I missed your PR when I did mine... I had been thinking about it prior to your work and missed this one when I started mine.

connector/oidc/oidc.go
```diff
@@ -184,6 +187,11 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 		c.PromptType = "consent"
 	}
 
+	pkceVerifier := ""
+	if c.EnablePKCE {
```
nabokihms, line 191: Do you know if we need an option for this? We can rely on what the provider exposes via the discovery endpoint.

```
  "code_challenge_methods_supported": [
    "S256",
    "plain"
  ],
```

If the `code_challenge_methods_supported` section is present, we can use PKCE.

connector/oidc/oidc.go
```diff
@@ -265,6 +275,12 @@ func (c *oidcConnector) LoginURL(s connector.Scopes, callbackURL, state string)
 	if s.OfflineAccess {
 		opts = append(opts, oauth2.AccessTypeOffline, oauth2.SetAuthURLParam("prompt", c.promptType))
 	}
+
+	if c.pkceVerifier != "" {
+		c.pkceVerifier = oauth2.GenerateVerifier()
+		opts = append(opts, oauth2.S256ChallengeOption(c.pkceVerifier))
```
nabokihms, line 281: S256 can be unsupported by the provider. It is required to either clarify that Dex only supports the S256 challenge or provide support for others. 

We can rename the option `EnablePKCE` to `PKCEChallenge` instead of the boolean variable and let users specify the type of challenge they want to use.

nabokihms (changes requested):
I have some questions regarding the implementation. The idea lgtm!

connector/oidc/oidc.go
nabokihms: @vixus0 my bad, I overlooked the info from your first post
> Dynamic configuration using the OAuth2 discovery metadata isn't possible as the coreos/go-oidc library currently doesn't expose code_challenge_methods_supported in the Provider.

nabokihms: So, after checking the docs once again, there is an option to extract the provider's metadata by using https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#Provider.Claims

connector/oidc/oidc.go
```diff
@@ -265,6 +275,12 @@ func (c *oidcConnector) LoginURL(s connector.Scopes, callbackURL, state string)
 	if s.OfflineAccess {
 		opts = append(opts, oauth2.AccessTypeOffline, oauth2.SetAuthURLParam("prompt", c.promptType))
 	}
+
+	if c.pkceVerifier != "" {
+		c.pkceVerifier = oauth2.GenerateVerifier()
```
cameronbrunner, line 280: Isn't this setting just one verifier for all clients?  It appears that if 2 clients were logging in at the same time the first one's verifier would be overwritten by the seconds preventing a proper verifier exchange (line 317-320).

johnvan7:
Hi, I've tested this feature and it works well!
LGTM !
#2253

nabokihms:
This pr cannot be merged I’m afraid:
- it has conflicts
- I’m still waiting for requested fixes

If somebody wants to finish the work, feel free to do this

vixus0:
Unfortunately I wasn't able to dedicate any more time to these changes, but I'm glad the feature was picked up again!

Closing in favour of #3777.

Issue #3194, opened 2023-11-15
Support Upstream OIDC Providers without a ClientSecret


cameronbrunner:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

When using the OIDC connector Dex should support a secure method for distributing a common configuration file that could be used by multiple instances of a given application.  Presently this is not possible as the OIDC connector requires both a ClientID and ClientSecret leading to either the distribution of the ClientSecret to all application instances or a unique ClientID for application instances.  The former is insecure and the later is unfeasible with significantly large numbers of application instances.

### Proposed Solution

OIDC classifies client applications into two categories, Confidential and Public.  Dex presently behaves only as a Confidential application and thus requires a ClientID and ClientSecret.  I propose that support for Public be added as an option to the OIDC connector.

I believe all that is required is:
1) Add a new configuration option to enable Public mode for connectors.
2) When public mode is enabled allow for an empty ClientSecret
3) When public mode is enabled add a S256 Challenge option when generating the OIDC AuthCodeURL.
4) Use the appropriate verifier when running 'Exchange' later on in the flow.

The majority fo the actual work is handled in the downstream golang oidc code and was added in April of this year:

https://github.com/golang/oauth2/issues/603
https://github.com/golang/go/issues/59835 

Some sample code.

Add something like this to the LoginURL function (https://github.com/dexidp/dex/blob/e41a28bf27225ab503eb9feef4feedd03bb4ac71/connector/oidc/oidc.go#L253)
```golang
    if c.Public {
        verifier := oauth2.GenerateVerifier()
        opts = append(opts, oauth2.S256ChallengeOption(verifier))
        // Add state and verifier to a connector level cache with a short expiration.  Referenced in Exchange later.
       }
```

And this to HandleCallback (https://github.com/dexidp/dex/blob/e41a28bf27225ab503eb9feef4feedd03bb4ac71/connector/oidc/oidc.go#L291C18-L291C18)
```golang
    state := q.Get("state")
    // Look up state in cache
    var tok *oauth2.Token
    var err error
    if cachedVerifier == nil {
        tok, err = c.oauth2Config.Exchange(ctx, code)
    } else {
[...]

PR #3195, opened 2023-11-16
Issue 3194 - Add support for Public OIDC Providers

cameronbrunner:
Add S256 verification support to the OIDC connector.

#### Overview

This implementation is a concrete solution to what I reported in issue https://github.com/dexidp/dex/issues/3194.  It adds a new mode to the OIDC connector allowing Dex to operated as an OAuth2 'Public Application' secured by a S256 PKCE verifier.

This change increases the use cases Dex can address.  With Public Application support Dex can support a configuration that has no secrets and can be reused across multiple instances of a given application.

I think the biggest question regarding the PR is the cache choice and behavior. 

Yes.  Configuration options in the OIDC connector.

```release-note
* Add new Public Application mode to the OIDC connector.  This mode can be configured by setting the connector option 'Public' to true.  Additionally 'maxConcurrentPublicConnections' option can be used to control the max number of concurrent inflight authentications.
```


nabokihms:
Fixed by https://github.com/dexidp/dex/pull/3777

PR #3250, opened 2023-12-28
SAML Initiated Flow

MostafaAmer2200:
#### Overview

This was a requested feature in PR #1514. I've rebased the branch now everything should be up to date. In addition to some tests. 

This PR will allow users to authenticate using an IdP initiated flow. Before this feature, this only allows authentication through SP flow. This feature will allow to have both. 

This addresses #1514 which was never merged since it became outdated. There seems to be a demand for this feature from four years ago and I hope this resolve. Happy to take any feedback to get this moving forward :)


api/v2/api.proto
```diff
@@ -14,6 +14,13 @@ message Client {
   bool public = 5;
   string name = 6;
   string logo_url = 7;
+  SamlInitiatedConfig saml_initiated = 8;
```
nabokihms, line 17: Adding settings for a particular connector into generic structures is not good. Introducing a special handler for SAML is okay, but adding these options to ALL connectors could be better.

Documentation/connectors/saml.md
```diff
@@ -0,0 +1,152 @@
+# Authentication through SAML 2.0
```
nabokihms, line 1: The documentation has been moved out of this repo.

api/v2/api_grpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
 // - protoc-gen-go-grpc v1.3.0
-// - protoc             v3.21.12
+// - protoc             v3.20.3
```
nabokihms, line 4: It should not be changed.

nabokihms (changes requested):

Documentation/connectors/saml.md
MostafaAmer2200: Done

api/v2/api_grpc.pb.go
MostafaAmer2200: Regenerated it using the correct versions

Documentation/connectors/saml.md
MostafaAmer2200: Will add the docs in the correct repo once this has been merged

erhudy:
@nabokihms what can we do to get this moving again? We've been using this in production for several months now and it would be very nice to get it integrated into mainline Dex.

PR #3267, opened 2024-01-10
Override OIDC provider discovered claims

nabokihms:
#### Overview

The feature allows to override specific endpoints discovered by the go-oidc library.

The exemplary case is when you want to request the token internally in Kubernetes by using the service discovery DNS-based name without accessing the public address of an IdP.

Closes https://github.com/dexidp/dex/pull/2147

All the provider options could be added to be able to be overridden in the future.

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -219,6 +219,9 @@ func (c *Config) Open(id string, logger log.Logger) (conn connector.Connector, e
 		cancel()
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


sagikazarmark (approved):
LGTM

Maybe suggest extreme care when using this option (maybe even log it) as it can be a vulnerability when not properly configured.

nabokihms:
Adding a warning does make sense to me. Added.

Issue #3180, opened 2023-11-04
Version always contains the `-dirty` suffix


lht:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.37.0

### Storage Type

Kubernetes

### Installation Type

Official container image

### Expected Behavior

`dex version` reports the tagged version or git commit without the `-dirty` suffix if the git repository is "clean".

### Actual Behavior

```shell
docker run --rm -it --entrypoint /bin/sh  dexidp/dex:v2.37.0 -c 'dex version'
Dex Version: v2.37.0-dirty
Go Version: go1.20.4
Go OS/ARCH: linux amd64
```
Or, with latest
```
docker run --rm -it --entrypoint /bin/sh  dexidp/dex -c 'dex version'
Dex Version: e41a28bf27225ab503eb9feef4feedd03bb4ac71-dirty
Go Version: go1.21.3
Go OS/ARCH: linux amd64
```

### Additional Information

Some [files](https://github.com/dexidp/dex/blame/e41a28bf27225ab503eb9feef4feedd03bb4ac71/.dockerignore#L1C7-L2) are not included in Docker build context, and scripts/git-version consider files are deleted.

PR #3276, opened 2024-01-16
Propagate Dex version from build args

nabokihms:
#### Overview

Propagate the image version from the image metadata annotations to Dex. It will help us to avoid depending on git analysis by the script and take the date from the CI system.

Closes https://github.com/dexidp/dex/issues/3180

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,10 @@ RUN go mod download
 
 COPY . .
 
-RUN make release-binary
+# Propagate Dex version from build args to the build environment
+ARG VERSION
+RUN ([ "$VERSION" != "" ] || export $VERSION) && make release-binary
+
 RUN xx-verify /go/bin/dex && xx-verify /go/bin/docker-entrypoint
 
 FROM alpine:3.19.0 AS stager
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -29,7 +29,10 @@ RUN go mod download
 
 COPY . .
 
-RUN make release-binary
+# Propagate Dex version from build args to the build environment
+ARG VERSION
+RUN ([ "$VERSION" != "" ] || export $VERSION) && make release-binary
```
sagikazarmark, line 34: I don't think you need to export it. It should be available as an environment variable.

nabokihms: Tried this without exporting, and it didn't work 🤷 

It works if I do `ENV VERSION=$VERSION`, but returns an empty version if there is no build arg.

sagikazarmark: Here is a dummy draft that proofs that it should be working: https://github.com/dexidp/dex/pull/3277

If it doesn't, something else is going on.

Can you please check again?

nabokihms: Yep, seems like you are right. I tried the most minimalistic script, and it worked. I will revert the `expose` command instructions.

sagikazarmark (approved):

Issue #3335, opened 2024-02-07
Token Exchange uses wrong tokenType for /userinfo request


MrDeerly:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.38.0

### Storage Type

etcd

### Installation Type

Official Helm chart

### Expected Behavior

When performing a tokenExchange, Dex will reuse the tokenType set in `subject_token_type` for requesting the userInfo from the upstream IdP (https://github.com/dexidp/dex/blob/master/connector/oidc/oidc.go#L436).

This will lead to errors as neither `urn:ietf:params:oauth:token-type:access_token` nor `urn:ietf:params:oauth:token-type:id_token` are valid token types for the oidc /userinfo endpoint as per RFC6750.

### Actual Behavior

Dex follows the oidc spec and uses `Bearer` as token type for `/userinfo` requests.

https://openid.net/specs/openid-connect-core-1_0.html#UserInfo (5.3.1)

> The Access Token obtained from an OpenID Connect Authentication Request MUST be sent as a Bearer Token, per Section 2 of [OAuth 2.0 Bearer Token Usage](https://openid.net/specs/openid-connect-core-1_0.html#RFC6750) [RFC6750].

PR #3336, opened 2024-02-07
fix(tokenExchange): use correct token type for userInfo requests

MrDeerly:
#### Overview

This small changes fixes an issue in the token exchange flow. It explicitly sets the tokenType used for `GET /userinfo` to `Bearer` to be conform with the OIDC spec.

Without this change the tokenType would be set to either `urn:ietf:params:oauth:token-type:access_token` or `urn:ietf:params:oauth:token-type:id_token`, which doesn't satisfy the OIDC spec.

> The Access Token obtained from an OpenID Connect Authentication Request MUST be sent as a Bearer Token, per Section 2 of [OAuth 2.0 Bearer Token Usage](https://openid.net/specs/openid-connect-core-1_0.html#RFC6750) [RFC6750].

https://openid.net/specs/openid-connect-core-1_0.html#UserInfo (5.3.1)

Closes https://github.com/dexidp/dex/issues/3335

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -433,7 +433,10 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 	// We immediately want to run getUserInfo if configured before we validate the claims.
 	// For token exchanges with access tokens, this is how we verify the token.
 	if c.getUserInfo {
-		userInfo, err := c.provider.UserInfo(ctx, oauth2.StaticTokenSource(token))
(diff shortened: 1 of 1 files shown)


nabokihms:
@MrDeerly thanks! It does make sense. Because there are no integration tests, I will test this PR manually. If everything is okay, I will bump the merge button.

nabokihms (approved):
Tested it and it works as expected, so, LGTM! Thank you for the fix.

PR #3373, opened 2024-02-23
OIDC connector: Allow specifying empty prompt type

nabokihms:
#### Overview

Enhanced the OIDC connector to allow specifying an empty promptType parameter. Previously, the default behavior always appended 'consent' if promptType was not specified. This adjustment was necessary due to variations in default behaviors across certain Identity Providers (IDPs).

https://cloud-native.slack.com/archives/C01HF2G1R34/p1708423095710099


connector/oidc/oidc.go
```diff
@@ -76,7 +76,7 @@ type Config struct {
 	UserNameKey string `json:"userNameKey"`
 
 	// PromptType will be used fot the prompt parameter (when offline_access, by default prompt=consent)
-	PromptType string `json:"promptType"`
+	PromptType *string `json:"promptType"`
```
sagikazarmark, line 79: If this is a deviation from the spec (even if due to some IdP implementing it improperly) we should clearly state that both in the code and documentation.

sagikazarmark (approved):
One comment, otherwise LGTM

connector/oidc/oidc.go
nabokihms: It doesn't seem like an absolute deviation. This is not fixed anywhere that IdP must ask for consent if the prompt is empty by default. The RFC only says how to behave if the prompt is provided.

For me, it seems like consent is the Dex default only (which we state in the doc and the code).

sagikazarmark: SGTM

PR #3432, opened 2024-03-25
Update Distroless to Debian 12

haydentherapper:
#### Overview

Updated the Distroless image to Debian 12

gcr.io/distroless/static is Debian 10, which is quite outdated. Updated to the Debian 12 version of the static image, and used the nonroot tag since the root user isn't needed.

We've been running a version of Dex in production with this image without issue.


nabokihms:
Thanks @haydentherapper.

LGTM, I'm just curious: is there any problem with the debian 10 static image? I always thought that being distroless makes the image distro-independent.

nabokihms (approved):

loosebazooka:
There are some files in static that are pulled from debian repos like CA certs and such.

loosebazooka:
also minor correction, the un-debian-versioned tag points to debian11, but we're going to start deprecating those images this year at some point.

PR #3458, opened 2024-04-08
use the default cosign version from the action

cpanato:
#### Overview

- use the default cosign version from the action

not pinning the cosign version helps to be always up-to-date and depends only in the action update


cpanato:
> Is it pinned in the action? I've burnt myself with silent upgrades more than I care to admit.

yes, it is pinned, but if you want to do another PR to update the cosign version is fine as well, just trying to reduce the overhead

sagikazarmark:
> yes, it is pinned

That's enough for me

sagikazarmark (approved):

Issue #3450, opened 2024-04-04
[google connector] prompt=consent on all offline_access scopes


TheRealNoob:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.39.0

### Storage Type

Kubernetes

### Installation Type

Official Helm chart

### Expected Behavior

Dex should omit the `prompt` field when using the Google connector.  Or allow it to be configurable using `config.promptType`.

### Actual Behavior

If someone opens an authentication flow against Dex with `scope=["offline_access"]` then dex includes `prompt=consent` on its request to Google every time ([this code](https://github.com/dexidp/dex/blob/master/connector/google/google.go#L172-L174)).  This is adds unnecessary dialogue boxes for the user.  Per Google's [docs](https://developers.google.com/identity/openid-connect/openid-connect#re-consent):
> You can prompt the user to re-authorize your app by setting the [prompt](https://developers.google.com/identity/openid-connect/openid-connect#prompt) parameter to consent in your [authentication request](https://developers.google.com/identity/openid-connect/openid-connect#sendauthrequest). When prompt=consent is included, the consent screen is displayed every time your app requests authorization of scopes of access, even if all scopes were previously granted to your Google APIs project. For this reason, include prompt=consent only when necessary.

Per [these docs](https://developers.google.com/identity/protocols/oauth2/web-server#creatingclient) (see the table) if you omit the `prompt` field, then Google will prompt on first consent and not after.

nabokihms:
Making it configurable it a good improvement.

TheRealNoob:
@nabokihms it seems it's not possible to omit the `prompt` field via `promptType`.  Perhaps there's another way?  Or if it's added to the Google connector, it would only make sense that it would have to added to all?

abhisek:
Looks like a simple fix to me that involves:

1. Introduce a new configuration parameter, essentially a boolean toggle
2. Based on the config, include or skip adding `oauth2.SetAuthURLParam("prompt", "consent")` in the `LoginURL`

[...]

PR #3475, opened 2024-04-13
feat: Add support for configurable prompt type for Google connector

abhisek:
#### Overview

Add support for configurable prompt type for Google Connector. Reference https://github.com/dexidp/dex/issues/3450

- Allows the user to explicitly configure the value of `prompt` parameter in login URL generated by Google connector

Ref https://github.com/dexidp/dex/issues/3450

The original ask was to avoid the user experience of consent screen every time authentication flow goes through `dex` into Google by skipping `prompt=consent` flag. As per OIDC specification, there is no valid value for `prompt` param to achieve this. So this PR just allows the ability to set the value of `prompt` parameter through configuration. Setting an explicit `""` (empty) value for prompt should be functionally equivalent for skipping it for Google connector

### DIFF
--- a/connector/google/google.go
+++ b/connector/google/google.go
@@ -58,6 +58,10 @@ type Config struct {
 
 	// If this field is true, fetch direct group membership and transitive group membership
 	FetchTransitiveGroupMembership bool `json:"fetchTransitiveGroupMembership"`
+
+	// Optional value for the prompt parameter, defaults to consent when offline_access
+	// scope is requested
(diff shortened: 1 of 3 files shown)


nabokihms (commented):
Looks good to me, thanks! The only thing left is satisfying the linter.

Oh, and one more. Could you please also open a PR to the website repo with the doc fix? https://dexidp.io/docs/connectors/google/

nabokihms (approved):
Thanks a lot. Great contribution!

PR #3502, opened 2024-05-03
use log/slog for structured logging

### LINKED ISSUE #2020: Introduce structured logging

seankhliao:
#### Overview

Switches pkg/log and logrus for the standard library's slog logger.

This is more or less a mechanical replacement of the internal `log.Logger` interface and `logrus` logger with the standard library's slog.
This appears to be the general community consensus on a standard logging interface, where slog.Logger is used as the common interface, and implementations can switch out the `slog.Handler` if necessary.

Connectors have their type and id added as attributes.
Log lines are lowercased for consistency.

Closes #2020

If dex is used as a library, this will be a breaking change.
Log line output will be different from before.


cmd/dex/serve.go
```diff
@@ -98,22 +97,22 @@ func runServe(options serveOptions) error {
...
-		version,
-		runtime.Version(),
-		runtime.GOOS,
-		runtime.GOARCH,
+	logger.Info(
+		"Version info",
+		"dex_version", version,
+		"go_version", runtime.Version(),
+		"go_os", runtime.GOOS,
+		"go_arch", runtime.GOARCH,
```
sagikazarmark, line 105: nit: I wonder if using slog's group feature here would make sense. As far as I understand, that basically means using dot-notated labels (ie. `go.version` instead of `go_version`).

https://pkg.go.dev/log/slog#hdr-Groups

cmd/dex/serve.go
```diff
@@ -508,53 +507,29 @@ func runServe(options serveOptions) error {
...
 	case "", "text":
-		formatter.f = &logrus.TextFormatter{DisableColors: true}
+		slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
+			Level: level,
+		})
 	case "json":
-		formatter.f = &logrus.JSONFormatter{}
+		slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
+			Level: level,
+		})
```
sagikazarmark, line 527: Although I don't think it makes a lot of difference, this changes where logs go (stdout instead of stderr).

We should at least highlight it somewhere.

WDYT @nabokihms ?

connector/atlassiancrowd/atlassiancrowd.go
```diff
@@ -80,16 +80,16 @@ type crowdAuthenticationError struct {
 }
 
 // Open returns a strategy for logging in through Atlassian Crowd
-func (c *Config) Open(_ string, logger log.Logger) (connector.Connector, error) {
+func (c *Config) Open(id string, logger *slog.Logger) (connector.Connector, error) {
 	if c.BaseURL == "" {
 		return nil, fmt.Errorf("crowd: no baseURL provided for crowd connector")
 	}
-	return &crowdConnector{Config: *c, logger: logger}, nil
+	return &crowdConnector{Config: *c, logger: logger.With("connector_type", "atlassiancrowd", "connector_id", id)}, nil
```
sagikazarmark, line 87: Same thing about connectors: maybe use `slog.Group`?

sagikazarmark (commented):
Thank you @seankhliao, I really appreciate you tackling this.

I'm mostly satisfied with how things look. I made a few comments, curious what you think.

cmd/dex/serve.go
seankhliao: I changed it to stdout. I just wasn't sure which one it should have gone to.

cmd/dex/serve.go
seankhliao: I changed it to use groups.
It'll be `go.version` in the text output, and in json `"go": {"version": ...}`

connector/atlassiancrowd/atlassiancrowd.go
seankhliao: updated to use groups too

cmd/dex/serve.go
nabokihms: Can be a problem, but mentioning it in changelog should be fine enough.

connector/ldap/ldap.go
```diff
@@ -188,12 +188,12 @@ func parseScope(s string) (int, bool) {
...
 // group to user matching implementations.
 // See "Config.GroupSearch.UserMatchers" comments for the details
-func userMatchers(c *Config, logger log.Logger) []UserMatcher {
+func userMatchers(c *Config, logger *slog.Logger) []UserMatcher {
 	if len(c.GroupSearch.UserMatchers) > 0 && c.GroupSearch.UserMatchers[0].UserAttr != "" {
 		return c.GroupSearch.UserMatchers
 	}
 
-	log.Deprecated(logger, `LDAP: use groupSearch.userMatchers option instead of "userAttr/groupAttr" fields.`)
+	slog.Warn(`use "groupSearch.userMatchers" option instead of "userAttr/groupAttr" fields`, "type", "deprecated", "connector", "ldap")
```
nabokihms, line 196: Should we use the logger here? Otherwise the logline will not be formatted as JSON. 

BTW what is type? Maybe we can use a boolean label like `"deprecated": true` similar to Kubernrtes?

cmd/dex/serve.go
seankhliao: sorry I meant I changed it back to stderr

connector/ldap/ldap.go
seankhliao: I was a bit unsure about it since the old deprecation messages logged through a different logger.
I was thinking of the type of log message when I originally went with type=deprecated.

I've changed it to use the component's logger and type=deprecated -> deprecated=true (plus 2 other places where deprecations were logged).

nabokihms (approved):

seankhliao:
i rebased to resolve a conflict with new additions to the google connector

nabokihms:
@sagikazarmark from my point of view, looks good enough to be merged. Anything else from your side?

sagikazarmark (approved):
LGTM thanks @seankhliao

yardenshoham:
Hi how to set it to log debug statements?

Previously it was
```yaml
logger:
  level: debug
```

PR #3506, opened 2024-05-03
feat: allow domain names or IDs in keystone connector

cardoe:
OpenStack Keystone allows a user to authenticate against a domain. That domain can be specified either as the domain ID or the domain name when authenticating. The domain ID is a UUID or the special "default" domain ID so key off of that when deciding what to submit to the keystone API. Collapsed the code to share the domainKeystone struct by utilizing omitempty to skip unset fields.

### DIFF
--- a/connector/keystone/keystone.go
+++ b/connector/keystone/keystone.go
@@ -10,11 +10,13 @@ import (
 	"log/slog"
 	"net/http"
 
+	"github.com/google/uuid"
+
 	"github.com/dexidp/dex/connector"
 )
 
 type conn struct {
-	Domain        string
+	Domain        domainKeystone
 	Host          string
 	AdminUsername string
 	AdminPassword string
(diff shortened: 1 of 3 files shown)


nabokihms (changes requested):
@cardoe thanks, idea looks great. Could you check why tests didn't pass?
```
--- FAIL: TestValidUserLogin (3.44s)
    --- PASS: TestValidUserLogin/test_with_email_address (1.00s)
    --- PASS: TestValidUserLogin/test_without_email_address (0.99s)
    --- FAIL: TestValidUserLogin/test_with_domain_ID (0.51s)
    --- FAIL: TestValidUserLogin/test_with_domain_Name (0.52s)
```

cardoe:
It's cause the dex project has actual good tests and is testing against a real keystone while I fudged it in the tests and had hand added the data into my local keystone. So 👍 on having a good test environment. I'll clean this up and improve the test coverage. Thanks again for the review.

cardoe:
I've now tested it with the rocky container used in the tests as well as my local keystone.

cardoe:
I've also rebased and fixed up for the log changes that happened recently.

nabokihms (approved):
Looks great. Thanks for your contribution, @cardoe!

Issue #3519, opened 2024-05-13
Option to override jwks_uri for OIDC


sohgaura:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

Please provide option to override jwks_uri. Currently it considers value fetched from .well-known/openid-configuration of issuer.

### Proposed Solution

Please add option to override jwksurl similar to already provided option for authurl and tokenurl.

nabokihms:
closed by https://github.com/dexidp/dex/pull/3543

PR #3527, opened 2024-05-13
Add ability to override JWKS URL for OIDC connector

nabokihms:
#### Overview

Add an option to override discovered JWKS endpoint for OIDC connector

closes https://github.com/dexidp/dex/issues/3519

The test to check the change would have been cumbersome if we had implemented it (because the library doesn't expose the JWKS URL).

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -107,6 +107,9 @@ type ProviderDiscoveryOverrides struct {
 	// AuthURL provides a way to user overwrite the Auth URL
 	// from the .well-known/openid-configuration authorization_endpoint
 	AuthURL string `json:"authURL"`
+	// JWKSURL provides a way to user overwrite the JWKS URL
+	// from the .well-known/openid-configuration jwks_uri
+	JWKSURL string `json:"jwksURL"`
 }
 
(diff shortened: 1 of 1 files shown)


connector/oidc/oidc.go
```diff
@@ -107,6 +107,9 @@ type ProviderDiscoveryOverrides struct {
 	// AuthURL provides a way to user overwrite the Auth URL
 	// from the .well-known/openid-configuration authorization_endpoint
 	AuthURL string `json:"authURL"`
+	// JWKSURL provides a way to user overwrite the JWKS URL
+	// from the .well-known/openid-configuration jwks_uri
+	JWKSURL string `json:"jwksURL"`
 }
 
 func (o *ProviderDiscoveryOverrides) Empty() bool {
```
sohgaura, line 115: Pls do a check for JWKSURL as well, current condition only checks authurl and tokenurl to override values. This will not work if only JWKSURL is passed. 
Add required changes to [oidc_test.go](https://github.com/dexidp/dex/blob/master/connector/oidc/oidc_test.go) for consistency.

nabokihms:
Closing in favour of https://github.com/dexidp/dex/pull/3543

PR #3554, opened 2024-05-29
Prevent double clicking of the submit button on login

pcorliss:
#### Overview

Disables the submit button on login to prevent clicking multiple times while waiting for a response from the form submission on login.

When connecting to a very slow provider a response can take several seconds, without a visual indicator users click the submit button several times. After multiple clicks when a response is returned it's often a 500 error page within our setup. Better to prevent the user from submitting the form multiple times and indicating that by disabling the button.

We can solve this issue internally by having a custom frontend documented here https://dexidp.io/docs/guides/templates/ but this felt like a good thing to share more widely to help others.

If necessary the following can be rewritten inline to avoid loading a unique JS file.

### DIFF
--- a/web/static/form.js
+++ b/web/static/form.js
@@ -1,4 +1,0 @@
-document.querySelector('form').onsubmit = function(e) {
-    var el = document.querySelector('#submit-login');
-    el.setAttribute('disabled', 'disabled');
-};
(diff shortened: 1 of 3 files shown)


nabokihms (changes requested):
Overall the idea looks cool. Just a couple of questions.

PR #3558, opened 2024-05-31
feat: new Prometheus metrics build_info

nlamirault:
#### Overview

Add a new metric `dex_build_info`, with `version` which will be set to the Git tag version.

```
$ curl -s http://127.0.0.1:5558/metrics | grep build
# HELP dex_build_info A metric with a constant '1' value labeled by version from which Dex was built.
# TYPE dex_build_info gauge
dex_build_info{goarch="arm64",goversion="go1.22.2",version="43a9a144a710803b6a3036368cec3bec9f6f02de-dirty"} 1 
```

### DIFF
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -51,6 +51,16 @@ type serveOptions struct {
 	grpcAddr      string
 }
 
+var (
+	buildInfo = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "build_info",
+			Help: "A metric with a constant '1' value labeled by version from which Dex was built.",
+		},
+		[]string{"version", "goversion", "goarch"},
+	)
+)
+
 func commandServe() *cobra.Command {
 	options := serveOptions{}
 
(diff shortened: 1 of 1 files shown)


nabokihms (changes requested):
Some suggestion to make the metrics match Kubernetes build info labels:
1. `go_version` instead of `goversion` (for readability) 
2. `platform` instead of`goarch`, because this label is usually used to understand where you can run the software

Overall, PR looks great! Thanks for your effort.

nabokihms:
@nlamirault could you please fix the lint error?

nlamirault:
I don't know how to do that:

```
└─ ➜ go fmt cmd/dex/serve.go

└─ ➜ make lint
golangci-lint has version 1.59.1 built with go1.22.4 from 1a55854 on 2024-06-08T22:05:49Z
WARN [config_reader] The configuration option `run.skip-dirs` is deprecated, please use `issues.exclude-dirs`.
WARN [config_reader] The configuration option `linters.gci.local-prefixes` is deprecated, please use `prefix()` inside `linters.gci.sections`.
cmd/dex/serve.go:54: File is not `gofumpt`-ed (gofumpt)
var (
        buildInfo = prometheus.NewGaugeVec(
                prometheus.GaugeOpts{
                        Name:      "build_info",
                        Namespace: "dex",
                        Help:      "A metric with a constant '1' value labeled by version from which Dex was built.",
                },
                []string{"version", "go_version", "platform"},
        )
make: *** [lint] Error 1
```

nabokihms:
It should help:
```
make help
make fix
```

nabokihms (approved):

PR #3584, opened 2024-06-17
fix: always retrieve github emails when `preferredEmailDomain` is set

thesuperzapper:
## Overview

As discussed in https://github.com/dexidp/website/pull/176, even when `preferredEmailDomain` (which was added by https://github.com/dexidp/dex/pull/2740) is set on the GitHub connector, if the user has a __public__ email on their GitHub profile, it will take precedence over any private one which might match the `preferredEmailDomain` selector.

This is problematic for organizations which want to use the email filter to assign permissions to employees based on their corporate email, as they must tell users to remove any public emails from their profile.

## What this PR does / why we need it

Simply makes it so that we always preform an email check when `preferredEmailDomain` is non empty.

## Special notes for your reviewer

N/A

### DIFF
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -531,9 +531,10 @@ func (c *githubConnector) user(ctx context.Context, client *http.Client) (user, 
 		return u, err
 	}
 
-	// Only public user emails are returned by 'GET /user'. u.Email will be empty
-	// if a users' email is private. We must retrieve private emails explicitly.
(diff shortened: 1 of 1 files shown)


nabokihms:
This is a breaking change, so we must mention it in the RM. Overall, it looks decent. Thanks, @thesuperzapper

nabokihms (approved):

PR #3589, opened 2024-06-20
Gracefully handle no args being passed to entrypoint

rgmz:
#### Overview

This is a minor change to prevent a panic in an admittedly niche edge case.

This PR gracefully handles no args being passed to the entrypoint binary.
```sh
# before
./docker-entrypoint
panic: runtime error: index out of range [0] with length 0
goroutine 1 [running]:
main.run({0xc000014140?, 0xc00006e740?, 0x0?}, 0x60?, 0x0?)
/usr/local/src/dex/cmd/docker-entrypoint/main.go:56 +0x5c5
main.main()
/usr/local/src/dex/cmd/docker-entrypoint/main.go:21 +0x65

# after
./docker-entrypoint 
error: no args passed to entrypoint
```

This can occur if someone builds upon the upstream dex image (e.g., adding internal company certificates) and inadvertently breaks the entrypoint/cmd.

### DIFF
--- a/cmd/docker-entrypoint/main.go
+++ b/cmd/docker-entrypoint/main.go
@@ -17,6 +17,10 @@ func main() {
 	// Note that this docker-entrypoint program is args[0], and it is provided with the true process
 	// args.
 	args := os.Args[1:]
+	if len(args) == 0 {
+		fmt.Println("error: no args passed to entrypoint")
+		os.Exit(1)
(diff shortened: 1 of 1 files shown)


nabokihms (approved):

nabokihms:
Good catch! Thanks, @rgmz

PR #3598, opened 2024-07-01
Added Discovery to grpc

koendelaat:
#### Overview

Adds the discovery information to the API

Allow API clients to retrieve important information regarding dex config (e.g. `issuerUrl`)

#### Use-case:

API is being used to create `clients`, however to use this client the `issuerUrl` is required


koendelaat:
@nabokihms, Can you add the required labels?

nabokihms:
@koendelaat sure I can! Could you please share why do we need this feature?

koendelaat:
> @koendelaat sure I can! Could you please share why do we need this feature?

This allows for API clients to get information about the Dex server.

Especially when used in Infrastructure as Code (e.g. Terraform), this information can be used to provision other resources that might require this information (e.g. `issuerUrl` required by oauth2-proxy to do token validation).

I've extended the terraform provider with this functionality based on this PR, see https://github.com/koendelaat/terraform-provider-dexidp/ and https://registry.terraform.io/providers/koendelaat/dexidp/latest/docs/data-sources/discovery

nabokihms (approved):
Overall, it looks great. Thanks @koendelaat

PR #3636, opened 2024-07-18
added support to set default scopes to static clients

dkhachyan:
#### Overview

Added the ability to set a  list of default scopes for static clients.

Adds the ability to specify a list default scopes for static clients, which will be included in the request in addition to `openid`


nabokihms:
@dkhachyan cool feature!
However, we cannot merge it without adding dynamic client support.

Issue #3632, opened 2024-07-17
CVE-2024-24790


MoeBensu:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

v2.40.0

### Storage Type

Kubernetes

### Installation Type

Official container image

### Expected Behavior

Vulnerability-free docker image

### Actual Behavior

[CVE-2024-24790](https://nvd.nist.gov/vuln/detail/CVE-2024-24790) has been published against the go stdlib net/netip and is found by trivy in docker image v2.40:

<img width="1511" alt="cve" src="https://github.com/user-attachments/assets/03c2c862-0d20-4c67-9203-b3d16b8af83f">

### Steps To Reproduce

trivy image --ignore-unfixed --exit-code 1 --severity CRITICAL ghcr.io/dexidp/dex:v2.40.0

nabokihms:
Thanks, we will fix this in the upcoming release

MoeBensu:
@nabokihms Thanks! Is it already known when the upcoming release will come out?

nabokihms:
It is scheduled for next week

MoeBensu:
@nabokihms Thank you for the fixes that were merged today into master (#3637). 

According to trivy, the CVE-2024-24790 existed previously in v2.40.0-latest at
- usr/local/bin/dex (gobinary)
- usr/local/bin/docker-entrypoint (gobinary)
- usr/local/bin/gomplate (gobinary)

After #3637, it got fixed, but remains existing at usr/local/bin/gomplate (gobinary)
<img width="1477" alt="Screenshot 2024-07-26 at 09 47 44" src="https://github.com/user-attachments/assets/77f3e22d-3b19-4c79-aadc-ff8b8a68b571">

PR #3652, opened 2024-07-26
Bump gomplate 4.0.1

MoeBensu:

Fixes https://github.com/dexidp/dex/issues/3632 by bumping gomplate from v3.11.8 to v4.0.0

This is breaking change bump in gomplate and requires testing. This is just a suggestion from me to fix the CVE. The local setup is not working well in my machine and I can't find the right docs to pursue that further, therefore I will set it to draft till tested by somebody else.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ ARG TARGETOS
 ARG TARGETARCH
 ARG TARGETVARIANT
 
-ENV GOMPLATE_VERSION=v3.11.8
+ENV GOMPLATE_VERSION=v4.0.1
 
 RUN wget -O /usr/local/bin/gomplate \
   "https://github.com/hairyhenderson/gomplate/releases/download/${GOMPLATE_VERSION}/gomplate_${TARGETOS:-linux}-${TARGETARCH:-amd64}${TARGETVARIANT}" \


Dockerfile
```diff
@@ -47,7 +47,7 @@ ARG TARGETOS
 ARG TARGETARCH
 ARG TARGETVARIANT
 
-ENV GOMPLATE_VERSION=v3.11.8
+ENV GOMPLATE_VERSION=v4.0.0
```
nabokihms, line 50:
```suggestion
ENV GOMPLATE_VERSION=v4.0.1
```
There is a new release already.

https://github.com/hairyhenderson/gomplate/releases/tag/v4.0.1

nabokihms:
The [number of breaking changes and deprecations](https://github.com/hairyhenderson/gomplate/releases/tag/v4.0.0) really blows my mind.

@sagikazarmark I think it starts to get boring updating complete every time CVE is found in go runtime. Maybe we can come back and discuss dex config and its templating once again.

Dockerfile
MoeBensu: Since it was a breaking change release. I thought to better go with 4.0.0, in case community has more experience with. Anyway, I bumped it now to v4.0.1

nabokihms (approved):

Issue #1788, opened 2020-08-31
Add IP addresses of clients to log messages


nabokihms:
Sometimes it is helpful to see users' IP addresses in successful or failed authorization request log entries. This behavior can be achieved by using simple nginx in front of dex and looking into nginx logs, but it would be more convenient and simpler to see addresses in dex logs.

Maybe it's better to add this feature only for debug logging mode.
For example:
```
{"level":"info", "client_address":"192.168.0.1", "msg":"login successful: connector \"what-is-it\", username=\"Maksim Nabokikh\", preferred_username=\"maksim.nabokikh\", email=\"maksim.nabokikh@example.com\", groups=[]","time":"2020-08-31T14:41:31Z"}
```

sagikazarmark:
I'd definitely make that optional.....because GDPR.

sagikazarmark:
Note: need to be able to set trusted proxies to properly detect client IPs.

nabokihms:
closed by https://github.com/dexidp/dex/pull/3661

PR #3661, opened 2024-07-30
Enrich Dex logs with real IP and request ID

nabokihms:
#### Overview

- Include client's real IP in logs.
- Capture and log unique request IDs.
- Update middleware to extract and forward this data.
- Enhance debugging and traceability. Improves insight and troubleshooting efficiency.

https://github.com/dexidp/dex/issues/1788

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"net/netip"
 	"os"
 	"slices"
 	"strings"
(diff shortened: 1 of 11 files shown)


cmd/dex/config.go
```diff
@@ -182,15 +183,36 @@ type OAuth2 struct {
...
+	TLSMinVersion  string         `json:"tlsMinVersion"`
+	TLSMaxVersion  string         `json:"tlsMaxVersion"`
+	AllowedOrigins []string       `json:"allowedOrigins"`
+	AllowedHeaders []string       `json:"allowedHeaders"`
+	ClientRemoteIP ClientRemoteIP `json:"clientRemoteIP"`
+}
+
+type ClientRemoteIP struct {
+	Header            string   `json:"header"`
+	TrustedProxyCIDRs []string `json:"trustedProxyCIDRs"`
```
sagikazarmark, line 200: How about just calling this `TrustedProxies`? Sounds simpler and I think people will still understand they need to pass CIDRs in there.

cmd/dex/config.go
```diff
@@ -182,15 +183,36 @@ type OAuth2 struct {
...
+}
+
+func (cr *ClientRemoteIP) ToParsedCIDRs() (string, []netip.Prefix, error) {
+	if cr == nil {
+		return "", nil, nil
+	}
```
sagikazarmark, line 206:
```suggestion
	}

```

Note to self: add more linter rules.

cmd/dex/config.go
```diff
@@ -182,15 +183,36 @@ type OAuth2 struct {
...
+	trusted := make([]netip.Prefix, 0, len(cr.TrustedProxyCIDRs))
+	for _, cidr := range cr.TrustedProxyCIDRs {
+		ipNet, err := netip.ParsePrefix(cidr)
+		if err != nil {
+			return "", nil, fmt.Errorf("failed to parse CIDR %q: %v", cidr, err)
+		}
+		trusted = append(trusted, ipNet)
+	}
```
sagikazarmark, line 214:
```suggestion
	}

```

cmd/dex/config.go
```diff
@@ -182,15 +183,36 @@ type OAuth2 struct {
...
+	AllowedHeaders []string       `json:"allowedHeaders"`
+	ClientRemoteIP ClientRemoteIP `json:"clientRemoteIP"`
+}
+
+type ClientRemoteIP struct {
+	Header            string   `json:"header"`
+	TrustedProxyCIDRs []string `json:"trustedProxyCIDRs"`
+}
+
+func (cr *ClientRemoteIP) ToParsedCIDRs() (string, []netip.Prefix, error) {
```
sagikazarmark, line 203: Why return the header here?

server/server.go
```diff
@@ -84,6 +87,9 @@ type Config struct {
 
 	// Headers is a map of headers to be added to the all responses.
 	Headers http.Header
+	// Header to extract real ip from.
```
sagikazarmark, line 90:
```suggestion

	// Header to extract real ip from.
```

server/server.go
```diff
@@ -682,3 +729,18 @@ func (s *Server) getConnector(id string) (Connector, error) {
 
 	return conn, nil
 }
+
+type logRequestKey string
+
+const (
+	LogRequestKeyRequestID logRequestKey = "request_id"
+	LogRequestKeyRemoteIP  logRequestKey = "client_remote_addr"
```
sagikazarmark, line 737:
```suggestion
	RequestKeyRequestID logRequestKey = "request_id"
	RequestKeyRemoteIP  logRequestKey = "client_remote_addr"
```

Drop the log part: we don't use it for anything else, but at the point of insertion it doesn't really matter IMO.

server/server.go
```diff
@@ -682,3 +729,18 @@ func (s *Server) getConnector(id string) (Connector, error) {
...
+)
+
+func SetLogRequestID(ctx context.Context) context.Context {
```
sagikazarmark, line 740:
```suggestion
func WithRequestID(ctx context.Context) context.Context {
```

sagikazarmark (commented):
LGTM, couple suggestions

cmd/dex/config.go
nabokihms: Ok, it does make sense. Fixed.

cmd/dex/config.go
nabokihms: At first, I thought this method could prepare everything for IP address detection. If you think this is odd, I can just remove it, if you don't mind. 

Fixed.

sagikazarmark (approved):

cmd/dex/config.go
sagikazarmark: I think I like the latest version better. If there is need for some reason to pack all that into a single method, I'd call it something like `prepareClientIPDetection`.

But as I said, I like the latest version.

PR #3699, opened 2024-08-16
Fix Microsoft connector emailToLowercase for refresh token

### LINKED ISSUE #3283: Microsoft connector - emailToLowercase only respected for initial access token
### Preflight Checklist

### Version

2.37.0

### Storage Type

Kubernetes

### Installation Type

Official Helm chart

### Expected Behavior

Logging into Dex with Microsoft as backend connector transforms the camelcase email from Microsoft AD correctly into a lowercase email. When the access token expires and a new one is issued using the refresh token, the email address should be transformed again. 
After refreshing the access token, Kubernetes API-Server doesn't authorize the token since the email doesn't match the RoleBinding anymore.  

### Actual Behavior

When the access token expires and a new one is issued using the refresh token, the email address is not transformed again. 

### Steps To Reproduce

_No response_

### Additional Information

_No response_

### Configuration

_No response_

### Logs

_No response_

hur:
Fixes #3283

#### Overview

Respect the `emailToLowercase` config in the `Refresh()` function

Updates the `Refresh` function to also consider the config option `emailToLowercase`.

I have added unit tests. I have not tested in practice - I don't have a access to a Microsoft tenant.

### DIFF
--- a/connector/microsoft/microsoft_test.go
+++ b/connector/microsoft/microsoft_test.go
@@ -1,6 +1,7 @@
 package microsoft
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


connector/microsoft/microsoft_test.go
```diff
@@ -101,6 +102,67 @@ func TestUserIdentityFromGraphAPI(t *testing.T) {
...
+
+	identity, err := c.HandleCallback(connector.Scopes{Groups: false, OfflineAccess: true}, req)
+	expectNil(t, err)
+	expectEquals(t, identity.Username, "Jane Doe")
+	expectEquals(t, identity.UserID, "S56767889")
+	expectEquals(t, identity.PreferredUsername, "")
+	expectEquals(t, identity.Email, "jane.doe@example.com")
+	expectEquals(t, identity.EmailVerified, true)
+	expectEquals(t, len(identity.Groups), 0)
+
```
hur, line 155: I didn't add a check for `expectEquals(t, identity.ConnectorData, expectedConnectorData)` (like in gitea_test.go, for example) as I wasn't sure if there's a good way to mock the Expiry time.

ollie-kane-CB:
Hit this in production today. Setup: `microsoft` connector with `emailToLowercase: true`, ID token consumed by the EKS API server (`oidc.usernameClaim=email`, `oidc.usernamePrefix=dex:`), and a `kind: User` ClusterRoleBinding listing the lowercase subject.

Initial login behaves correctly — `HandleCallback` lowercases. Every silent refresh afterward emits the raw-cased `userPrincipalName` from Graph, so the kubectl call's User subject flips to upper/mixed case and silently misses the binding. The user goes from cluster-admin to "forbidden" with nothing visible at the connector layer. EKS RBAC's case-sensitive subject match is the trap — we'd accumulated a stale uppercase-subject entry in our `dex-admins` CRB before realizing what was actually happening.

The diff here (4 lines in `Refresh`, mirroring `HandleCallback`) resolves it cleanly. We've been running an equivalent patch as a fork on top of v2.45.1 since today; kubectl calls riding the silent-refresh path now keep the lowercase casing.

Any chance of getting a reviewer on this? Happy to help shepherd — including adding a Refresh-path table test if that's a blocker (we wrote one for our fork and it's straightforward against the existing `newTestServer` harness).

PR #3700, opened 2024-08-16
chore: update nix-direnv from 1.5.0 to 3.0.5

hur:
#### Overview

Update `nix-direnv` in the `.envrc` file to the newest version.

The current nix-direnv version has issues on MacOS due to BSD sed differing from GNU sed. This has been fixed in nix-direnv https://github.com/nix-community/nix-direnv/issues/120 
Updating to the newest nix-direnv version fixes this issue.


sagikazarmark (approved):

hur:
Added sign-off to commits for DCO

PR #3701, opened 2024-08-17
Add retry mechanism for storage initialization

MoeBensu:
- Implement retry logic for storage connection attempts
- Add configuration options for retry attempts and delay
- Default to 5 attempts with 5 second delay if not specified


MoeBensu:
@nabokihms I would like to request a review here, if possible! Or how do you like to plan this in the next release?

nabokihms (changes requested):

Issue #1547, opened 2019-09-18
TOTP support


ravjeetsinghb:
I could not find TOTP integration for dex? Does it support or can the feature be included in future releases?

Thanks

PR #1270, opened 2018-08-06
Add One Time Password support.

kepkin:

Issue #352, opened 2016-03-01
Two-factor authentication


ericchiang:
We don't have an immediate need for this but it would be a very cool thing to add.

Perhaps have some way of integrating with the local connector?

ibotty:
I am interested in using u2f with dex. There are some notes on using u2f with OIDC from yubico: https://www.yubico.com/wp-content/uploads/2015/08/Yubico-U2F-and-OIDC-Final.pdf

If given guidance, I might be able to devote some time to it.

bobbyrullo:
@ibotty : that would be wonderful. 

I agree with @ericchiang that it sounds like a local connector addition. Consider the different deployment options: requiring 2-factor for all users, user's choice, or not enabled. Also consider how general does this need to be: if it works with yubikey will it work with other vendors?

The way to proceed is to post a proposal here and once we reach agreement you can move forward with an implementation. If you need more guidance don't hesitate to ask!

ibotty:
U2f is a vendor-neutral standard. It would also work with cheap 5$ dongles.

I will try to do a proposal draft next week.

ibotty:
I did not get to it a few weeks ago. Sorry. I don't know, when I will have time. So nobody should be discouraged to work on it!

wyattanderson:
Would be awesome to see U2F support, but we'd also love to have [TOTP](https://en.wikipedia.org/wiki/Time-based_One-time_Password_Algorithm). We use the LDAP connector, though, so if it could be configurable to work with connectors that don't already provide 2FA, that would be neat.

remohammadi:
I'm also interested to work on this.

To implement it as a local connector addition, I think we need to rethink about the connector config format. Because in the current format, connectors are ORed with each other (authorizing through one of them is sufficient). I can think of these options:
* To make it possible to define a pipeline of connectors in the config. (overkill?)
* Another flag for dex-worker, `-2fa-connectors` for example, to set a list of connectors as the second authenticator.
* Have a combined connector (local+TOTP for example)

Which option does make sense to be the base for the proposal?

ericchiang:
Going to copy and past a comment I added over in #763 

> Might be something that can be implemented out of scope of dex. E.g. user logs into app through dex, then when they try to do something sensitive the app itself triggers a second factor.
>
> When to trigger a second factor is hard to express in a general way.

[...]

PR #3712, opened 2024-08-26
Two-Factor authentication (TOTP)

nabokihms:
#### Overview
This pull request implements two-factor authentication (2FA) in Dex. The 2FA data is securely stored within the `OfflineSessions` object, enhancing security for connectors that lack built-in 2FA support, such as LDAP and local connectors. Upon first login, users will save their 2FA settings using a QR code, after which they will use the saved 2FA for subsequent logins. Below is an example configuration for enabling 2FA:

```yaml
# Configuration for the two-factor authentication
twoFactorAuthn:
  issuer: "dex"
  connectors:
    - mock
```

Enhancing Dex with 2FA adds an additional layer of security, making unauthorized access significantly more difficult. This is particularly valuable for connectors like LDAP and local connectors that do not inherently support 2FA. By implementing 2FA, we align Dex with industry best practices for identity management, meet higher security compliance requirements, and ensure better protection for user data, thereby building greater trust with our users.

image

image


nabokihms:
closes https://github.com/dexidp/dex/issues/352

nabokihms:
closes https://github.com/dexidp/dex/issues/1547

nabokihms:
closes https://github.com/dexidp/dex/pull/1270

sambonbonne:
I'm not a maintainer nor a reviewer so I'm not sure this is the best play to ask this, but would it be possible to display the "textual" code below the QR code?

This is useful when you can't scan the QR code, for example when your TOTP application is directly on your computer.

lanord:
Do we know if we are going to see this merge in the near future ? This would be a great feature to see deployed. As more and more security requirement ask for 2FA on auth provider.

nabokihms:
@sambonbonne good addition, thanks!
@lanord I am willing to merge it this year after figuring out some API nuances.

Kludex:
@nabokihms Is there a way to fund the work in this PR?

shaohme:
Nice feature to have indeed.

Have you considered supporting slapo-otp(5) when you started the project?

thebasementcoders:
Hi, is TOTP still planned for this year? My team would like to use this amazing feature with local connectors. :pray: 

A small update: I've taken a look at the PR and noticed that `totp.Generate` is always called with an empty secret option and generally it is good to have it randomized, but for simpler use cases (e.g. in-memory storage), it might be worth supporting [a static secret](https://github.com/pquerna/otp/blob/v1.5.0/totp/totp.go#L186-L188) in order to keep `offlineSessions.TOTP` idempotent between restarts. Kind of similar to how static passwords are already handled in Dex.

braginini:
Would love to see this one merged as we are planning to integrate Dex in [NetBird](https://github.com/netbirdio/netbird).

jjzazuet:
Super cool feature right here. Any chances of merging? Thanks!

nabokihms:
For all waiting the feature. We are working now on the proposal, which will make this PR be merged in the future.

You can see the progress here https://github.com/dexidp/dex/issues/4560

server/handlers.go
```diff
@@ -760,6 +760,36 @@ func (s *Server) finalizeLogin(ctx context.Context, identity connector.Identity,
...
+	}
+	if len(mfaChain) > 0 {
+		// Redirect to MFA verification for the first authenticator in the chain.
+		v := url.Values{}
+		v.Set("req", authReq.ID)
+		v.Set("hmac", hmacParam)
+		v.Set("authenticator", mfaChain[0])
+		returnURL := path.Join(s.issuerURL.Path, "/totp/verify") + "?" + v.Encode()
+		return returnURL, false, nil
+	}
```
Jabejixo, line 816: The naming suggests a "chain" of multiple authenticators, but only `mfaChain[0]` is ever used. There's no logic to proceed to subsequent authenticators after the first one completes. The comment "first authenticator in the chain" implies there should be a second, third, etc

server/mfa.go
```diff
@@ -0,0 +1,295 @@
...
+				return
+			}
+
+			secret = &storage.MFASecret{
+				AuthenticatorID: authenticatorID,
+				Type:            "TOTP",
+				Secret:          generated.String(),
+				Confirmed:       false,
+				CreatedAt:       s.now(),
+			}
```
Jabejixo, line 149: Unconfirmed MFA secrets are stored but never cleaned up. If a user starts enrollment multiple times without completing it, stale secrets accumulate

server/mfa.go
```diff
@@ -0,0 +1,295 @@
...
+		if err != nil {
+			s.logger.ErrorContext(ctx, "failed to load TOTP key", "err", err)
+			s.renderError(r, w, http.StatusInternalServerError, "Internal server error.")
+			return
+		}
+
+		if !totp.Validate(code, generated.Secret()) {
+			s.renderTOTPPage(secret, true, totpProvider.issuer, authReq.ConnectorID, w, r)
+			return
+		}
```
Jabejixo, line 188: With a 30-second validity window, an attacker could potentially brute-force valid codes without any rate limiting. While Dex currently lacks rate limiting across all endpoints, TOTP is particularly vulnerable due to its limited keyspace.

server/mfa.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// getConnectorType returns the type of the connector with the given ID.
+func (s *Server) getConnectorType(connectorID string) string {
+	conn, err := s.storage.GetConnector(context.Background(), connectorID)
+	if err != nil {
+		return ""
+	}
+	return conn.Type
+}
```
Jabejixo, line 364: The parent function `mfaChainForClient` already receives a `ctx context.Context` parameter and uses it for `GetClient`. However, `getConnectorType` ignores it and creates a new `context.Background()`.

t
```diff
@@ -0,0 +1,5 @@
+export DEX_POSTGRES_DATABASE=dex
+export DEX_POSTGRES_USER=postgres
+export DEX_POSTGRES_PASSWORD=postgres
+export DEX_POSTGRES_HOST=0.0.0.0
+export DEX_POSTGRES_PORT=5432
```
Jabejixo: the remainder of the local testing?

server/handlers.go
nabokihms: Yeah, forgot about this part, it was a stab code. Will fix it.

server/mfa.go
nabokihms: Good point.

server/mfa.go
nabokihms: Yeah, good point. The only way for now is to use external rate limiters. I will add a comment to the code about it.

server/mfa.go
nabokihms: My bad, will fix

t
nabokihms: oh yeah this is my test file, thanks

server/mfa.go
nabokihms: I checked the code, and it seems like a common problem. If we often change MFa ids - we accumulate codes for old mfas. If we change client ids - we accumulate refresh tokens for them. 

It should be a separate issue to clean stale map entries in the storage, I believe.

server/mfa.go
```diff
@@ -0,0 +1,354 @@
...
+		provider, ok := s.mfaProviders[authID]
+		if ok && provider.EnabledForConnectorType(connectorType) {
+			chain = append(chain, authID)
+		}
+	}
+	return chain, nil
+}
+
+// getConnectorType returns the type of the connector with the given ID.
+func (s *Server) getConnectorType(ctx context.Context, connectorID string) string {
```
sagikazarmark: Shouldn't this return an error?

Transient database errors currently return an empty string as well, causing EnabledForConnectorType to return false even when MFA is enabled for that connector.

server/server.go
```diff
@@ -537,6 +548,7 @@ func newServer(ctx context.Context, c Config) (*Server, error) {
 	// "authproxy" connector.
 	handleFunc("/callback/{connector}", s.handleConnectorCallback)
 	handleFunc("/approval", s.handleApproval)
+	handleFunc("/totp/verify", s.handleMFAVerify)
```
sagikazarmark: Shouldn't this be `/mfa/verify` instead? The PR claims potential future support for WebAuthn.

server/mfa.go
```diff
@@ -0,0 +1,354 @@
...
+				authReq, err = s.storage.GetAuthRequest(ctx, authReq.ID)
+				if err != nil {
+					s.logger.ErrorContext(ctx, "failed to get auth request", "err", err)
+					s.renderError(r, w, http.StatusInternalServerError, "Login error.")
+					return
+				}
+				s.sendCodeResponse(w, r, authReq)
+				return
+			}
+		}
```
sagikazarmark: Isn't this already duplicated somewhere? Can we reuse?

web/templates/totp_verify.html
```diff
@@ -0,0 +1,36 @@
...
+  {{ if not (eq .QRCode "") }}
+  <p>Scan the QR code below using your authenticator app, then enter the code.</p>
+  <div style="text-align: center; margin: 1em 0;">
+    <img src="data:image/png;base64,{{ .QRCode }}" alt="QR code" width="200" height="200"/>
+  </div>
+  {{ else }}
+  <p>Enter the code from your authenticator app.</p>
+  <p class="theme-heading">{{ .Issuer }}</p>
+  {{ end }}
+  <form method="post" action="{{ .PostURL }}">
```
sagikazarmark, line 14: I know we don't do that in other places, but this is a security sensitive form that would probably benefit from CSRF.

sagikazarmark (commented):
I had some comments, but it probably needs a closer look to see if they are valid, hence just comments.

server/server.go
nabokihms: I thought about making it a separate path, but this is ok, we can use it for WebAuthn also

web/templates/totp_verify.html
nabokihms: There is a kind of CSRF protection using HMAC. It's generated server-side with a secret key and validated on submission, so a forged request from an external origin would fail verification. Same principle as a synchronizer token, just delivered via the URL rather than a hidden form field.

We use it all over critical forms. I can open an issue to add a different CSRF mechanism, but not in the scope of this PR.

server/mfa.go
nabokihms: Replaced with a helper

server/server.go
nabokihms: Renamed to mfa

server/mfa.go
nabokihms: Added an error

storage/storage.go
```diff
@@ -330,12 +337,22 @@ type RefreshTokenRef struct {
 	LastUsed  time.Time
 }
 
+// MFASecret stores the enrollment state and secret for an MFA authenticator.
+type MFASecret struct {
+	AuthenticatorID string    `json:"authenticatorID"`
+	Type            string    `json:"type"`
+	Secret          string    `json:"secret"`
```
AlwxSin, line 346: Secret stores string without encryption. I don't think that it's a Dex problem but we should mention it somewhere in docs.

server/mfa.go
AlwxSin: Maybe we can use `ui.BlockedUntil` with 1 second window for now?

server/mfa.go
```diff
@@ -0,0 +1,363 @@
...
+
+		s.sendCodeOrRedirectToApproval(w, r, authReq, returnURL)
+
+	default:
+		s.renderError(r, w, http.StatusBadRequest, "Unsupported request method.")
+	}
+}
+
+func (s *Server) renderTOTPPage(secret *storage.MFASecret, lastFail bool, issuer, connectorID string, w http.ResponseWriter, r *http.Request) {
```
AlwxSin, line 251: I think we should add `Cache-Control: no-store` header here, so browser wouldn't cache the page.

server/mfa.go
```diff
@@ -0,0 +1,363 @@
...
+	if len(s.mfaProviders) == 0 {
+		return nil, nil
+	}
+
+	client, err := s.storage.GetClient(ctx, clientID)
+	if err != nil {
+		return nil, err
+	}
+
+	// nil means "not set" — fall back to default.
```
AlwxSin, line 334: This behavior is fragile. Due to serialization `null` and `[]string` between json, sql/k8s we can lose information. Maybe separate field like `MFAChainSet bool`?

server/mfa.go
```diff
@@ -0,0 +1,363 @@
...
+				s.logger.ErrorContext(ctx, "failed to store MFA secret", "err", err)
+				s.renderError(r, w, http.StatusInternalServerError, "Internal server error.")
+				return
+			}
+		}
+
+		s.renderTOTPPage(secret, false, totpProvider.issuer, authReq.ConnectorID, w, r)
+
+	case http.MethodPost:
+		// TODO(nabokihms): this endpoint should be proteted with a rate limit (like the auth endpoint).
```
AlwxSin:
```suggestion
		// TODO(nabokihms): this endpoint should be protected with a rate limit (like the auth endpoint).
```

server/mfa.go
```diff
@@ -0,0 +1,363 @@
...
+		// TOTP has a limited keyspace (6 digits) with a 30-second validity window,
+		// making it particularly vulnerable to brute-force without rate limiting.
+		// This endpoint should be protected similarly to the auth/login endpoints.
+		//
+		// For now the best way is to use external rate limitting solutions.
```
AlwxSin:
```suggestion
		// For now the best way is to use external rate limiting solutions.
```

server/mfa.go
```diff
@@ -0,0 +1,363 @@
...
+		if ok && provider.EnabledForConnectorType(connectorType) {
+			chain = append(chain, authID)
+		}
+	}
+	return chain, nil
+}
+
+// getConnectorType returns the type of the connector with the given ID.
+func (s *Server) getConnectorType(ctx context.Context, connectorID string) (string, error) {
+	conn, err := s.storage.GetConnector(ctx, connectorID)
```
AlwxSin: Shouldn't we use `s.getConnector` to avoid extra DB calls?

cmd/dex/serve.go
```diff
@@ -759,3 +762,24 @@ func loadTLSConfig(certFile, keyFile, caFile string, baseConfig *tls.Config) (*t
...
+		switch auth.Type {
+		case "TOTP":
+			var cfg TOTPConfig
+			if err := json.Unmarshal(auth.Config, &cfg); err != nil {
+				logger.Error("failed to parse TOTP config", "id", auth.ID, "err", err)
+				continue
+			}
+			providers[auth.ID] = server.NewTOTPProvider(cfg.Issuer, auth.ConnectorTypes)
+			logger.Info("MFA authenticator configured", "id", auth.ID, "type", auth.Type)
+		}
```
AlwxSin, line 838: Consider adding log entry about unknown types

server/templates.go
```diff
@@ -325,6 +328,21 @@ func (t *templates) approval(r *http.Request, w http.ResponseWriter, authReqID,
...
+	if lastWasInvalid {
+		w.WriteHeader(http.StatusUnauthorized)
+	}
+	data := struct {
+		PostURL   string
+		Invalid   bool
+		Issuer    string
+		Connector string
+		QRCode    string
+		ReqPath   string
```
AlwxSin, line 354: Couldn't find `ReqPath` in template. Is it used somewhere? Or is it needed?

nabokihms: Used here https://github.com/dexidp/dex/blob/master/web/templates/header.html so we need it for all pages

cmd/dex/serve.go
nabokihms: Ok, make sense

server/mfa.go
nabokihms: Yes, why not

server/mfa.go
nabokihms: It is destinction between set and unset, I think it is fine, we use the same logic in other places.

i don't like two fields because.

Normal, I can understand this
```
mfaChainSet: true
defaultChainSet: []
```

Ignored, because mfaChainSet is not set
```
defaultChainSet: []
```

Error? Cannot be set and not provided.
```
mfaChainSet: true
```

since it works (but is fragile) i'd like to leave it as is, just interface-wise

storage/storage.go
nabokihms: Added a comment, will add to the doc in the future

server/mfa.go
nabokihms: No objections, added

server/mfa.go
nabokihms: Added a type to the server connector and switched to the `s.getConnector`

cmd/dex/serve.go
nabokihms: Added

AlwxSin (approved):

nabokihms:
No API to recover mfa and no recovery codes for now, but I still think it is ok to be merged now. We will add the API later.

PR #3770, opened 2024-09-30
fix unmarshaling of expanded environment variables with special characters

### LINKED ISSUE #3689: When using DEX_EXPAND_ENV, environment values with " or \ break the resulting JSON
### Preflight Checklist

### Version

2.29.0, main, doesn't matter

### Storage Type

etcd

### Installation Type

Binary

### Expected Behavior

User can supply password (or other config via environment variable, while using DEX_EXPAND_ENV) in the config YAML, and the values would be safely converted into JSON.

### Actual Behavior

Unmashaling JSON converted from connector YAML config fails:

```console
=== RUN   TestUnmarshalConfigWithEnvExpand
    config_test.go:445: failed to decode config: error unmarshaling JSON: parse connector config: invalid character 'd' after object key:value pair
```

### Steps To Reproduce

1. Apply the following diff to `config_test.go`:

```diff
diff --git a/cmd/dex/config_test.go b/cmd/dex/config_test.go
index c6d37cb0..87a78f80 100644
--- a/cmd/dex/config_test.go
+++ b/cmd/dex/config_test.go
@@ -273,7 +273,7 @@ func checkUnmarshalConfigWithEnv(t *testing.T, dexExpandEnv string, wantExpandEn
 	os.Setenv("DEX_FOO_USER_PASSWORD", "$2a$10$33EMT0cVYVlPy6WAMCLsceLYjWhuHpbz5yuZxu/GAFj03J9Lytjuy")
 	// For os.ExpandEnv ($VAR -> value_of_VAR):
 	os.Setenv("DEX_FOO_POSTGRES_HOST", "10.0.0.1")
-	os.Setenv("DEX_FOO_OIDC_CLIENT_SECRET", "bar")
+	os.Setenv("DEX_FOO_OIDC_CLIENT_SECRET", `abc"def\ghi`)
 	if dexExpandEnv != "UNSET" {
 		os.Setenv("DEX_EXPAND_ENV", dexExpandEnv)
 	} else {
```

2. It will fail in `make test` with:

```console
=== RUN   TestUnmarshalConfigWithEnvExpand
    config_test.go:445: failed to decode config: error unmarshaling JSON: parse connector config: invalid character 'd' after object key:value pair
```

3. The root cause is in function at cmd/dex/config.go:341. The call to os.ExpandEnv on L362 is unaware of the JSON context in which the variables are being expanded, and has a comment about it already. JSON enforces that values are enclosed in quotes, hence extra unescaped quotes or escape characters in the fields make the resulting JSON invalid.

### Additional Information

These JSON illegal characters can appear in passwords, so the real use-case this came up was LDAP connector `bindPW` field. OIDC secret is used for reproduction as triggering it has a single line diff.

### Configuration

```yaml
config_test.go has the config.
```

### Logs

```shell
=== RUN   TestUnmarshalConfigWithEnvExpand
    config_test.go:443: failed to decode config: error unmarshaling JSON: parse connector config: invalid character 'd' after object key:value pair
--- FAIL: TestUnmarshalConfigWithEnvExpand (0.00s)
```

tuminoid:
#### Overview

If we expand environment values directly with os.ExpandEnv() over whole config, we might end up in a situation where the environment variable has escape characters that break the resulting JSON, and unmarshalling fails. Instead of expanding the entire config with single call, we recurse through the config and expand the values in leaves one by one.

Closes #3689


tuminoid:
/cc @nabokihms 
Here is the PR for fixing the escaping of special characters from #3689 using the approach we agreed on.

Suggested label is kind/bug.

It should be noted that this may cause issues for existing configs that have passwords with escapes due previous implementation requiring them. This must not be many installations, but maybe good to note regardless.

nabokihms (commented):
LGTM! Could you also add it to the storage config?
https://github.com/dexidp/dex/blob/6c7766249fe9b7e0af8601f1ad178b9a454198f1/cmd/dex/config.go#L316

tuminoid:
> LGTM! Could you also add it to the storage config?
> 
> https://github.com/dexidp/dex/blob/6c7766249fe9b7e0af8601f1ad178b9a454198f1/cmd/dex/config.go#L316

I implemented this and then started adding test for it, but there doesn't seem to be any value that can actually hold escaped strings, like there is on connector side. PostgresHost seems to be only string there, and there is AFAIK no legit case of putting escapes in. If we fix the parsing, it means the parsed URL will be invalid as a host string, instead of parsing the config failing, thus deferring the inevitable failure from parsing to connecting. Which I'd say is not what we want, would you agree?

tuminoid:
> > LGTM! Could you also add it to the storage config?
> > https://github.com/dexidp/dex/blob/6c7766249fe9b7e0af8601f1ad178b9a454198f1/cmd/dex/config.go#L316
> 
> I implemented this and then started adding test for it, but there doesn't seem to be any value that can actually hold escaped strings, like there is on connector side. PostgresHost seems to be only string there, and there is AFAIK no legit case of putting escapes in. If we fix the parsing, it means the parsed URL will be invalid as a host string, instead of parsing the config failing, thus deferring the inevitable failure from parsing to connecting. Which I'd say is not what we want, would you agree?

Ah nevermind. I noticed the test config only tests Postgres related configs, but I see there is at least Etcd backend which has password option as well. I will implement the same for storage and add Etcd related config test.

nabokihms:
@tuminoid thank you for your patience

tuminoid:
> @tuminoid thank you for your patience

No worries, thanks for taking in the fix!

PR #3771, opened 2024-09-30
allow empty redirect uris for confidential clients

kcolford:
#### Overview

Allow confidential clients to not require pre-registering their redirect URIs.

The standards do not require that confidential clients have preregistered redirect URIs, that requirement is only for public clients. While confidential clients SHOULD use pre-configured redirect URIs, they are not essential their security. We also will not adopt wildcards in redirect URIs as per #448 since that would be an explicit violation of the standard.

### DIFF
--- a/server/oauth2.go
+++ b/server/oauth2.go
@@ -477,6 +477,55 @@ func (s *Server) parseAuthorizationRequest(r *http.Request) (*storage.AuthReques
 		codeChallengeMethod = codeChallengeMethodPlain
 	}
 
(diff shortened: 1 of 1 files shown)


kcolford:
@nabokihms for review

TomHellier:
Just chiming in to say that this is a really useful addition to dex, and its a shame the previous PRs:
* #448 
* #1783  

were all closed without merging

kcolford:
@sagikazarmark could you please review this PR?

PR #3778, opened 2024-10-03
feat: also allow localhost equivalent IP addresses

dsonck92:
#### Overview

Instead of only checking for "localhost", also validate through `net.ParseIP` + `IsLoopback` whether the host is numerically localhost

Currently, `localhost` is already used as a special case that allows any port and http scheme if no particular redirectURIs are configured. When integrating with ownClouds desktop client, it actually sends an arbitrary `http://127.0.0.1:xyz` redirect uri. Considering that the go standard library already has the functionality to check if a given IP is a loopback address, we can extend the `== "localhost"` with also a check for `IsLoopback()` which should allow the ownCloud desktop client to be accepted.

### DIFF
--- a/server/oauth2.go
+++ b/server/oauth2.go
@@ -668,7 +668,8 @@ func validateRedirectURI(client storage.Client, redirectURI string) bool {
 		return true
 	}
 
-	// verify that the host is of form "http://localhost:(port)(path)" or "http://localhost(path)"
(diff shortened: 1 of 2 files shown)


nabokihms (approved):

nabokihms:
Great improvement! Thanks.

Issue #1405, opened 2019-02-15
Missing "email_verified" in OIDC claim


framillien:
Hello,

When an OIDC provider return claim with "email" but without "email_verified" field, this is assumed to be false by Dex, and newly generated claim is with "email_verified=false".

Some enterprise providers return claims without "email_verified", when they had no usage of emails verification in enrollement process. Like Okta, CloudFoundry, ...

When this field is missing, we need to assume it to a successfull validation instead of validation error - or at least let it unchanged (missing) in generated claim.

In my case, OAuth2_proxy talking directly to my provider (Okta) works (Oauth2_proxy assume a missing field is valid), but when I plug Dex between OAuth2_proxy and my provider (to allow multiple providers) authentication fail.

Some links:
- PR for optional "email_verified" on kubernetes: https://github.com/kubernetes/kubernetes/issues/59496
- OAuth2_proxy handle email_verified has optional: https://github.com/pusher/oauth2_proxy/blob/master/providers/oidc.go#L119

I do a trivial patch in OIDC connector to workaround this (if verified is missing then verifie is True), following Kubernetes & OAuth2Proxy practices. I can post it if you want.

ericchiang:
> Some enterprise providers return claims without "email_verified", when they had no usage of emails verification in enrollement process. Like Okta, CloudFoundry, ...

Setting email_verified to false seems like the correct behavior in these cases. I realize it's incontinent, but being able to fake an email is a serious security issue, so I don't think it's appropriate to assume email_verified=True

framillien:
Hello, providers needing emails validation set this field. It's only missing if email validation has no meaning for this enrollement process (like employee enrollement). I link upper same discution in kubernetes connector (I assume they need a good security too ;) )

But yes, maybe it's better to add a config parameter to let choose by oidc provider ?

ericchiang:
A InsecureSkipEmailVerified config option seems fine

framillien:
I think more about changing default value rather than skipping it. 

defaultEmailVerified={Value of email_verified if missing in OIDC claim}

?

ericchiang:
This is now merged https://github.com/dexidp/dex/pull/1417

PR #3811, opened 2024-10-25
feat(oauth): add new insecureSkipEmailVerified property

dariusbakunas:
#### Overview

Same option as it was added for oidc connector here: https://github.com/dexidp/dex/pull/1417

Reasons are exactly the same is in https://github.com/dexidp/dex/issues/1405, we are using Oauth connector instead, which currently lacks this option. Our internal IDP does not have 'email_verified' property on /oauth/v1/userinfo endpoint.

Documentation update here: https://github.com/dexidp/website/pull/188


cardoe (approved):

cardoe:

dariusbakunas:
Great! Will test it soon, if not today, will try to do it early next week, thank you!

dariusbakunas:
Did anything else change from 2.41 to 2.44 with oauth connector? For some reason when I switch to this image, I get "Bad Request: Resource does not exist" when my browser is redirected from `/auth/..` to `/auth/oracleoauth..` where oracleoauth is my oauth connector id.

cardoe:
> Did anything else change from 2.41 to 2.44 with oauth connector? For some reason when I switch to this image, I get "Bad Request: Resource does not exist" when my browser is redirected from `/auth/..` to `/auth/oracleoauth..` where oracleoauth is my oauth connector id. 

Sounds like the connector id isn't defined?

dariusbakunas:
Hm, but it is, I did not change my configs, we used our custom build image with my change in and it was working. As soon as I switch to this image I get that error, I will poke around

PR #3818, opened 2024-10-30
add hidden prop to connector config

johnvan7:
#### Overview

Added config property to hide a connector in the login web page, for example if it should be used only for m2m.

As requested in #3145 and #4231

Adds connector's config property "hidden" (default false).

#### Related
Discussion: Can I hide a connector? #3145


cardoe (approved):
This would be a great addition.

cardoe:
@nabokihms and @sagikazarmark what do you think of this?

cardoe:

johnvan7:
Hi @cardoe, I've tested ghcr.io/cardoe/dex:v2.44.90 with the connector property "hidden: true" and it works as expected!

storage/storage.go
```diff
@@ -374,6 +374,8 @@ type Connector struct {
 	// However, fixing this requires migrating Kubernetes objects for all previously created connectors,
 	// or making Dex reading both tags and act accordingly.
 	Config []byte `json:"email"`
+	// It specifies if the connector should be hidden in the login web page.
+	Hidden bool `json:"hidden"`
```
sagikazarmark, line 378: Unfortunately, this isn't enough for the storage layer.

Try adding it to this test: https://github.com/dexidp/dex/blob/master/storage/conformance/conformance.go#L616

Conformance tests will fail and you will see that you need to change.

Generally, any schema based storage will have to be updated as well.

johnvan7: Hi @sagikazarmark, I updated the db schema but the tests still fail due to this issue #4461

0x2b3bfa0:
@sagikazarmark, since https://github.com/dexidp/dex/issues/4461 was resolved and https://github.com/dexidp/dex/pull/3818/commits/9f5271bf184912f9f1326d837c797cad45ec2d5c pushed, is this pull request worth another review? 😇

Issue #2504, opened 2022-05-05
Support the configuration of additional auth request parameters on OIDC connector


roddyherries:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

Some IdPs support login features that are accessible via additional parameters on an authorization request. The current OIDC connector implementation provides no mechanism to configure additional authorization request parameters and hence the IdP specific features are out of reach when using DEX.

For example, when authorizing with Auth0 a client can pass an `organisation` parameter to tell Auth0 which organization the user must authenticate via:

```
HTTP/1.1 302 Found
  Location: https://server.example.com/authorize?
    response_type=code
    &scope=openid profile email
    &client_id=1234
    &state=abcde
    &redirect_uri=https://client.example.org/callback
    &organisation=my_org
```

### Proposed Solution

Extend the existing OIDC configuration with new map element `additionalAuthRequestParams`. 

```
connectors:
- type: oidc
  id: oidc
  name: OIDC
  config:
    … existing config 
    additionalAuthRequestParams:
      paramKey1: value1
      paramKey2: value2
```

Enhance `LoginURL` to add these parameters to the auth request:

```
if len(c.additionalAuthRequestParams) > 0 {
    for k, v := range c.additionalAuthRequestParams {
	opts = append(opts, oauth2.SetAuthURLParam(k, v))
    }
}
```

nabokihms:
It looks like an excellent addition to the OIDC connector! Thanks for the idea.

The only concern about implementation is that this feature requires splitting command line parameters into two lists: params maintained by dex and additional params. 
* Dex should validate that provided additional auth request params do not interfere with the first list.
* It should be covered by tests that Dex itself uses only parameters from the first list.

roddyherries:
Hi @nabokihms - yes, agreed on that additional check for splatting standard OIDC params. I've implemented this on a fork for our own internal use for now - I'll raise a PR if that helps?

nabokihms:
Yeah, sure. I think it is worth seeing the code and reviewing it.

roddyherries:
Hi @nabokihms - could you clarify something please ...

[...]

PR #3831, opened 2024-11-07
Support auth params

a-nych:
#### Overview

This PR is a refresh of https://github.com/dexidp/dex/pull/2631 by @roddyherries.

Some IdPs support login features that are accessible via additional parameters on an authorization request. The current OIDC connector implementation provides no mechanism to configure additional authorization request parameters and hence the IdP specific features are out of reach. This PR adds support to for such additional params to the OIDC connector.

Adds a new "additionalAuthRequestParams" map to the OIDC Config struct
Populates the map from config at connector "open"
Adds any additional auth params to the auth LoginURL
Prevents mis-configuration attempts to override OIDC specified auth params.

Issue: https://github.com/dexidp/dex/issues/2504


mardonner:
Hi, so, what's the current state of this? Also interested in this. If I understood correctly, the suggested changes from the review in #2631 are applied in here? What would be the next steps? Rebase, fix linter issues, request review?

PR #3857, opened 2024-11-21
Generate access tokens for implicit & hybrid flows only when needed

mfila:
#### Overview

Avoid access token generation when `response_type` is either `id_token` (for implicit flow) or `code id_token` (for hybrid flow).

This PR aims to make dex better follow what is stated in the following sections of the OpenID Connect Core1.0 specification:

- [3.2.2.5. Successful Authentication Response](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse) (for implicit flow)
- [3.3.2.5. Successful Authentication Response](https://openid.net/specs/openid-connect-core-1_0.html#HybridAuthResponse) (for hybrid flow)

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -717,8 +717,6 @@ func (s *Server) sendCodeResponse(w http.ResponseWriter, r *http.Request, authRe
 			}
 		case responseTypeToken:
 			implicitOrHybrid = true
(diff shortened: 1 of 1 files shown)


nabokihms (approved):
@mfila sorry for the long delay, looks good.

PR #3868, opened 2024-12-04
feat: set resource revision for connectors

### LINKED ISSUE #3855: Updates to connectors via API are not reflected at runtime
### Preflight Checklist

### Version

2.41.1

### Storage Type

etcd

### Installation Type

Official Helm chart

### Expected Behavior

Updating an existing connector via the API should reflect the changes at runtime.

### Actual Behavior

The config is not reloaded at runtime and changes are only reflected on a restart.

### Steps To Reproduce

1. Start dex
2. Use a storageConnector (i.e. one that has been configured using the API) to log into an application
3. Use the API to update a field of the connector, such as scope or insecureSkipEmailVerified
4. Use the storageConnector again. The changes made in step 3 will not be reflected
5. Restart dex
6. The changes are now reflected

### Additional Information

This is highly likely due to the fact that the connector's `ResourceVersion` field is not incremented on changes to the connector. We checked etcd directly and found `ResourceVersion` on the connector to be an empty string.

The update function that sets updated fields can be seen here, the `ResourceVersion` field is not touched: https://github.com/dexidp/dex/blob/master/server/api.go#L466

In `UpdateConnector`, `SetResourceVersion` is called, apparently expecting that the field was set before as part of the updater function: https://github.com/dexidp/dex/blob/master/storage/ent/client/connector.go#L76

Lastly, `Server.getConnector` checks this field to determine if a connector needs to be reloaded, which will always end up comparing the empty string against the empty string: https://github.com/dexidp/dex/blob/master/server/server.go#L734

In this related issue, seemingly the same issue was fixed for the Kubernetes storage by relying on the `ObjectMeta.ResourceVersion` field instead: https://github.com/dexidp/dex/issues/1672

### Configuration

_No response_

### Logs

_No response_

nrwiersma:
#### Overview

This PR adds resource versions to connectors that are API controlled but default. It does not require the Connector to use the resource revision, so that Kubernetes still functions. 

This fixes an issue with connectors where updates to a connector via the API does not get applied to the runtime use of the connector. This seems to have been previously fixed for Kubernetes specifically, but this is a more general fix.

Fixes #3855

I see no obvious way to test this, but if I am missing something, please let me know.

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"strconv"
 
 	"golang.org/x/crypto/bcrypt"
 
@@ -430,10 +431,11 @@ func (d dexAPI) CreateConnector(ctx context.Context, req *api.CreateConnectorReq
 	}
 
 	c := storage.Connector{
-		ID:     req.Connector.Id,
-		Name:   req.Connector.Name,
-		Type:   req.Connector.Type,
-		Config: req.Connector.Config,
+		ID:              req.Connector.Id,
+		Name:            req.Connector.Name,
+		Type:            req.Connector.Type,
+		ResourceVersion: "1",
+		Config:          req.Connector.Config,
 	}
 	if err := d.s.CreateConnector(ctx, c); err != nil {
 		if err == storage.ErrAlreadyExists {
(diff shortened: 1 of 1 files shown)


nabokihms:
I didn't find how it fixes the original problem, but still think that the change is ok.

nrwiersma:
> I didn't find how it fixes the original problem, but still think that the change is ok.

The issue is that updates to a Connector in the API are ignored because the ResourceVersion, which is empty, does not change, and therefore needs a restart of Dex to get the updated Connector. By adding and incrementing the ResourceVersion field, this solves the observed issue.

nabokihms (approved):

0x0dr1y:
Hey there @sagikazarmark @nabokihms,

can you let us know when we can expect the next Dex release?
Thanks!

PR #3874, opened 2024-12-05
Microsoft connector: add support for Federated Identity to avoid secrets

### LINKED ISSUE #3852: Microsoft connector - support for using signed assertions from managed identity instead of client secrets
### Preflight Checklist

### Problem Description

Using Dex with the Microsoft connector with an Entra ID (Azure AD) app requires setting a client secret in the configuration. Applications are generally discouraged to work with pre-shared keys (like client secrets) whenever possible, as they can pose a security risk.

Apps that are running in Azure, or in environments that can be federated with Entra (e.g. [Workload Identity Federation](https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation), that works on Kubernetes clusters hosted anywhere, as well as a bunch of other places including Google Cloud, AWS, etc), can leverage managed identity / workload identity to perform a token exchange instead of requiring pre-shared credentials.

### Proposed Solution

Support for obtaining a token from Entra ID using signed assertions.

When invoking the `/token` endpoint to exchange an authorization code for an access token, it should be possible to use a [signed assertion](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-access-token-with-a-certificate-credential) instead of a client secret. The assertion can come from a certificate, or can come from a token issued by Managed Identity / Workload Identity.

### Alternatives Considered

_No response_

### Additional Information

This feature request would greatly improve the security of environments that use Dex, as they would be able to run without any pre-shared credentials, which are susceptible to being leaked.

ItalyPaleAle:
Fixes #3852

Maintaining shared secrets in the application's configuration is not good, and should be avoided as possible. This includes the client secret used by the Microsoft connector to exchange authorization codes for access tokens using Entra ID (fka Azure AD).

It is possible to avoid using secrets by leveraging a [signed assertion](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-access-token-with-a-certificate-credential), which comes from managed identity (when the app is running on Azure) or [Workload Identity](https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation) (when the app is running on a supported platform, including K8s anywhere).

This is configured with a new configuration key `federatedIdentity` which can have 3 values (all case-insensitive):

- `ManagedIdentity` -> Uses a system-assigned managed identity
- `ManagedIdentity=00000000-0000-0000-0000-000000000000` -> Uses a user-assigned managed identity, where the value is the client ID of the MI
- `WorkloadIdentity` -> uses Workload Identity

Note: it's very hard to write a CI test for this as it requires support on the infrastructure


richshadman:
@ItalyPaleAle @cardoe any chance on revisiting this? Having this option would make our Argo implementations much better - no more secret management for auth.

cardoe:
I don't have an environment to test. I'm not a maintainer, but I'm happy to review when it passes tests.

lepouletsuisse:
@ItalyPaleAle Any chance to fix the tests? I think Workload Identities would be a great addition to dex for Microsoft Azure's users

ItalyPaleAle:
@lepouletsuisse The "failed test" is that snyk detected a new dependency (I suppose), so not something I can fix.

I am honestly not going to be able to drive this PR forward, especially since it doesn't seem maintainers have shown much interest...

richshadman:
great news!

PR #3886, opened 2024-12-14
Add CSS for example app to make it prettier

nabokihms:
#### Overview

Add inline CSS code to make the example app looking less ugly and improve experience of users interacting with it.

image
image


KNIF:
Can someone please update the package "example-app" (ghcr.io/dexidp/example-app) to include these CSS changes?
The last version has been pushed 4 years ago.

PR #3950, opened 2025-01-27
Add authproxy preferred_username header

### LINKED ISSUE #3876: Authproxy should have a separate header for username and preferred_username
### Preflight Checklist

### Problem Description

According to https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims, user name is "End-User's full name in displayable form" (display name), and preferred_username is "Shorthand name by which the End-User wishes to be referred to at the RP, such as janedoe or j.doe. This value MAY be any valid JSON string including special characters such as @, /, or whitespace."

Even though preferred username can have special characters, some implementations don't handle this very well, and they have semantically different meanings (username vs display name).

### Proposed Solution

The [authproxy connector](https://github.com/dexidp/dex/blob/master/connector/authproxy/authproxy.go) should have separate headers for specifying these values, such as a `X-Remote-User-Name` header in addition to the existing `X-Remote-User`.

### Alternatives Considered

_No response_

### Additional Information

_No response_

kotx:
#### Overview

Allows proxies to pass `X-Remote-User-Name` header to set preferred username. Also adds tests.
Perhaps authproxy should be extended to allow any claims, instead of adding headers for preferred username, picture, etc? But that's out of scope for this PR.

Resolves #3876.

Defaults to the value of `X-Remote-User` ([user name](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims)), to preserve the existing default. But we may want to change this because preferred username = shorthand, and user name = display name.

### DIFF
--- a/connector/authproxy/authproxy.go
+++ b/connector/authproxy/authproxy.go
@@ -19,11 +19,12 @@ import (
 // Headers retrieved to fetch user's email and group can be configured
 // with userHeader and groupHeader.
 type Config struct {
-	UserIDHeader string   `json:"userIDHeader"`
-	UserHeader   string   `json:"userHeader"`
-	EmailHeader  string   `json:"emailHeader"`
-	GroupHeader  string   `json:"groupHeader"`
-	Groups       []string `json:"staticGroups"`
+	UserIDHeader   string   `json:"userIDHeader"`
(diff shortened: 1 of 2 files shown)


nabokihms (approved):

nabokihms:
Looks good and straightforward. Thanks!

PR #4000, opened 2025-02-20
Fix Makefile to Support Spaces in Paths

EthanDieterich:
#### Overview
In this branch, I fixed issues my team and I encountered with the Makefile, where it would not function correctly when spaces were present in the path. The issue was traced back to how `PATH` and `SHELL` were being set, leading to failures when running `make` in directories with spaces in their names.

- This PR ensures that paths are **properly quoted** in the Makefile to prevent issues caused by spaces in directory names.
- Updates how `SHELL` passes the `PATH` variable to ensure compatibility with `/bin/sh` while keeping the correct environment settings.
- This fix is necessary for users working in directories with spaces (e.g., `/home/user/My Project/`), where `make` previously failed due to improper path handling.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
 export PATH := $(abspath bin/protoc/bin/):$(abspath bin/):${PATH}
-export SHELL := env PATH=$(PATH) /bin/sh
 
 OS = $(shell uname | tr A-Z a-z)
 


jsquyres:
Can this PR be reviewed?  We have found that we need this for environments that have a PATH with a space.

Thanks!

sagikazarmark (approved):

sagikazarmark:
I wonder if this change is the reason why tests are failing

jsquyres:
> I'm guessing that we also need:

I added a commit to this PR that added the `chmod +x ...`.  Could you approve the github actions again to see if that problem at least got fixed?

I haven't had time to dig into the 2nd failure yet, but I offhand wonder if that failed because protoc previously didn't run...?

jsquyres:
Oops -- had the wrong `chmod` path in there.  Fixed.

jsquyres:
Sorry to bug -- could you authorize the github actions again?  Thanks!

jsquyres:
It took some digging, but I tracked down what was causing CI to fail, and re-evaluate the solution of this PR.  I have pushed a new commit with a full explanation.  Here's the commit message:

Makefile: Do not export SHELL
    
The intent for exporting the SHELL variable is to ensure that all
Makefile target comands are run with the amended PATH value that is
set via the "export PATH := ..." directive.
    
However, exporting PATH value is all that is necessary to ensure that
all target commands run with the amended PATH.  It is not necessary to
*also* export a SHELL command that explicitly sets the PATH value.
    
Specifically: setting SHELL to use "env" to set the PATH environment
variable gets difficult if the PATH includes whitespace.  make's
assignment right hand side parsing does not treat quotes as delimiters
of tokens.  For example, the following directive in a Makefile results
in the FOO variable having 3 tokens, not 2:
    
FOO := bar "baz yow"
    
Therefore, achieving the overall goal of amending the PATH for all
Makefile target commands -- even if the PATH includes whitespace -- is
best achieved simply by not exporting the SHELL.
    
Signed-off-by: Ethan Dieterich <ethandieterich@gmail.com>
Signed-off-by: Jeff Squyres <jsquyres@cisco.com>

sagikazarmark (approved):

sagikazarmark:
Congrats on the 4000th PR. 🎉

Issue #4025, opened 2025-03-05
LDAP Client Certificate Authentication is Ignored in Dex


crafa2:
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [x] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.42.0

### Storage Type

Kubernetes

### Installation Type

Official Helm chart

### Expected Behavior

If a client certificate is provided (clientCert and clientKey are set), Dex should use it for authentication instead of defaulting to an anonymous bind.
Dex should only attempt an anonymous bind if neither a client certificate nor bindDN/bindPW are provided.

### Actual Behavior

When a client certificate (clientCert and clientKey) is provided, Dex drops it before performing the LDAP query.
If bindDN and bindPW are not set, Dex defaults to an anonymous bind, ignoring the client certificate.
This results in failed authentication unless the LDAP server allows anonymous binds.
Even when a certificate is provided, there is no way to use it for authentication since Dex does not pass it along.

### Steps To Reproduce

1. Configure Dex with an LDAP connector that relies on **client certificate authentication**.  
2. Ensure no `bindDN` and `bindPW` are set, expecting the certificate to be used.  
3. Observe that Dex attempts an **anonymous bind** instead of using the client certificate.  
4. As a result, authentication fails unless anonymous binds are allowed on the LDAP server.

### Configuration

```yaml
issuer: https://rke2-controlplane:32000
    logger:
      level: debug
    storage:
      type: Kubernetes
      config:
        inCluster: true
    web:
      https: 0.0.0.0:5554
      tlsCert: /etc/dex/tls/tls.crt
      tlsKey: /etc/dex/tls/tls.key
    connectors:
    - type: ldap
      name: LDAP
      id: ldap
      config:
        host: ldap.google.com:636
        insecureSkipVerify: true
        
        # Certificats LDAP
        rootCA: /etc/dex/ldap/ldap_ca.crt
        clientCert: /etc/dex/ldap/client.crt
        clientKey: /etc/dex/ldap/client.key
        #No bindDN and no bindPW
        userSearch:
[...]

PR #4026, opened 2025-03-05
fix(ldap): Support client certificate authentication without fallback to anonymous bind

crafa2:
#### Overview

This PR fixes an issue where Dex incorrectly drops client certificates before performing an LDAP bind, leading to unintended anonymous authentication instead of certificate-based authentication.

- Ensures that if a client certificate is provided, Dex does not perform an anonymous bind that could override the certificate-based authentication.
- Simplifies the binding logic to make it more readable and maintainable.
- Addresses cases where `bindDN` and `bindPW` are empty while a certificate is provided.

Closes #[<4025>](https://github.com/dexidp/dex/issues/4025).

- This change assumes that if a client certificate is provided, it should take precedence over an anonymous or credential-based bind.
- The updated logic ensures that Dex behaves consistently when handling different authentication methods.
- Let us know if additional tests are required to validate different LDAP configurations.


nabokihms:
Is this a breaking change? We probably need a new test to cover this case.

crafa2:
This change **should not** be a breaking change for existing configurations because it **only modifies behavior when a client certificate is explicitly provided and no bindDN/bindPW is set**.  

- If **no certificate is provided**, Dex continues to behave as before:  
  - If `bindDN` and `bindPW` are set → Dex performs a credential-based bind.  
  - If `bindDN` and `bindPW` are empty → Dex falls back to an anonymous bind.  

- If **a certificate is provided**:  
  - If `bindDN` and `bindPW` are **also set**, Dex **still performs the credential-based bind**, keeping the previous behavior. 
  - If `bindDN` and `bindPW` are **empty**, Dex now **uses the certificate authentication instead of falling back to an anonymous bind** (which was unintended behavior before).  

So effectively, this fix just ensures that **when a client certificate is used, it actually gets used for authentication instead of being dropped**.

crafa2:
@nabokihms Do you want me to add a test to cover this case?

alexberry:
I would very much appreciate this being merged, as I am attempting to use google LDAP, which requires _either_ Certificates _or_ Certificates & BindDN to be used - this ticket appears to suggest that dex not support two levels of authentication for this, and so this merge would enable me to fix my issue by dropping bindDN

PR #4046, opened 2025-03-13
Microsoft Connector: Update group scope to groupMember.read.all

### LINKED ISSUE #3989: Microsoft Connector: Overly-permissive scope
Are we able to make this a more granular permission requirement as our team that manages Microsoft Entra is not comfortable delegating **Directory.Read.All** 

If it is only used for auth and group membership, are the following not sufficient?
 
* user.read
* group.read.all
* groupMember.read.all

References:

From the documentation (https://dexidp.io/docs/connectors/microsoft/) 
> when registering dex application on https://apps.dev.microsoft.com/ add an explicit Directory.Read.All permission to the list of Delegated Permissions

Coded here:

https://github.com/dexidp/dex/blob/487717d61e479b44a1fad5262ab617d6d2508f52/connector/microsoft/microsoft.go#L37-L39

ap0phi5:
#### Overview

This is a small change to reduce the default permission scope for a Microsoft connector from the overly-permissive `directory.read.all` down to `groupMember.read.all`.

Closes #3989


ap0phi5:
Anyone able to allow this to run the CI tests?

ap0phi5:
@nabokihms If it is within your purview, could you let the tests run on this please?
The details look a bit old but I'm assuming this tenant is still available?
https://github.com/dexidp/dex/blob/master/connector/microsoft/microsoft_test.go#L20-L23

ap0phi5:
Bump. Is there anything more I can do to help this PR be considered?

cardoe (approved):
Reading through what is needed @ap0phi5 is correct that the lesser permission is good enough.

sagikazarmark:
Looks like CI is stuck. In any case, I wonder if it would make sense to make this configurable. @cardoe is telling me that it requires reapproval which may be a problem for users.

We could leave the default the old value for a while and after a grace period change it to the new one.

sagikazarmark:
Actually, I've been thinking about something like Kubernetes' feature flag for a while now. This would be a prime candidate.

Geethree:
It would be really great to get this merged!

PR #4055, opened 2025-03-24
Add a SAML connector option to allows SAML request to be sent using RedirectBinding

Thanhphan1147:
#### Overview

Some IDP ( like Canonical's SSO ) systematically drops incoming POST requests. This PR adds the option for Dex to use HTTP redirect to send its SAML request, providing more flexibility.

This PR adds a new attribute `bindingType` to the SAML connector configuration, the default value is `post` so this won't change the behavior of existing configs. Updating this value to `redirect` will prompt Dex to send the SAML request via a HTTP Get redirect instead.

### DIFF
--- a/connector/connector.go
+++ b/connector/connector.go
@@ -81,7 +81,7 @@ type SAMLConnector interface {
 	//
 	// POSTData should encode the provided request ID in the returned serialized
 	// SAML request.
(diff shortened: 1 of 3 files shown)


Thanhphan1147:
Hi @nabokihms, can I ask you to take a look at this? thanks!

Issue #3849, opened 2024-11-15
Support LDAP login based on sAMAccountName instead of the email


anuviswa:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

Is there a way to login a user based on the sAMAccountName instead of the email id using the LDAP connector? There are many legacy customers who would like to logon users using their configured sAMAccountName. 

### Proposed Solution

Support LDAP login based on sAMAccountName in addition to email based login.

jsquyres:
We would be happy to create a PR for this functionality if the Dex community would be interested in it.  Should we do so?

@nabokihms @sagikazarmark Similar to https://github.com/dexidp/dex/pull/1058#issuecomment-2491635779, I'm pinging you because it looks like you both have recent commits in Dex, and you are both listed in [MAINTAINERS](https://github.com/dexidp/dex/blob/master/MAINTAINERS).

nabokihms:
Didn't get the cause of the issue. Could you please educate me what you want to change in Dex to make it work?

anuviswa:
Thanks @nabokihms for your response. 

Currently in the dex config file, the userSearch section can be configured with "username: mail" to search for the user based on the email. Similarly if "username: sAMAccountName" is specified, the user search will happen based on the sAMAccountName attribute. But there is no way to do the user search based on either mail OR sAMAccountName. 

Any filter specified in the "filter" section gets AND' with the username filter as seen here: https://github.com/dexidp/dex/blob/38309ed94d31a35a73e1377b63cac5f4faab6e0d/connector/ldap/ldap.go#L419. 

Could we add a special filter for username that allows OR: ||(<attr1>=<username>)(<attr2>=<username>)

PR #4061, opened 2025-03-29
feat(ldap): allow specifying multiple attributes on username input

yardenshoham:
#### Overview

In some use cases, it may be desirable to log in using either a username or an email address. Administrators no longer need to select a single field and can specify multiple fields to be recognized as "username. "

Changes the `userSearch.username` to a comma-separated list of attributes.

- Closes https://github.com/dexidp/dex/issues/3849

This change is backward-compatible. Tests have been added.

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -34,10 +34,10 @@ import (
 //       bindDN: uid=serviceaccount,cn=users,dc=example,dc=com
 //       bindPW: password
 //       userSearch:
-//         # Would translate to the query "(&(objectClass=person)(uid=<username>))"
+//         # Would translate to the query "(&(objectClass=person)(!(uid=<username>)|(mail=<username>)))"
 //         baseDN: cn=users,dc=example,dc=com
 //         filter: "(objectClass=person)"
-//         username: uid
+//         username: uid,mail
 //         idAttr: uid
 //         emailAttr: mail
 //         nameAttr: name
@@ -108,8 +108,8 @@ type Config struct {
 		// Optional filter to apply when searching the directory. For example "(objectClass=person)"
 		Filter string `json:"filter"`
 
-		// Attribute to match against the inputted username. This will be translated and combined
-		// with the other filter as "(<attr>=<username>)".
+		// Attributes (comma-separated) to match (OR)against the inputted username. This will be translated and combined
+		// with the other filter as "(!(<attr1>=<username>)|(<attr2>=<username>))".
 		Username string `json:"username"`
 
 		// Can either be:
@@ -414,7 +414,21 @@ func (c *ldapConnector) identityFromEntry(user ldap.Entry) (ident connector.Iden
 }
 
 func (c *ldapConnector) userEntry(conn *ldap.Conn, username string) (user ldap.Entry, found bool, err error) {
-	filter := fmt.Sprintf("(%s=%s)", c.UserSearch.Username, ldap.EscapeFilter(username))
+	var filter string
+	escapedUsername := ldap.EscapeFilter(username)
+
+	// Split username attribute by comma to support multiple search attributes
+	usernameAttrs := strings.Split(c.UserSearch.Username, ",")
+
+	attrFilters := make([]string, 0, len(usernameAttrs))
+	for _, attr := range usernameAttrs {
+		attr = strings.TrimSpace(attr)
+		if attr != "" {
+			attrFilters = append(attrFilters, fmt.Sprintf("(%s=%s)", attr, escapedUsername))
+		}
+	}
+	filter = fmt.Sprintf("(|%s)", strings.Join(attrFilters, ""))
+
 	if c.UserSearch.Filter != "" {
 		filter = fmt.Sprintf("(&%s%s)", c.UserSearch.Filter, filter)
 	}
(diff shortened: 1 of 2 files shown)


connector/ldap/ldap.go
```diff
@@ -34,10 +34,10 @@ import (
 //       bindDN: uid=serviceaccount,cn=users,dc=example,dc=com
 //       bindPW: password
 //       userSearch:
-//         # Would translate to the query "(&(objectClass=person)(uid=<username>))"
+//         # Would translate to the query "(&(objectClass=person)(!(uid=<username>)|(mail=<username>)))"
```
anuviswa:
```suggestion
//         # Would translate to the query "(&(objectClass=person)(|(uid=<username>)(mail=<username>)))"

```

connector/ldap/ldap.go
```diff
@@ -108,8 +108,8 @@ type Config struct {
 		// Optional filter to apply when searching the directory. For example "(objectClass=person)"
 		Filter string `json:"filter"`
 
-		// Attribute to match against the inputted username. This will be translated and combined
-		// with the other filter as "(<attr>=<username>)".
+		// Attributes (comma-separated) to match (OR)against the inputted username. This will be translated and combined
+		// with the other filter as "(!(<attr1>=<username>)|(<attr2>=<username>))".
```
anuviswa:
```suggestion
		+// with the other filter as "(|(<attr1>=<username>)(<attr2>=<username>))".
```

connector/ldap/ldap.go
```diff
@@ -414,7 +414,21 @@ func (c *ldapConnector) identityFromEntry(user ldap.Entry) (ident connector.Iden
...
+	usernameAttrs := strings.Split(c.UserSearch.Username, ",")
+
+	attrFilters := make([]string, 0, len(usernameAttrs))
+	for _, attr := range usernameAttrs {
+		attr = strings.TrimSpace(attr)
+		if attr != "" {
+			attrFilters = append(attrFilters, fmt.Sprintf("(%s=%s)", attr, escapedUsername))
+		}
+	}
+	filter = fmt.Sprintf("(|%s)", strings.Join(attrFilters, ""))
```
anuviswa:
```suggestion
	if len(attrFilters) == 1 {
        filter = attrFilters[0]  // Skip OR wrapper for single attribute
    } else {
        filter = fmt.Sprintf("(|%s)", strings.Join(attrFilters, ""))
    }
```

anuviswa: @yardenshoham This is great! Thanks for implementing this. I have a couple of minor comments, but otherwise it looks great to me. What is needed to get this merged? Happy to pitch in here with anything that's needed.

yardenshoham: I added your changes. I am just waiting for a review from a dex maintainer.

anuviswa:
@nabokihms  @sagikazarmark  - Could you please review and consider this PR? It addresses https://github.com/dexidp/dex/issues/3849.

connector/ldap/ldap.go
```diff
@@ -419,7 +419,25 @@ func (c *ldapConnector) identityFromEntry(user ldap.Entry) (ident connector.Iden
 }
 
 func (c *ldapConnector) userEntry(conn *ldap.Conn, username string) (user ldap.Entry, found bool, err error) {
-	filter := fmt.Sprintf("(%s=%s)", c.UserSearch.Username, ldap.EscapeFilter(username))
+	var filter string
+	escapedUsername := ldap.EscapeFilter(username)
+
+	// Split username attribute by comma to support multiple search attributes
+	usernameAttrs := strings.Split(c.UserSearch.Username, ",")
```
nabokihms: The usernameAttrs should be parsed on config loading, not on every request.

connector/ldap/ldap.go
```diff
@@ -110,8 +110,8 @@ type Config struct {
 		// Optional filter to apply when searching the directory. For example "(objectClass=person)"
 		Filter string `json:"filter"`
 
-		// Attribute to match against the inputted username. This will be translated and combined
-		// with the other filter as "(<attr>=<username>)".
+		// Attributes (comma-separated) to match (OR)against the inputted username. This will be translated and combined
+		// with the other filter as "(|(<attr1>=<username>)(<attr2>=<username>))".
 		Username string `json:"username"`
```
nabokihms: Honestly, string-typed APIs aren't good. It's so much better to add a new field, like `usernames` or just change the `username` behavior to be `string | []string`

yardenshoham:
Thanks for the review!

Accepting both username and email on login forms is standard practice — Okta, Azure AD, Google Workspace, Keycloak, and GitLab all do this. Users genuinely expect it, and forcing them to remember which specific attribute is configured adds unnecessary friction.

Regarding the collision concern: when multiple entries are returned, dex rejects the login, so this is fail-closed by design. And in any reasonably managed directory, a user's `cn` matching another user's `mail` is extremely unlikely.

As for making LDAP data consistent, many dex users don't control their directory (e.g., corporate AD managed by a different team), so restructuring the schema to work around a single application isn't realistic.

The feature is opt-in, backward-compatible, and uses standard LDAP OR filters. It gives admins flexibility without requiring upstream directory changes.

PR #4112, opened 2025-04-29
chore: bump server apiVersion

skrobul:
#### Overview

Bumps `apiVersion` so that API clients can correctly discover availability of a certain new API calls.

Inline documentation states that apiVersion should be increased every time a new call is added to the API. Unfortunately, this process was not followed in https://github.com/dexidp/dex/pull/3598 where GetDiscovery API call was added.
It also was not increased after adding GetClient in https://github.com/dexidp/dex/pull/2972.


sagikazarmark:
I don't remember why that was introduced. We certainly don't add APIs without implementation so it shouldn't be a problem.

In any case, the api version was bumped in a recent PR.

PR #4143, opened 2025-05-21
added UNIX socket support for serve and ldap connector

shaohme:
#### Overview

UNIX socket support for serve() and LDAP connector.

I like use UNIX sockets when no TCP sockets are needed. For simplicity and safety.


cmd/dex/serve.go
```diff
@@ -88,6 +88,17 @@ func commandServe() *cobra.Command {
...
+// try detect the intended socket type from address string
+func getSocketType(address string) string {
+	_, resErr := net.ResolveTCPAddr("tcp", address)
+	if resErr == nil {
+		return "tcp"
+	} else {
+		// assume UNIX socket
+		return "unix"
+	}
+}
```
nabokihms, line 117: I'm ok with the feature, but it feels like this approach is fragile:
- It may incorrectly classify an invalid Unix path as TCP if it resembles a TCP-style format.
- It does not account for the possibility that the address is invalid for both TCP and Unix, yet still returns "unix", thereby masking the error.
- It may unnecessarily perform address parsing even in straightforward situations.

And it calls resolve. 

Let's rewrite the function without resolve, and it also demands a unit test with all kinds of possible addresses, e.g.:
```go
examples := []string{
    "tcp://127.0.0.1:8080",
    "unix:///tmp/my.sock",
    "127.0.0.1:9000",
    "/var/run/app.sock",
    "./socket.sock",
    "relative.sock",
    "example.com:80",
    "unix://./run/rel.sock",
}
```

shaohme:
Sounds all reasonable. I've made a test and altered the function.

PR #4144, opened 2025-05-21
feat: Add ModifyGroupNames claimMutation to oidc connector

peschmae:
#### Overview

This PR adds a `claimModification` to the oidc connector, to allow prefixes/suffixes of group names.
Similar things were already requested in #918 and other issues.

Since the middleware component is still pending, and in the meantime `claimModifications` were introduced in #3056 for the oidc connector, this PR just adds another modification.

This partially solves #918 but only for oidc connectors

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -104,6 +104,7 @@ type Config struct {
 	ClaimMutations struct {
 		NewGroupFromClaims []NewGroupFromClaims `json:"newGroupFromClaims"`
 		FilterGroupClaims  FilterGroupClaims    `json:"filterGroupClaims"`
+		ModifyGroupNames   ModifyGroupNames     `json:"modifyGroupNames"`
 	} `json:"claimModifications"`
 }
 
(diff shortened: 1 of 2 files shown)


marians:
@peschmae Thanks for this contribution (from a Dex user running a fork because of this)! Does this cover both the initial token issuing as well as refreshing of a token?

connector/oidc/oidc.go
```diff
@@ -573,6 +584,11 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 		}
 	}
 
+	// add prefix/suffix to groups
+	for i, group := range groups {
+		groups[i] = c.groupsPrefix + group + c.groupsSuffix
+	}
```
nabokihms: NIT
```suggestion
    if c.groupsPrefix != "" || c.groupsSuffix != "" {
		for i, group := range groups {
			groups[i] = c.groupsPrefix + group + c.groupsSuffix
		}
	}
```

nabokihms (approved):
One minor suggestion, otherwise LGTM

connector/oidc/oidc.go
peschmae: Added the nitpick, and ammended the commit

peschmae:
> @peschmae Thanks for this contribution (from a Dex user running a fork because of this)! Does this cover both the initial token issuing as well as refreshing of a token?

This should work in both cases, as the `createIdentity` method is called from `oidc.Refresh` as well

PR #4145, opened 2025-05-21
:bug: remove extra method="get" from device-code template

tuminoid:
#### Overview

In device-code template, same form tag has method="post", then followed by method="get".. Chrome can handle that as post just fine, but in headless environments you might be using Lynx, which uses GET: it results in 400 Bad Request from Dex server.

```html
<div class="theme-panel">
  <h2 class="theme-heading">Enter User Code</h2>
  <form method="post" action="/device/auth/verify_code" method="get">
```

Remove the extra method="get" from the form.

### DIFF
--- a/web/templates/device.html
+++ b/web/templates/device.html
@@ -2,7 +2,7 @@
 
 <div class="theme-panel">
   <h2 class="theme-heading">Enter User Code</h2>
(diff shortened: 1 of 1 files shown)


nabokihms (approved):

nabokihms:
Nice catch! Thanks.

PR #4153, opened 2025-05-27
feat(connectors): add a go-plugin connector to support external connectors

### LINKED ISSUE #1907: External connectors

maxboone:
#### Overview

This PR adds a connector named `plugin` that allows the use of out-of-tree external connectors over RPC using go-plugin.

Closes https://github.com/dexidp/dex/issues/1907


cardoe:
I think this is a great idea. We could probably discuss it on the Slack channel.

maxboone:
> I think this is a great idea. We could probably discuss it on the Slack channel.

I discussed it a while ago with @nabokihms on Slack, iirc the maintainers prefer to go with a gRPC based solution, of which there is a another PR that could be re-done.

Do bring it up though, as I'd love for plugin functionality to make it to main!

sagikazarmark:
Personally, I don't see why we couldn't have both mechanisms, especially since the gRPC based solution seems a bit harder to implement. (AFAIK go-plugin is also gRPC based)

But @nabokihms has been spearheading this effort, so he may have more insights on the matter.

Also, I've been recently chatting with someone about HC's go-plugin and he mentioned a few issues with it, but I can't recall exactly what they were. Could be it just didn't work for their specific use case.

maxboone:
> Personally, I don't see why we couldn't have both mechanisms, especially since the gRPC based solution seems a bit harder to implement. (AFAIK go-plugin is also gRPC based)

It supports `net/rpc` and gRPC indeed, but will take a look at the docs how to enable both.

> But @nabokihms has been spearheading this effort, so he may have more insights on the matter.

Sounds good, I'm afk on vacation at the moment but can ping @nabokihms on Slack when I get back to see how I can contribute. 

If `go-plugin` ends up the choice for the matter, it's probably nice to build a connector that serves as example for third-party integrations (and maybe later "core" plugins that get built in by default). What might be a connector / integration that we would like to see but don't want to ship as core Dex right now?

typed-sigterm:
Hi, how is it going? There're tens of PRs that suggests new connectors now. Can we have this or some them them merged :)

PR #4159, opened 2025-05-29
Fix: Allow server startup with partial connector failures

manojVivek:
#### Overview

Improves server resilience by allowing Dex to start even when some connectors fail to initialize. Previously, a single misconfigured connector would prevent the entire server from starting.

This change prevents a single misconfigured connector from blocking the entire Dex server, improving availability in production environments where some identity providers may be temporarily unavailable or misconfigured.


brancz:
cc @sagikazarmark @nabokihms

cmd/dex/serve.go
```diff
@@ -44,10 +44,11 @@ type serveOptions struct {
...
 	// Flags
-	webHTTPAddr   string
-	webHTTPSAddr  string
-	telemetryAddr string
-	grpcAddr      string
+	webHTTPAddr                string
+	webHTTPSAddr               string
+	telemetryAddr              string
+	grpcAddr                   string
+	continueOnConnectorFailure bool
```
nabokihms: Let's configure this feature through a feature flag instead. I believe this is the behaviour we want to enable by default in the future.

https://github.com/dexidp/dex/tree/master/pkg/featureflags

nabokihms: We can default it to false for the next release with keeping in mind changing it to true in the future

server/server.go
```diff
@@ -325,12 +329,22 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
 		return nil, errors.New("server: no connectors specified")
 	}
 
+	var failedCount int
 	for _, conn := range storageConnectors {
 		if _, err := s.OpenConnector(conn); err != nil {
+			failedCount++
+			if c.ContinueOnConnectorFailure {
```
nabokihms, line 336: image
image

When I try to click on the connector, if it is not opened at startup, the error is quite bizarre. Let's return something that will guide the user.

nabokihms (changes requested):
Nice, I tested the PR manually, and everything seems fine. I have a couple of suggestions, but otherwise it is good enough to be merged. 

Some thoughts about the solution:

My initial idea was to experiment with a Kubernetes-like approach to handle OIDC providers. It retries discovery endpoint requests till a successful response in the background. 

However, there are disadvantages, like it is not possible to distinguish misconfigured connectors and non-responding providers. Additionally, it adds much more changes to the code.

server/server.go
manojVivek: Returning `Connector failed to initialize.` now. Please let me know if you'd prefer something different.

nabokihms: looks fine to me

joelanford:
Just as a heads up, this change broke my homelab Kubernetes workflow.

I run [Authelia](https://www.authelia.com/) and dex in the same single-node cluster. I rebooted it recently and discovered that Dex was unable to connect to Authelia, despite both of them being up and running.

Turns out dex started up before Authelia did (Kubernetes is eventually consistent), so dex failed to setup its only configured connector (Authelia), and then required a manual restart. In earlier versions of dex, the process would have crashed, and kubernetes would have restarted it with exponential backoff until Authelia came up. Now, the error is silently ignored and no self-healing occurs.

I can obviously fix this myself by turning this flag back off in my environment, so no big deal!

But big +1 on the ideas in #3274 and #2789.

PR #4164, opened 2025-06-02
Improve signature verification logic

ahacker1-securesaml:
#### Overview

- Update verifyResponseSig, so that it returns only signed assertion contents
- Removed spurious business logic checks on a Response element (since response signature can be removed anyway)
- Updated test case to reflect the changes

Resolves security concerns around: https://github.com/dexidp/dex/discussions/1884
Previously, the verifyResponseSig returned contents from the original SAML Response document. We must instead return contents from only the signed parts of the document i.e. parts verified with the XML Signature library.
This PR does that.

In refactoring, we stopped verifying the business logic on the SAML Response. This is because the SAML Response is not guaranteed to be signed, i.e. IdPs MUST only sign the assertion.
Furthermore, any signatures on the SAML Response can be removed by a malicious attacker (what's left would be solely the assertion's signature).
https://stackoverflow.com/questions/13234618/should-i-require-idps-to-sign-saml2-sso-responses/13245092#13245092

The connector now receives only the SAML assertion.
A test case broke i.e. a response signed, with a bad Status (AuthnFailed) The underlying assertions were correct though.
It would be an extremely rare case to handle because an IdP wouldn't issue a SAML assertion if authn failed.
Therefore, to balance code readability, the decision was made to remove the test case.

### DIFF
--- a/connector/saml/saml.go
+++ b/connector/saml/saml.go
@@ -303,51 +303,22 @@ func (p *provider) HandlePOST(s connector.Scopes, samlResponse, inResponseTo str
 		return ident, errors.Wrap(xrvErr, "validating XML response")
 	}
 
-	// Root element is allowed to not be signed if the Assertion element is.
-	rootElementSigned := true
+	var rawSignedAssertion []byte
+
 	if p.validator != nil {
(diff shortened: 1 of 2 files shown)


ahacker1-securesaml:
Superseded by: #4165

PR #4138, opened 2025-05-18
google: add support for groups access without service account

devodev:
#### Overview

Add support for retrieving groups the user is a member of without needing to configure a Service Account with Domain-Wide Delegation.

The Google connector has support for retrieving groups the user is a member of using the Admin SDK API. This API requires using an authenticated client sourced from a service account granted domain-wide delegation. This account would be highly-privileged and give access to all Workspace users, which is less than ideal.

This change adds an alternative method to retrieve group membership using the Cloud Identity API. When using the `SearchDirectGroups` API endpoint, we can provide an access token that has been authorized with the appropriate Cloud Identity API scope and search for direct and transitive membership.

In contrast to the existing method where the user is expected to specify the `groups` scope, this one relies on the user setting one of the required Cloud Identity scope explicitly:
- https://www.googleapis.com/auth/cloud-platform
- https://www.googleapis.com/auth/cloud-identity.groups
- https://www.googleapis.com/auth/cloud-identity.groups.readonly

This is similar to how Grafana provides this feature: https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/google/#configure-team-synchronization

Alternatively, we could also decide to support this feature by providing an explicit way in the config to choose which method to use.

Fixes: #3517

```release-note
Google connector now supports retrieving group information without a service account.
```

#### Testing

I added a basic unit test that mirrors the coverage that the existing `TestGetGroups` provide.

Additionally, I tested the new changes with a free version of Google Workspace + Google Cloud

I configured ArgoCD with the following:
```yaml
configs:
  cm:
    dex.config: |
      connectors:
        - id: google
          name: Google
          type: google
          config:
            issuer: https://accounts.google.com
            clientID: clientID
            clientSecret: clientSecret
            scopes: ["openid", "email", "profile", "https://www.googleapis.com/auth/cloud-identity.groups.readonly"]
            fetchTransitiveGroupMembership: true
  rbac:
    policy.default: role:readonly
    policy.csv: |
      g, admins@devodev.cloud, role:admin
    scopes: "[groups]"
dex:
  enabled: true
  image:
    repository: devodev/dex
    tag: v2.42.1-alpine-connectors-google-sa-less-groups
    imagePullPolicy: Always
```

To validate transitive group membership, I created the following group hierarchy:
- group: `admins@devodev.cloud`
  - group: `argocd-admins@devodev.cloud`
    - user: `alex@devodev.cloud`

And after login, I see the following log (+ admin role assigned):
```
[...]

PR #4169, opened 2025-06-04
feat: [Google connector] use Cloud Identity API for fetching groups

midu-git:
A new Google connector option, `useCloudIdentityApi`, has been introduced.
If the value is `true`, dex will use cloud identity api to fetch
groups. In particular, no user impersonation happens. The logic to
obtain the credentials is based on Application Default Credentials.
Alternatively, the user is allowed to pass a path to a credentials JSON
file using `serviceAccountFilePath`. In both cases, the principal
linked to the credentials requires group read rights. In case
of a Service Account, a custom admin role with this right need to be
created in Google Workspace, and the Service Account needs to be
assigned to this role.

Moreover, Workload Identity Federation is supported as Application Default
Credentials supports this use case. Make sure to include
`Service Account Token Creator` for the linked service account in that
case.

Until now, user impersonation was required to fetch groups for a certain user. In this PR, the user is allowed to opt-in for using cloud identity api which doesn't require user impersonation. Moreover, as Application Default Credentials are being used, Workload Identity Federation is supported with this PR, allowing interactions with Google Services from external sources, e.g. AWS, without requiring to provide any sensitive credential data on the caller side, c.f. [here](https://cloud.google.com/iam/docs/workload-identity-federation).

```release-note
A new Google connector option, `useCloudIdentityApi`, has been introduced.
If the value is `true`, dex will use cloud identity api to fetch
groups. In particular, no user impersonation happens. The logic to
obtain the credentials is based on Application Default Credentials.
Alternatively, the user is allowed to pass a path to a credentials JSON
file using `serviceAccountFilePath`. In both cases, the principal
described linked to the credentials requires group read rights. In case
of a Service Account, a custom admin role with this right need to be
created in Google Workspace, and the Service Account needs to be
assigned to this role.

Moreover, Workload Identity Federation is supported as Application Default
Credentials supports this use case. Make sure to include
`Service Account Token Creator` for the linked service account in that
case.
```

### DIFF
--- a/connector/google/google.go
+++ b/connector/google/google.go
@@ -17,6 +17,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	admin "google.golang.org/api/admin/directory/v1"
+	"google.golang.org/api/cloudidentity/v1"
 	"google.golang.org/api/impersonate"
 	"google.golang.org/api/option"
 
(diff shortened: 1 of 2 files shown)


midu-git:
@nabokihms any chance looking into this PR or anything I can do supporting the review process?

Vad1mo:
Isn't this similar to https://github.com/dexidp/dex/pull/4138?
@devodev

devodev:
> Isn't this similar to #4138? @devodev

@Vad1mo Yeah pretty much

PR #4200, opened 2025-07-01
fix(server): respond with unauthorized if failed to authenticate

aljoshare:
#### Overview

It changes the response from 500 - Internal Server Error to 401 - Unauthorized if the Dex is not able to authenticate a user, e.g. because the user is not in any of the required groups.

This PR is necessary because 401 - Unauthorized is imho a better response than 500 - Internal Server Error in this case. For example, this part of the code gets executed if a user is not in any of the required groups or if a refresh token is not valid, which is in my opinion a case where the user expects a 401 instead of a 500.

If this part is also executed in other cases I missed and where we need the 500, I could also try to change the code so that it differentiates between those cases.


cardoe:
Could this path affect any of the RFC response paths? Is there an expectation of a 500 somewhere?

nabokihms:
Reasonable question, but I think there is no RFC because the callback endpoint is part of the relying party, and the RP can act as it wishes.

My question is: can we distinguish these two errors from other possible errors?
- user is not in any of the required groups
- refresh token is not valid

aljoshare:
Sorry, I was on vacation. I'll check :v:

nabokihms:
@aljoshare looked through the changes, and the PR seems good. Feel free to undraft it.

aljoshare:
> @aljoshare looked through the changes, and the PR seems good. Feel free to undraft it.

Thanks for the quick response! I updated the PR again and changed it to 403, because I think that 403 is more accurate. The user is authenticated but not authorized.

> If authentication credentials were provided in the request, the
   server considers them insufficient to grant access.
https://datatracker.ietf.org/doc/html/rfc7231#page-59

What do you think?

nabokihms:
Yeah that's even better, thanks.

Issue #3496, opened 2024-04-26
gRPC endpoint for ListClients


MrFoxTrot:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

Api already have List endpoints for passwords and refresh tokens.
When someone install dex with storage backend (for example sql) and decides to add clients via gRPC there is no way to list all existing clients

### Proposed Solution

Add gRPC endpoint for ListClients

eLLIkin:
+1

ifurs:
It would be great to have that feature too for my automation tool https://github.com/ifurs/dexus-vault

daemonfire300:
Added a PR for it: https://github.com/dexidp/dex/pull/4202

PR #4202, opened 2025-07-01
feat: grpc api list clients

daemonfire300:
refers to https://github.com/dexidp/dex/issues/3496

#### Overview

Adds the ability to List existing clients using the gRPC API and adds minimal testing for it.

Adds the ability to List existing clients using the gRPC API and adds minimal testing for it.
I think the motivation is self-emergent given that we can create and delete things we should naturally be able to also get an overview of thise things.

Closes https://github.com/dexidp/dex/issues/3496

_none_

### DIFF
--- a/server/api_test.go
+++ b/server/api_test.go
@@ -735,7 +735,7 @@ func TestMissingConnectorsCRUDFeatureFlag(t *testing.T) {
 }
 
 func TestListClients(t *testing.T) {
-	logger := slog.New(slog.NewTextHandler(io.Discard, &slog.HandlerOptions{}))
+	logger := slog.New(slog.DiscardHandler)
 
 	s := memory.New(logger)
 	client := newAPI(s, logger, t)


daemonfire300:
going to
 fix this tomorrow.

daemonfire300:
Not sure what's up with the Lint CI Job, when I run `make lint` locally it works but the CI says

```
Error: server/api_test.go:738:41: undefined: io (typecheck)
```

Which does not make any sense.

daemonfire300:
@nabokihms I cannot re-run the failing CI steps and I cannot reproduce the failing CI steps locally. `io` is literally there and I just added a new test block. So I am struggling to see why this fails in the CI. Am I missing something very obvious?

daemonfire300:
Ok, I ran the lint job with `act` locally (github action runner) and it also does not fail. I think it's the CI. But I am happy to be corrected.

daemonfire300:
`act -j lint`



grafik

daemonfire300:
To me this feels very much like ghost in the machine in the CI.

daemonfire300:
@nabokihms found the issue. The fork was 5 months behind. Rebased.

cardoe (approved):
This is a good addition to the API and would make implementing some other addons a plus.

daemonfire300:
Is there a way how I can help out to speed this PR up? It's rather undercomplex and I would prefer not to fork the repository instead.

@JoelSpeed
@nabokihms
@sagikazarmark
@bonifaido
@rithujohn191 
@justaugustus

cardoe:
I think this is a good addition. The api/v2/api.proto has an api = 2 that should be bumped per the comments since you are adding something.

PR #4203, opened 2025-07-02
fix: device code should not require scope

cardoe:
#### Overview

As per RFC8628 section 3.1, https://datatracker.ietf.org/doc/html/rfc8628#section-3.1 the scope is optional. Since dex always requires at least 'openid', default the value to comply with the RFC.

Fixes compliance with RFC8628.


erikgb (approved):
/lgtm

PR #4204, opened 2025-07-02
fix: device code pending HTTP response

### LINKED ISSUE #4240: Device Code support fails to comply with RFC8628
### Preflight Checklist

### Version

current

### Storage Type

Kubernetes

### Installation Type

Official Helm chart

### Expected Behavior

Conforming RFC8628 behavior...

As per RFC8628 section 3.1, https://datatracker.ietf.org/doc/html/rfc8628#section-3.1 the scope is optional.

As per RFC8628 section 3.5, https://datatracker.ietf.org/doc/html/rfc8628#section-3.5 the authorization_pending response should extend RFC6749 section 5.2, https://datatracker.ietf.org/doc/html/rfc6749#section-5.2 which specifies that the HTTP response code should be 400, Bad Request.

### Actual Behavior

Code that expects a conforming RFC8628 implementation fails.

### Steps To Reproduce

_No response_

### Additional Information

_No response_

### Configuration

```yaml

```

### Logs

```shell

```

cardoe:
#### Overview

As per RFC8628 section 3.5, https://datatracker.ietf.org/doc/html/rfc8628#section-3.5 the authorization_pending response should extend RFC6749 section 5.2, https://datatracker.ietf.org/doc/html/rfc6749#section-5.2 which specifies that the HTTP response code should be 400, Bad Request.

RFC8628 compliance.

Partially fixes #4240

### DIFF
--- a/server/deviceflowhandlers.go
+++ b/server/deviceflowhandlers.go
@@ -249,7 +249,7 @@ func (s *Server) handleDeviceToken(w http.ResponseWriter, r *http.Request) {
 		if slowDown {
 			s.tokenErrHelper(w, deviceTokenSlowDown, "", http.StatusBadRequest)
 		} else {
-			s.tokenErrHelper(w, deviceTokenPending, "", http.StatusUnauthorized)
+			s.tokenErrHelper(w, deviceTokenPending, "", http.StatusBadRequest)
 		}
 	case deviceTokenComplete:
 		codeChallengeFromStorage := deviceToken.PKCE.CodeChallenge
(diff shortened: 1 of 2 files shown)


cardoe:
@nabokihms I am not sure what I need to do to pass the label check on this PR as I don't have permissions to modify the labels. Any advice?

I believe my two PRs for RFC8628 compliance are pretty straight forward fixes.

cardoe:
Adding a linked bug report with hopes that makes this check pass.

PR #4263, opened 2025-08-09
fix: trim the issuer URL trailing slash for server "Discovery" link

### LINKED ISSUE #4036: Remove Trailing / from issuer url
### Preflight Checklist

### Version

2.42.0

### Storage Type

In-memory

### Installation Type

Official Helm chart

### Expected Behavior

no 404s when viewing anything other than the root

### Actual Behavior

when you set the config like this

```
config:
  issuer: https://<host>/
```

there is a double / which results in 404s.

ex: viewing the known hosts results in
https://<host>///.well-known/openid-configuration and a 404 is returned

removing the / from the issuer fixes this:

```
config:
  issuer: https://<host>
```

however it would be simpler to remove a trailing / for more painless deployments

### Steps To Reproduce

1. deploy dex using the official helm chart, adding the issuer url as shown above with the trailing slash `issuer: https://<host>/`
2. go to the UI and click on discovery. this will result in a 404
3. remove the trailing /
4. test again and the 404 is gone

### Additional Information

_No response_

### Configuration

```yaml
config:
  issuer: https://<hostname>/
  storage:
    type: memory
  enablePasswordDB: true
  staticPasswords:
    - email: "jason@example.com"
      hash: "<redacted>"
      username: "jason"
      userID: "0"
```

### Logs

```shell

```

vizv:
#### Overview

Currently the homepage "Discovery" link doesn't work with issuer that ends with a slash, see #4036.

Instead of trim the trailing slash for issuer URL, this fix trims the issuer trailing slash when construct the discovery URL.
Closes #4036 without breaking existing deployments.


server/server.go
```diff
@@ -464,7 +464,7 @@ func newServer(ctx context.Context, c Config, rotationStrategy rotationStrategy)
 			<h1>Dex IdP</h1>
 			<h3>A Federated OpenID Connect Provider</h3>
 			<p><a href=%q>Discovery</a></p>`,
-			s.issuerURL.String()+"/.well-known/openid-configuration")
+			strings.TrimSuffix(strings.s.issuerURL.String(), "/")+"/.well-known/openid-configuration")
```
nabokihms:
```suggestion
			s.issuerURL.JoinPath(".well-known", "openid-configuration").String()
```

vizv: Yes, that's even better! I'll update the patch

cardoe (approved):

PR #4363, opened 2025-10-09
Do not wrap Kubernetes Address in brackets

nabokihms:
#### Overview

Wrap only ipv6 addresses in brackets

In the latest go versions (1.25.2, 1.24.8) the [bug was fixed](https://github.com/golang/go/issues/75712), and now Dex fails with an error

```
failed to initialize storage: cannot get kubernetes version: parse "https://[10.1.1.0]:443/version": invalid IPv6 host
```

### DIFF
--- a/storage/kubernetes/client.go
+++ b/storage/kubernetes/client.go
@@ -32,6 +32,17 @@ import (
 	"github.com/dexidp/dex/storage/kubernetes/k8sapi"
 )
 
+const (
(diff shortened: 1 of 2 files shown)


nabokihms:
Tested this PR on prod clusters, everything seems fine so far. Merging.

alamorre:
Hello! Do we know when an official release with this fix is coming out?
Appreciate any info - happy holidays!

PR #4370, opened 2025-10-15
fix: add dedicated go.mod for /v2 to fix /v2 imports

### LINKED ISSUE #4222: import of v2 dex release versions fails due to no /v2 in module name
### Preflight Checklist

### Version

2.43.1

### Storage Type

Postgres

### Installation Type

Custom container image

### Expected Behavior

The standard approach to v2 modules and tagging is to add /v2 to the module name.
This enables usage of v2 dex within other packages, eg. 

 go get github.com/dexidp/dex/v2@v2.43.1 

See https://go.dev/ref/mod

However you have not added /v2 to the module name so this fails.
SImilarly you havent added a /v2 folder with a duplicate go.mod with /v2 in the module name to allow both v2 and older to be installed.

Having v2 import blocked feels a little broken.  Especially when you install latest and it gives you a copy of dex from 2018.

The sort of problems that arise are that it leaves pseudo versions as the only way to install dex.

The problem with pseudo versions is that vulnerability scanning tools such as govuln read the version as v0.0.0 and so wrongly flag up the installation as insecure. Listing every vulnerability dex ever had.

NOTE: I notice if I fix the /v2 via a local replace edited copy of dex, I can install release versions, but the go mod tidy gets into dependency resolution issues. These are not present for pseudo versions.
Have you deliberately blocked v2 import due to this problem? Maybe if so at least add documentation of this?

### Actual Behavior

The latest tag is:
git log -1 --format=%ai v2.43.1
2025-05-22 12:01:51 +0300
... and the latest commit is from June.
But the module name is still without the v2 index/go.mod

Hence you can only install as a branch or a specific git hash -> a pseudo version.

If you try to get a v2 release version it fails...
go get  github.com/dexidp/dex/v2@v2.43.1
go: github.com/dexidp/dex@v2.43.1: invalid version: module contains a go.mod file, so module path must match major version ("github.com/dexidp/dex/v2")

OR if you manually update the go.mod and run it ...
without /v2

go.mod:12:9: require github.com/dexidp/dex: version "v2.43.1" invalid: should be v0 or v1, not v2
go: errors parsing go.mod:

with /v2

github.com/dexidp/dex/v2@v2.43.1: go.mod has non-.../v2 module path "github.com/dexidp/dex" (and .../v2/go.mod does not exist) at revision v2.43.1

If you use latest you get
go: added github.com/dexidp/dex v2.13.0+incompatible

So for some reason it has been pinned to only install as a version for the version from 2018 ...
git log -1 --format=%ai v2.13.0
2018-11-06 21:27:28 -0800

### Steps To Reproduce

In any environment try to import a dex v2 version from after 2018

go get  github.com/dexidp/dex/v2@v2.43.1

### Additional Information

_No response_

### Configuration

```yaml

```

### Logs

```shell

```

TylerGillson:
#### Overview

Applies the [major version subdirectory](https://go.dev/ref/mod#glos-major-version-subdirectory) approach to fixing dex's currently broken usage of major version release tags without a corresponding major version in go.mod.

This approach maintains backward compatibility with current projects importing pseudo versions.

Closes #4222

`govulncheck` should stop flagging https://pkg.go.dev/vuln/GO-2024-2476#versions-note once this is merged and a new release is published.


kulaginds:
@nabokihms Hi! Can you review this simple PR which fixed go mod v2 compatibility?

TylerGillson:
Hey @nabokihms. Is there anything I can do to move this forward?

nabokihms:
@TylerGillson I will test this PR, missed it completely. If standard go tooling works from the root, it seems ok, but maybe we need just to migrate imports... I will test and return back with the verdict.

nabokihms:
Ok, I tested. It seems like we cannot move forward with this PR. 
1. How do we keep both go.mod files in sync? Probably, we do not need to because it is a fake gomod, but it confuses users
2. How will Dependabot work? It should probably update the main go mod and skip the /v2
3. Imports like github.com/dexidp/dex/v2/server still do not work, because v2 is just a stub

If we want to properly solve the problem with v2, I would like to solve it the right way, but it is a big breaking change for many

TylerGillson:
@nabokihms you’re right.. my apologies. I’ll close this PR. A `v2/go.mod` without packages under `v2/` doesn’t fix `github.com/dexidp/dex/v2/...` imports and doesn't add any value. I misunderstood the Go major version subdirectory guidance.

The root issue here is that dex uses `v2.x` tags on the `github.com/dexidp/dex` root module, which doesn't end with `v2`. This results in `+incompatible` Go imports and the vulnerability flagging I originally raised.

There are three options:

1. Keep the current module path, but revert the latest release to the `v1.x` tag line.
1. Add a proper `github.com/dexidp/dex/v2` module with all code under `v2/` and update all imports (large, breaking change).
1. Do nothing 😢

PR #4381, opened 2025-10-17
feat: DEP #32 "Remember me" for Password-based Connectors

### LINKED ISSUE #32: "Remember me" when logging in

daemonfire300:
#### Overview

kind/enhancement

Build concept/solution for https://github.com/dexidp/dex/issues/32

For details refer to **`docs/enhancements/remember-me-2025-10-19-#32.md`**

Closes https://github.com/dexidp/dex/issues/32


daemonfire300:
@nabokihms @sagikazarmark I think you have to add a label for me because I cannot do that. Would be appreciated.

FEATURE-NOTES.md
```diff
@@ -0,0 +1,27 @@
+Goal is to come up with a compact and minimal design for https://github.com/dexidp/dex/issues/32
```
daemonfire300, line 1: TODO: Remove this file, this was just my scratch pad to gather my thoughts.

nabokihms:
I looked through this PR briefly. As was mentioned, this feature requires the sessions concept and the way of storing them. There are other possible features that can benefit from it, like SSO or 2FA, but moving towards a simple in-memory store will make this thing unextendible in the future.Let's discuss the session implementation first and then move forward with the remember me feature.

daemonfire300:
> I looked through this PR briefly. As was mentioned, this feature requires the sessions concept and the way of storing them. There are other possible features that can benefit from it, like SSO or 2FA, but moving towards a simple in-memory store will make this thing unextendible in the future.Let's discuss the session implementation first and then move forward with the remember me feature.

In general I agree with your notion, so it's a yes but I would have the following suggestion(s):

1. Intentionally did it in memory only to avoid complexity, because this did not cost that much and does not introduce such a big maintenance burden. So I would be in favor to have a two phase solution, if possible, where we have a in memory remember version out first (soon-ish) and have 80% of the benefit of 20% of the work and then use that cushion to work on a generic/extended session concept.
2. Would you like to have an initial Teamspeak/Discord/Signal/Matrix/Google Meet call between us or more people to get on common ground and then start writing proposals?

daemonfire300:
@nabokihms any update?

daemonfire300:
happy to setup a call or any kind of "forum" / documents to discuss this. Also still think this is an intermediary solution.

cc 
@JoelSpeed 
@sagikazarmark 
@bonifaido 
@rithujohn191
@justaugustus

steve-todorov:
Hey, 

What's the state of this? Any progress? Whats blocking this?

daemonfire300:
Afaik superseded by https://github.com/dexidp/dex/issues/4560 @nabokihms correct me if I am wrong though.

nabokihms:
@daemonfire300 I'm not sure whether it is fully superseded, but I will check this later

PR #4383, opened 2025-10-17
feat: OAuth 2.0 Dynamic Client Registration Protocol (RFC 7591)

ByteBaker:
#### Overview

Implements RFC 7591 OAuth 2.0 Dynamic Client Registration Protocol, allowing clients to register dynamically at runtime via HTTP API.

- `POST /register` endpoint for client registration
- Request/response types compliant with RFC 7591
- Support for both confidential and public clients (PKCE)
- Optional authentication via initial access token
- Default values per RFC spec (`grant_types`, `response_types`, etc.)
- Validation of grant types and response types
- Discovery endpoint includes `registration_endpoint` field


cardoe:
Hello @ByteBaker I'm curious if you have an opinion on my approach in #4988 ?

Issue #4382, opened 2025-10-17
Possible regression: kubernetes client IPv6 resolution


JGuinegagne:
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [x] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

2.44.0

### Storage Type

Kubernetes

### Installation Type

Official container image

### Expected Behavior

Kubernetes go client within the `dex` image should correctly resolve the Kubernetes API internal service.

Previous image `ghcr.io/dexidp/dex:v2.43.1` not impacted.

### Actual Behavior

Kubernetes go client treats the IPv4 service Kubernetes service as an IPv6 address. As a result the pod go in crash loop.

### Steps To Reproduce

environment: Kubernetes cluster on AWS EKS which uses single stack IP on IPv4
```bash
kubectl get service kubernetes -o yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: <>
  labels:
    component: apiserver
    provider: kubernetes
  name: kubernetes
  namespace: default
  resourceVersion: <>
  uid: <>
spec:
  clusterIP: 172.20.0.1
  clusterIPs:
  - 172.20.0.1
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: https
    port: 443
    protocol: TCP
    targetPort: 443
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```

### Configuration

```yaml
`dex` deployment configuration]

apiVersion: apps/v1
kind: Deployment
metadata:
  name: dex
  namespace: {{ .Values.namespace }}
  labels: <omitted>
spec:
  replicas: 1
  selector: <omitted>
  template:
    metadata:
      labels: <omitted>
    spec:
      serviceAccountName: dex
      containers:
      - name: dex
        image: ghcr.io/dexidp/dex:latest
        command: ["/usr/local/bin/dex", "serve", "/etc/dex/config/config.yaml"]
        env: <omitted>
        ports: <omitted>
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
[...]

PR #4388, opened 2025-10-20
fix(storage/kubernetes): Only wrap IPv6 addresses in brackets

rene-dekker:
#### Overview

The Kubernetes client code was wrapping all IP addresses (both IPv4 and IPv6) in square brackets when constructing the API server URL. This was based on an incorrect assumption that IPv4 addresses in brackets are valid in a URL.

As a result, the container crashes immediately when built with newer versions of golang.

Recent versions of Go (1.25.2 and later) have stricter URL parsing that conforms to RFC 3986. This causes a failure when running Dex in a Kubernetes environment where the KUBERNETES_SERVICE_HOST is an IPv4 address, leading to the error "invalid IPv6 host".

This commit changes the logic to only wrap IPv6 addresses in square brackets. It uses `net.JoinHostPort`, which is a convenience function made for this specific purpose. This ensures that URLs are correctly formatted for both IPv4 and IPv6 addresses (and even if KUBERNETES_SERVICE_HOST would be a dns name), fixing the incompatibility with newer Go versions.

- It fixes https://github.com/dexidp/dex/issues/4382 which was issued by @JGuinegagne

### DIFF
--- a/storage/kubernetes/client.go
+++ b/storage/kubernetes/client.go
@@ -528,16 +528,8 @@ func getInClusterConnectOptions(host, port string) (k8sapi.Cluster, error) {
 		)
 	}
 
-	// we need to wrap IPv6 addresses in square brackets
-	// IPv4 used to work with square brackets, but it was fixed in the latest Go versions
-	// https://github.com/golang/go/issues/75712
-	ipAddr := net.ParseIP(host)
-	if ipAddr != nil && ipAddr.To4() == nil {
-		host = "[" + host + "]"
-	}
-
 	cluster := k8sapi.Cluster{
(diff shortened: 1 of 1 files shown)


rene-dekker:
@sagikazarmark or @ericchiang or @nabokihms

Would this suffice, it was verified in a kubeadm cluster. Or does the change need test coverage?

rene-dekker:
I see after rebasing that another commit was made by @nabokihms  to address the same problem. I think my fix is still a better choice, as we are using the utility maintained by the net package to solve the problem. Therefore I am not yet closing the PR.

Issue #827, opened 2017-02-28
Provide binary releases


ericchiang:
Would improve our ability to get people up and running quickly.

Dentrax:
I think this is an important issue. I had to compile binaries myself. Maybe goreleaser? Any thoughts on this?

cc'in for aweraness @sagikazarmark @nabokihms

sagikazarmark:
What's your use case that requires prebuilt binaries?

fflorent:
I come with a use case: some applications for [Yunohost](https://yunohost.org/) (an OS for self-hosters) relies on the [dex package](https://github.com/YunoHost-Apps/dex_ynh) to provide authentication for apps that require the use of OIDC for that while Yunohost provides an LDAP.

Yunohost apps don't rely on Docker images, or containerization.

Self-hosters may use low-power or old machine, which makes compiling an heavy process.

Providing a Dex binary would benefit for these apps and for these self-hosters.

The below issue to tracks the idea of extracting the binary from the Docker image, but it would be much more convenient to directly have a binary downloadable from a Release:
https://github.com/YunoHost-Apps/dex_ynh/issues/40

fflorent:
BTW, I understand this is not your top priority, would you welcome a PR for that?

joonas:
I'll echo @fflorent's comments that if this was something that you'd be willing to accept contributions for, I for one would be interested in contributing this.

In terms of use case, we use Dex as part of our (non-container-based) local development setup, and would love to have the binaries more readily available. Right now we're looking at using [`mise`](https://mise.jdx.dev/) for managing our development dependencies, and unfortunately I've not yet been able to make [any of the backends](https://mise.jdx.dev/dev-tools/backends/#backends) work for Dex.

If Dex provided binaries as part of the release, it would become installable via [`aqua`](https://aquaproj.github.io/), [`github`](https://mise.jdx.dev/dev-tools/backends/github.html) and [`ubi`](https://mise.jdx.dev/dev-tools/backends/ubi.html) backends basically for free.

PR #4392, opened 2025-10-23
chore: remove api/v2 replace directive

joonas:
#### Overview

This PR gets rid of the replace directive in the `go.mod` for the `github.com/dexidp/dex` module.

Doing so has the benefit of making it possible to `go install github.com/dexidp/dex/cmd/dex`, which at least slightly addresses the underlying need from https://github.com/dexidp/dex/issues/827, though an officially provided binary release would of course be better.

This change addresses the error message you'll receive if you try to `go install github.com/dexidp/dex/cmd/dex`:
```shell
$ go install --tags '!cgo' github.com/dexidp/dex/cmd/dex@2dce75009acfcd9df77d57f2d144aae999a4c569
go: github.com/dexidp/dex/cmd/dex@2dce75009acfcd9df77d57f2d144aae999a4c569 (in github.com/dexidp/dex@v0.0.0-20250829100637-2dce75009acf):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

It also bumps the `github.com/dexidp/dex/api/v2` dependency to `v2.4.0` since parts of the code rely on types introduced in the `github.com/dexidp/dex/api/v2` release `v2.4.0`.

This has been previously brought up in [a discussion](https://github.com/dexidp/dex/discussions/3102).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -124,6 +124,4 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-replace github.com/dexidp/dex/api/v2 => ./api/v2
-
 tool entgo.io/ent/cmd/ent


cardoe (approved):

cardoe:

cardoe:
@nabokihms we should pull this in

nabokihms:
@cardoe honestly, I'm not sure. Dex is not intended to be used as a library or be deployed as a separate binary. If we want to provide both, we have to add more tests, run tests for other OS etc. 

Having a replace makes it much harder to manage API package as a dependency. Makes releases harder.

It seems like the initial request in this issue goes beyond just removing the replace.

cardoe:
> Having a replace makes it much harder to manage API package as a dependency. Makes releases harder.

That's what my thinking was. Not for using it as a library. Just to simplify maintenance.

joonas:
@nabokihms 

> Dex is not intended to be used as a library or be deployed as a separate binary.

I can certainly appreciate this, and to be honest with you this isn't the use case I had in mind for making this PR.

Instead, as I mentioned in my original description for the PR, my goal with this change was to simply enable users to install Dex via the standard means (`go install github.com/dexidp/dex/cmd/dex`) so that we can more easily manage Dex as part of our local development setup where in my case at least it's not desirable to run our dependencies as/in containerized processes.

I'm not asking or expecting the Dex project to provide or maintain pre-built artifacts for non-supported OS (in my case, macOS), and if I were to come across any issues specific to my use of Dex on macOS, I would certainly be happy to contribute fixes back to the project to address my own issues.

Unfortunately as it stands, the `replace` directive in `go.mod` prevents me from managing the local development Dex binary, which is a real bummer.

I would encourage you to reconsider whether it's worth actively/knowingly prevent end-users wanting to use Dex in this way from doing so.

cardoe:
You could rebase this as just a removal of the replace directive.

joonas:
> You could rebase this as just a removal of the replace directive.

That's fair, it'd gone stale since I opened this.

PR #4408, opened 2025-11-07
feat(gitlab): full group retrieval

CorentinPtrl:
#### Overview

Implements full group retrieval for the GitLab connector. and adds all the mappings of GitLab access levels into group-permission suffixes.

Implement fetching all groups and per-group membership from the GitLab API so Dex can return the complete set of groups for a user (not only those in the `/oauth/userinfo` response). See upstream discussion: https://gitlab.com/gitlab-org/gitlab/-/issues/340905. The filtering via `groups` will prevent overly large tokens. Full group retrieval now requires the `read_api` scope.


cardoe (approved):

cardoe:

nabokihms:
Hello! Thanks for the PR. I'm ok with the feature if it is toggleable.

PR #4443, opened 2025-12-11
chore: extend example configs for idEnv and public

cardoe:
Add a few more examples in the configs for staticClient configuration using idEnv and public flags.


cardoe:
@sagikazarmark This is a pretty boring one but hopefully helps end users.

nabokihms (approved):
Thanks!

Issue #4444, opened 2025-12-11
support RFC8414 for OAuth 2.0 server metadata


cardoe:
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

[RFC8414](https://www.rfc-editor.org/rfc/rfc8414) has come out to provide a well-known endpoint for OAuth 2.0. This is mostly intended for MCPs so this would be a bit of a forward looking feature.

### Proposed Solution

Implement the well-known endpoint at /.well-known/oauth-authorization-server

johnvan7:
@cardoe I've create a PR for this feature: #4446 .

PR #4446, opened 2025-12-12
Add support for the OAuth 2.0 Authorization Server Metadata Endpoint (RFC 8414)

johnvan7:
#### Overview

This PR adds support for the OAuth 2.0 Authorization Server Metadata endpoint defined in [RFC 8414](https://www.rfc-editor.org/rfc/rfc8414). 
Alongside the existing OIDC discovery document, this PR exposes a dedicated well‑known endpoint for OAuth 2.0 metadata.

- Implements the endpoint /.well-known/oauth-authorization-server as specified in RFC 8414.
- Introduces separate discovery structures for OIDC and OAuth2 to ensure correct metadata for each protocol.
- Adds full test coverage for both discovery endpoints.

This feature enables better interoperability with systems and MCPs that rely on OAuth 2.0 Authorization Server Metadata, making Dex more compliant with modern standards and forward‑looking integrations.

Related issue: 

- #4444

### DIFF
--- a/connector/oidc/oidc_test.go
+++ b/connector/oidc/oidc_test.go
@@ -862,6 +862,17 @@ func setupServer(tok map[string]interface{}, idTokenDesired bool) (*httptest.Ser
 		})
 	})
 
+	mux.HandleFunc("/.well-known/oauth-authorization-server", func(w http.ResponseWriter, r *http.Request) {
+		url := fmt.Sprintf("http://%s", r.Host)
+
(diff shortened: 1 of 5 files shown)


cardoe:
@johnvan7 looks reasonable. Can you squash the commits into 1? At a minimum things won't compile until all the commits are together. The last one could be split off if we really wanted but practically I don't think it makes sense.

johnvan7:
@cardoe Thanks for your advice, I did the squash in one commit.

PR #4448, opened 2025-12-12
fix: device callback URL needs to handle a /

### LINKED ISSUE #4242: Device Callback URI is broken if Issuer URI path is "/"
### Preflight Checklist

### Version

2.43.1

### Storage Type

SQLite

### Installation Type

Official container image

### Expected Behavior

Dex should intelligently join the path segments resulting in `/device/callback`.

### Actual Behavior

Dex uses simple string concatenation, resulting in `//device/callback` which has a different meaning to browsers.

### Steps To Reproduce

1. Set issuer to any host with path "/"
2. Start device auth flow
3. Finish device auth flow in browser
4. Get redirected to http://device/callback (bug)

### Additional Information

Maybe it would be enough to use `url.JoinPath` here?

https://github.com/dexidp/dex/blob/745e1114f341e849f3b0edde45b39c14017deaf8/server/oauth2.go#L493-L495

I don't know if it would break anything else.

Our workaround for now is to use an issuer URI without path

### Configuration

```yaml
issuer: http://localhost:5556
      web:
        http: 0.0.0.0:5556
        allowedOrigins: ['*']
      grpc:
        addr: 0.0.0.0:5557
      storage:
        type: sqlite3
      oauth2:
        skipApprovalScreen: true
      staticClients:
        - id: cli
          name: 'CLI App'
          public: true
          redirectURIs:
          - '/device/callback'
      enablePasswordDB: true
      staticPasswords:
        - email: "admin@example.com"
          # bcrypt hash of the string "password": $(echo password | htpasswd -BinC 10 admin | cut -d: -f2)
          username: "admin"
      expiry:
        idTokens: "5m"
```

### Logs

```shell

```

cardoe:
If the issuer path ends with a / the URL will be built wrong so we should instead use the helper function to ensure the path is built correctly. fixes #4242.

### DIFF
--- a/server/oauth2.go
+++ b/server/oauth2.go
@@ -491,7 +491,7 @@ func (s *Server) parseAuthorizationRequest(r *http.Request) (*storage.AuthReques
 		return nil, newDisplayedErr(http.StatusBadRequest, "Unregistered redirect_uri (%q).", redirectURI)
 	}
 	if redirectURI == deviceCallbackURI && client.Public {
-		redirectURI = s.issuerURL.Path + deviceCallbackURI
+		redirectURI = s.absPath(deviceCallbackURI)
 	}
 
 	// From here on out, we want to redirect back to the client with an error.


cardoe:
This is part of my v2.44.90 testing build.

nabokihms:
Thanks, good catch!

PR #4453, opened 2025-12-17
Update distroless base image to debian13

loosebazooka:
`debian13` is the current active debian distroless build. debian12 images will be updated till sept 2026 though. Just easier to get ahead of this before support is dropped.


nabokihms:
Thanks, that one is good.

PR #4464, opened 2026-01-07
fix: handle url fragments for redirects

### LINKED ISSUE #4462: URL fragment is lost when using username/password form
### Preflight Checklist

### Version

master

### Storage Type

SQLite

### Installation Type

Binary

### Expected Behavior

When you visit an URL with fragment, and get redirected to dex to authenticate, the fragment should be part of the url when redirected back to the original URL

Example: A user navigates to `https://example.com/#/about/me`. It is redirected for authentication to dex IDP for ldap authentication. When successfully authenticated it will end up being redirected to `https://example.com/`

### Actual Behavior

When you visit an URL with fragment, and get redirected to dex to authenticate, form interaction (password.html) will remove (or not include) the fragment in form action.

The fragment should be part of the form post so that it is present in the url when redirected back to the original URL

### Steps To Reproduce

1. Checkout master
2. make build
3. Run dex `./bin/dex serve examples/config-dev.yaml`
4. Run example app `cd examples && go run ./example-app`
5. Go to `http://127.0.0.1:5555/` click login
6. Add `#foobar` to the URL and force refresh
8. Click "Login with email" and observe the fragment is gone
9. Add `#foobar` to the URL and force refresh
10. Login using static user credentials `admin@example.com` and `password` and observe the fragment is gone when redirected.

### Additional Information

This issue exists because the URL fragment stays in the browser, it's a client side part. So with the first url request from the browser the server can have many redirects and the fragment will stay. But when there is a form to submit something (user/pass) and the fragment is not included it will be lost.

The redirect url is created on the server and it is expected that this does not include the URL fragment.

oauth2-proxy has a similar fix for this problem. https://github.com/oauth2-proxy/oauth2-proxy/blob/0100ca9403399716586f56a08c784fc690c6fc79/pkg/app/pagewriter/sign_in.html#L113

--

I'm using dex + ldap as IDP with oauth2-proxy to protect an upstream webserver. I only have one connector and the auth selection page is skipped.

### Configuration

```yaml
examples/config-dev.yaml
```

### Logs

```shell

```

hvoort:
#### Overview

Includes original URL fragment throughout auth process to keep it in the original redirect. 

Closes #4462

Including the URL fragment on the pages that do form posts to different URL (not to self) and anchor links to providers.  

What are implications of adding the hash to the URLs on the login.html page. Do some providers use the hash and if so this will mess up the url fragment. Maybe we should only add it to the password.html page.  

Tested manually using the following flow:
1. Run dex `./bin/dex serve examples/config-dev.yaml`
2. Run example app `cd examples && go run ./example-app`
3. Go to `http://127.0.0.1:5555/` click login
4. Add `#foobar` to the URL and force refresh
5. Click "Login with email" and observe the fragment is still there
6. Add `#foobar` to the URL and force refresh
7. Login using static user credentials `admin@example.com` and `password` and observe the fragment is still there after being redirected.

### DIFF
--- a/web/templates/login.html
+++ b/web/templates/login.html
@@ -2,7 +2,7 @@
 
 <div class="theme-panel">
   <h2 class="theme-heading">Log in to {{ issuer }} </h2>
-  <div>
+  <div id="providers">
     {{ range $c := .Connectors }}
       <div class="theme-form-row">
         <a href="{{ $c.URL }}" target="_self">
@@ -18,7 +18,7 @@
 
 <script type="text/javascript">
   if (window.location.hash) {
-    var buttons = document.querySelectorAll('div.theme-form-row a');
+    var buttons = document.querySelectorAll('#providers a');
     buttons.forEach(function(button) {
       button.href = button.href + window.location.hash;
     });


web/templates/login.html
```diff
@@ -16,4 +16,13 @@ <h2 class="theme-heading">Log in to {{ issuer }} </h2>
   </div>
 </div>
 
+<script type="text/javascript">
+  if (window.location.hash) {
+    var buttons = document.querySelectorAll('div.theme-form-row a');
```
nabokihms:
``` suggestion
    var buttons = document.querySelectorAll('div.theme-form-row a');
```

NIT: Can we make the selector more straightforward? Maybe we can add ID to the password button.

hvoort: Do you mean to only change the button that is linking to a password connector, or all connectors? I think this applies to all connectors. Because a press on any provider button will not include the url fragment, if not explicitly injected.

If you want to support all connectors, I can make it more specific by adding a class to the anchor elements "provider-anchor" and update the query selector to `a.provider-anchor`. 
(edit: or maybe better to add an id attribute "providers" to the parent div and then use "#providers a")

Or do you have something else in mind?

nabokihms: You are selecting div by a class, and I think more about selecting div by id or something. 

`getElementById('myId')` or `querySelectorAll('div#myId')`

hvoort: Allright. I added id `providers` to the container div and now select all anchors found in the providers element using `#providers a`. 

I cannot set unique id attribute on anchors because we loop through connectors. Unless we use the connector's identifier (if it exists) as id, with the downside we have to loop over these identifiers in javascript  as well. To get these identifiers into javascript we will have to mix the template language and javascript, and this is asking for trouble if you ask me.

Let me know what you think

PR #4496, opened 2026-01-28
feat: gitlab support custom rootCAData

Jabejixo:
#### Overview

Add optional `rootCAData` support for the GitLab connector to improve TLS connectivity to self-hosted GitLab instances, with proxy support and a 30s request timeout.

Self-hosted GitLab deployments (especially in isolated/private environments) often use certificates issued by an internal CA that is not trusted by the system trust store. In those cases the GitLab connector fails TLS verification and cannot complete the OAuth flow (token exchange and subsequent API calls).

This PR:
- Adds an optional `rootCAData` field to the GitLab connector config to provide a PEM CA chain for TLS verification.
- Ensures HTTP requests respect environment proxy settings.
- Adds a 30s request timeout to avoid hanging calls.
- Keeps backward-compatible error semantics for invalid PEM input by returning `gitlab: invalid rootCAData`.
- Adds tests using a local HTTPS server and CA fixtures to verify:
  - successful callback flow when `rootCAData` is provided
  - expected TLS failure when it is not
  - correct error handling for invalid `rootCAData`

- The change is opt-in: existing GitLab configs without `rootCAData` behave the same.
- Test certificates/keys are stored under `connector/gitlab/testdata/` and are used only in unit tests (no runtime impact).


nabokihms (approved):

nabokihms:
Thanks, good improvement!

PR #4512, opened 2026-02-06
feat: Add Vault signer for JWT

nabokihms:
#### Overview

This pull request introduces a new, pluggable signing interface for JWT tokens in the server, allowing for both local and Vault-based signing. The changes refactor how signing keys are managed and accessed throughout the codebase, replacing direct storage access with a new Signer abstraction, and update configuration, server initialization, and key verification logic accordingly.

Outsourcing the signing function to an external provider is crucial for security reasons. Signer abstraction allows us to add more signers in the future.

Docs https://dexidp.io/docs/configuration/tokens/#token-signing-configuration

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -115,7 +115,7 @@ require (
 	golang.org/x/sync v0.19.0 // indirect
 	golang.org/x/sys v0.41.0 // indirect
 	golang.org/x/text v0.34.0 // indirect
-  golang.org/x/time v0.14.0 // indirect
+	golang.org/x/time v0.14.0 // indirect
 	golang.org/x/tools v0.41.0 // indirect
 	golang.org/x/tools/go/packages/packagestest v0.1.1-deprecated // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20251202230838-ff82c1b0f217 // indirect


server/handlers.go
```diff
@@ -125,6 +124,14 @@ func (s *Server) constructDiscovery() discovery {
 		},
 	}
 
+	// Determine signing algorithm from signer
+	signingAlg, err := s.signer.Algorithm(context.Background())
```
AlwxSin: @nabokihms maybe should we pass existing context from `newServer` func?

nabokihms: Fixed

nabokihms:
@sagikazarmark could you please take a look? I plan to add integration tests in future releases.

server/signer_vault.go
```diff
@@ -0,0 +1,327 @@
...
+	"hash"
+
+	"github.com/go-jose/go-jose/v4"
+	vault "github.com/openbao/openbao/api/v2"
+)
+
+// VaultSignerConfig holds configuration for the Vault signer.
+type VaultSignerConfig struct {
+	Addr    string `json:"addr"`
+	Token   string `json:"token"`
```
sagikazarmark, line 27: Using VAULT_ADDR and VAULT_TOKEN are quite standard to configure these details. Not to mention, plain text token in a config file isn't really secure.

server/signer_vault.go
```diff
@@ -0,0 +1,327 @@
...
+	"crypto/sha512"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"hash"
+
+	"github.com/go-jose/go-jose/v4"
+	vault "github.com/openbao/openbao/api/v2"
```
sagikazarmark, line 21: While the API remains compatible for the moment, I don't think we have any guarantees it's gonna stay that way in the future.

There is a potential risk that without any integration tests, a library update or a newer version of Vault/OpenBao may silently break.

The two potential options I see:

- Support Vault and OpenBao separately using their own libraries
- Provide detailed integration tests to make sure it works with both

I'm okay delaying this to a subsequent PR though.

server/signer_vault.go
nabokihms: I predicted this question. Currently, in other places, we solve the problem with env vars. The feature is intended to be used with them like this:

```yaml
 signer:
   type: vault
   vault:
     addr: $VAULT_ADDR
     token: $VAULT_TOKEN
     keyName: dex-key
```

server/signer_vault.go
nabokihms: This one I didn't predict. If you are ok, then next PR I will provide integration tests for both openbao and vault. In the future, if there is a problem with their APIs not being a match, we will be able to split them.

server/signer_vault.go
nabokihms: I can add this to the config example instead of placeholders.

nabokihms: Ok, I gave env variables another think. I will add the code to pick defaults from these variables.

nabokihms: Fixed

nabokihms:
@sagikazarmark if you are ok, let's merge this, and I will provide integration tests.

server/signer_vault.go
sagikazarmark: Fine by me, but we should probably highlight this in the documentation somehow?

nabokihms: Yes, I will provide the doc

nabokihms:
Documentation https://dexidp.io/docs/configuration/tokens/#token-signing-configuration

PR #4513, opened 2026-02-06
feat: support claim-based group name modification with separators

lorissenneca:
**Overview**
This PR adds the ability to reference OIDC claims when modifying group names as prefixes and suffixes, moving beyond static string values. This enables more dynamic group naming strategies based on claim data.

**Problem**
Previously, group name prefixes and suffixes could only be static strings. Organizations need the ability to incorporate claim values (e.g., organization name, environment) dynamically into group names.

**Solution**
Introduces a new ValueOrClaim union type that allows specifying either:

- A literal string (Value field)
- A reference to an OIDC claim (Claim field)
- An optional separator (Separator field) applied directionally

This replaces the previous string-based approach with a more flexible, extensible pattern.

This basically extends the functionality of #4144 to allow to use other claims as prefix & suffix.

Special notes for your reviewer

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -190,10 +190,19 @@ type FilterGroupClaims struct {
 	GroupsFilter string `json:"groupsFilter"`
 }
 
+// ValueOrClaim represents a value that can be either a literal string or a reference to a claim.
+// If Claim is set, its value from the claims map will be used.
+// Otherwise, Value (the literal string) will be used.
+type ValueOrClaim struct {
+	Value     string `json:"value,omitempty"`
+	Claim     string `json:"claim,omitempty"`
+	Separator string `json:"separator,omitempty"`
+}
+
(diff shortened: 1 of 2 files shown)


connector/oidc/oidc.go
```diff
@@ -190,10 +190,19 @@ type FilterGroupClaims struct {
...
+	Claim     string `json:"claim,omitempty"`
+	Separator string `json:"separator,omitempty"`
+}
+
 // ModifyGroupNames allows to modify the group claims by adding a prefix and/or suffix to each group.
 type ModifyGroupNames struct {
-	Prefix string `json:"prefix"`
-	Suffix string `json:"suffix"`
+	Prefix ValueOrClaim `json:"prefix"`
+	Suffix ValueOrClaim `json:"suffix"`
```
nabokihms, line 205: This looks like a braking change. Could you introduce a new option without changing the behavior of the current ones? It is much more preferable.

PR #4526, opened 2026-02-11
feat: add name and emailVerified fields for static passwords

Jabejixo:
#### Overview

Add support for **`name`** and **`emailVerified`** for statically configured users (`staticPasswords`) and propagate them through the local password DB connector into standard OIDC ID token claims.

- **Adds new optional config fields** for `staticPasswords`:
  - `name: string` → OIDC **`name`** claim (when `profile` scope is requested). Defaults to `username` when empty for backward compatibility.
  - `emailVerified: *bool` → OIDC **`email_verified`** claim (when `email` scope is requested). Defaults to `true` when nil for backward compatibility.
  - `preferredUsername: string` → OIDC **`preferred_username`** claim (when `profile` scope is requested).
- **Persists these attributes** in `storage.Password` with proper JSON tags and `omitempty` for `emailVerified` to avoid breaking existing serializations.
- **Adds resolver functions** in the local password DB connector:
  - `resolvePasswordName(p)` — returns `Name` if set, falls back to `Username`
  - `resolvePasswordEmailVerified(p)` — returns `*EmailVerified` if set, defaults to `true`
- **Updates local password login/refresh** so `passwordDB` returns `connector.Identity` with these fields correctly resolved in both `Login()` and `Refresh()` methods.
- **Adds tests** validating:
  - config parsing for the new fields (`config_test.go`: `TestUnmarshalConfig` with `name`, `emailVerified: false`, `preferredUsername`)
  - end-to-end ID token claims via password grant with local connector (`handlers_test.go`: `TestHandlePassword_LocalPasswordDBClaims`)
- **Updates config examples** (`config.dev.yaml`, `config.yaml.dist`, `examples/config-dev.yaml`, `examples/k8s/dex.yaml`) with documentation for new fields.
- **Backward compatible**:
  - existing configs without new fields keep working unchanged
  - `name` defaults to `username` (existing behavior preserved)
  - `emailVerified` defaults to `true` when not specified (existing behavior preserved)
  - `*bool` pointer type for `EmailVerified` distinguishes "not set" from `false`

- The `name` field maps to `connector.Identity.Username` which then flows to `storage.Claims.Username` and finally to `tok.Name` in the ID token (via `oauth2.go` scope=profile handling). This reuses the existing claim pipeline without requiring changes to `storage.Claims` or storage backends.
- `EmailVerified` uses `*bool` (pointer) in `storage.Password` to distinguish between "not configured" (nil → default true) and "explicitly set to false". This is important for backward compatibility.
- The `Refresh()` method in `passwordDB` also uses the resolver functions, ensuring that updated `name`/`emailVerified` values are reflected in refreshed tokens.

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1281,7 +1281,9 @@ func TestPasswordDB(t *testing.T) {
 	s.CreatePassword(ctx, storage.Password{
 		Email:             "jane@example.com",
 		Username:          "jane",
+		Name:              "Jane Doe",
 		PreferredUsername: "jane-public",
+		EmailVerified:     boolPtr(false),
 		UserID:            "foobar",
 		Groups:            []string{"team-a", "team-a/admins"},
 		Hash:              h,
@@ -1301,10 +1303,10 @@ func TestPasswordDB(t *testing.T) {
 			password: pw,
 			wantIdentity: connector.Identity{
 				Email:             "jane@example.com",
-				Username:          "jane",
+				Username:          "Jane Doe",
 				PreferredUsername: "jane-public",
 				UserID:            "foobar",
-				EmailVerified:     true,
+				EmailVerified:     false,
 				Groups:            []string{"team-a", "team-a/admins"},
 			},
 		},
--- a/storage/conformance/conformance.go
+++ b/storage/conformance/conformance.go
@@ -447,6 +447,10 @@ func (n byEmail) Len() int           { return len(n) }
 func (n byEmail) Less(i, j int) bool { return n[i].Email < n[j].Email }
 func (n byEmail) Swap(i, j int)      { n[i], n[j] = n[j], n[i] }
 
+func boolPtr(v bool) *bool {
+	return &v
+}
+
 func testPasswordCRUD(t *testing.T, s storage.Storage) {
 	ctx := t.Context()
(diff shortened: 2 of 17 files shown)


nabokihms (changes requested):
In your previous PR you also added some changes to the storage https://github.com/dexidp/dex/pull/4456/

Do we need to add these new claims as well?

```
func (Password) Fields() []ent.Field {
	return []ent.Field{
		field.Text("email").
			SchemaType(textSchema).
			StorageKey("email"). // use email as ID field to make querying easier
			NotEmpty().
			Unique(),
		field.Bytes("hash"),
		field.Text("username").
			SchemaType(textSchema).
			NotEmpty(),
		field.Text("preferred_username").
			SchemaType(textSchema).
			Default(""),
		field.Text("user_id").
			SchemaType(textSchema).
			NotEmpty(),
		field.JSON("groups", []string{}).
			Optional(),
	}
}
```
This is an example of the database storage schema. Claims are missing.

And I assume because tests didn't fail, new claims should be added to the conformance tests.

Jabejixo:
@nabokihms You're right, thanks for catching that! I've added name and email_verified to all storage backends and updated tests.

nabokihms (approved):

PR #4529, opened 2026-02-11
fix: clean up in-memory connector before create

### LINKED ISSUE #4528: gRPC Connector API references stale in-memory copies
### Preflight Checklist

### Version

2.44.0

### Storage Type

SQLite

### Installation Type

Official container image

### Expected Behavior

Recently I have been heavily using the Connector gRPC API to dynamically add and remove connectors. Sometimes the same connector is removed and added with the same ID but different configuration e.g. `getUserInfo = false` vs `getUserInfo = true`. Encountered weird behaviour where the Dex server seemingly re-used earlier settings. 

### Actual Behavior

Expecting robust and consistent representation of Connectors when managing them through gRPC API

### Steps To Reproduce

1. deploy Dex
2. add Connector with id `a` via gRPC and `getUserInfo = true`
3. remove Connector `a` via gRPC
4. add Connect with id `a` again via gRPC, but this time without `getUserInfo` config (= false?)
5. observe Dex still behaves as if `getUserInfo = true`

### Additional Information

_No response_

### Configuration

`N/A`

### Logs

`N/A`

loafoe:
#### Overview

Fixes stale references to Connector when re-adding under same id

I've investigated and fixed the issue where connectors provisioned via the gRPC API would retain old configurations after being deleted and recreated.

The root cause was that Dex's Server maintains an in-memory cache of opened connectors. When a connector is retrieved, Dex compares its ResourceVersion in storage with the one in the cache. Since the gRPC CreateConnector call hardcodes the ResourceVersion to "1", deleting and recreating a connector with the same ID resulted in matching versions, causing the server to continue using the stale, cached connector object.

I implemented the following changes to resolve this:
1. Added `CloseConnector(id string)` to `Server`: This new method safely removes a connector from the server's in-memory cache.
2. Updated gRPC Handlers: I modified the CreateConnector and DeleteConnector handlers in server/api.go to call CloseConnector. This ensures the cache is invalidated whenever a connector is created or removed, forcing the server to re-initialize it with the latest configuration on its next use.

All existing server tests passed.

Fixes #4528

### DIFF
--- a/server/api_cache_test.go
+++ b/server/api_cache_test.go
@@ -91,7 +91,7 @@ func TestConnectorCacheInvalidation(t *testing.T) {
 	}
 
 	pc2 := c2.Connector.(connector.PasswordConnector)
-	
+
 	// If the fix works, it should now use the second password.
 	_, valid2, err := pc2.Login(ctx, connector.Scopes{}, "user", "second-password")
 	if err != nil || !valid2 {
@@ -125,7 +125,7 @@ func TestConnectorCacheInvalidation(t *testing.T) {
 	}
 
 	pc3 := c3.Connector.(connector.PasswordConnector)
-	
+
 	_, valid3, err := pc3.Login(ctx, connector.Scopes{}, "user", "third-password")
 	if err != nil || !valid3 {
 		t.Errorf("failed to login with third password, UpdateConnector might be missing cache invalidation")


server/api.go
```diff
@@ -470,6 +470,10 @@ func (d dexAPI) CreateConnector(ctx context.Context, req *api.CreateConnectorReq
 		return nil, fmt.Errorf("create connector: %v", err)
 	}
 
+	if d.server != nil {
```
nabokihms, line 474: Didn't get this. Why do we need to close the created connector?

loafoe: Because when the connector is added under the same id again later it may inherit settings from the original one, causing unexpected behaviour

nabokihms: Still hard on this one. We create a connector and then delete a connector. Shouldn't we delete it before creating?

loafoe: So we manage Connectors via gRPC using this [Crossplane provider](https://marketplace.upbound.io/providers/loafoe/provider-dex/v1.10.0) which leverages the Dex gRPC API (`DEX_API_CONNECTORS_CRUD=true`)

What I'm seeing is that if we add and remove a connector with the same `id` but with different `config` [settings](https://github.com/loafoe/provider-dex) the original settings seems to be preserved, at least partially. By deleting the entry from the internal cache we get rid of this issue, that's what this PR does.

nabokihms: As I see it if the connector was correctly deleted after your PR it will be also deleted from the connectors map. After creating a connector we assume that this connector (even with the same id) is not in the connectors map anymore because it was deleted from the map by calling the delete api method. The only reason closing the connector on create is some safety precautions, but in this case it should be clarified by the comment in the code.

loafoe: @nabokihms added a test that demonstrates the issue. The test proves that without the CloseConnector calls in CreateConnector and DeleteConnector, the server would reuse a  stale in-memory instance of the connector if the ResourceVersion happened to match.

loafoe:
@nabokihms added a test that demonstrates the issue. The test proves that without the CloseConnector calls in CreateConnector and DeleteConnector, the server would reuse a stale in-memory instance of the connector if the ResourceVersion happened to match. Also rebased to latest `master`

nabokihms:
@loafoe I tested the test

1. With no changes it fails
```
➜  dex git:(bugfix/connector-crud) ✗ git diff | cat
diff --git a/server/api.go b/server/api.go
index 62782b44..724c4807 100644
--- a/server/api.go
+++ b/server/api.go
@@ -470,10 +470,6 @@ func (d dexAPI) CreateConnector(ctx context.Context, req *api.CreateConnectorReq
                return nil, fmt.Errorf("create connector: %v", err)
        }
 
-       if d.server != nil {
-               d.server.CloseConnector(req.Connector.Id)
-       }
-
        return &api.CreateConnectorResp{}, nil
 }
 
@@ -542,11 +538,6 @@ func (d dexAPI) DeleteConnector(ctx context.Context, req *api.DeleteConnectorReq
                d.logger.Error("api: failed to delete connector", "err", err)
                return nil, fmt.Errorf("delete connector: %v", err)
        }
-
-       if d.server != nil {
-               d.server.CloseConnector(req.Id)
-       }
-
        return &api.DeleteConnectorResp{}, nil
 }
 
➜  dex git:(bugfix/connector-crud) ✗ go test ./server/...
--- FAIL: TestConnectorCacheInvalidation (0.00s)
    api_cache_test.go:98: failed to login with second password, cache might still be stale
    api_cache_test.go:103: unexpectedly logged in with first password, cache is definitely stale
FAIL
FAIL    github.com/dexidp/dex/server    1.101s
?       github.com/dexidp/dex/server/internal   [no test files]
ok      github.com/dexidp/dex/server/signer     (cached)
FAIL
```
2. With closing only on deleting it passes
```
➜  dex git:(bugfix/connector-crud) ✗ git diff | cat
diff --git a/server/api.go b/server/api.go
index 62782b44..eafff924 100644
--- a/server/api.go
+++ b/server/api.go
@@ -470,10 +470,6 @@ func (d dexAPI) CreateConnector(ctx context.Context, req *api.CreateConnectorReq
                return nil, fmt.Errorf("create connector: %v", err)
        }
 
-       if d.server != nil {
-               d.server.CloseConnector(req.Connector.Id)
-       }
-
        return &api.CreateConnectorResp{}, nil
 }
 
➜  dex git:(bugfix/connector-crud) ✗ go test ./server/...
ok      github.com/dexidp/dex/server    1.145s
?       github.com/dexidp/dex/server/internal   [no test files]
ok      github.com/dexidp/dex/server/signer     (cached)
```
3. With closing only on creating it also passes (I understand why, because it was not closed properly on deleting, we need to close it on creating)

```
➜  dex git:(bugfix/connector-crud) ✗ git diff | cat      
diff --git a/server/api.go b/server/api.go
index 62782b44..72750811 100644
--- a/server/api.go
+++ b/server/api.go
@@ -543,10 +543,6 @@ func (d dexAPI) DeleteConnector(ctx context.Context, req *api.DeleteConnectorReq
                return nil, fmt.Errorf("delete connector: %v", err)
        }
 
-       if d.server != nil {
-               d.server.CloseConnector(req.Id)
-       }
-
        return &api.DeleteConnectorResp{}, nil
 }

➜  dex git:(bugfix/connector-crud) ✗ go test ./server/...
ok      github.com/dexidp/dex/server    1.143s
?       github.com/dexidp/dex/server/internal   [no test files]
ok      github.com/dexidp/dex/server/signer     (cached)
```


We need to close on creating only if someone deletes the connector from the database manually (not through the API), e.g., by deleting a row from Postgres or a resource from Kubernetes. This is ok, we can leave closing on creating, but I just want a proper comment in the code for this specific case.

loafoe:
> We need to close on creating only if someone deletes the connector from the database manually (not through the API), e.g., by deleting a row from Postgres or a resource from Kubernetes. This is ok, we can leave closing on creating, but I just want a proper comment in the code for this specific case.

thanks @nabokihms for going through the test code. I've removed the close on delete and left it in the create path with a comment. All tests still pass.

server/api_cache_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// 5. Load it again
+	c2, err := serv.getConnector(ctx, connID)
+	if err != nil {
+		t.Fatalf("failed to get connector second time: %v", err)
+	}
+
+	pc2 := c2.Connector.(connector.PasswordConnector)
+	
+	// If the fix works, it should now use the second password.
```
nabokihms, line 95:
```suggestion
	pc2 := c2.Connector.(connector.PasswordConnector)

	// If the fix works, it should now use the second password.
```

Issue #681, opened 2016-11-11
cmd/dex: ensure config fields are properly populated on dex.


rithujohn191:

ericchiang:
Another one, ensure that connector names aren't empty: https://github.com/coreos/dex/pull/782

PR #4531, opened 2026-02-12
feat: Disallow unknown config fields

nabokihms:
#### Overview

This pull request introduces a new feature flag to optionally enforce strict config validation by disallowing unknown fields in JSON configuration files. It also refactors config unmarshalling logic to use a centralized function that can toggle this strictness based on the feature flag. Additionally, it removes YAML tags from various struct definitions, standardizing on JSON tags only.

It is hard to debug misspellings in the config.

There is a known limitation that only the first error is shown. Other solutions are more complicated.

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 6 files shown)


nabokihms:
Connected https://github.com/dexidp/dex/issues/681

PR #4532, opened 2026-02-12
feat: refactor signer configuration with local and vault options

nabokihms:
#### Overview

1. Moved signer configs to the dedicated `signer` package
2. Added deprecated warning and moved the signing keys rotation interval settings to the local signer config
3. Added handling of specific signing options through the `config` field, like we already do for storage and connectors

Refactoring before the release


nabokihms:
Since it is just a refactoring and the release is coming, I'm going to merge this.

PR #4553, opened 2026-02-16
feat: enhance git-version script to generate pseudo-versions

nabokihms:
#### Overview

Generate a pseudo version instead of using just a branch name.

Use next minor version as the pseudo tag

Example:
```
➜  dex git:(master) ✗ make build > /dev/null 2>&1 && ./bin/dex version
Dex Version: v2.45.0-20260216164129-7d4414ea
Go Version: go1.25.0
Go OS/ARCH: darwin arm64
```

The main issue is that when we build images on branches, the tag is equal to the branch name, which is not good. 

image

Security scanners trigger alerts on the pseudo version added to buildinfo by golang compiler.

### DIFF
--- a/.github/workflows/artifacts.yaml
+++ b/.github/workflows/artifacts.yaml
@@ -52,6 +52,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-tags: true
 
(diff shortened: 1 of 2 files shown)


nabokihms:
Now Dex reports a future pseudo version for the master branch
```
docker run ghcr.io/dexidp/dex:master dex version
Dex Version: v2.45.0-20260216171632-955142ba
Go Version: go1.26.0
Go OS/ARCH: linux arm64
```

PR #4580, opened 2026-02-24
fix(mysql): quote `groups` reserved word in query replacer

backkem:
Fixes #4579

`groups` is a reserved word in MySQL >= 8.0.2. Migration 13 (from #4456) fails with:

```
migration 13 statement 2 failed: Error 1064 (42000): ... right syntax to use near 'groups blob'
```

Additionally, since MySQL auto-commits DDL, statement 1 (`preferred_username`) is persisted
despite the transaction rollback, leaving the migration in an unrecoverable state on restart.

### Fix

Add `groups` to the existing reserved word quoting regex in `flavorMySQL.queryReplacers`,
alongside `keys` which is already handled.

### Recovery

Users stuck with a half-applied migration 13 need to drop the orphaned column before restarting:

```sql
ALTER TABLE password DROP COLUMN preferred_username;
```

### DIFF
--- a/storage/sql/sql.go
+++ b/storage/sql/sql.go
@@ -95,7 +95,7 @@ var (
 			// For compound indexes (with two keys) even less.
 			{matchLiteral("text"), "varchar(384)"},
 			// Quote keywords and reserved words used as identifiers.
-			{regexp.MustCompile(`\b(keys)\b`), "`$1`"},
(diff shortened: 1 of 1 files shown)


backkem:
cc @nabokihms . IDK if you're open to cutting a patch release for something like this?

nabokihms:
Thanks, we will wait a little longer for the other bugs and then release a patch.

dextor1944-create:
nice

PR #4583, opened 2026-02-25
feat(oauth2): add client credentials flow with opt-in config flag

### LINKED ISSUE #3660: Implement client credentials flow
### Preflight Checklist

### Problem Description

We have an application, precisely https://docs.k8s.anynines.com/docs/platform-operator/central-management-cluster-setup/oidc-keycloak, that needs the client credentials flow to work.

### Proposed Solution

Implement client credentials flow.

### Alternatives Considered

_No response_

### Additional Information

Maybe also allow dex to run without any connector for the people only using client credentials flow.

matzegebbe:
#### Overview

Add OAuth2 `client_credentials` grant type support, gated behind an opt-in `clientCredentialsEnabled` config flag.

Dex has no built-in support for the `client_credentials` grant, which is the standard OAuth2 flow ([RFC 6749 Section 4.4](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4)) for service-to-service authentication where no end-user is involved. This has been a long-standing community request.

This PR implements the grant and gates it behind a new `clientCredentialsEnabled` config flag that defaults to `false`, following the same pattern used by `passwordConnector` for the `password` grant. This ensures no behavior change for existing deployments.

**Grant behavior:**
- Requires a confidential client (public clients are rejected)
- Authenticates via client ID + secret (HTTP Basic auth)
- Supports `openid`, `email`, `profile`, `groups` scopes
- Returns an ID token when `openid` scope is requested
- Rejects `offline_access` and `federated:id` scopes (no refresh tokens for M2M)
- Claims are derived from the client itself: `aud` = client ID, `sub` = encoded client ID + connector ID, `name`/`preferred_username` = client name (with `profile` scope)

**Configuration:**
```yaml
oauth2:
  clientCredentialsEnabled: true
```

**Security considerations:**
- Opt-in only: the grant is never exposed unless explicitly enabled
- Confidential clients only: public clients cannot use this grant
- No refresh tokens: `offline_access` scope is rejected, limiting token lifetime
- Service identity: the `sub` claim is the client ID, not a user

**Testing:**
- `go build ./...` compiles cleanly
- `go test ./server/... -run TestHandleClientCredentials` — 7 cases pass
- `go test ./server/... -run TestServerSupportedGrants` — 6 cases pass
- `go test ./server/... -run TestHandleDiscovery` — passes
- `go test ./server/...` — full server suite passes
- Manual test: start dex with `clientCredentialsEnabled: true`, request token via `curl`

**Manual verification:**

1. Enable the flag in `examples/config-dev.yaml`:
```yaml
oauth2:
  clientCredentialsEnabled: true
```

2. Build and run dex:
```bash
go run ./cmd/dex serve examples/config-dev.yaml
```

3. Request an access token:
```bash
curl -X POST http://127.0.0.1:5556/dex/token \
  -u "example-app:ZXhhbXBsZS1hcHAtc2VjcmV0" \
  -d "grant_type=client_credentials"
```

4. Request an access token + ID token:
```bash
curl -X POST http://127.0.0.1:5556/dex/token \
  -u "example-app:ZXhhbXBsZS1hcHAtc2VjcmV0" \
  -d "grant_type=client_credentials&scope=openid+profile"
```

Closes https://github.com/dexidp/dex/issues/3660

The gating mechanism works the same way as the `password` grant: `client_credentials` is included in the default allowed grant types list, but only gets added to `allSupportedGrants` (and thus passes the intersection filter in `newServer()`) when `ClientCredentialsEnabled` is `true`. Without the flag, the grant type is silently filtered out.

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -22,6 +23,15 @@ import (
 	"github.com/dexidp/dex/storage/memory"
 	"github.com/dexidp/dex/storage/sql"
 )
+
+func configUnmarshaller(b []byte, v interface{}) error {
+	if featureflags.ConfigDisallowUnknownFields.Enabled() {
+		return json.Unmarshal(b, v)
+	}
+	dec := json.NewDecoder(bytes.NewReader(b))
+	dec.DisallowUnknownFields()
+	return dec.Decode(v)
+}
 
 // Config is the config format for the main application.
 type Config struct {
@@ -109,7 +119,7 @@ func (p *password) UnmarshalJSON(b []byte) error {
 		HashFromEnv       string   `json:"hashFromEnv"`
 		Groups            []string `json:"groups"`
 	}
-	if err := json.Unmarshal(b, &data); err != nil {
+	if err := configUnmarshaller(b, &data); err != nil {
 		return err
 	}
 	*p = password(storage.Password{
@@ -333,7 +343,7 @@ func (s *Storage) UnmarshalJSON(b []byte) error {
 		Type   string          `json:"type"`
 		Config json.RawMessage `json:"config"`
 	}
-	if err := json.Unmarshal(b, &store); err != nil {
+	if err := configUnmarshaller(b, &store); err != nil {
 		return fmt.Errorf("parse storage: %v", err)
 	}
 	f, ok := storages[store.Type]
@@ -346,7 +356,7 @@ func (s *Storage) UnmarshalJSON(b []byte) error {
 		data := []byte(store.Config)
 		if featureflags.ExpandEnv.Enabled() {
 			var rawMap map[string]interface{}
-			if err := json.Unmarshal(store.Config, &rawMap); err != nil {
+			if err := configUnmarshaller(store.Config, &rawMap); err != nil {
 				return fmt.Errorf("unmarshal config for env expansion: %v", err)
 			}
 
@@ -363,7 +373,7 @@ func (s *Storage) UnmarshalJSON(b []byte) error {
 			data = expandedData
 		}
 
-		if err := json.Unmarshal(data, storageConfig); err != nil {
+		if err := configUnmarshaller(data, storageConfig); err != nil {
 			return fmt.Errorf("parse storage config: %v", err)
 		}
 	}
@@ -462,7 +472,7 @@ func (c *Connector) UnmarshalJSON(b []byte) error {
 
 		Config json.RawMessage `json:"config"`
 	}
-	if err := json.Unmarshal(b, &conn); err != nil {
+	if err := configUnmarshaller(b, &conn); err != nil {
 		return fmt.Errorf("parse connector: %v", err)
 	}
 	f, ok := server.ConnectorsConfig[conn.Type]
@@ -475,7 +485,7 @@ func (c *Connector) UnmarshalJSON(b []byte) error {
 		data := []byte(conn.Config)
 		if featureflags.ExpandEnv.Enabled() {
 			var rawMap map[string]interface{}
-			if err := json.Unmarshal(conn.Config, &rawMap); err != nil {
+			if err := configUnmarshaller(conn.Config, &rawMap); err != nil {
 				return fmt.Errorf("unmarshal config for env expansion: %v", err)
 			}
 
@@ -492,7 +502,7 @@ func (c *Connector) UnmarshalJSON(b []byte) error {
 			data = expandedData
 		}
 
-		if err := json.Unmarshal(data, connConfig); err != nil {
+		if err := configUnmarshaller(data, connConfig); err != nil {
 			return fmt.Errorf("parse connector config: %v", err)
 		}
 	}
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -161,6 +161,8 @@ type OAuth2 struct {
 	AlwaysShowLoginScreen bool `json:"alwaysShowLoginScreen"`
 	// This is the connector that can be used for password grant
 	PasswordConnector string `json:"passwordConnector"`
+	// If enabled, the server will support the client_credentials grant type
+	ClientCredentialsEnabled bool `json:"clientCredentialsEnabled"`
 }
 
 // Web is the config format for the HTTP server.
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -161,8 +161,6 @@ type OAuth2 struct {
 	AlwaysShowLoginScreen bool `json:"alwaysShowLoginScreen"`
 	// This is the connector that can be used for password grant
 	PasswordConnector string `json:"passwordConnector"`
-	// If enabled, the server will support the client_credentials grant type
-	ClientCredentialsEnabled bool `json:"clientCredentialsEnabled"`
 }
 
 // Web is the config format for the HTTP server.
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -279,6 +279,9 @@ func runServe(options serveOptions) error {
 	if c.OAuth2.PasswordConnector != "" {
 		logger.Info("config using password grant connector", "password_connector", c.OAuth2.PasswordConnector)
 	}
+	if c.OAuth2.ClientCredentialsEnabled {
+		logger.Info("config client credentials grant enabled")
+	}
 	if len(c.Web.AllowedOrigins) > 0 {
 		logger.Info("config allowed origins", "origins", c.Web.AllowedOrigins)
 	}
@@ -356,6 +359,7 @@ func runServe(options serveOptions) error {
 		SkipApprovalScreen:         c.OAuth2.SkipApprovalScreen,
 		AlwaysShowLoginScreen:      c.OAuth2.AlwaysShowLoginScreen,
 		PasswordConnector:          c.OAuth2.PasswordConnector,
+		ClientCredentialsEnabled:   c.OAuth2.ClientCredentialsEnabled,
 		Headers:                    c.Web.Headers.ToHTTPHeader(),
 		AllowedOrigins:             c.Web.AllowedOrigins,
 		AllowedHeaders:             c.Web.AllowedHeaders,
--- a/cmd/dex/serve.go
+++ b/cmd/dex/serve.go
@@ -614,6 +614,7 @@ func applyConfigOverrides(options serveOptions, config *Config) {
 	if len(config.OAuth2.GrantTypes) == 0 {
 		config.OAuth2.GrantTypes = []string{
 			"authorization_code",
+			"client_credentials",
 			"implicit",
 			"password",
 			"refresh_token",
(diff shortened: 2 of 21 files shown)


cmd/dex/config.go
```diff
@@ -161,6 +161,8 @@ type OAuth2 struct {
 	AlwaysShowLoginScreen bool `json:"alwaysShowLoginScreen"`
 	// This is the connector that can be used for password grant
 	PasswordConnector string `json:"passwordConnector"`
+	// If enabled, the server will support the client_credentials grant type
+	ClientCredentialsEnabled bool `json:"clientCredentialsEnabled"`
```
nabokihms: We already have `GrantTypes` option. How is this option different? How do they coexist?

cmd/dex/serve.go
```diff
@@ -610,6 +614,7 @@ func applyConfigOverrides(options serveOptions, config *Config) {
 	if len(config.OAuth2.GrantTypes) == 0 {
 		config.OAuth2.GrantTypes = []string{
 			"authorization_code",
+			"client_credentials",
```
nabokihms: We can disable this by default and add a feature flag to enable it by default instead of a new option.

server/handlers.go
```diff
@@ -1461,6 +1463,97 @@ func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request, cli
...
+	if len(invalidScopes) > 0 {
+		s.tokenErrHelper(w, errInvalidScope, fmt.Sprintf("Client can't request scope(s) %q", invalidScopes), http.StatusBadRequest)
+		return
+	}
+
+	// Build claims from the client itself — no user involved.
+	claims := storage.Claims{
+		UserID:            client.ID,
+		Username:          client.Name,
+		PreferredUsername: client.Name,
```
nabokihms: Is this valid for client credentials?

server/handlers.go
```diff
@@ -1461,6 +1463,97 @@ func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request, cli
...
+	}
+
+	connID := "client_credentials"
```
nabokihms: This is a weak spot. The value must be reserved. Otherwise it will overlap with connectors.

nabokihms (changes requested):
Hello! There are some points to resolve before merging.

cmd/dex/config.go
matzegebbe: You're right. Having both is redundant. I initially considered adding a dedicated feature flag (`DEX_CLIENT_CREDENTIALS_ENABLED`) as a hard kill-switch, but that would mean operators need to configure two things (env var + grantTypes), which is unnecessarily complex.

Instead I have removed the `ClientCredentialsEnabled` config option entirely and now rely on the existing `grantTypes` mechanism. `client_credentials` is always *supported* by the server (added to `allSupportedGrants`), but it's **not included in the default `grantTypes` list**. Operators opt in by explicitly adding `"client_credentials"` to their `grantTypes` config - the same way configure any other grant type.

This keeps gating simple and consistent with how other grants work.

cmd/dex/serve.go
matzegebbe: Done. I have **removed `client_credentials` from the default `grantTypes` list** in `applyConfigOverrides()`. The grant is now always *supported* but never *active* unless the operator explicitly lists it in their config

server/handlers.go
matzegebbe: Fair point for a pure M2M flow there is no "user", so unconditionally populating `Username` and `PreferredUsername` is misleading. I've changed this to be **scope-conditional**:

- `UserID` is **always** set to `client.ID` (this is the service identity).
- `Username` and `PreferredUsername` are **only** set to `client.Name` when the `profile` scope is explicitly requested.

This follows the OIDC spec pattern where profile claims are tied to the `profile` scope. If a client just requests `openid`, they get `sub` (the client ID encoded with the connector ID) but no name claims.

If you do not like it we can remove it; During my testing I though it can be handy to have the `name` of the client

server/handlers.go
matzegebbe: Whoopsie.  True. A user could create a connector with ID `"client_credentials"` and break the assumption. I've addressed this in two ways:

1. **Changed the connector ID to `"__client_credentials"`** (double-underscore prefix). This is a reserved namespace for internal use.

2. **Added validation in `CreateConnector` (gRPC API)** that rejects any connector ID starting with `__`

examples/config-dev.yaml
```diff
@@ -100,10 +100,11 @@ telemetry:
 #   format: "text" # can also be "json"
 
 # Default values shown below
-# oauth2:
+#oauth2:
```
nabokihms, line 103:
```suggestion
# oauth2:
```

server/server.go
```diff
@@ -254,6 +254,8 @@ func newServer(ctx context.Context, c Config) (*Server, error) {
 		allSupportedGrants[grantTypePassword] = true
 	}
 
+	allSupportedGrants[grantTypeClientCredentials] = true
```
nabokihms, line 257: My idea to wrap it with a feature gate is that we open an attack surface by introducing the new grant. 
I'd like to do this in three steps:
1. Introduce the feature flag smth like `client_credential_grant_enabled_by_default` with the false default in the next dex release
2. After one or two releases, set the default to true (which allows users to still make it false easily via the feature flag without changing their configs).
3. Remove the feature flag.

It gives users a grace period between 2 and 3.

server/handlers.go
```diff
@@ -1461,6 +1463,104 @@ func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request, cli
...
+		if scope == scopeProfile {
+			claims.Username = client.Name
+			claims.PreferredUsername = client.Name
+			break
+		}
+	}
+
+	connID := "__client_credentials"
+
+	accessToken, expiry, err := s.newAccessToken(ctx, client.ID, claims, scopes, "", connID)
```
nabokihms: nonce is missed here, should be retrieved from the form, I guess

Somthing like this `nonce := q.Get("nonce")`

server/api.go
```diff
@@ -439,6 +440,10 @@ func (d dexAPI) CreateConnector(ctx context.Context, req *api.CreateConnectorReq
 		return nil, errors.New("no id supplied")
 	}
 
+	if strings.HasPrefix(req.Connector.Id, "__") {
```
nabokihms: Connectors can be registered through the config file, so we also need a validation on config parsing.
```
		if c.ID == "" || c.Name == "" || c.Type == "" {
			return fmt.Errorf("invalid config: ID, Type and Name fields are required for a connector")
		}
```

I started thinking that maybe leaving the connector ID empty is the better way instead of using a special value.

nabokihms (changes requested):

server/server.go
matzegebbe: fully agree on the phased rollout and I can sleep way better :D 

Added `DEX_CLIENT_CREDENTIAL_GRANT_ENABLED_BY_DEFAULT` (default `false`) in `pkg/featureflags/set.go`. The `client_credentials` grant is always registered in `allSupportedGrants` so it can be explicitly enabled via `oauth2.grantTypes` in config, but the default code path (no explicit grant config) now skips it unless the feature flag is set to `true`.

1. Start dex **without** the feature flag (default behavior):
   ```bash
   go run ./cmd/dex/ serve examples/config-dev.yaml
   ```
   Check the discovery endpoint — `client_credentials` should **not** appear in `grant_types_supported`:
   ```bash
   curl -s http://127.0.0.1:5556/dex/.well-known/openid-configuration | jq .grant_types_supported
   ```

2. Start dex **with** the feature flag:
   ```bash
   DEX_CLIENT_CREDENTIAL_GRANT_ENABLED_BY_DEFAULT=true go run ./cmd/dex/ serve examples/config-dev.yaml
   ```
   client_credentials` should now appear in `grant_types_supported`.
   ```bash
    curl -s http://127.0.0.1:5556/dex/.well-known/openid-configuration | jq .grant_types_supported
    [
    "authorization_code",
    "client_credentials",
    "refresh_token",
    "urn:ietf:params:oauth:grant-type:device_code",
    "urn:ietf:params:oauth:grant-type:token-exchange"
    ]
   ```

server/handlers.go
matzegebbe: image

Good catch, fixed. The nonce is now read from the form via `r.Form.Get("nonce")` and passed to both `newAccessToken` and `newIDToken`.

server/api.go
matzegebbe: Agreed, using an empty connector ID is cleaner and avoids the need for reserved prefix validation entirely. Changed `connID` from `"__client_credentials"` to `""` in the handler, removed the `__` prefix check from `CreateConnector` in `api.go` (along with the now-unused `strings` import), and removed the associated test.

server/server.go
```diff
@@ -254,6 +255,8 @@ func newServer(ctx context.Context, c Config) (*Server, error) {
...
```
matzegebbe, line 257: Note: `client_credentials` is always registered in `allSupportedGrants` because it only requires a client with a secret - unlike `password`, which needs a configured PasswordConnector to function at all. Whether `client_credentials` is enabled by default is controlled by the `DEX_CLIENT_CREDENTIAL_GRANT_ENABLED_BY_DEFAULT` feature flag in `cmd/dex/serve.go`.

server/handlers.go
```diff
@@ -1461,6 +1463,106 @@ func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request, cli
...
+		if scope == scopeProfile {
+			claims.Username = client.Name
+			claims.PreferredUsername = client.Name
+			break
+		}
+	}
+
+	nonce := r.Form.Get("nonce")
+
+	connID := ""
```
nabokihms, line 1549:
```suggestion
	// Empty connector ID is unique for cluster credentials grant
	// Creating connectors with an empty ID with the config and API is prohibited
	connID := ""
```

pkg/featureflags/set.go
```diff
@@ -17,4 +17,8 @@ var (
 
 	// ConfigDisallowUnknownFields enables to forbid unknown fields in the config while unmarshaling.
 	ConfigDisallowUnknownFields = newFlag("config_disallow_unknown_fields", false)
+
+  // ClientCredentialGrantEnabledByDefault enables the client_credentials grant type by default
+	// without requiring explicit configuration in oauth2.grantTypes.
+	ClientCredentialGrantEnabledByDefault = newFlag("client_credential_grant_enabled_by_default", false)
```
nabokihms:
```suggestion
	// ClientCredentialGrantEnabledByDefault enables the client_credentials grant type by default
	// without requiring explicit configuration in oauth2.grantTypes.
	ClientCredentialGrantEnabledByDefault = newFlag("client_credential_grant_enabled_by_default", false)
```

nabokihms (approved):
@sagikazarmark, since the feature is opt-in and not enabled by default, I see no problem merging this. If you have no objections, I will merge the PR.

@matzegebbe thank you for the PR.

matzegebbe:
Hi @nabokihms,

thanks again for reviewing and merging this PR!

Do you already have an idea when the next Dex release is planned? I’d like to know when this feature might become available in an official release.

Thanks a lot!

nabokihms:
Late May / beginning of June

orange-hbenmabrouk:
> Late May / beginning of June

I guess the release got postponed ? is there an estimate date ?
Thanks

jarrettprosser:
I'm also interested in a release with this feature added, thanks!

tmiddlet2666:
I'm also interested in a release with this feature, very useful.

PR #4611, opened 2026-03-05
feat(oauth2): implement Identity Assertion JWT (ID-JAG) issuance (DEP #4600)

kanywst:
#### Overview

Implement ID-JAG (Identity Assertion JWT) token issuance via Token Exchange, per DEP #4600.

Adds a new branch to the existing Token Exchange flow that issues ID-JAG tokens (`typ: oauth-id-jag+jwt`) when
`requested_token_type=urn:ietf:params:oauth:token-type:id-jag`. This enables cross-domain access brokered by the IdP,
as specified in draft-ietf-oauth-identity-assertion-authz-grant-02.

Key points:

- Disabled by default; enabled via `oauth2.tokenExchange.tokenTypes`
- Per-client `idJAGPolicies` in `staticClients` control allowed audiences/scopes (default-deny)
- Validates subject_token audience against client_id, rejects public clients
- OIDC Discovery extended with `identity_chaining_requested_token_types_supported`
- Prometheus counters and structured logging for observability

Phase 2 (accepting ID-JAG as upstream identity via RFC 7523) is out of scope.

Connected DEP: https://github.com/dexidp/dex/pull/4600

Config example in `config.yaml.dist` and `examples/config-dev.yaml`.


kanywst:
Hi @nabokihms @AlwxSin,

DEP [#4600 ](https://github.com/dexidp/dex/pull/4600)has been merged, and this is the implementation PR. Everything described in the DEP is covered here.

Would appreciate a review when you get a chance.

Thanks!

server/handlers.go
```diff
@@ -1814,6 +1837,193 @@ func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request, cli
...
+	if s.idJAGRequestsTotal != nil {
+		s.idJAGRequestsTotal.WithLabelValues(result).Inc()
+	}
+
+	s.tokenErrHelper(w, errType, errDesc, status)
+}
+
+// extractJWTSubAndAud extracts the "sub" and "aud" claims from a JWT without
+// verifying the signature. The aud claim may be a string or []string.
+func extractJWTSubAndAud(token string) (sub string, aud []string, err error) {
```
AlwxSin: I'm concerned that `extractJWTSubAndAud` doesn't verify signature. 
* `subject_token` is an ID token issued by this same server and should be verified against the server's own signing keys. Otherwise an attacker can impersonate any user
* Also, this func doesn't check whether the subject token has expired. An expierd ID token can be exchanged for a fresh ID-JAG indefinitely

server/policy.go
```diff
@@ -0,0 +1,98 @@
...
+			matched = p
+			break
+		}
+		if p.ClientID == "*" && matched == nil {
+			matched = p
+		}
+	}
+
+	if matched == nil {
+		return PolicyResult{
```
AlwxSin, line 46: The function returns both a non-nill `error` and sets `Denied: true`, which is confusing. I'm not sure which returned variable I should look at.
In Go convention, errors should signal unexpected failures, not business-logic denials.

server/signer/vault.go
```diff
@@ -179,6 +179,88 @@ func (v *vaultSigner) Sign(ctx context.Context, payload []byte) (string, error)
 	return fmt.Sprintf("%s.%s.%s", headerB64, payloadB64, signatureB64), nil
 }
 
+func (v *vaultSigner) SignWithType(ctx context.Context, payload []byte, tokenType string) (string, error) {
```
AlwxSin, line 101: Thi func is almost identical to `Sign` func.
Consider to separate common func logic
```
func Sign() {
  return sign(false)
}

func SignWithType() {
  return sign(true)
}

func sign(withType bool) {}
```

server/server.go
```diff
@@ -432,6 +467,25 @@ func newServer(ctx context.Context, c Config) (*Server, error) {
 
 		c.PrometheusRegistry.MustRegister(requestCounter, durationHist, sizeHist)
 
+		// ID-JAG metrics.
```
AlwxSin: Consider adding a check if ID-JAG is enabled. Otherwise there are phantom metrics that never increments, adding noise to monitoring dashboards

server/handlers.go
```diff
@@ -829,6 +836,13 @@ func (s *Server) finalizeLogin(ctx context.Context, identity connector.Identity,
 		userIdentity = &ui
 	}
 
+	// Skip approval if user already consented to the requested scopes for this client.
```
AlwxSin: This check seems unrelated to token exchange. It modifies core auth flow

server/oauth2.go
```diff
@@ -298,6 +300,62 @@ func (s *Server) newAccessToken(ctx context.Context, clientID string, claims sto
...
 }
 
+// idJAGTyp is the JWT "typ" header value for ID-JAG tokens.
+const idJAGTyp = "oauth-id-jag+jwt"
+
+// idJAGClaims is the JWT payload for an ID-JAG token.
+type idJAGClaims struct {
+	Issuer   string `json:"iss"`
+	Subject  string `json:"sub"`
+	Audience string `json:"aud"`
```
AlwxSin, line 313: I don;t know how to follow the RFC in a best way.
RFC allows `aud` to be either a string or an array. Current draft assumes only a string. So, if the ID-JAG draft allows multi-audience in the future, this will need a breaking change.

server/policy.go
kanywst: Fixed in c07a63af. Denials now return `Denied: true` with nil error

server/signer/vault.go
kanywst: Done. Fixed in 9fd8d05f.

server/server.go
kanywst: Fixed in 4c6664d7.

server/handlers.go
kanywst: Removed.  Fixed in cc4bd0f9.

server/oauth2.go
kanywst: Added a comment noting that audience is a single string per the current draft (draft-ietf-oauth-identity-assertion-authz-grant-02). 

If multi-audience is added in a future revision, we can update then. Fixed in 9ce3053d.

server/handlers.go
kanywst: Thanks. 

Fixed in cc4bd0f9. This verifies signature, expiry, issuer, and audience.

server/policy.go
AlwxSin: Now no code path returns a non-nil error. Can we we change signature to `func evaluateIDJAGPolicy(policies []TokenExchangePolicy, clientID, audience string, scopes []string) PolicyResult {}` without an error?

server/server.go
```diff
@@ -149,6 +155,21 @@ type Config struct {
...
 
+// TokenExchangeConfig holds configuration for Token Exchange support.
+type TokenExchangeConfig struct {
+	TokenTypes []string `json:"tokenTypes"`
+}
+
+// IDJAGEnabled reports whether the ID-JAG token type is enabled.
+func (c TokenExchangeConfig) IDJAGEnabled() bool {
+	for _, t := range c.TokenTypes {
+		if t == "urn:ietf:params:oauth:token-type:id-jag" {
```
AlwxSin: There is a const `tokenTypeIDJAG` in `oauth2.go`

server/handlers.go
```diff
@@ -1814,6 +1830,138 @@ func (s *Server) handleTokenExchange(w http.ResponseWriter, r *http.Request, cli
...
+	}
+
+	// connector_id is required for identifying the upstream connector.
+	if connectorID == "" {
+		s.idJAGReject(ctx, w, "rejected", errInvalidRequest, "Missing required parameter connector_id for ID-JAG token exchange.", http.StatusBadRequest,
+			"client_id", client.ID, "connector_id", connectorID, "audience", audience, "resource", resource, "requested_scope", requestedScope, "reason", "missing_connector_id")
+		return
+	}
+
+	if _, err := s.getConnector(ctx, connectorID); err != nil {
```
AlwxSin, line 1955: Please, add a comment explaining why the result from `s.getConnector` is unused. I figured it out only after checking log message

server/server.go
kanywst: Fixed in https://github.com/dexidp/dex/pull/4611/changes/98ed3522ecaf72c91701bbb0d37773fb443c2f81

server/handlers.go
kanywst: Thanks. Added a comment in https://github.com/dexidp/dex/pull/4611/changes/f397972fa1b9da604e53f77372c8882f7da72e08

AlwxSin (approved):
Great job. Thanks.

Issue #4231, opened 2025-07-22
Declare connectors as "hidden" or "backend-only"


jeengbe:
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

In Dex, all configured connectors are currently shown on the login "select connector" screen, regardless of their intended usage. This becomes problematic when some connectors are only used for backend purposes, such as token exchange, and should not be visible to end users. Additionally, these backend-only connectors often do not require redirect URIs, yet Dex expects a full redirect flow for all connectors.

For example, the following additional Google connector is needed to exchange Google's identity/access tokens:

```yaml
id: google-te
name: Google (Token Exchange)
type: oidc
config:
  issuer: https://accounts.google.com
  clientID: $GOOGLE_OAUTH_CLIENT_ID
  clientSecret: $GOOGLE_OAUTH_CLIENT_SECRET
  redirectURI: nil
```

### Proposed Solution

Introduce a `backendOnly: true` flag (or similar) in the connector configuration. Connectors marked as backend-only would:

- Be hidden from the "select connector" login screen.
- Not require a specified redirect URI.
- Only be used in programmatic authentication flow.

This would allow administrators to cleanly separate internal token flow connectors from user-facing ones, improving UX and reducing confusion.

### Alternatives Considered

- Specifying a `connector_id` in each client to skip the connector selection screen. While this avoids the UI screen, it doesn’t prevent backend-only connectors from appearing if no `connector_id` is specified.
- Maintaining separate Dex instances for backend vs. frontend use cases, which increases infrastructure complexity.

### Additional Information

This feature would be especially useful in environments where Dex acts as a federation point between multiple identity systems.\
I would be interested in implementing this.

danxmoran:
This would be a big help for us as well!

@jeengbe are you still planning to work on this implementation?

cardoe:
This is implemented in #3818 which I've pulled into my proposed v2.44.90 at ghcr.io/cardoe/dex:v2.44.90

nabokihms:
The case should be covered by https://github.com/dexidp/dex/pull/4619 now.

cardoe:
[...]

PR #4619, opened 2026-03-06
feat(connector): connectors for grants

nabokihms:
#### Overview   

Add per-connector grantTypes configuration that restricts which OAuth2 grant types each connector can be used with. Connectors without grantTypes configured remain unrestricted (backwards compatible).

- New grantTypes field on connectors (config, storage, gRPC API)
- Filtering in handleAuthorization (login page) based on the current flow's grant type
- Rejection in handleConnectorLogin, handleTokenExchange, handlePasswordGrant, and handleRefreshToken when the connector doesn't allow the requested grant type
- Error page when all connectors are filtered out instead of showing an empty list
- SQL migration, ent schema update, Kubernetes type update
- Conformance tests covering GrantTypes round-trip through all storage backends                                                                                     
   
Different connectors often serve different purposes, but today Dex exposes all connectors to all flows unconditionally. This creates UX and security issues:
                                                                                                                                                                                                                                                                                                                                                                                                                   
- Hide irrelevant connectors from the login page. An organization uses an OIDC connector for browser-based SSO and a separate OIDC connector configured for token exchange with a Kubernetes cluster. Without this feature, both connectors appear on the login page, confusing end users who see a "Kubernetes Token Exchange" button they should never click.
- Restrict a connector to token exchange only. A platform team configures a connector specifically for service-to-service token exchange - e.g., exchanging Kubernetes service account tokens for Dex tokens. This connector should never accept interactive logins, but today any client can initiate an authorization code flow against it.
- Disable refresh tokens for specific connectors. A security-sensitive connector (e.g., for a contractor IdP) should issue short-lived tokens only. By omitting refresh_token from grantTypes, the connector cannot be used to obtain refresh tokens, enforcing session time limits regardless of client configuration.
- Separate password grant connectors. An LDAP connector is configured as the `passwordConnector` for legacy CLI tools, but the same connector also appears on the login page. With `grantTypes: ["password"]`, it is hidden from the browser UI while remaining available for password grant requests.


nabokihms:
Connected https://github.com/dexidp/dex/issues/4231

server/api.go
```diff
@@ -487,7 +488,7 @@ func (d dexAPI) UpdateConnector(ctx context.Context, req *api.UpdateConnectorReq
 		return nil, errors.New("no email supplied")
 	}
 
-	if len(req.NewConfig) == 0 && req.NewName == "" && req.NewType == "" {
+	if len(req.NewConfig) == 0 && req.NewName == "" && req.NewType == "" && len(req.NewGrantTypes) == 0 {
```
AlwxSin: UpdateConnector can’t clear restrictions (can’t set grantTypes to “unrestricted”). Is that how it was intended?

server/handlers.go
AlwxSin, line 1031: Refresh token issuance is not restricted by grantTypes (only refresh exchange is). But there is the check at `server/refreshhandlers.go:205`

AlwxSin, line 1313: Same thing about grantType restrictions

server/server.go
```diff
@@ -57,6 +57,20 @@ const LocalConnector = "local"
 type Connector struct {
 	ResourceVersion string
 	Connector       connector.Connector
+	GrantTypes      []string
+}
+
+// GrantTypeAllowed checks if the given grant type is allowed for this connector.
+func (c Connector) GrantTypeAllowed(grantType string) bool {
```
AlwxSin: Duplicated logic here and at `storage/storage.go:397`. Maybe put it in helper func?

server/handlers_test.go
```diff
@@ -1060,6 +1060,46 @@ func TestHandleTokenExchange(t *testing.T) {
 	}
 }
 
+func TestHandleTokenExchangeConnectorGrantTypeRestriction(t *testing.T) {
```
AlwxSin, line 1063: No tests for  `/auth ` login page filtering + “no connectors available” path

server/server.go
nabokihms: Moved it to a dedicated helper function, we do not need to identical methods :+1:

server/handlers.go
nabokihms: Oh my, this is a good catch! Fixed and added a comment.

nabokihms: Same here.

server/api.go
nabokihms: This is a mistake. Migrated to a different API design with which Dex can distinguish an empty slice and an absent slice.

server/handlers_test.go
nabokihms: Added more tests.

AlwxSin (approved):

leogasparini:
Hi @sagikazarmark @AlwxSin when is this planned to be released?

PR #4640, opened 2026-03-13
feat: Add Kerberos support

Jabejixo:
#### Overview

Add Kerberos/SPNEGO Single Sign-On support for the LDAP connector, enabling seamless authentication for users in Active Directory/Kerberos environments without requiring password entry.

This PR adds optional Kerberos (SPNEGO) authentication to the LDAP connector. When enabled, Dex can authenticate users via the `Authorization: Negotiate` header on the password login endpoint, completing the OIDC flow without rendering the password form.

**Why we need it:**
- Enterprise environments with Active Directory heavily rely on Kerberos for SSO
- Users on domain-joined machines can authenticate transparently without entering credentials
- Reduces friction and improves security by leveraging existing Kerberos infrastructure

**Key features:**
- Validates SPNEGO tokens using a service keytab
- Maps Kerberos principals to LDAP usernames (`localpart`, `sAMAccountName`, or `userPrincipalName`)
- Optional realm validation
- Configurable fallback to password form when SPNEGO fails
- Full compatibility with existing LDAP connector functionality

**Configuration example:**
```yaml
connectors:
- type: ldap
  id: corp-ldap
  config:
    # ... existing LDAP config ...
    kerberos:
      enabled: true
      keytabPath: /etc/dex/krb5.keytab
      expectedRealm: EXAMPLE.COM
      usernameFromPrincipal: localpart
      fallbackToPassword: false
```

- The `SPNEGOAware` interface in `connector/spnego.go` is designed to be generic, allowing other connectors to implement SPNEGO in the future if needed
- All Kerberos-specific code is isolated in `connector/ldap/kerberos.go` for maintainability
- Comprehensive test coverage included with mock validators


connector/ldap/kerberos.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+		return tokb, true
+	}
+	return nil, false
+}
+
+// LDAP connector SPNEGO integration
+
+// krbLookupUserHook allows tests to inject a user entry without LDAP queries.
+var krbLookupUserHook func(c *ldapConnector, username string) (ldap.Entry, bool, error)
```
nabokihms: Nit: Can we avoid using a global var here, even if it is private? Can we make it a struct field?

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,332 @@
...
+			p = principal[:i]
+		}
+		return strings.ToLower(p)
+	}
+}
+
+// gokrb5 implementation of KerberosValidator
+
+// context key used by gokrb5 to store credentials in the context
+var ctxCredentialsKey interface{} = "github.com/jcmturner/gokrb5/v8/ctxCredentials"
```
nabokihms: Why is it a var and an interface? Didn't follow the logic.

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,332 @@
...
+	kt     *keytab.Keytab
+	logger *slog.Logger
+}
+
+func newGokrb5ValidatorWithLogger(keytabPath string, logger *slog.Logger) (KerberosValidator, error) {
+	kt, err := keytab.Load(keytabPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load keytab: %w", err)
+	}
+	if fi, err := os.Stat(keytabPath); err != nil || fi.IsDir() {
```
nabokihms: Why do we check file after loading keytab and not before?

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,332 @@
...
+		return "", "", false, fmt.Errorf("no credentials in context")
+	}
+	return id.UserName(), id.Domain(), true, nil
+}
+
+func (v *gokrb5Validator) Challenge(w http.ResponseWriter) { writeNegotiateChallenge(w) }
+
+// ContinueToken attempts to continue the SPNEGO handshake and returns a response token
+// (to be placed into WWW-Authenticate: Negotiate <b64>) if available.
+func (v *gokrb5Validator) ContinueToken(r *http.Request) ([]byte, bool) {
```
nabokihms: Looks like the method has duplicate logic with ValidateRequest

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,332 @@
...
+}
+
+// writeNegotiateChallenge writes a standard 401 Negotiate challenge.
+func writeNegotiateChallenge(w http.ResponseWriter) {
+	w.Header().Set("WWW-Authenticate", "Negotiate")
+	w.WriteHeader(http.StatusUnauthorized)
+}
+
+// mapPrincipal maps a Kerberos principal to LDAP username per configuration.
+func mapPrincipal(principal, realm, mapping string) string {
```
nabokihms: realm is unused? Let's add a mappiing with realm, e.g., principal@realm.

nabokihms (changes requested):
Overall, looks good. Some minor suggestions. And there are also conflicts.

connector/ldap/kerberos.go
Jabejixo: Actually, realm is already validated via ExpectedRealm config. Removed unused parameter.

connector/ldap/kerberos.go
Jabejixo: Agreed, extracted common logic into helper methods

connector/ldap/kerberos.go
Jabejixo: My bad, fixed

connector/ldap/kerberos.go
Jabejixo: Fair point: it's gokrb5's context key. Converted to typed const with explanation

connector/ldap/kerberos.go
Jabejixo: Good catch, moved to struct field

nabokihms (approved):

nabokihms:
All good from my side here. @sagikazarmark do you have any objections to merging this?

KubaFajny:
We tracked a Kerberos/SPNEGO issue in the LDAP connector implementation from branch `add-kerberos-supportt` and ended up with a working fix on our side.

**What we observed**
- Firefox completed the Kerberos login flow.
- Chrome/Edge did not, even with browser allowlist/policy configured.
- With additional tracing, we found three separate issues in the current connector implementation.

**Errors observed in the connector flow**
- `KRB Error: (34) KRB_AP_ERR_REPEAT Request is a replay - replay detected`
- `kerberos: credentials missing in context`
- `err="no credentials in context"`

**Issues found**

1. `AcceptSecContext()` was effectively invoked twice for the same authenticated request.
   - First in `ValidateRequest()`
   - Then again in `ContinueToken()`
   - This correlated with the replay-style failure:
     - `KRB_AP_ERR_REPEAT Request is a replay - replay detected`

2. Credentials were read from the returned context using a typed key:
   - `type ctxCredentialsKeyType string`
   - `const ctxCredentialsKey ctxCredentialsKeyType = "github.com/jcmturner/gokrb5/v8/ctxCredentials"`
   - In practice, this did not match how credentials were stored in the context, so `ctx.Value(...)` returned `nil`.
   - That matched the runtime errors:
     - `kerberos: credentials missing in context`
     - `err="no credentials in context"`
   - Using a plain `string` key resolved the issue for us.

3. On the initial request without `Authorization: Negotiate`, the server was returning a continuation token immediately:
   - `WWW-Authenticate: Negotiate <token>`
   - Firefox tolerated this, but Chrome/Edge did not start SPNEGO from that response.
   - Returning a bare initial challenge instead:
     - `WWW-Authenticate: Negotiate`
     fixed Chrome/Edge interoperability for us.

**What fixed it**
- Refactor validation so the main request path performs one `AcceptSecContext()` call only.
- Return a continuation token only when the client already sent `Authorization: Negotiate ...` and `AcceptSecContext()` returns `ContinueNeeded`.
- Read credentials from context using a plain string key, with optional fallback for the old typed key.

**Effective behavior that worked**
1. No auth header:
   - return bare `401 + WWW-Authenticate: Negotiate`
2. Client sends `Authorization: Negotiate ...`
   - call `AcceptSecContext()` once
   - if `ContinueNeeded`, return `401 + WWW-Authenticate: Negotiate <response-token>`
   - if `Complete`, extract credentials from context and continue

With these changes applied, the connector is now deployed and fully working for our client in production-like use.

I am attaching the patch that implements these changes.

https://gist.github.com/KubaFajny/4d1c7c8f3816591642eb397b62b67ed6

Jabejixo:
Thanks @KubaFajny - you were right on all three points, and what we ended up doing matches your patch pretty closely: single AcceptSecContext path, plain-string ctxCredentialsKey, bare Negotiate on the initial response.

Retested end-to-end in a test cluster against Chrome (with AuthServerAllowlist), Firefox and curl --negotiate, through to the code -> id_token exchange. Chrome now completes SPNEGO instead of looping on 401.

Branch rebased on current master and pushed!

connector/spnego.go
```diff
@@ -0,0 +1,22 @@
...
+// SPNEGOAware is an optional extension for connectors that can authenticate
+// users via Kerberos SPNEGO on the initial GET to the password login endpoint.
+//
+// If handled is true and ident is non-nil, the caller should complete the
+// OAuth flow as with a successful password login. If handled is true and
+// ident is nil, the implementation has already written an appropriate
+// response (e.g., 401 with WWW-Authenticate: Negotiate) and the caller should
+// return without rendering the password form. If handled is false, proceed
+// with the legacy password form flow.
+type SPNEGOAware interface {
```
AlwxSin, line 20: Consider adding `var _ connector.SPNEGOAware = (*ldapConnector)(nil)` check at `ldapConnector`

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,317 @@
...
+
+	sp := v.createSPNEGOService(r)
+	authed, ctx, status := sp.AcceptSecContext(tok)
+
+	switch status.Code {
+	case gssapi.StatusComplete:
+		// Handled below.
+	case gssapi.StatusContinueNeeded:
+		v.logger.Info("kerberos: continuation needed", "message", status.Message)
+		tokb, err := base64.StdEncoding.DecodeString(spnegoIncompleteKRB5B64)
```
AlwxSin, line 189: `spnegoIncompleteKRB5B64` is a constant. Micro optimization - decode it once on startup or using `sync.Once`

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,317 @@
...
+// ctxCredentialsKey is the context key used by gokrb5 to store credentials.
+//
+// gokrb5/v8/spnego declares this as an untyped string const, which is passed
+// into context.WithValue and therefore stored with concrete type `string`.
+// To retrieve credentials the key MUST be a plain string (not a typed alias),
+// otherwise context.Value returns nil because keys are compared by both type
+// and value.
+//
+//nolint:revive,staticcheck // gokrb5 stores credentials under a plain string key; we must match it.
+const ctxCredentialsKey = "github.com/jcmturner/gokrb5/v8/ctxCredentials"
```
AlwxSin, line 60: Well, it's an undocumented implementation detail of gokrb5. It's possible that library can change this key even within a minor version. We should add a check (some kind of `init` or build-time assert), that verifies the matches.

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	mapped := mapPrincipal(res.Principal, c.krbConf.UsernameFromPrincipal)
+	c.logger.Info("kerberos principal mapped",
+		"principal", res.Principal, "realm", res.Realm, "mapped_username", mapped)
+
+	userEntry, err := c.lookupKerberosUser(ctx, mapped)
+	if err != nil {
+		c.logger.Error("kerberos user lookup failed", "principal", res.Principal, "mapped", mapped, "err", err)
+		return nil, true, fmt.Errorf("ldap: user lookup failed for kerberos principal %q: %v", res.Principal, err)
```
AlwxSin, line 263:
```suggestion
		return nil, true, fmt.Errorf("ldap: user lookup failed for kerberos principal %q: %w", res.Principal, err)
```
Is it on purpose?

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,317 @@
...
+	ident, err := c.identityFromEntry(userEntry)
+	if err != nil {
+		c.logger.Info("failed to build identity from LDAP entry after kerberos SPNEGO", "err", err)
+		return nil, true, err
+	}
+	if s.Groups {
+		groups, err := c.groups(ctx, userEntry)
+		if err != nil {
+			c.logger.Info("failed to query groups after kerberos SPNEGO", "err", err)
+			return nil, true, fmt.Errorf("ldap: failed to query groups: %v", err)
```
AlwxSin, line 276:
```suggestion
			return nil, true, fmt.Errorf("ldap: failed to query groups: %w", err)
```

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+		ident.Groups = groups
+	}
+
+	// No password means no user bind happened; only populate ConnectorData when
+	// OfflineAccess was requested so refresh can query groups later.
+	if s.OfflineAccess {
+		refresh := refreshData{Username: mapped, Entry: userEntry}
+		if data, mErr := json.Marshal(refresh); mErr == nil {
+			ident.ConnectorData = data
```
AlwxSin, line 286: Is it ok to assign `nil` data to a `ident.ConnectorData` on error?

server/handlers.go
```diff
@@ -606,6 +606,43 @@ func (s *Server) handlePasswordLogin(w http.ResponseWriter, r *http.Request) {
...
+				if ident != nil {
+					redirectURL, canSkipApproval, err := s.finalizeLogin(r.Context(), *ident, authReq, conn.Connector)
+					if err != nil {
+						s.logger.ErrorContext(r.Context(), "failed to finalize login", "err", err)
+						s.renderError(r, w, http.StatusInternalServerError, "Login error.")
+						return
+					}
+
+					if canSkipApproval {
+						authReq, err = s.storage.GetAuthRequest(ctx, authReq.ID)
```
AlwxSin, line 629: Mix of contexts. Here is a `ctx` but surrounding code uses `r.Context()`

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,317 @@
...
+func writeNegotiateContinuation(w http.ResponseWriter, token []byte) {
+	w.Header().Set("WWW-Authenticate", "Negotiate "+base64.StdEncoding.EncodeToString(token))
+	w.WriteHeader(http.StatusUnauthorized)
+}
+
+// mapPrincipal maps a Kerberos principal to LDAP username per configuration.
+// Supported mappings:
+//   - "localpart" / "samaccountname": extracts username before @ (default)
+//   - "userprincipalname": uses full principal as-is
+func mapPrincipal(principal, mapping string) string {
```
AlwxSin, line 88: `mapPrincipal` lowercases the result which may break LDAP lookups in case-sensitive directories

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+		creds, ok := ident.(*credentials.Credentials)
+		if !ok {
+			innerErr = fmt.Errorf("kerberos: unexpected identity type %T", ident)
+			return
+		}
+		id = creds
+	})
+
+	rec := httptest.NewRecorder()
```
AlwxSin: It's weird to see `httptest` package in production code. Add a comment, why `httptest` used here, please

connector/ldap/ldap.go
```diff
@@ -344,6 +422,17 @@ type ldapConnector struct {
 	usernameAttrs []string
 
 	logger *slog.Logger
+
+	// Kerberos/SPNEGO support. krbEnabled gates TrySPNEGO; krb is nil until
+	// the keytab has been loaded successfully in Open().
+	krbEnabled bool
```
AlwxSin: Does `krbEnabled` field needed? You can gate support with `krb != nil`

connector/ldap/kerberos.go
```diff
@@ -0,0 +1,248 @@
...
+		copyRecorded(rec, w)
+		return nil, true, nil
+	}
+
+	if c.krbConf.ExpectedRealm != "" && !strings.EqualFold(c.krbConf.ExpectedRealm, id.Domain()) {
+		c.logger.Info("kerberos: realm mismatch", "expected", c.krbConf.ExpectedRealm, "actual", id.Domain())
+		if c.krbConf.FallbackToPassword {
+			return nil, false, nil
+		}
+		w.Header().Set("WWW-Authenticate", "Negotiate")
```
AlwxSin: There are `copyRecorded` use above. But here you write into `w`. Is it intended?

examples/ldap/config-ldap.yaml
```diff
@@ -59,6 +59,15 @@ connectors:
...
       nameAttr: cn
 
+    # Optional Kerberos (SPNEGO) SSO. When enabled, Dex will challenge with
+    # WWW-Authenticate: Negotiate on GET and skip the password form on success.
+    #kerberos:
+    #  enabled: true
+    #  keytabPath: /etc/dex/krb5.keytab
+    #  expectedRealm: EXAMPLE.COM
+    #  usernameFromPrincipal: sAMAccountName # or userPrincipalName or localpart
+    #  fallbackToPassword: false
```
AlwxSin, line 69: Consider adding `spn`, `keytabPrincipal` and `maxClockSkew` parameters to the example

AlwxSin (approved):

PR #4641, opened 2026-03-13
fix: use case-insensitive comparison for GitHub orgs to fix RBAC #2499

### LINKED ISSUE #2499: Org definition using incorrect case can silently fail to pull github groups.
### Preflight Checklist

### Version

v2.30.0-dirty

### Storage Type

Kubernetes

### Installation Type

Custom Helm chart

### Expected Behavior

A warning would be logged if the defined org within the github connector does not match the case used in the login name (as returned by the github api).

### Actual Behavior

Defining `A1Example` in the config when the org name is `a1example` will allow successful authentication but with an empty group list, and subsequently fail to apply any group specific access policies.

### Steps To Reproduce

_No response_

### Additional Information

This was discovered using dex bundled within the following helm chart:
https://artifacthub.io/packages/helm/argo/argo-cd/3.29.5#dex

### Configuration

```yaml
argo-cd:
  server:
    configEnabled: true
    config:
      url: https://argo.a1example.com
      dex.config: |-
        connectors:
        - type: github
          id: github
          name: GitHub
          config:
            clientID: $GITHUB_CLIENT_ID
            orgs:
            - name: A1Example
```

### Logs

```shell
Correct case:
`time="2022-04-26T15:02:55Z" level=info msg="login successful: connector \"github\", username=\"Fil Ivovic\", preferred_username=\"fivovic\", email=\"f.ivovic@a1example.co.uk\", groups=[\"a1example:admins\"]"`

Incorrect case:
`time="2022-04-26T15:02:55Z" level=info msg="login successful: connector \"github\", username=\"Fil Ivovic\", preferred_username=\"fivovic\", email=\"f.ivovic@a1example.co.uk\", groups=[]"`
```

venkatamutyala:
#### Overview
Previously, configuring `orgs: [{name: "MyOrg"}]` when the actual org was `myorg` would silently return empty groups, causing users to be denied access despite being valid org members

adds case-insensitive comparison for GitHub organization names

Closes #2499

For testing, I did a docker build, updated my dex install to use this custom image, i changed my github org to have mixed casing and then tested the login via dex github connector and had no issues.

### DIFF
--- a/connector/github/github.go
+++ b/connector/github/github.go
@@ -699,8 +699,8 @@ type org struct {
 // which inserts a bearer token as part of the request.
 func (c *githubConnector) teamsForOrg(ctx context.Context, client *http.Client, orgName string) ([]string, error) {
 	apiURL, groups := c.apiURL+"/user/teams", []string{}
+	warnedCaseMismatch := false
 	for {
-		// https://developer.github.com/v3/orgs/teams/#list-user-teams
 		var (
 			teams []team
 			err   error
@@ -710,7 +710,17 @@ func (c *githubConnector) teamsForOrg(ctx context.Context, client *http.Client, 
 		}
 
 		for _, t := range teams {
-			if t.Org.Login == orgName {
+			// GitHub org names are case-insensitive; the API returns canonical (lowercase) form.
+			// https://docs.github.com/en/rest/teams/teams?apiVersion=2026-03-10#list-teams-for-the-authenticated-user
+			if strings.EqualFold(t.Org.Login, orgName) {
+				if !warnedCaseMismatch && t.Org.Login != orgName {
+					c.logger.Warn(
+						"org name in config differs in case from GitHub canonical form; use lowercase in config to keep group claims consistent",
+						"org", orgName,
+						"canonical", t.Org.Login,
+					)
+					warnedCaseMismatch = true
+				}
 				groups = append(groups, c.teamGroupClaims(t)...)
 			}
 		}
--- a/connector/github/github_test.go
+++ b/connector/github/github_test.go
@@ -126,6 +126,32 @@ func TestUserGroupsWithTeamNameAndSlugFieldConfig(t *testing.T) {
 	})
 }
 
+// tests case-insensitive org matching and the warn log on case mismatch
+func TestTeamsForOrgCaseInsensitive(t *testing.T) {
+	s := newTestServer(map[string]testResponse{
+		"/user/teams": {
+			data: []team{
+				{Name: "team-1", Org: org{Login: "myorg"}},
+				{Name: "team-2", Org: org{Login: "myorg"}},
+				{Name: "team-3", Org: org{Login: "otherorg"}},
+			},
+		},
+	})
+	defer s.Close()
+
+	var logBuf strings.Builder
+	logger := slog.New(slog.NewTextHandler(&logBuf, &slog.HandlerOptions{Level: slog.LevelWarn}))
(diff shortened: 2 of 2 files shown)


connector/github/github.go
```diff
@@ -706,7 +706,7 @@ func (c *githubConnector) teamsForOrg(ctx context.Context, client *http.Client,
 		}
 
 		for _, t := range teams {
-			if t.Org.Login == orgName {
+			if strings.EqualFold(t.Org.Login, orgName) {
```
nabokihms, line 715: Seems good to me, but let's add the comment on why the comparison here is case-insensitive. 

Also, it's kind of a breaking change for people who, for some reason, used an org with an upper case in their configs. Users from these orgs can log in now, but they could not before this change.

nabokihms (changes requested):
I have one comment and a question. The original issue was about the org name, but this PR fixes team names, not the org name. Am I missing something?

venkatamutyala:
> I have one comment and a question. The original issue was about the org name, but this PR fixes team names, not the org name. Am I missing something?

@nabokihms  I think the function is a little misleading. It's fetching the users teams across all orgs and then filters them by org name. It's the comparison on org name where it's silently failing because it's case sensitive. I updated the link to the API call as well in my most recent commit

connector/github/github.go
venkatamutyala: @nabokihms Added a comment and updated the link as well to go straight to the relevant API call.

RE: breaking change. 

Potentially. I would think if you didn't want an org to have access you would have removed the org entirely rather than relying on the current case mismatch as security.

venkatamutyala: @nabokihms thoughts?

nabokihms: Nit: let's make the comment a bit more specific:

```
// GitHub org names are case-insensitive; the API returns canonical (lowercase) form.
// [link-to-the-API-doc]
```

connector/github/github.go
```diff
@@ -706,7 +706,8 @@ func (c *githubConnector) teamsForOrg(ctx context.Context, client *http.Client,
 		}
 
 		for _, t := range teams {
-			if t.Org.Login == orgName {
+			// case-insensitive check to avoid silent failures when config casing is different than GitHub
+			if strings.EqualFold(t.Org.Login, orgName) {
```
nabokihms, line 715: Could you also add tests for this edge-case?

nabokihms, line 715: The warning is required if there is a case mismatch.
1. In the final groups, the ID token org name will be in lowercase, so we need to warn users that it was converted to lowercase
2. we need to suggest using lowercase orgs in the config 

The warning will solve both.

nabokihms (changes requested):
Some suggestions

connector/github/github.go
venkatamutyala: Done. @nabokihms

connector/github/github.go
venkatamutyala: @nabokihms Added. Please let me know if you have any feedback on it.

venkatamutyala: Added. Let me know if you have any feedback on the wording.

venkatamutyala:
@nabokihms sorry for the delay. Been a busy month. Just made the requested changes. Please take another look when you have a moment.

venkatamutyala:
@nabokihms 🙏🏾

venkatamutyala:
Hey @nabokihms any chance you have time to give this a review?

Issue #4560, opened 2026-02-18
Implement Auth Sessions


nabokihms:
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.

### Problem Description

This issue tracks the implementation of native browser sessions in Dex. Currently, Dex operates statelessly regarding the browser, relying on refresh tokens for long-term access. To support standard OIDC features like SSO, prompt handling, and improved UX, we propose introducing a server-side session entity coupled with encrypted browser cookies.

#### Implementation
- Proposal: https://github.com/dexidp/dex/pull/4561

### Proposed Solution

This enhancement aims to bridge the gap in OIDC conformance and enable the following capabilities.

OIDC Features & Logic
- [ ] Single Sign-On (SSO)
  - Modify /auth handler to detect valid session cookies.
  - Skip connector selection if a valid session exists.
  - SSO for Password Connectors: Create session upon successful credential validation.
  - SSO for Callback Connectors: Create session upon successful upstream callback (storing upstream tokens if applicable).

- [ ] Prompt Parameter Handling
  - Support prompt=none: Return error if no session, otherwise return code immediately.
  - Support prompt=login: Ignore existing session and force re-authentication.

- [ ] ID Token Hint
  - Support id_token_hint: Validate the hint against the current session's subject.

User Experience & Lifecycle
- [ ] Remember Me
  - Add "Remember Me" checkbox to login views.
  - Consent Persistence
  - Store granted scopes per client.
  - Skip approval view if scopes are already granted for the specific client.

- [ ] Logout
  - Implement RP-Initiated Logout.
  - Logic to revoke the AuthSession and cascade revocation to associated Refresh Tokens.

- [ ] Security & MFA Foundation
  - 2FA State Management
  - Implement logic to store temporary 2FA state/codes linked to the session before final token issuance.

### Additional Information

Relates Issues/PRs:
- Remember Me:
  - https://github.com/dexidp/dex/pull/4381
  - https://github.com/dexidp/dex/issues/32

- ID Token Hint:
  - https://github.com/dexidp/dex/pull/1125
  - https://github.com/dexidp/dex/issues/990
  - https://github.com/dexidp/dex/pull/1508

- Logout:
[...]

PR #4703, opened 2026-03-31
feat: disconnect upstream refreshing

nabokihms:
#### Overview

PR14: Upstream refresh logic

Connected https://github.com/dexidp/dex/issues/4560

Nobody refreshes upstream on every refresh request. Dex used to do this only because it didn't save user claims, but now we have a cache that updates on every login. There are risks of ending up with stale data, so I'd like to discuss the approach.

### DIFF
--- a/server/refreshhandlers.go
+++ b/server/refreshhandlers.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/dexidp/dex/connector"
+	"github.com/dexidp/dex/pkg/featureflags"
 	"github.com/dexidp/dex/server/internal"
 	"github.com/dexidp/dex/storage"
 )
@@ -105,6 +106,10 @@ func (r *refreshError) Error() string {
 
 func newInternalServerError() *refreshError {
 	return &refreshError{msg: errInvalidRequest, desc: "", code: http.StatusInternalServerError}
+}
+
+func newUpstreamRefreshError(desc string) *refreshError {
+	return &refreshError{msg: errInvalidGrant, desc: desc, code: http.StatusBadGateway}
 }
 
 func newBadRequestError(desc string) *refreshError {
@@ -271,7 +276,7 @@ func (s *Server) refreshWithConnector(ctx context.Context, rCtx *refreshContext,
 		newIdent, err := refreshConn.Refresh(ctx, parseScopes(rCtx.scopes), ident)
 		if err != nil {
 			s.logger.ErrorContext(ctx, "failed to refresh identity", "err", err)
-			return ident, newInternalServerError()
+			return ident, newUpstreamRefreshError(err.Error())
 		}
 
 		return newIdent, nil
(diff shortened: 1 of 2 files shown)


server/refreshhandlers.go
```diff
@@ -107,6 +108,10 @@ func newInternalServerError() *refreshError {
 	return &refreshError{msg: errInvalidRequest, desc: "", code: http.StatusInternalServerError}
 }
 
+func newUpstreamRefreshError(desc string) *refreshError {
+	return &refreshError{msg: errInvalidGrant, desc: desc, code: http.StatusBadGateway}
```
AlwxSin: Is it ok to use `5xx` error here? Standard OAuth clients treat `5xx` as retryable server errors.
Also, every other use of `errInvalidGrant` uses `http.StatusBadRequest` (which should be by spec)

server/refreshhandlers.go
```diff
@@ -271,7 +276,7 @@ func (s *Server) refreshWithConnector(ctx context.Context, rCtx *refreshContext,
 		newIdent, err := refreshConn.Refresh(ctx, parseScopes(rCtx.scopes), ident)
 		if err != nil {
 			s.logger.ErrorContext(ctx, "failed to refresh identity", "err", err)
-			return ident, newInternalServerError()
+			return ident, newUpstreamRefreshError(err.Error())
```
AlwxSin: It's dangerous to pass internal error as a desc. There is possible a leak of sensitive data. I.e. `ldap: refresh for username %q expected DN %q got %q`

server/refreshhandlers.go
```diff
@@ -373,7 +392,26 @@ func (s *Server) updateRefreshToken(ctx context.Context, rCtx *refreshContext) (
...
+			if cachedIdentity != nil {
+				s.logger.WarnContext(ctx, "upstream refresh failed, using cached identity from last login",
+					"err", rerr, "user_id", cachedIdentity.Claims.UserID, "connector_id", rCtx.storageToken.ConnectorID)
+				ident = connector.Identity{
+					UserID:            cachedIdentity.Claims.UserID,
+					Username:          cachedIdentity.Claims.Username,
+					PreferredUsername: cachedIdentity.Claims.PreferredUsername,
+					Email:             cachedIdentity.Claims.Email,
+					EmailVerified:     cachedIdentity.Claims.EmailVerified,
+					Groups:            cachedIdentity.Claims.Groups,
```
AlwxSin: nit. `ConnectorData` is lost (which is fine because of failed upstream). But code looks like it missed one field. Consider adding a comment or explicit `ConnectorData: nil`

AlwxSin (commented):
The DEP explicitly says `UpdateUserIdentity` should be called after a successful `RefreshConnector.Refresh()` to keep claims in sync. 
So call upstream, update `UserIdentity` on success, fallback on failure (what DEP proposed) - this PR is close but missing the "update on success" part. 
@nabokihms what exactly do you want to discuss?

nabokihms:
The main concern is a new behaviour. It is inline with other providers like keycloak, ory or auth0, but for Dex it is a new thing. It is questionable...

nabokihms:
It is also questionable do we really want to update in offline refreshing... maybe we can define a better rule when to call refresh in upstream (and do we really need to?)

AlwxSin:
Ok, I dove in and here are my thoughts.
Correct me if I am wrong somewhere.

| Option | Desc | Problem |
| ---- | ----- | ----- |
| A - Before | Always call upstream, fail on error | Break UX when upstream is unavailable |
| B - This PR | Call upstream, on error fallback to cache | Silent degradation, non-obvious behavior |
| C - Clean | Do not call upstream at all when `sessions=on` | Rely entirely on session TTL |
| D - Hard | Call upstream on schedule | More complex, but more flex |

Option B creates unclear semantics - upstream refresh sometiomes works, then is silently ignored. If we are willing to absorb the upstream error anyway, option C is more logical - simply dont' call `Refresh()` when sessions are enabled. It's simpler, predictable and fairer to the user/admin.

nabokihms:
Most providers do C. The only concern is that we will update user data less frequently... Users will have to log out and log in again to receive new claims, or wait till the session expires. 
 
 However, C is definitely better than B.

server/refreshhandlers.go
AlwxSin, line 477: Second call to storage after `s.updateRefreshToken`. Consider populate `ident` with `LastLogin`

AlwxSin (approved):

PR #4742, opened 2026-04-13
connector/saml: implement SP-Initiated Single Logout via LogoutCallbackConnector

Jabejixo:
#### Overview

Add SAML Single Logout (SLO) support to the SAML connector by implementing
the `LogoutCallbackConnector` interface from #4674.

The SAML connector does not support Single Logout - when a user logs out
through Dex, the upstream IdP session stays active.

This PR adds two methods to the SAML connector:

- **`LogoutURL`** - builds a `<LogoutRequest>` with HTTP-Redirect binding
  (DEFLATE + base64) using `NameID` and `SessionIndex` captured during login.
- **`HandleLogoutCallback`** - validates the IdP's `<LogoutResponse>` for both
  HTTP-Redirect (GET) and HTTP-POST bindings with binding-aware signature
  validation.

New config fields: `sloURL` (IdP's SLO endpoint) and
`insecureSkipSLOSignatureValidation`. Existing `ca`/`caData` certificates
are reused - no additional configuration required.

- No changes to server-side code - uses the existing `LogoutCallbackConnector`
  interface called by `logout.go`.
- `NameID`, `NameIDFormat`, and `SessionIndex` are persisted in `ConnectorData`
  during `HandlePOST` alongside the existing refresh token fields.
- Fixed a latent bug: `nameID.Format` XML tag was missing `attr` — it is an
  attribute, not a child element.
- HTTP-Redirect signature validation (`Signature`/`SigAlg` query params) is
  implemented separately from embedded XML signatures.

### DIFF
--- a/connector/saml/saml.go
+++ b/connector/saml/saml.go
@@ -3,14 +3,19 @@ package saml
 
 import (
 	"bytes"
+	"compress/flate"
 	"context"
+	"crypto/rand"
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
 	"encoding/pem"
 	"encoding/xml"
 	"fmt"
+	"io"
 	"log/slog"
+	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"sync"
@@ -83,6 +88,15 @@ type Config struct {
 	CAData []byte `json:"caData"`
 
 	InsecureSkipSignatureValidation bool `json:"insecureSkipSignatureValidation"`
+
+	// SLOURL is the IdP's Single Logout Service URL (HTTP-Redirect binding).
+	// If empty, SLO is not available for this connector.
+	SLOURL string `json:"sloURL"`
+
+	// InsecureSkipSLOSignatureValidation skips signature validation on SLO responses.
+	// This is insecure and should only be used for testing or when the IdP
+	// does not sign LogoutResponses.
+	InsecureSkipSLOSignatureValidation bool `json:"insecureSkipSLOSignatureValidation"`
 
 	// Assertion attribute names to lookup various claims with.
 	UsernameAttr string `json:"usernameAttr"`
@@ -164,6 +178,9 @@ func (c *Config) openConnector(logger *slog.Logger) (*provider, error) {
 		logger:        logger,
 
 		nameIDPolicyFormat: c.NameIDPolicyFormat,
+
+		sloURL:                             c.SLOURL,
+		insecureSkipSLOSignatureValidation: c.InsecureSkipSLOSignatureValidation,
 	}
 
 	if p.nameIDPolicyFormat == "" {
@@ -189,7 +206,8 @@ func (c *Config) openConnector(logger *slog.Logger) (*provider, error) {
 		}
 	}
 
-	if !c.InsecureSkipSignatureValidation {
+	needsSLOSigValidation := c.SLOURL != "" && !c.InsecureSkipSLOSignatureValidation
+	if !c.InsecureSkipSignatureValidation || needsSLOSigValidation {
 		if (c.CA == "") == (c.CAData == nil) {
 			return nil, errors.New("must provide either 'ca' or 'caData'")
 		}
@@ -233,8 +251,9 @@ func (c *Config) openConnector(logger *slog.Logger) (*provider, error) {
 }
 
 var (
-	_ connector.SAMLConnector    = (*provider)(nil)
-	_ connector.RefreshConnector = (*provider)(nil)
+	_ connector.SAMLConnector           = (*provider)(nil)
+	_ connector.RefreshConnector        = (*provider)(nil)
+	_ connector.LogoutCallbackConnector = (*provider)(nil)
 )
 
 type provider struct {
@@ -259,12 +278,15 @@ type provider struct {
 
 	nameIDPolicyFormat string
 
+	sloURL                             string
+	insecureSkipSLOSignatureValidation bool
+
 	logger *slog.Logger
 }
 
-// cachedIdentity stores the identity from SAML assertion for refresh token support.
-// Since SAML has no native refresh mechanism, we cache the identity obtained during
-// the initial authentication and return it on subsequent refresh requests.
+// cachedIdentity stores the identity from SAML assertion for refresh token support
+// and SLO (Single Logout). The NameID/NameIDFormat/SessionIndex fields are used
+// to build a SAML LogoutRequest when the user logs out.
 type cachedIdentity struct {
 	UserID            string   `json:"userId"`
 	Username          string   `json:"username"`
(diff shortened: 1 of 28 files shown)


connector/saml/saml.go
```diff
@@ -711,3 +754,265 @@ func before(now, notBefore time.Time) bool {
...
+	if err != nil {
+		rawResp = compressed
+	}
+
+	byteReader := bytes.NewReader(rawResp)
+	if xrvErr := xrv.Validate(byteReader); xrvErr != nil {
+		return fmt.Errorf("saml slo: %w", xrvErr)
+	}
+
+	if !p.insecureSkipSLOSignatureValidation {
```
nabokihms: A few checks we do on the login side that I think we want here too:
- `InResponseTo` - otherwise, a captured valid response could be replayed
- `Destination` and `Issuer` — to prevent recipient/issuer confusion
- Freshness on `IssueInstant`

`InResponseTo` is the important one — would need to stash the outgoing request ID in `LogoutState`. WDYT?

connector/saml/saml.go
```diff
@@ -711,3 +754,265 @@ func before(now, notBefore time.Time) bool {
 func after(now, notOnOrAfter time.Time) bool {
 	return now.After(notOnOrAfter.Add(allowedClockDrift))
 }
+
+// newRequestID generates a random ID suitable for SAML request IDs.
+func newRequestID() string {
+	buf := make([]byte, 16)
+	if _, err := io.ReadFull(rand.Reader, buf); err != nil {
+		panic("crypto/rand failed: " + err.Error())
```
nabokihms: NIT: most of the connector returns errors rather than panicking. Mind making this `(string, error)` to match?

connector/saml/saml.go
```diff
@@ -711,3 +754,265 @@ func before(now, notBefore time.Time) bool {
...
+		if err := json.Unmarshal(connectorData, &ci); err != nil {
+			return "", fmt.Errorf("saml: failed to unmarshal connector data for logout: %v", err)
+		}
+	}
+
+	if ci.NameID == "" {
+		return "", nil
+	}
+
+	req := &logoutRequest{
```
nabokihms, line 937: Most IdPs reject unsigned `LogoutRequest`s in production. You've already got the redirect-binding sign logic for the verify path, so reusing it here should be straightforward. Could be a follow-up if we want to keep this PR focused, but probably worth flagging in the docs either way.

Did you get a chance to test against a real IdP? Curious if it accepted the unsigned request.  Internally, we discussed that you tested with Keycloak, but...

connector/saml/saml.go
```diff
@@ -164,6 +180,9 @@ func (c *Config) openConnector(logger *slog.Logger) (*provider, error) {
 		logger:        logger,
 
 		nameIDPolicyFormat: c.NameIDPolicyFormat,
+
+		sloURL:                             c.SLOURL,
+		insecureSkipSLOSignatureValidation: c.InsecureSkipSLOSignatureValidation,
```
nabokihms: Is there a case where someone trusts login signatures but not logout (or vice versa)? If not, maybe just reuse the existing `InsecureSkipSignatureValidation`? Two near-identically-named flags tend to cause support pain.

nabokihms (changes requested):
Before we merge, two things I'd like us to sort out:

1. The outgoing `LogoutRequest` isn't signed, and most IdPs reject unsigned ones by default.
2. `HandleLogoutCallback` is missing a few checks that login-side already does — `InResponseTo`, `Issuer`, `Destination`. Without `InResponseTo` matching, a captured response could be replayed.

Left some other minor suggestions.

connector/saml/saml.go
```diff
@@ -711,3 +831,393 @@ func before(now, notBefore time.Time) bool {
...
+	if r.Method == http.MethodGet {
+		samlResponse = r.URL.Query().Get("SAMLResponse")
+	} else {
+		if err := r.ParseForm(); err != nil {
+			return fmt.Errorf("saml slo: failed to parse form: %v", err)
+		}
+		samlResponse = r.FormValue("SAMLResponse")
+	}
+
+	if samlResponse == "" {
```
AlwxSin, line 1038: Does it secure? In case of misconfig or when attacker strips the parameter it means that simple GET with valid session cookie can force-logout user without any SAML validation.

server/logout.go
```diff
@@ -192,10 +192,19 @@ func (s *Server) handleLogoutCallback(w http.ResponseWriter, r *http.Request) {
...
 	// Let the connector validate the upstream logout response if it supports it.
+	// Prefer StatefulLogoutCallbackConnector (replays ls.ConnectorState — e.g.
+	// SAML's outgoing LogoutRequest ID for InResponseTo correlation) and fall
+	// back to the simpler LogoutCallbackConnector for stateless connectors.
 	if ls.ConnectorID != "" {
 		conn, err := s.getConnector(ctx, ls.ConnectorID)
 		if err == nil {
-			if logoutConn, ok := conn.Connector.(connector.LogoutCallbackConnector); ok {
+			switch logoutConn := conn.Connector.(type) {
+			case connector.StatefulLogoutCallbackConnector:
```
AlwxSin: In case of OIDC ignore errors is best effort.  But in case of SAML there are crypto checks. 
For potential replay attacks defense is in decorative state (only logging).

connector/saml/saml.go
```diff
@@ -711,3 +831,393 @@ func before(now, notBefore time.Time) bool {
...
+
+	compressed, err := base64.StdEncoding.DecodeString(samlResponse)
+	if err != nil {
+		return fmt.Errorf("saml slo: failed to decode SAMLResponse: %v", err)
+	}
+
+	// HTTP-Redirect binding uses DEFLATE compression; HTTP-POST does not.
+	// Try to inflate; if it fails, treat the data as uncompressed XML.
+	rawResp, err := io.ReadAll(flate.NewReader(bytes.NewReader(compressed)))
+	if err != nil {
```
AlwxSin: Does the comment mean that for GET we should return an error when inflate fails?

AlwxSin (approved):

connector/saml/saml.go
Jabejixo: Implemented all four checks

connector/saml/saml.go
Jabejixo: Done: `newRequestID` now returns `(string, error)`, and `LogoutURLWithState` propagates the error.

connector/saml/saml.go
Jabejixo: Added HTTP-Redirect signing for `LogoutRequest` via `sloSigningKey/sloSigningKeyData`. The original Keycloak test used unsigned requests with “Client signature required” disabled
a fully signed Keycloak setup also requires signing `AuthnRequest`, which Dex does not support yet

connector/saml/saml.go
Jabejixo: Agree, the separate SLO flag was removed

PR #4743, opened 2026-04-13
fix: implement id token hint on RP-Initiated logout

jnfrati:
#### Overview

Hello again! Thanks to all the Dex team for taking the time to review my contributions :heart: 

The purpose of this PR is to introduce the `id_token_hint` when redirecting to upstream on RP-Initiated logout, while testing sessions we encountered the issue that some OIDC (e.g Pocket-id, Authentik) providers force-require the `id_token_hint` to be added when provided a `post_logout_redirect_uri`, this means that the chain RP -> OP(Dex) -> OP(Pocket-id) gets broken due to Dex not sending that extra parameter. 

Even though the current implementation of Dex is not "broken" as it's compliant with the specification, having this extra id_token_hint in the URL makes integration easier when working with a wide variety of third party services. Still, I understand that the intention might not be to send the id_token_hint always, so happy to apply any recommended approach on how to "flag" when to send this or not :raised_hands: 


This PR introduces:
- A new field `AuthSession.ConnectorData []byte`
- Adds logic to store the id_token received from upstream during login under `AuthSession.ConnectorData`
- Implements logic to add the query parameter `id_token_hint` in case connectorData is provided to `LogoutCallbackConnector.LogoutURL`, worth mentioning that:
  - The `id_token_hint` is added if it's present in the connectorData, meaning it does not depend on whether the `post_redirect_logout_uri` is present or not.
  - This is only implemented on oidc connector 

This enables a much smoother integration with OIDC providers that require the `id_token_hint` to validate the identity of the requesting party. 

#### Tests introduced

- `TestLogoutURL`: covers `IDToken` present/absent, malformed connector data, and hint emission without `post_logout_redirect_uri`.
- `TestTryUpstreamLogoutPrefersSessionConnectorData`: verifies precedence of auth-session data over offline-session data, plus fallback behavior.                                                                                                                                                                                                                                                                                                                                                                  
- `storage/conformance` `testAuthSessionCRUD` now asserts `ConnectorData` round-trips through both create and update.  

My assumption on this PR is that the logout process is only done if sessions exist, meaning that there's a higher chance of finding the `id_token` on the `AuthSession.ConnectorData` than in the `OfflineSession` object. This is why in logout, the auth session takes precedence over the offline session, but please let me know if this is a correct assumption!

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -228,6 +228,7 @@ var brokenAuthHeaderDomains = []string{
 // connectorData stores information for sessions authenticated by this connector
 type connectorData struct {
 	RefreshToken []byte
+	IDToken      []byte // raw upstream id_token JWT for RP-Initiated logout
 }
 
 // Detect auth header provider issues for known providers. This lets users
@@ -736,6 +737,9 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 	cd := connectorData{
 		RefreshToken: []byte(token.RefreshToken),
 	}
+	if rawIDToken, ok := token.Extra("id_token").(string); ok {
+		cd.IDToken = []byte(rawIDToken)
+	}
 
 	connData, err := json.Marshal(&cd)
 	if err != nil {
@@ -766,7 +770,7 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 // LogoutURL returns the upstream OIDC provider's end_session_endpoint URL.
 // Per the OIDC RP-Initiated Logout spec, the post_logout_redirect_uri parameter
 // tells the upstream where to redirect after logout.
-func (c *oidcConnector) LogoutURL(_ context.Context, _ []byte, postLogoutRedirectURI string) (string, error) {
+func (c *oidcConnector) LogoutURL(_ context.Context, rawConnectorData []byte, postLogoutRedirectURI string) (string, error) {
 	if c.endSessionURL == "" {
 		return "", nil
 	}
@@ -780,6 +784,16 @@ func (c *oidcConnector) LogoutURL(_ context.Context, _ []byte, postLogoutRedirec
 	if postLogoutRedirectURI != "" {
 		q.Set("post_logout_redirect_uri", postLogoutRedirectURI)
 		q.Set("client_id", c.oauth2Config.ClientID)
+	}
+	// Per the RP-Initiated Logout spec, id_token_hint is independently valid
+	// of post_logout_redirect_uri — include it whenever we have one.
+	if len(rawConnectorData) > 0 {
+		var cd connectorData
+		if err := json.Unmarshal(rawConnectorData, &cd); err == nil {
+			if len(cd.IDToken) > 0 {
+				q.Set("id_token_hint", string(cd.IDToken))
+			}
+		}
 	}
 	u.RawQuery = q.Encode()
 
--- a/server/logout.go
+++ b/server/logout.go
@@ -251,11 +251,19 @@ func (s *Server) tryUpstreamLogout(ctx context.Context, userID, connectorID stri
 	}
 
 	// Check that the session exists — we need it to store logout state.
-	_, err = s.storage.GetAuthSession(ctx, userID, connectorID)
+	session, err := s.storage.GetAuthSession(ctx, userID, connectorID)
 	if err != nil {
 		s.logger.DebugContext(ctx, "logout: no auth session for upstream logout, skipping",
 			"user_id", userID, "connector_id", connectorID)
 		return "", false
+	}
+
+	// The auth session connector data should keep an id_token that will be used as hint for RP-Initiated logout
+	if len(session.ConnectorData) > 0 {
+		connectorData = session.ConnectorData
+		s.logger.DebugContext(ctx, "logout: using auth_session.ConnectorData", "connector_id", connectorID)
+	} else if len(connectorData) == 0 {
+		s.logger.DebugContext(ctx, "logout: no connector data available", "connector_id", connectorID)
 	}
 
 	// Store logout parameters in the session.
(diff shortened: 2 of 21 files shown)


connector/oidc/oidc.go
```diff
@@ -781,6 +785,16 @@ func (c *oidcConnector) LogoutURL(_ context.Context, _ []byte, postLogoutRedirec
 		q.Set("post_logout_redirect_uri", postLogoutRedirectURI)
 		q.Set("client_id", c.oauth2Config.ClientID)
 	}
+	// Per the RP-Initiated Logout spec, id_token_hint is independently valid
+	// of post_logout_redirect_uri — include it whenever we have one.
+	if len(rawConnectorData) > 0 {
+		var cd connectorData
+		if err := json.Unmarshal(rawConnectorData, &cd); err == nil {
```
nabokihms, line 792: NIT: Let's log the error here for debugging

connector/oidc/oidc.go
```diff
@@ -736,6 +737,9 @@ func (c *oidcConnector) createIdentity(ctx context.Context, identity connector.I
 	cd := connectorData{
 		RefreshToken: []byte(token.RefreshToken),
 	}
+	if rawIDToken, ok := token.Extra("id_token").(string); ok {
+		cd.IDToken = []byte(rawIDToken)
+	}
```
nabokihms, line 742: Two issues here:
 
1) `rawIDToken` is already extracted and verified at the top of `createIdentity` (line 563). You can reuse it instead of extracting twice.
2) This is the core of my concern — `IDToken` gets packed into the same `connectorData` struct as `RefreshToken`, and this whole blob ends up in `AuthSession.ConnectorData`. That leaks the upstream refresh token into the auth session table.
I think we should separate session-scoped data from refresh-scoped data at the `connector.Identity` level. Something like:

```go
type Identity struct {
    // existing fields...
    ConnectorData []byte // refresh lifecycle → OfflineSession
    SessionData   []byte // session lifecycle → AuthSession
}
```
 
This is a struct, not an interface — adding a field is backward compatible. Connectors that don't set `SessionData` will have `nil`, no behavior change.
 
Then here you'd do:
 
```go
// ConnectorData — for refresh, as before
cd := connectorData{RefreshToken: []byte(token.RefreshToken)}
identity.ConnectorData = marshal(cd)
 
// SessionData — only what's needed for session lifecycle (logout)
if rawIDToken, ok := token.Extra("id_token").(string); ok {
    sd := sessionData{IDToken: rawIDToken}
    identity.SessionData = marshal(sd)
}
```
 
This way auth session never sees the refresh token. SAML connectors can store `SessionIndex` in `SessionData` later without touching the model.

server/session.go
```diff
@@ -261,6 +261,7 @@ func (s *Server) createOrUpdateAuthSession(ctx context.Context, r *http.Request,
 				old.ClientStates = make(map[string]*storage.ClientAuthState)
 			}
 			old.ClientStates[authReq.ClientID] = clientState
+			old.ConnectorData = authReq.ConnectorData
```
nabokihms, line 264: This overwrites `ConnectorData` on every session update. If we go with the `SessionData` approach above, this becomes `old.SessionData = authReq.SessionData` and carries only what the connector explicitly marked as session-scoped.
 
Same on line 293 (create path).

server/logout.go
```diff
@@ -251,13 +251,21 @@ func (s *Server) tryUpstreamLogout(ctx context.Context, userID, connectorID stri
...
-	_, err = s.storage.GetAuthSession(ctx, userID, connectorID)
+	session, err := s.storage.GetAuthSession(ctx, userID, connectorID)
 	if err != nil {
 		s.logger.DebugContext(ctx, "logout: no auth session for upstream logout, skipping",
 			"user_id", userID, "connector_id", connectorID)
 		return "", false
 	}
 
+	// The auth session connector data should keep an id_token that will be used as hint for RP-Initiated logout
+	if len(session.ConnectorData) > 0 {
```
nabokihms, line 262: The precedence logic (auth session over offline session) makes sense, but the comment says "should keep an id_token" — that's OIDC-specific language in generic server code. With `SessionData` this becomes cleaner:
 
```go
if len(session.SessionData) > 0 {
    connectorData = session.SessionData
}
```
 
No need to explain what's inside — the connector decides.

storage/storage.go
```diff
@@ -455,6 +455,10 @@ type AuthSession struct {
 	// upstream provider. The callback handler reads it back to complete the flow.
 	// Nil when no logout is in progress.
 	LogoutState *LogoutState
+
+	// Connector data is set during login, meant to store information from the
+	// upstream OIDC connector to be used later on logout (id_token)
+	ConnectorData []byte
```
nabokihms, line 461: The comment references OIDC and id_token specifically — this shouldn't be in the generic storage model. With `SessionData` the comment becomes something like "connector-provided data for the session lifecycle, set during login".

storage/conformance/conformance.go
```diff
@@ -1388,6 +1389,7 @@ func testAuthSessionCRUD(t *testing.T, s storage.Storage) {
 		UserAgent:      "TestBrowser/1.0",
 		AbsoluteExpiry: now.Add(24 * time.Hour),
 		IdleExpiry:     now.Add(1 * time.Hour),
+		ConnectorData:  []byte(`{"RefreshToken":"dGVzdA==","IDToken":"ZXlKaGJHY21PaUpTVXpJMU5pSjk="}`),
```
nabokihms, line 1392: Test data includes `RefreshToken` in auth session — this confirms the leak concern. With `SessionData` this should only contain session-scoped data (e.g. `{"IDToken":"..."}`).

nabokihms (changes requested):
Thanks for the PR.
 
I have concerns about the approach of storing full `ConnectorData` on `AuthSession`. The `connectorData` struct in the OIDC connector carries both `RefreshToken` and `IDToken`. Storing it as-is in the auth session means upstream refresh tokens now persist in a second table — that's an unnecessary expansion of the attack surface.
 
More fundamentally, `AuthSession` and `OfflineSession` have different lifecycles and different needs. They shouldn't share the same opaque blob. See my inline comments for a proposed alternative.

PR #4758, opened 2026-04-25
fix(refreshhandlers): preserve existing username

### LINKED ISSUE #4458: Claim `name` for Google connector is not present on tokens resulting from refreshing
### Preflight Checklist

### Version

2.44.0

### Storage Type

Kubernetes

### Installation Type

Official Helm chart

### Expected Behavior

I expected that ID tokens created from refreshing with the Google connector would have the `name` claim, like ID tokens resulting from completing a login flow e.g. exchanging a code for a token. I show the value of the `name` claim in the profile card of my web app and suddenly it changes to my fallback (the `email` claim) because newer ID tokens resulting from refreshing do not have the claim. My web app server is stateless and can't store this info anywhere.

### Actual Behavior

The claim `name` is not present on ID tokens resulting from refreshing.

### Steps To Reproduce

1. Set up Dex with the most basic Google connector configuration and a simple static client.
3. Use the client to perform the code exchange flow with the scopes `profile email offline_access` to get an access/ID token and a refresh token from Dex.
4. Observe the `name` claim is present.
5. Use the refresh token to get a new access/ID token.
6. Observe the `name` claim is not present.

### Additional Information

_No response_

### Configuration

```yaml

```

### Logs

```shell

```

hisamafahri:
#### Overview

The name claim is missing in ID tokens after using `refresh_token` grant with certain OIDC providers (e.g., Google). 

During initial authentication, Google returns the name field in the ID token. However, when refreshing tokens, Google's OAuth2 endpoint does not include the name field in the refreshed ID token. This causes the name claim to be lost in subsequent ID tokens issued by Dex.

Modified the refresh token update logic in `server/refreshhandlers.go` to preserve `username` and `preferred_username` from the stored refresh token when the connector returns empty values. The fix ensures that:

- If the connector returns an empty username but the stored token has one, the old value is preserved
- If the connector returns a username, it updates the stored value
- The preserved username is used in both the stored refresh token and the identity object returned to create new tokens

- I only test it with Google connector
- I test it both manually and added the test case in `server/refreshhandlers_test.go`
- Fixes #4458

### DIFF
--- a/connector/google/google.go
+++ b/connector/google/google.go
@@ -247,6 +247,15 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
 		return identity, fmt.Errorf("oidc: failed to decode claims: %v", err)
 	}
 
+	// Google sometimes do not return username and other claims. It is correct, according to OIDC spec.
+	// One option to solve this is to call the user endpoint, but Google throttles it more aggressive than
+	// the token endpoint. For concurrent refreshes it is an unwanted behavior.
+	// As a tradeoff, dex preserves previous username and preferred username if absent in the ide token
+	// as a way to keep the claims and do not call the userinfo endpoint.
+	if claims.Username == "" {
+		claims.Username = identity.Username
+	}
+
 	if len(c.hostedDomains) > 0 {
 		found := false
 		for _, domain := range c.hostedDomains {


nabokihms (changes requested):
First of all, thanks for digging into the problem. This is an odd (yet OIDC spec-compliant) behavior where Google sometimes returns the name on refreshing and sometimes doesn't.

I'm ok with the idea of preserving the username instead of calling the userinfo endpoint for each refresh, but the current PR touches not only the Google connector, but the refresh flow in general.

I would suggest only changing the Google connectors code, because everything is fine with other providers.

The following patch should do the trick:
```patch
diff --git a/connector/google/google.go b/connector/google/google.go
index 4a8599c0..68bdf18c 100644
--- a/connector/google/google.go
+++ b/connector/google/google.go
@@ -247,6 +247,15 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
        }
 
+       // Google sometimes do not return username and other claims. It is correct, according to OIDC spec.
+       // One option to solve this is to call the user endpoint, but Google throttles it more aggressive than
+       // the token endpoint. For concurrent refreshes it is an unwanted behavior.
+       // As a tradeoff, dex preserves previous username and preferred username if absent in the ide token
+       // as a way to keep the claims and do not call the userinfo endpoint.
+       if claims.Username == "" {
+               claims.Username = identity.Username
+       }
+
        if len(c.hostedDomains) > 0 {
                found := false
                for _, domain := range c.hostedDomains {

```

Thank you again for debugging this. The idea looks good.

hisamafahri:
adjusted @nabokihms

nabokihms (approved):

PR #4759, opened 2026-04-25
docs: document Docker gomplate configuration

### LINKED ISSUE #2645: Include links and information for gomplate
### Preflight Checklist

### Problem Description

DexIDP in the Docker environment (specifically [docker-entrypoint](https://github.com/dexidp/dex/blob/6da5187b474a967d4b096c5a137de048f00f6bbc/cmd/docker-entrypoint/main.go#L65-L80) ) uses gomplate under the hood to process configuration files and interpolation with env vars. This does not seem to be documented anywhere, and only by search though source is one able to determine valid and invalid ways to configure the docker environment for DexIDP. The gomplate docs are an invaluable resource for this information

### Proposed Solution

Expand information related to creating and using templated configurations in the Docker environment and pointing to the [gomplate](https://docs.gomplate.ca) as an additional documentation source.

### Alternatives Considered

None

### Additional Information

_No response_

AkashKumar7902:
#### Overview

Closes #2645.

- Documents that the official container entrypoint renders `.yaml`, `.tpl`, and `.tmpl` config arguments with gomplate before starting dex.
- Clarifies that this preprocessing is separate from dex's built-in `$VAR` config expansion.
- Notes that the Kubernetes example overrides the image entrypoint, so it uses dex config expansion rather than gomplate.

Validation:
- `go test ./cmd/docker-entrypoint -run TestRun -count=1`
- `go test ./cmd/dex -run "TestUnmarshalConfigWithEnv(NoExpand|Expand)$" -count=1`
- `git diff --check`

### DIFF
--- a/examples/k8s/dex.yaml
+++ b/examples/k8s/dex.yaml
@@ -25,6 +25,9 @@ spec:
       containers:
       - image: ghcr.io/dexidp/dex:v2.32.0
         name: dex
+        # This example runs the dex binary directly, so the Docker entrypoint's
+        # gomplate preprocessing is skipped. The $GITHUB_* values below are
+        # expanded by dex's built-in config environment expansion.
         command: ["/usr/local/bin/dex", "serve", "/etc/dex/cfg/config.yaml"]
 
         ports:


nabokihms (changes requested):
@AkashKumar7902, hello and thank you for submitting the PR!

The README file is not the place for this kind of a doc. Please consider moving it to https://dexidp.io/docs/configuration/

AkashKumar7902:
@nabokihms Thanks! You're right — I removed the README block. The PR now only adds a small inline comment in `examples/k8s/dex.yaml` explaining why gomplate doesn't apply when overriding the container `command` (which is a property of that specific example, not general doc).

I'll open a follow-up against the [docs site](https://github.com/dexidp/website) to add the gomplate / `DEX_EXPAND_ENV` content under `docs/configuration/`. PTAL.

cardoe (changes requested):
I don't find it helpful to have this in a comment. It would be better to include something in the docs site.

AkashKumar7902:
Thanks for the direction. I moved this explanation to the docs site instead: dexidp/website#213. Closing this code-repo PR since the remaining change is only the inline Kubernetes example comment that reviewers did not want here.

PR #4777, opened 2026-05-05
feat(google): fetch groups in parallel

### LINKED ISSUE #3929: Recursive fetching for Google Groups causes timeouts ( >1m auth req times)
### Preflight Checklist

### Version

2.41.1

### Storage Type

Kubernetes

### Installation Type

Official Helm chart

### Expected Behavior

This [recursive function](https://github.com/dexidp/dex/blob/v2.38.0/connector/google/google.go#L279) to fetch google groups makes requests for users with more than 100 groups take longer than standard timeout bounds ( > 30s)  

For instance here
https://github.com/argoproj/argo-cd/issues/21420#issuecomment-2596933212

Expected is to be able to not take this long in returning with the groups by either providing a parameter for depth to which groups should be fetched or not to give sub groups in the default response.

### Actual Behavior

The recursive fetching takes longer and causes timeouts.

### Steps To Reproduce

To Reproduce
- Ensure user has membership in more than 100 Google Workspace groups. Based on observations, users with fewer than 90 groups successfully complete the authentication flow, while those with more than 110 groups fail.

-  Ensure Dex is enabled with the Google connector and configured to fetch transitive group memberships as instructed in the Argo CD [documentation](https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/google/#openid-connect-plus-google-groups-using-dex).
  
- Attempt to log in to the Argo CD dashboard using Google as the authentication provider.

Though this the core way to reproduce but as pointed out the recursive code would exhibit this behavior for any client be it Argo or not.

### Additional Information

_No response_

### Configuration

```yaml

```

### Logs

```shell

```

codrutpanea:

Fetches nested (transitive) group membership in parallel (with a concurrency limit) instead of strictly sequential API calls, so large/deep group graphs complete faster.

In my tests the loading times went form 10-15 seconds down to 2-3 seconds. I tested the changes with 20 nested groups. And I'm using this fork for over a month with no problems.

Closes https://github.com/dexidp/dex/issues/3929

### DIFF
--- a/connector/google/google.go
+++ b/connector/google/google.go
@@ -8,7 +8,9 @@ import (
 	"log/slog"
 	"net/http"
 	"os"
+	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
@@ -16,6 +18,7 @@ import (
 	"golang.org/x/exp/slices"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
+	"golang.org/x/sync/errgroup"
 	admin "google.golang.org/api/admin/directory/v1"
 	"google.golang.org/api/impersonate"
 	"google.golang.org/api/option"
@@ -27,6 +30,10 @@ import (
 const (
 	issuerURL                  = "https://accounts.google.com"
 	wildcardDomainToAdminEmail = "*"
+
+	// defaultConcurrentGroupLookups is the limit used when Config.MaxConcurrentGroupLookups
+	// is zero or negative.
+	defaultConcurrentGroupLookups = 10
 )
 
 // Config holds configuration options for Google logins.
@@ -60,6 +67,10 @@ type Config struct {
 
 	// If this field is true, fetch direct group membership and transitive group membership
 	FetchTransitiveGroupMembership bool `json:"fetchTransitiveGroupMembership"`
+
+	// MaxConcurrentGroupLookups limits concurrent Admin Directory API calls when resolving
+	// transitive group membership. If zero or negative, the connector default limit applies.
+	MaxConcurrentGroupLookups int `json:"maxConcurrentGroupLookups"`
 
 	// Optional value for the prompt parameter, defaults to consent when offline_access
 	// scope is requested
@@ -119,6 +130,11 @@ func (c *Config) Open(id string, logger *slog.Logger) (conn connector.Connector,
 	}
 
 	clientID := c.ClientID
+	maxConcurrent := c.MaxConcurrentGroupLookups
+	if maxConcurrent <= 0 {
+		maxConcurrent = defaultConcurrentGroupLookups
+	}
+
 	return &googleConnector{
 		redirectURI: c.RedirectURI,
 		oauth2Config: &oauth2.Config{
@@ -138,6 +154,7 @@ func (c *Config) Open(id string, logger *slog.Logger) (conn connector.Connector,
 		serviceAccountFilePath:         c.ServiceAccountFilePath,
 		domainToAdminEmail:             c.DomainToAdminEmail,
 		fetchTransitiveGroupMembership: c.FetchTransitiveGroupMembership,
+		maxConcurrentGroupLookups:      maxConcurrent,
 		adminSrv:                       adminSrv,
 		promptType:                     promptType,
 	}, nil
@@ -159,6 +176,7 @@ type googleConnector struct {
 	serviceAccountFilePath         string
 	domainToAdminEmail             map[string]string
 	fetchTransitiveGroupMembership bool
+	maxConcurrentGroupLookups      int
 	adminSrv                       map[string]*admin.Service
 	promptType                     string
 }
@@ -272,8 +290,7 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
 
 	var groups []string
 	if s.Groups && len(c.adminSrv) > 0 {
-		checkedGroups := make(map[string]struct{})
-		groups, err = c.getGroups(claims.Email, c.fetchTransitiveGroupMembership, checkedGroups)
+		groups, err = c.getGroups(ctx, claims.Email, c.fetchTransitiveGroupMembership)
 		if err != nil {
 			return identity, fmt.Errorf("google: could not retrieve groups: %v", err)
 		}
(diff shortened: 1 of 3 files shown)


connector/google/google.go
```diff
@@ -300,50 +304,121 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
...
+
+	setError := func(err error) {
+		firstErrOnce.Do(func() {
+			firstErr = err
+			hasError.Store(true)
+		})
+	}
+
+	var traverse func(string)
+	traverse = func(groupEmail string) {
```
nabokihms: This whole block - WaitGroup + sync.Once + atomic.Bool + chan semaphore - is golang.org/x/sync/errgroup with SetLimit(n). It's already a transitive dep. Switching cuts ~40 lines, removes hand-rolled first-error logic, and gives us context cancellation for free. Please rewrite using errgroup.

connector/google/google.go
```diff
@@ -27,6 +30,7 @@ import (
 const (
 	issuerURL                  = "https://accounts.google.com"
 	wildcardDomainToAdminEmail = "*"
+	maxConcurrentGroupLookups  = 10
```
nabokihms: Please move this to Config (keeping 10 as the default). At 200ms per call this already pushes 3000 req/min, which is over the default Directory API per-user quota. Tenants with stricter quotas will see 429s and there's no retry path.

connector/google/google.go
```diff
@@ -300,50 +304,121 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
 // getGroups creates a connection to the admin directory service and lists
 // all groups the user is a member of
 func (c *googleConnector) getGroups(email string, fetchTransitiveGroupMembership bool, checkedGroups map[string]struct{}) ([]string, error) {
```
nabokihms: Once we're issuing parallel API calls, we need a context.Context here. createIdentity already has one - please thread it through getGroups - listGroupEmails - adminSrv.Groups.List().Context(ctx).Do(). Without it, a cancelled or timed-out login keeps burning Google Directory API quota until every in-flight goroutine returns.

connector/google/google.go
```diff
@@ -300,50 +304,121 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
...
-	var userGroups []string
-	var err error
-	groupsList := &admin.Groups{}
-	domain := c.extractDomainFromEmail(email)
-	adminSrv, err := c.findAdminService(domain)
+	if checkedGroups == nil {
```
nabokihms: checkedGroups is now dead - the function is no longer recursive, the closure inside traverse owns the dedup map, and the only call-site passes nil. Please drop the parameter.

nabokihms (changes requested):
Thanks for tackling this. Direction is right, but I'd like a few things addressed before merging.

codrutpanea:
I implemented the suggestions. I also deployed them in my environment to test the changes. Thanks for the feedback. 
Also feel free to change anything in this PR if you think it will be faster and easier to merge it.

connector/google/google.go
```diff
@@ -298,52 +315,102 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
...
-			}
+	if !fetchTransitiveGroupMembership || len(seeds) == 0 {
+		return userGroups, nil
+	}
 
-			checkedGroups[group.Email] = struct{}{}
-			// TODO (joelspeed): Make desired group key configurable
-			userGroups = append(userGroups, group.Email)
+	g, gctx := errgroup.WithContext(ctx)
+	g.SetLimit(c.maxConcurrentGroupLookups)
```
nabokihms: `errgroup.Group.Go()` under `SetLimit(N)` blocks the caller on g.sem <- token{} when N goroutines are already active. Since enqueue calls g.Go recursively from inside an already-running worker, the workers can hold all N slots while simultaneously trying to enqueue more - none can return to release a slot.

Minimal repro: `MaxConcurrentGroupLookups=1`, user -> A -> B. Worker for A fetches parents, calls enqueue(B) -> g.Go blocks on the full sem -> worker A can't return -> hang. Existing tests don't catch it because the graphs are small enough that workers find leaves and return before N concurrent enqueues happen.

The fundamental issue: SetLimit bounds goroutines, but what we actually need to bound is concurrent API calls.

connector/google/google.go
```diff
@@ -298,52 +315,102 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
...
+		if addGroup(groupEmail) {
+			seeds = append(seeds, groupEmail)
 		}
+	}
 
-		for _, group := range groupsList.Groups {
-			if _, exists := checkedGroups[group.Email]; exists {
-				continue
-			}
+	if !fetchTransitiveGroupMembership || len(seeds) == 0 {
```
nabokihms, line 347: Sort is applied below for the transitive branch but not here - so the two paths return groups in different orders depending on FetchTransitiveGroupMembership. Move `sort.Strings(userGroups)` above this early-return so both branches always return alphabetically sorted output.

connector/google/google_test.go
```diff
@@ -32,7 +33,8 @@ var (
 		"groups_2@dexidp.com": {{Email: "groups_0@dexidp.com"}},
 		"groups_0@dexidp.com": {},
 	}
-	callCounter = make(map[string]int)
+	callCounterMu sync.Mutex
```
nabokihms, line 37: Package-level globals here will silently break the day someone adds `t.Parallel()` to these tests - counts will leak between tests, mutex or not. Not blocking this PR, but worth a follow-up: have `testSetup()` allocate and return its own counter, pass it through.

connector/google/google.go
```diff
@@ -298,52 +315,102 @@ func (c *googleConnector) createIdentity(ctx context.Context, identity connector
...
+	domain := c.extractDomainFromEmail(userKey)
+	adminSrv, err := c.findAdminService(domain)
+	if err != nil {
+		return nil, err
+	}
+
+	groupEmails := []string{}
+	groupsList := &admin.Groups{}
+	for {
+		if err := ctx.Err(); err != nil {
```
nabokihms: NIT: redundant — Do() already respects the context attached via .Context(ctx), and any cancellation will surface as the Do() error on the next iteration. Can drop these three lines for readability. Not blocking.

nabokihms (changes requested):
My feedback is mostly addressed - errgroup, ctx, config knob, dead param gone. Nice.
One blocker: recursive `g.Go` under `SetLimit` deadlocks. Need a regression test for it.

Other minor stuff inline. Also please squash to one commit and rename to feat(google): - sort order changed and there's a new config field, not really a refactor anymore (and do not forget to sign the commit, otherwise we cannot merge this).

codrutpanea:
I think i addressed all the feedback, thank you

duckboy81:
bump?

PR #4780, opened 2026-05-07
feat(server): log successful token exchange requests

0x2b3bfa0:
#### Overview

Adds an info-level `"token exchange successful"` log line to `handleTokenExchange`, mirroring the existing `"login successful"` log that `finalizeLogin` emits for the authorization code flow.

The authorization code flow logs a detailed success line in `finalizeLogin` (`server/handlers.go`) with the resolved identity — `user_id`, `username`, `preferred_username`, `email`, `groups`, `connector_id`. This is invaluable for operators auditing who logged in and which groups they had at login time.

The RFC 8693 token exchange flow (`handleTokenExchange`) has no equivalent log line. It only emits error-level logs on failure, so successful CI/CD-style token exchanges are effectively invisible in operator logs — even at debug level, because there is nothing to print on the success path.

This PR adds a single `InfoContext` call after `TokenIdentity` returns, structured the same way as `finalizeLogin`:

- Same fields as `finalizeLogin` for consistency: `connector_id`, `user_id`, `username`, `preferred_username`, `email` (with the same `(unverified)` suffix convention), `groups`.
- Three additional fields specific to this grant: `client_id`, `subject_token_type`, `requested_token_type` — useful operational context when multiple clients or token types use the same connector.
- Different message (`"token exchange successful"` vs `"login successful"`) so log aggregators can distinguish the two flows without keying off `connector_id`.

Example output (JSON handler):

```json
{"time":"...","level":"INFO","msg":"token exchange successful",
 "connector_id":"gitlab-ci","client_id":"argo-cd",
 "user_id":"12345","username":"jane.doe","preferred_username":"",
 "email":"jane.doe@example.com","groups":["gitlab:environment:prod"],
 "subject_token_type":"urn:ietf:params:oauth:token-type:id_token",
 "requested_token_type":"urn:ietf:params:oauth:token-type:access_token"}
```

- Test added: `TestHandleTokenExchangeLogsSuccess` injects a JSON `slog` handler via the test server config, invokes the token exchange flow, and asserts the log line is emitted with the expected fields. Pattern matches `cmd/dex/excluding_handler_test.go` (`bytes.Buffer` + `NewJSONHandler` + `json.Unmarshal`).
- No API, storage, or behavior changes — this is purely observability.
- `gofmt` and `go vet` clean; full `go test ./server/...` passes.

### DIFF
--- a/server/handlers_test.go
+++ b/server/handlers_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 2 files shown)


nabokihms (approved):
Good improvement, thanks!

0x2b3bfa0:
@nabokihms, it looks like tests are failing; can you please re-run them? https://github.com/dexidp/dex/actions/runs/25509431993/job/74879206313?pr=4780#step:2:413

server/handlers_test.go
```diff
@@ -1741,6 +1742,56 @@ func TestHandleTokenExchange(t *testing.T) {
...
+			found = entry
+			break
+		}
+	}
+	require.NotNil(t, found, "expected \"token exchange successful\" log line, got: %s", logBuf.String())
+	require.Equal(t, "INFO", found["level"])
+	require.Equal(t, "mock", found["connector_id"])
+	require.Equal(t, "client_1", found["client_id"])
+	require.Equal(t, "0-385-28089-0", found["user_id"])
+	require.Equal(t, "Kilgore Trout", found["username"])
```
0x2b3bfa0, line 1788: TIL https://en.wikipedia.org/wiki/Kilgore_Trout 😄

nabokihms:
Thanks!

PR #4790, opened 2026-05-11
ci: improve behavior for first time forks

cardoe:
#### Overview

When someone creates a fork, the build of the master branch in GitHub Actions will fail immediately due to no trivy cache being available and the fact that they don't have a Docker Hub user/pass set as a CI secret.

This aims to improve contributor on boarding by making initial forks pass tests so that we can insist they pass tests locally before pushing.


nabokihms (approved):

cardoe:
Thanks for the quick review @nabokihms. Unfortunately due to that GitHub 500 the other container build needs to be restarted as well.

PR #4797, opened 2026-05-16
Expand .dockerignore to reduce unnecessary cache invalidation

### LINKED ISSUE #1968: Stop copying unneccesary files/folders during the building stage
`COPY .  . ` in the Dockefile is always a tricky thing. It copies more than dex needs, which leads to unwanted behavior.
1. `Dockerfile`: Changin anything makes rebuilding the dex.
2. `.git/`: Actually, dex needs git for building to extract version (and other things). However, any new branch, tag, or commit invalidates the downward layers of a docker image.
3. `examples/`: I think the idea is clear.
4. `*.md`: ...

* Add files/folders to .dockerignore (including .git folder).
* Do all git operations outside of the docker context and then send calculated values to docker build via args.

As an alternative, it is possible to migrate to another building tool, e.g., scaffold, bazel, werf, whatever. It looks like an overcomplication, though.

#### Additional context
Solving problems with unnecessary rebuilds will help us with caching layers in the future.

singhvishalkr:
Fixes #1968

The Dockerfile uses `COPY . .` which copies the entire source tree into the build context. This means any change to files that aren't needed for the build (like `.git/`, `docs/`, `examples/`) still invalidates the Docker layer cache.

This is particularly noticeable with `.git/` -- every new commit, branch, or tag causes a full rebuild even though the actual source code hasn't changed.

The fix expands `.dockerignore` to exclude:
- `.git/` and `.gitignore`
- `docs/`, `examples/`, and Markdown files
- Editor and CI configs (`.github/`, `.editorconfig`, etc.)
- Development-only files (`docker-compose*.yaml`, `config.dev.yaml`)
- The `Dockerfile` itself

This keeps the build context minimal and focused on what the image actually needs: Go source files, dependencies (`go.mod`/`go.sum`), `config.docker.yaml`, and `web/` templates.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,33 @@
 bin/
 tmp/
+
+# Git metadata - changes on every commit and invalidates Docker cache
+.git/
+.gitignore
+
+# CI/CD and GitHub files
+.github/
+
+# Documentation - not needed for runtime
+docs/
+*.md
+LICENSE
+
+# Examples and development configs
+examples/
+config.dev.yaml
+config.yaml.dist
+docker-compose*.yaml
+docker-compose*.dist
+
+# Editor and tool configs
+.editorconfig
+.envrc
+.golangci.yaml
+
+# Nix flakes
+flake.lock
+flake.nix
+
+# The Dockerfile itself
+Dockerfile


nabokihms:
@singhvishalkr thanks for the PR!

Could you please run `git rebase HEAD~1 --signoff` to pass the DCO?

singhvishalkr:
Signed off the commit and force-pushed 0dc9ccfe. The DCO check should rerun on the updated branch.

PR #4800, opened 2026-05-19
feat(gitlab): support inherited group claims

AntonKulyashov:
#### Overview

Add inherited group support to the GitLab connector.

GitLab's `/oauth/userinfo` response only includes groups where the user is a direct member. This means Dex can miss inherited group memberships and reject users when group filtering depends on subgroup membership inherited from a parent group.

This PR adds an `inheritedGroups` option to the GitLab connector. When enabled, Dex uses `/api/v4/groups` as the source of truth for group names so inherited memberships are included. To support that flow, the connector requests `read_api` in addition to the existing `read_user` and `openid` scopes.

This PR also keeps `getGroupsPermission` working with inherited groups. Dex first uses permission claims from `/oauth/userinfo` where possible, and only falls back to per-group effective membership lookups when needed. If GitLab returns `403` or `404` for one of those fallback permission lookups, Dex skips the suffix for that group instead of failing the entire authentication flow.

Tests were added for:
- scope selection
- inherited group loading
- pagination
- filtering with inherited groups
- permission enrichment with inherited groups
- graceful handling of forbidden permission lookups

### DIFF
--- a/connector/gitlab/gitlab.go
+++ b/connector/gitlab/gitlab.go
@@ -28,9 +28,11 @@ const (
 	// used to retrieve groups from /oauth/userinfo
 	// https://docs.gitlab.com/ee/integration/openid_connect_provider.html
 	scopeOpenID = "openid"
+)
 
+const (
+	//constants for inheritedGroups flag
 	inheritedGroupsPerPage = 100
-
 	accessLevelMinimalAccess = 5
 	accessLevelGuest         = 10
 	accessLevelPlanner       = 15
@@ -44,16 +46,24 @@ const (
 
 // Config holds configuration options for gitlab logins.
 type Config struct {
-	BaseURL             string   `json:"baseURL"`
-	ClientID            string   `json:"clientID"`
-	ClientSecret        string   `json:"clientSecret"`
-	RedirectURI         string   `json:"redirectURI"`
-	Groups              []string `json:"groups"`
-	UseLoginAsID        bool     `json:"useLoginAsID"`
-	GetGroupsPermission bool     `json:"getGroupsPermission"`
+	// BaseURL is the root URL of the GitLab instance. Defaults to https://gitlab.com.
+	BaseURL string `json:"baseURL"`
+	// ClientID is the OAuth client ID registered in GitLab.
+	ClientID string `json:"clientID"`
+	// ClientSecret is the OAuth client secret registered in GitLab.
+	ClientSecret string `json:"clientSecret"`
+	// RedirectURI is the callback URL configured for the GitLab OAuth application.
+	RedirectURI string `json:"redirectURI"`
+	// Groups limits logins to users who belong to at least one of the configured GitLab groups.
+	Groups []string `json:"groups"`
+	// UseLoginAsID uses the GitLab username as the Dex user ID instead of the numeric GitLab user ID.
+	UseLoginAsID bool `json:"useLoginAsID"`
+	// GetGroupsPermission appends role-qualified entries, such as group:owner, to the groups claim.
+	GetGroupsPermission bool `json:"getGroupsPermission"`
 	// When enabled, Dex uses /api/v4/groups as the source of truth for group names so
 	// inherited memberships are included as well. This requires GitLab's read_api scope.
 	InheritedGroups bool   `json:"inheritedGroups"`
+    // RootCAData is a PEM-encoded CA bundle used to trust custom TLS certificates on the GitLab instance.
 	RootCAData      []byte `json:"rootCAData,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


connector/gitlab/gitlab.go
```diff
@@ -23,9 +23,23 @@ import (
 const (
 	// read operations of the /api/v4/user endpoint
 	scopeUser = "read_user"
+	// read operations of the REST API, including /api/v4/groups
+	scopeReadAPI = "read_api"
 	// used to retrieve groups from /oauth/userinfo
 	// https://docs.gitlab.com/ee/integration/openid_connect_provider.html
 	scopeOpenID = "openid"
+
+	inheritedGroupsPerPage = 100
```
nabokihms: NIT: it is better to move the gitlab constant to a dedicated const block just for readability

connector/gitlab/gitlab.go
```diff
@@ -23,9 +23,23 @@ import (
 const (
 	// read operations of the /api/v4/user endpoint
 	scopeUser = "read_user"
+	// read operations of the REST API, including /api/v4/groups
+	scopeReadAPI = "read_api"
 	// used to retrieve groups from /oauth/userinfo
 	// https://docs.gitlab.com/ee/integration/openid_connect_provider.html
 	scopeOpenID = "openid"
+
```
nabokihms, line 32:
```suggestion
)

const (
```

connector/gitlab/gitlab.go
```diff
@@ -37,7 +51,10 @@ type Config struct {
 	Groups              []string `json:"groups"`
 	UseLoginAsID        bool     `json:"useLoginAsID"`
 	GetGroupsPermission bool     `json:"getGroupsPermission"`
-	RootCAData          []byte   `json:"rootCAData,omitempty"`
+	// When enabled, Dex uses /api/v4/groups as the source of truth for group names so
+	// inherited memberships are included as well. This requires GitLab's read_api scope.
```
nabokihms, line 64: NIT: The comment for single options doesn't describe how it works with other options. Could you please provide comments for other group-related options as well?

connector/gitlab/gitlab.go
```diff
@@ -105,12 +123,17 @@ type gitlabConnector struct {
...
 	// if set to true permissions will be added to list of groups
 	getGroupsPermission bool
+
+	// if set to true inherited groups will be retrieved from /api/v4/groups
+	inheritedGroups bool
 }
 
 func (c *gitlabConnector) oauth2Config(scopes connector.Scopes) *oauth2.Config {
-	gitlabScopes := []string{scopeUser}
+	gitlabScopes := []string{scopeUser, scopeOpenID}
```
nabokihms, line 143:
```suggestion
	gitlabScopes := []string{scopeUser}
```

nabokihms: Shouldn't be required without groups

connector/gitlab/gitlab.go
```diff
@@ -105,12 +123,17 @@ type gitlabConnector struct {
...
 	if c.groupsRequired(scopes.Groups) {
-		gitlabScopes = []string{scopeUser, scopeOpenID}
+		if c.inheritedGroups {
```
nabokihms, line 145:
```suggestion
		gitlabScopes = append(gitlabScopes, scopeOpenID)
		if c.inheritedGroups {
```

connector/gitlab/gitlab.go
```diff
@@ -313,42 +336,268 @@ type userInfo struct {
...
+			return true
+		}
+	}
+	return false
+}
+
+func accessLevelPermission(accessLevel int) (string, bool) {
+	switch accessLevel {
+	case accessLevelMinimalAccess:
+		return "minimal_access", true
```
nabokihms, line 651: Just a question, is it ok to use contant for numbers and not for strings?

connector/gitlab/gitlab.go
```diff
@@ -313,42 +336,268 @@ type userInfo struct {
...
+func permissionMatches(groupPath string, permissions []string) bool {
+	for _, permissionPath := range permissions {
+		if groupPath == permissionPath {
+			return true
+		}
+		if len(groupPath) > len(permissionPath) &&
+			groupPath[0:len(permissionPath)] == permissionPath &&
+			string(groupPath[len(permissionPath)]) == "/" {
+			return true
+		}
```
nabokihms:
```suggestion
		if len(groupPath) <= len(permissionPath) {
				continue		
		}
		if groupPath[0:len(permissionPath)] == permissionPath {
				continue	
		}
		if string(groupPath[len(permissionPath)]) != "/" {
				continue			
		}
		return true
```

This is a more human-readable variant, and this is probably better to comment on each case.

nabokihms (changes requested):
Hello @AntonKulyashov  and thanks for the PR.

The idea is ok, but I think we can work a little bit more on implementation. I've reviewed half of the PR. One generic recommendation is to align the naming of functions. Now the same logic is mixed in many places, and all methods are named like opaque methods. User info, user groups, api Groups, getGroups. getGroups, getGroupPermissions, getUserGreoup, etc.

Let's rename all the group methods and try to build a single logical chain.

connector/gitlab/gitlab.go
AntonKulyashov: as long as we don't use them anywhere else - it is fine

connector/gitlab/gitlab.go
AntonKulyashov: applied with comments

AntonKulyashov:
I've reimagined the functions' naming, merged some of them as well. Please take a look @nabokihms

nabokihms (approved):

nabokihms:
Thanks!

PR #4821, opened 2026-06-11
fix(google): allow "*" wildcard in hostedDomains to accept any hd claim

Shivam-nagar23:
#### Overview

Allow `"*"` as a wildcard entry in the Google connector's `hostedDomains` config so that any non-empty `hd` claim is accepted. This brings the post-token claim check in line with the existing `LoginURL` behavior, which already passes `hd=*` to Google.

When `hostedDomains: ["*"]` is configured, `LoginURL` correctly sets `hd=*` on the Google auth URL, which tells Google to restrict the picker to Workspace (hosted) accounts. However the post-token validation in `createIdentity` did a strict equality check between `claims.HostedDomain` and each configured entry, so the wildcard never matched a real domain and login failed with:

```
oidc: unexpected hd claim <domain>
```

This is a common ask for multi-tenant Workspace deployments where the operator wants to allow any corporate Google account (i.e. block personal `@gmail.com`) without enumerating every domain.

Changes:
- Extract the validation loop into a small `matchHostedDomain(claim, allowed)` helper.
- Treat a `"*"` entry as a wildcard that accepts any **non-empty** `hd` claim. An empty `hd` (personal Gmail) is still rejected when `*` is configured — that's the intent.
- Add table-driven unit tests covering: exact match, exact match within multiple, no match, wildcard accepting any non-empty `hd`, wildcard alongside explicit domains, wildcard rejecting empty `hd`, and empty claim with explicit domains.

- No config-schema change. `hostedDomains` is already `[]string`; this just makes `"*"` a meaningful value end-to-end.
- Backwards compatible: prior to this change, `"*"` was never a valid `hd` claim value, so the new branch only fires for operators who explicitly opt in.
- The existing `LoginURL` logic already uses `"*"` as the wildcard when multiple domains are configured (`connector/google/google.go` line ~180), so this PR makes the two sides of the validation consistent.

```release-note
google: allow `"*"` in `hostedDomains` to accept any non-empty `hd` claim.
```

### DIFF
--- a/connector/google/google.go
+++ b/connector/google/google.go
@@ -168,6 +168,23 @@ func (c *googleConnector) Close() error {
 	return nil
 }
 
+// matchHostedDomain reports whether the given Google "hd" claim is allowed by
+// the configured hostedDomains list. An entry of "*" acts as a wildcard that
+// accepts any non-empty hd claim, allowing operators to require that users
+// authenticate with a Workspace account (any domain) rather than a personal
+// account, without enumerating every domain.
+func matchHostedDomain(claim string, allowed []string) bool {
+	for _, domain := range allowed {
+		if domain == "*" && claim != "" {
+			return true
+		}
+		if claim == domain {
+			return true
+		}
+	}
+	return false
+}
+
 func (c *googleConnector) LoginURL(s connector.Scopes, callbackURL, state string) (string, []byte, error) {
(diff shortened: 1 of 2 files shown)


Shivam-nagar23:
can any of the maintainers look at this please @sagikazarmark @lstoll @vito @fajran @SEJeff @

PR #4830, opened 2026-06-11
fix: retry refresh token updates on serialization failures (SQL & ent)

RonnanSouza:
#### Overview

When several clients call `UpdateRefreshToken` — which performs a read-modify-write inside a SERIALIZABLE transaction — concurrent rotation of the same token makes the database abort one transaction with a serialization failure, e.g.:

​```
time=2026-06-10T09:34:44.629Z level=ERROR msg="failed to update refresh token" err="update refresh token: pq: could not serialize access due to concurrent update"
​```

Neither SQL storage backend retried, so the abort surfaced to clients as a 500.

Add a bounded, jittered retry around the refresh-token update in **both** SQL backends (`storage/sql` and `storage/ent`):
- Up to 8 retries with jittered backoff; each attempt re-runs the transaction in a fresh read-modify-write, so it's safe.
- Per-driver detection of transient failures via `errors.As`: Postgres `40001` / `40P01`, MySQL `1213` / `1205`.
- The retry is an application-level wrapper in both backends, since neither Go's `database/sql` nor [ent](https://github.com/ent/ent/issues/2341) retries transactions natively.
- Re-enables the refresh-token concurrency conformance tests for Postgres / MySQL (and MySQL 8 for ent).

- I've decided to set reasonable hard coded values for the retries and backoff, but we can expose to be configurable if you folks think it is necessary
- Also, I've done this fix only for update refresh token as it is where I faced the issue and I think where the API is more vulnerable, but I can port to cover all the endpoints

### DIFF
--- a/storage/sql/crud.go
+++ b/storage/sql/crud.go
@@ -51,7 +51,7 @@ type jsonEncoder struct {
 func (j jsonEncoder) Value() (driver.Value, error) {
 	b, err := json.Marshal(j.i)
 	if err != nil {
-		return nil, fmt.Errorf("marshal: %v", err)
+		return nil, fmt.Errorf("marshal: %w", err)
 	}
 	return b, nil
 }
@@ -69,7 +69,7 @@ func (j jsonDecoder) Scan(dest interface{}) error {
 		return fmt.Errorf("expected []byte got %T", dest)
 	}
 	if err := json.Unmarshal(b, &j.i); err != nil {
-		return fmt.Errorf("unmarshal: %v", err)
+		return fmt.Errorf("unmarshal: %w", err)
 	}
 	return nil
 }
@@ -91,7 +91,7 @@ func (c *conn) GarbageCollect(ctc context.Context, now time.Time) (storage.GCRes
 
 	r, err := c.Exec(`delete from auth_request where expiry < $1`, now)
 	if err != nil {
-		return result, fmt.Errorf("gc auth_request: %v", err)
+		return result, fmt.Errorf("gc auth_request: %w", err)
 	}
 	if n, err := r.RowsAffected(); err == nil {
 		result.AuthRequests = n
@@ -99,7 +99,7 @@ func (c *conn) GarbageCollect(ctc context.Context, now time.Time) (storage.GCRes
 
 	r, err = c.Exec(`delete from auth_code where expiry < $1`, now)
 	if err != nil {
-		return result, fmt.Errorf("gc auth_code: %v", err)
+		return result, fmt.Errorf("gc auth_code: %w", err)
 	}
 	if n, err := r.RowsAffected(); err == nil {
 		result.AuthCodes = n
@@ -107,7 +107,7 @@ func (c *conn) GarbageCollect(ctc context.Context, now time.Time) (storage.GCRes
 
 	r, err = c.Exec(`delete from device_request where expiry < $1`, now)
 	if err != nil {
-		return result, fmt.Errorf("gc device_request: %v", err)
+		return result, fmt.Errorf("gc device_request: %w", err)
 	}
 	if n, err := r.RowsAffected(); err == nil {
 		result.DeviceRequests = n
@@ -115,7 +115,7 @@ func (c *conn) GarbageCollect(ctc context.Context, now time.Time) (storage.GCRes
 
 	r, err = c.Exec(`delete from device_token where expiry < $1`, now)
 	if err != nil {
-		return result, fmt.Errorf("gc device_token: %v", err)
+		return result, fmt.Errorf("gc device_token: %w", err)
 	}
 	if n, err := r.RowsAffected(); err == nil {
 		result.DeviceTokens = n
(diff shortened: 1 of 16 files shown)


storage/sql/crud.go
matheuscscp, line 349: Running `updater(r)` inside a retry can be unsafe in many scenarios... The least I would require for a feature like this would be to make it opt-in

matheuscscp:
I'd also consider making the hard-coded parameters inputs with defaults (i.e. the number of retries, backoff, jitter). Maybe the hard-coded / defaults are good for some users, but others may need to fine-tune...

storage/sql/crud.go
RonnanSouza: > Running `updater(r)` inside a retry can be unsafe in many scenarios... The least I would require for a feature like this would be to make it opt-in

I don't see how this can be unsafe, if the it re-opens a fresh transaction and re-reads state each attempt. Also, I couldn't think of a scenario where a straight 500 is better than a retry with a reasonable timeout. But if you folks thinks this is necessary to have it merged, I'm happy to adapt

RonnanSouza:
Hey Pimenta, thanks for the review.

> I'd also consider making the hard-coded parameters inputs with defaults (i.e. the number of retries, backoff, jitter). Maybe the hard-coded / defaults are good for some users, but others may need to fine-tune...

I set this way because I saw a similar approach in the [k8s retry mechanism](https://github.com/dexidp/dex/blob/master/storage/kubernetes/storage.go#L900). But I can make it customizable with the enable/disable config from the previous comment.

storage/sql/crud.go
matheuscscp: Running `updater(r)` multiple times can have all sorts of consequences depending on the connector being used. This introduces a large pool of failure modes.

  * For OIDC connectors backed by providers with refresh-token rotation, the first attempt can successfully exchange the upstream refresh token and receive a new one, then lose the SQL transaction. The retry can call the provider again with the old upstream refresh token. Providers like Okta may reject that as reuse, and may invalidate the newly issued token chain.

  * The updater mutates captured state outside the transaction (`newToken`, `lastUsed`, `ident`, `rerr`). If attempt 1 reaches the connector and attempt 2 takes a different branch after re-reading the DB
  row, the final response/offline-session update can be based on state produced by a failed attempt.

  * Even when the upstream refresh call is technically reusable, retrying it can duplicate side effects: extra token endpoint calls, rate-limit pressure, duplicate audit/security events, and slower refresh
  responses under contention.
  
I'd strongly advise not turning on this code by default. In Flux, when we change default functionality that can have a wide impact like this, we ship it under a feature flag. If we don't hear back about problems in that feature and we think it's a good behavior to have by default, we flip the flag's default value later after a few quarters. I think this is what a GA CNCF project with 10k stars like Dex should do for a feature with this impact surface.

RonnanSouza: Fair, I did change the implementation to have a flag to enable the retry, but I'm not sure this is the best way to tackle that, can you PTAL?

matheuscscp (approved):
LGTM! 🚀 

Thanks @RonnanSouza for addressing my concerns 🤗 

For the opt-in to work as a canary we need to tell people about it, can you pls open a PR in the website repo for adding docs? 🙏 Thanks ❤️

RonnanSouza:
> LGTM! 🚀
> 
> Thanks @RonnanSouza for addressing my concerns 🤗
> 
> For the opt-in to work as a canary we need to tell people about it, can you pls open a PR in the website repo for adding docs? 🙏 Thanks ❤️

Yey, thanks for the time reviewing it! Sure, I'll open a PR for the website and for the helm chart for it.

storage/sql/crud.go
alcantarajulio: This one needs a retry too, under concurrent refreshes of the same session this transaction also aborts with 40001. I ran this PR against a wide concurrent refresh burst and all the remaining 500s came from this function. Switching it to `execTxWithRetry` zeroed them out.

matheuscscp:
@alcantarajulio Just to confirm: your diff since Ronnan's last commit is only the following, right? If so, LGTM

```diff
diff --git a/storage/sql/crud.go b/storage/sql/crud.go
index e7829fab..e87c9717 100644
--- a/storage/sql/crud.go
+++ b/storage/sql/crud.go
@@ -774,7 +774,7 @@ func (c *conn) CreateOfflineSessions(ctx context.Context, s storage.OfflineSessi
 }
 
 func (c *conn) UpdateOfflineSessions(ctx context.Context, userID string, connID string, updater func(s storage.OfflineSessions) (storage.OfflineSessions, error)) error {
-	return c.ExecTx(func(tx *trans) error {
+	return c.execTxWithRetry(func(tx *trans) error {
 		s, err := getOfflineSessions(ctx, tx, userID, connID)
 		if err != nil {
 			return err
```

alcantarajulio:
@matheuscscp Yes, that's the only change.

rsafonseca:
@sagikazarmark , any chance you could spare a few moments to take a look at this 🙏

PR #4831, opened 2026-06-12
feat(signer): support EdDSA for the local token signer

### LINKED ISSUE #4442: Support ES256 or EdDSA key for signing tokens
### Preflight Checklist

### Problem Description

We use Dex integrated with ArgoCD, where tokens are stored as cookies and are subject to a 4KB limit. Since the token must contain all of a user's group memberships, we frequently approach this limit.

Currently, Dex uses RS256 for signing, which adds about 344 bytes of overhead for the signature alone. It would be great if Dex could support ES256 or EdDSA for signing. This would reduce the signature size to roughly 87 bytes, freeing up approximately 257 bytes. We consider this a significant saving given the strict 4KB limit.

In addition to the smaller signature size, we believe other users may prefer these algorithms for their speed or specific security requirements.

### Proposed Solution

Make signing key algorithm ES256 orEdDSA by default, or make them available by configuration

### Alternatives Considered

_No response_

### Additional Information

_No response_

somaz94:
#### Overview

Adds EdDSA (Ed25519) as a signing algorithm for the local token signer, completing #4442. The local signer already supported RS256 and ES256, so this fills in the remaining algorithm requested in the issue.

EdDSA produces much smaller signatures than RS256, which matters for setups with a tight token-size budget — the issue describes Argo CD storing tokens as cookies under a 4 KB limit.

- `server/signer/rotation.go`: generate an Ed25519 key when the configured algorithm is `EdDSA`. `server/signer/utils.go` already maps Ed25519 keys to `jose.EdDSA` and `signPayload` is algorithm-generic, so signing, JWKS publication, and the discovery `id_token_signing_alg_values_supported` value already work once a key is generated.
- `server/signer/local.go` and `cmd/dex/config.go`: accept `EdDSA` in the local-signer config validation.
- `server/oauth2.go`: add `EdDSA → SHA-512` to `hashForSigAlg` so `at_hash`/`c_hash` can be computed for EdDSA id_tokens (Ed25519 uses SHA-512 internally).

It also fixes a verifier gap that EdDSA exposes: the `id_token_hint`, `/userinfo`, and `/token/introspect` verifiers build `oidc.NewVerifier` with an `oidc.Config` that does not set `SupportedSigningAlgs`. go-oidc then defaults to `RS256` only and rejects ES256/EdDSA Dex-issued tokens with `malformed jwt` before `signerKeySet.VerifySignature` ever runs. These three sites now pass the same algorithm list `signerKeySet` already uses, extracted into a shared `supportedSigningAlgs`.

> Note: this verifier gap also affects ES256 today, not just EdDSA. I included the fix here because EdDSA cannot work end-to-end without it, but I'm happy to split it into a separate PR if you'd prefer to keep this one signer-only.

Closes #4442

- Tests:
  - EdDSA cases added to the local-signer table tests, plus a new `TestRotationStrategyForAlgorithm`.
  - `TestAccessTokenHashEdDSA` for the `at_hash` mapping.
  - `TestSignerKeySetWithEdDSALocalSigner` and an end-to-end `TestValidateIDTokenHintEdDSA` that drives the real verifier path. The latter fails against the unpatched verifier with `oidc: malformed jwt: unexpected signature algorithm "EdDSA"; expected ["RS256"]`.
- `go test ./server/... ./cmd/...`, `golangci-lint run`, and `gofmt -l` are all clean. No new dependencies (Ed25519 is stdlib).

### DIFF
--- a/cmd/dex/config.go
+++ b/cmd/dex/config.go
@@ -545,7 +545,7 @@ func normalizeLocalSignerConfig(c *signer.LocalConfig) error {
 		c.Algorithm = jose.RS256
 		return nil
 	}
-	if c.Algorithm == jose.RS256 || c.Algorithm == jose.ES256 {
+	if c.Algorithm == jose.RS256 || c.Algorithm == jose.ES256 || c.Algorithm == jose.EdDSA {
 		return nil
 	}
 	return fmt.Errorf("unsupported local signer algorithm %q", c.Algorithm)
--- a/server/authflow/request.go
+++ b/server/authflow/request.go
@@ -162,7 +162,8 @@ type PKCEConfig struct {
 // token so callers can extract Subject, Audience, etc.
 func (h *Handler) validateIDTokenHint(ctx context.Context, hint string) (*oidc.IDToken, error) {
 	verifier := oidc.NewVerifier(h.IssuerURL.String(), &signer.KeySet{Signer: h.Signer}, &oidc.Config{
-		SkipExpiryCheck: true,
+		SupportedSigningAlgs: signer.SupportedSigningAlgStrings(),
+		SkipExpiryCheck:      true,
 		// SkipClientIDCheck is set because the hint may originate from any client that
 		// Dex issued a token to — the caller does not know the expected audience in advance.
 		// The signature verification via signer.KeySet already guarantees the token was
--- a/server/introspection/introspection.go
+++ b/server/introspection/introspection.go
@@ -172,6 +172,12 @@ func (h *Handler) guessTokenType(ctx context.Context, token string) (TokenTypeEn
 		SkipExpiryCheck:   true,
 		SkipIssuerCheck:   true,
 
+		// go-oidc validates the token's alg against SupportedSigningAlgs before
+		// honoring InsecureSkipSignatureCheck; without this it defaults to RS256
+		// only and rejects ES256/EdDSA access tokens, mis-classifying them as
+		// refresh tokens.
+		SupportedSigningAlgs: signer.SupportedSigningAlgStrings(),
+
 		// We skip signature checks to avoid database calls;
 		InsecureSkipSignatureCheck: true,
 	}
@@ -271,7 +277,10 @@ func (h *Handler) introspectRefreshToken(ctx context.Context, token string) (*In
 }
 
 func (h *Handler) introspectAccessToken(ctx context.Context, token string) (*Introspection, error) {
-	verifier := oidc.NewVerifier(h.Issuer, &signer.KeySet{Signer: h.Signer}, &oidc.Config{SkipClientIDCheck: true})
+	verifier := oidc.NewVerifier(h.Issuer, &signer.KeySet{Signer: h.Signer}, &oidc.Config{
+		SupportedSigningAlgs: signer.SupportedSigningAlgStrings(),
+		SkipClientIDCheck:    true,
+	})
 	idToken, err := verifier.Verify(ctx, token)
 	if err != nil {
 		return nil, newIntrospectInactiveTokenError()
--- a/server/logout/logout.go
+++ b/server/logout/logout.go
@@ -352,8 +352,9 @@ func (h *Handler) deleteAuthSession(ctx context.Context, userID, connectorID str
 // audience are intentionally skipped per RP-Initiated Logout.
 func (h *Handler) validateIDTokenHint(ctx context.Context, hint string) (*oidc.IDToken, error) {
 	verifier := oidc.NewVerifier(h.IssuerURL.String(), &signer.KeySet{Signer: h.Signer}, &oidc.Config{
-		SkipExpiryCheck:   true,
-		SkipClientIDCheck: true,
+		SupportedSigningAlgs: signer.SupportedSigningAlgStrings(),
+		SkipExpiryCheck:      true,
+		SkipClientIDCheck:    true,
 	})
 	return verifier.Verify(ctx, hint)
 }
--- a/server/server_introspection_test.go
+++ b/server/server_introspection_test.go
@@ -11,12 +11,15 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-jose/go-jose/v4"
 	"github.com/stretchr/testify/require"
 
 	"github.com/dexidp/dex/server/internal"
 	"github.com/dexidp/dex/server/introspection"
+	"github.com/dexidp/dex/server/signer"
 	"github.com/dexidp/dex/server/tokens"
 	"github.com/dexidp/dex/storage"
+	"github.com/dexidp/dex/storage/memory"
 )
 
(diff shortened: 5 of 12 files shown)


nabokihms (changes requested):
`guessTokenType` (introspectionhandler.go) wasn't touched — still no `SupportedSigningAlgs`. go-oidc checks the alg before honoring `InsecureSkipSignatureCheck`, so it defaults to RS256 and an EdDSA token fails to parse. `guessTokenType` then returns `RefreshToken`, and a valid EdDSA access token comes back `active: false`. The `introspectAccessToken` fix never runs for it (same for ES256).

Same one-line fix as the other sites + an e2e introspection test with EdDSA.

somaz94:
Thanks, good catch. That was exactly the path I missed.

Added `SupportedSigningAlgs: supportedSigningAlgStrings()` to the `guessTokenType` verifier config in introspectionhandler.go, so an EdDSA (or ES256) access token is no longer mis-classified as a refresh token.

Also added `TestHandleIntrospectEdDSA`, which drives the full `/token/introspect` endpoint with an EdDSA-signed access token and asserts `active: true` (it fails without the fix). Local run:

```
--- PASS: TestHandleIntrospect (0.01s)
--- PASS: TestHandleIntrospectEdDSA (0.00s)
ok  github.com/dexidp/dex/server
```

Pushed in ed39215. Ready for another look.

somaz94:
Thanks for catching this @nabokihms — you're right that `guessTokenType` was the missing piece.

Fixed: `guessTokenType` in `introspectionhandler.go` now sets `SupportedSigningAlgs: supportedSigningAlgStrings()` on its verifier config (same one-line fix as the other sites), so go-oidc no longer defaults to RS256-only and mis-classifies a valid EdDSA/ES256 access token as a refresh token. Added `TestHandleIntrospectEdDSA` — an e2e introspection test that signs an access token with an EdDSA local signer and asserts it comes back `active: true` (and that `guessTokenType` classifies it as an access token).

Also rebased on `master` (the recent server refactor moved `handleUserInfo` into `server/userinfo.go`; relocated the verifier change there). Full EdDSA suite green.

somaz94:
Rebased onto `master` (`8524e3b1`). The conflict was substantial — the `server` package split landed in between, so this is a port rather than a mechanical rebase, and it changed the shape of the fix for the better.

**What moved.** The three files this PR touched (`server/oauth2.go`, `server/userinfo.go`, `server/introspectionhandler.go`) are gone; their contents now live in `server/oauth2/`, `server/userinfo/`, `server/introspection/`, and `signerKeySet` became the exported `signer.KeySet` in `server/signer/keyset.go`.

**What that improved.** Previously the algorithm list and its string helper were unexported in `server/oauth2.go`, next to one of its four consumers. Since `signer.KeySet.VerifySignature` is the thing that actually parses the JWS, the list now lives with it as `signer.SupportedSigningAlgs` / `signer.SupportedSigningAlgStrings()`, and `VerifySignature` consumes its own list instead of a duplicated literal. Happy to move it elsewhere if you'd rather it sat in `tokens`.

**One extra site.** There are now five `oidc.NewVerifier` call sites, not four — `server/logout/logout.go` arrived with the split and had the same RS256-only default. It's wired up here too.

Also `hashForSigAlg` moved to `server/tokens/claims.go`, so the `jose.EdDSA: sha512.New` entry went with it.

**On the introspection test.** `guessTokenType` is now unexported inside `server/introspection`, so the direct unit call from the old test is no longer reachable from package `server`. I kept the end-to-end half, which is the stronger assertion anyway — it drives the real `/token/introspect` endpoint and therefore exercises `guessTokenType` on the actual path. Verified it's a real regression test, not a passing no-op: dropping just the `SupportedSigningAlgs` line from `guessTokenType` makes it fail with `{"active": false}`, which is the original bug.

`gofmt`, `go build ./...`, `go vet ./server/... ./cmd/...` and `go test ./server/... ./cmd/...` are all clean locally.

Your earlier `guessTokenType` point is carried over intact; nothing about that changed in the port.

somaz94:
@nabokihms Thanks for the catch — you were right, and both parts are now addressed.

**1. `guessTokenType` now sets `SupportedSigningAlgs`.**
Exactly as you described: go-oidc validates `alg` before honoring
`InsecureSkipSignatureCheck`, so it defaulted to RS256 and rejected the EdDSA
token, and `guessTokenType` fell through to `RefreshToken`. Fixed in
`server/introspection/introspection.go`, with a comment recording the reasoning
so it does not regress:

```go
// go-oidc validates the token's alg against SupportedSigningAlgs before
// honoring InsecureSkipSignatureCheck; without this it defaults to RS256
// only and rejects ES256/EdDSA access tokens, mis-classifying them as
// refresh tokens.
SupportedSigningAlgs: signer.SupportedSigningAlgStrings(),
```

I applied the same to `introspectAccessToken`, which had the identical gap.
As you noted, this covers ES256 as well, not just EdDSA.

**2. Added the e2e introspection test.**
`TestHandleIntrospectEdDSA` drives the real `/token/introspect` endpoint with a
server configured with an EdDSA signer and asserts the access token comes back
active. I verified it is a genuine regression test — reverting only the
`guessTokenType` line makes it fail with exactly the symptom you predicted:

```
- "token_use": "access_token"
+ "active": false
```

**Note on the diff size:** the PR was re-based onto the `server/` decomposition
(#4910/#4911/#4915), so the changes now live in `server/introspection/`,
`server/userinfo/`, `server/authflow/`, and `server/tokens/` instead of the
files that were removed. The EdDSA signer core in `server/signer/` is unchanged.

CI is green and there are no unresolved threads. Could you take another look
when you have a moment? The review is still showing your earlier
`CHANGES_REQUESTED`.

somaz94:
@nabokihms this has been ready since Jul 31 and I'd value another look when you have time.

Both halves of your review are addressed:

1. **`guessTokenType` now sets `SupportedSigningAlgs`.** You were right about the mechanism — go-oidc validates `alg` *before* honoring `InsecureSkipSignatureCheck`, so it defaulted to RS256, failed to parse the EdDSA token, and `guessTokenType` fell through to `RefreshToken`, mis-classifying a valid EdDSA access token. It now passes `SupportedSigningAlgs: supportedSigningAlgStrings()`, the same one-line fix as the other verifier sites.
2. **Test coverage** — `TestHandleIntrospectEdDSA` drives the full `/token/introspect` endpoint with an EdDSA-signed access token rather than testing the helper in isolation.

One thing worth flagging since it changes what you'd be re-reading: on Jul 29 I **ported** this across the `server` package split rather than rebasing it. The three files the PR originally touched (`server/oauth2.go`, `server/userinfo.go`, `server/introspectionhandler.go`) no longer exist; the fix now lives in `server/introspection/` and friends. So the current diff doesn't match what you reviewed in June.

CI is green. Happy to rework anything if you'd prefer a different shape.

PR #4832, opened 2026-06-12
feat: add logout confirmation page

Lutz-Pfannenschmidt:
#### Overview

Add a logout confirmation page when no id_token_hint is provided on GET /logout.

Previously, GET /logout without an id_token_hint immediately logged the user out. Any cross-site image or link `` could silently terminate a user's session.
This PR adds a confirmation step: when id_token_hint is empty and the method is GET, a confirmation page is shown asking "Do you want to log out?" with a POST button. Only POST requests actually perform the logout. Logout with a valid id_token_hint is unaffected (proceeds immediately per OIDC spec).

Relates to #4560 (Implement Auth Sessions). 

The implementation follows the OIDC RP-Initiated Logout spec recommendation to require explicit user confirmation when the RP cannot identify the user session from the request.

### DIFF
--- a/server/logout.go
+++ b/server/logout.go
@@ -44,6 +44,17 @@ func (s *Server) handleLogout(w http.ResponseWriter, r *http.Request) {
 	postLogoutRedirectURI := r.FormValue("post_logout_redirect_uri")
 	state := r.FormValue("state")
 
+	// When no id_token_hint is provided and this is a GET request,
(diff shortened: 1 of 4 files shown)


nabokihms:
Looks like a duplicate of #4734, which adds the same logout confirmation page and is still open.

Lutz-Pfannenschmidt:
i must be blind, sorry for the duplication ^^

PR #4891, opened 2026-07-14
refactor(server): split handlers.go into focused per-flow files

nabokihms:
#### Overview

Split `server/handlers.go` (2110 lines) into focused per-flow files, with behaviour preserved 1:1.

`handlers.go` had grown to 2110 lines mixing discovery, the browser login flow, the token endpoint and every grant, which made flows hard to find and cross-cutting checks (like connector authorization) easy to miss. It is split into focused files — `discovery.go`, `authorize.go`, `login.go`, `approval.go`, `token.go`, `userinfo.go`, and one `grant_*.go` per grant — and connector authorization is consolidated in `connectorauth.go` with a documented invariant so the checks live in one obvious place. Test files are split to match the source layout.

Pure code movement, no behaviour change: the set of declaration lines is identical to the old `handlers.go` (only a doc comment was added), and the existing suite passes unchanged (same 176 tests). Two commits: the source split, then the test split. `refreshhandlers.go` / `refreshhandlers_test.go` are renamed `grant_refresh.go` / `grant_refresh_test.go` for consistency with the other grants.


nabokihms:
Only rearranging the code between files. First iteration.

PR #4897, opened 2026-07-15
connector/oidc: allow overriding userinfo and device auth URLs

### LINKED ISSUE #4110: Allow override oidc userinfo and deviceauth url
### Preflight Checklist

### Problem Description

Hello, first of all, thanks for making this awesome software.

I would like to propose additonal overridable fields for oidc connector related to urls setting, which are userinfo and device auth url.

The main reason: recently i need to integrate internal oidc provider that only listen to http but misconfigure the discovery setting where it declares all url using https. Sadly the oidc provider is not something i can control and the oidc client can only take discovery url.
I can makes pr if it make sense.

Thanks

### Proposed Solution

Allow issuer url and device auth url to be overidden just like auth, token, and jwks url

### Alternatives Considered

_No response_

### Additional Information

_No response_

Kunalbehbud:
Closes #4110

The OIDC connector's `providerDiscoveryOverrides` can already override the token, auth and JWKS URLs discovered from `.well-known/openid-configuration`, but not the userinfo or device authorization endpoints.

As reported in #4110, this is a problem when a provider advertises those endpoints incorrectly in its discovery document (e.g. it declares `https` URLs when only `http` is reachable) and the provider's discovery configuration can't be changed.

This adds `userInfoURL` and `deviceAuthURL` to `ProviderDiscoveryOverrides` and applies them in `getProvider`, mirroring the existing `tokenURL` / `authURL` / `jwksURL` overrides. `oidc.ProviderConfig` already carries both fields, so this only wires the override branches.

Scope note: the issue text mentions "userinfo and device auth url", so this covers exactly the two discovery URLs that weren't yet overridable. The issuer URL is configured separately and isn't a discovery override.

### Testing

Extended `TestProviderOverride` with a case that overrides both new URLs: the device auth URL is asserted via `provider.Endpoint().DeviceAuthURL`, and the userinfo override is verified behaviourally — a second test server returns a distinct subject and the connector's `UserInfo` call is shown to hit the overridden endpoint. Runs offline (`httptest`), no infra.

```
go test ./connector/oidc/ -run TestProviderOverride
```

### DIFF
--- a/connector/oidc/oidc.go
+++ b/connector/oidc/oidc.go
@@ -136,13 +136,19 @@ type ProviderDiscoveryOverrides struct {
 	// JWKSURL provides a way to user overwrite the JWKS URL
 	// from the .well-known/openid-configuration jwks_uri
 	JWKSURL string `json:"jwksURL"`
+	// UserInfoURL provides a way to override the UserInfo URL
+	// from the .well-known/openid-configuration userinfo_endpoint
+	UserInfoURL string `json:"userInfoURL"`
+	// DeviceAuthURL provides a way to override the Device Authorization URL
+	// from the .well-known/openid-configuration device_authorization_endpoint
+	DeviceAuthURL string `json:"deviceAuthURL"`
 	// EndSessionURL provides a way to override the end_session_endpoint
 	// from the .well-known/openid-configuration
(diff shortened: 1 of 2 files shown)


nabokihms (approved):
Thanks! Good improvemnt.

PR #4949, opened 2026-07-31
fix(server): give refresh token reuse a default grace window

nabokihms:
#### Overview

Default `expiry.refreshTokens.reuseInterval` to `3s` instead of `0`.

Rotation is enabled by default, and with `reuseInterval` unset the reuse window is zero, so a refresh token is claimable exactly once. Any client that does not see the response to its refresh — a request timeout, a dropped connection, or a second replica of an HA relying party refreshing the same stored token — retries with a token dex has already rotated, gets `invalid_grant`, and has to send the user back through login.

`Rotate` already handles this properly: inside the reuse window it returns the token the first call minted, so the retry is idempotent. Only the default was missing, and `3s` is the value the sample configuration has suggested all along.

`reuseInterval: "0s"` keeps the previous behavior for anyone who wants a token claimable exactly once.

The trade-off is a 3 second window in which an already-rotated token still works. Dex does not revoke the token family on reuse detection today, so the practical difference is that a retry succeeds rather than logging the user out.

### DIFF
--- a/config.yaml.dist
+++ b/config.yaml.dist
@@ -91,7 +91,7 @@ web:
 #   idTokens: "24h"
 #   refreshTokens:
 #     disableRotation: false
-#     reuseInterval: "3s"
+#     reuseInterval: "3s" # default, "0s" disables the reuse grace window
 #     validIfNotUsedFor: "2160h" # 90 days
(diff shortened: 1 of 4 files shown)


nabokihms:
A questionable PR. I will let it rest for a bit.

PR #4964, opened 2026-08-09
feat(google): implement TokenIdentity for RFC 8693 token exchange

### LINKED ISSUE #3768: Support token exchange in Google connector
### Preflight Checklist

### Problem Description

We are performing machine authentication by exchanging a Google access token (from an impersonated service account) for a dex-issued JWT, as in [these docs](https://dexidp.io/docs/guides/token-exchange/).
 
The Google connector does not appear to support token exchange, but the OIDC connector configured with Google does. 

The limitation of using the OIDC connector is that we cannot get group claims; it appears that the Google connector handles those as a special case.

### Proposed Solution

Support token exchange in the Google connector. I think it should be possible given that token exchange is possible with Google via the OIDC connector, and would enable group claims when doing so.

### Alternatives Considered

_No response_

### Additional Information

_No response_

pujitha24:
#### Overview

Adds RFC 8693 token-exchange support to the Google connector by implementing `TokenIdentity`.

The Google connector only implemented `CallbackConnector` and `RefreshConnector`, so it couldn't be used with the token-exchange grant. Users needing machine authentication with group claims (e.g. exchanging a Google access token from an impersonated service account for a dex-issued JWT) had to fall back to the OIDC connector configured against Google, which does not return group claims since the Google connector handles group membership as a special case via the Admin Directory API.

This PR adds a `TokenIdentity` method to the Google connector, mirroring the existing OIDC connector implementation. It verifies a Google ID Token (subjectTokenType `urn:ietf:params:oauth:token-type:id_token`) or loads the identity from Google's userinfo endpoint for an access token (`urn:ietf:params:oauth:token-type:access_token`), then reuses the connector's existing hosted-domain check and group-fetching logic, which was extracted from `createIdentity` into a shared `identityFromClaims` helper. Since `TokenIdentity` has no `Scopes` parameter, groups are always attempted and gated by connector configuration, matching the approach already used by the GitLab connector's `TokenIdentity` (see https://github.com/dexidp/dex/pull/4606).

Closes #3768

- Validation: `go build ./...` passes; `go vet ./connector/google/... ./server/...` passes with no findings; `gofmt -l` on both changed files is clean; `go test ./connector/google/...` and `go test -race ./connector/google/... ./server/grants/...` pass, including new tests `TestIdentityFromClaims`, `TestIdentityFromClaimsHostedDomain`, and `TestTokenIdentityUnknownTokenType` covering the hosted-domain check, group fetch/filter, and unknown-token-type error paths.
- `golangci-lint` could not be run in this sandbox (the pinned v2.4.0 panics against the sandbox's Go 1.26 toolchain, unrelated to this change).
- Not validated: an end-to-end exchange against real Google OAuth tokens, since no live Google credentials are available in this environment. The `id_token` and `access_token` branches delegate to the now shared, unit-tested `identityFromClaims` helper but are not exercised by a test that performs a real network call to Google.

---
AI assistance: this change was drafted with Claude Code.

### DIFF
--- a/connector/google/google.go
+++ b/connector/google/google.go
@@ -131,6 +131,7 @@ func (c *Config) Open(id string, logger *slog.Logger) (conn connector.Connector,
 		verifier: provider.Verifier(
 			&oidc.Config{ClientID: clientID},
 		),
+		provider:                       provider,
 		logger:                         logger,
 		cancel:                         cancel,
 		hostedDomains:                  c.HostedDomains,
@@ -144,14 +145,16 @@ func (c *Config) Open(id string, logger *slog.Logger) (conn connector.Connector,
 }
 
 var (
-	_ connector.CallbackConnector = (*googleConnector)(nil)
-	_ connector.RefreshConnector  = (*googleConnector)(nil)
+	_ connector.CallbackConnector      = (*googleConnector)(nil)
+	_ connector.RefreshConnector       = (*googleConnector)(nil)
+	_ connector.TokenIdentityConnector = (*googleConnector)(nil)
 )
 
 type googleConnector struct {
 	redirectURI                    string
 	oauth2Config                   *oauth2.Config
 	verifier                       *oidc.IDTokenVerifier
+	provider                       *oidc.Provider
 	cancel                         context.CancelFunc
 	logger                         *slog.Logger
 	hostedDomains                  []string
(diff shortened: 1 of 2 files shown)


pujitha24:
This has been rebased and green for a week now — happy to make any changes that would help move review along, just let me know.

PR #4971, opened 2026-08-12
fix(keystone): retry federation auth redirects, surface retryable session errors

mridulgain:
## Summary

KAAP-2297: Keystone's federation auth endpoint intermittently returns a 302 redirect instead of a token, which the `keystonefed` connector today surfaces as an opaque 500. This adds:

- A bounded retry (3 attempts, 150ms apart) around the federation auth call when it redirects, since this has so far only reproduced as a transient session-lookup miss rather than a genuinely invalid session.
- A generic `connector.RetryableError` interface, implemented by a new `federationSessionError`, returned only if the redirect persists past retries.
- `handleConnectorCallback` now special-cases `RetryableError` to render `401` with an actionable message, instead of the generic `500 Failed to authenticate: ...`.

**Advantage:** most real occurrences of this redirect resolve transparently within the retry window, so users no longer hit an error page at all for the common case. When it does exhaust retries, the message tells the user to just log in again instead of looking like a server failure.

## Test plan

- `go test ./connector/keystone/... ./server/...`, `go vet ./...`, `go build ./...` all pass, including new unit tests for the redirect-retries-then-`RetryableError` path and a non-redirect (503) error staying non-retryable.
- Verified end-to-end against the real `mg-du` dev environment (Dex/Keystone/Shibboleth, not mocked): a local Dex build hitting the live federation endpoint with a stale session cookie retried 3x then returned 401 with the new message; the same flow with a freshly-valid session cookie completed the login successfully with no retries triggered.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mridulgain:
Opened in error against the wrong repo (internal fork PR, not intended for upstream). Closing immediately, apologies for the noise.

Issue #4017, opened 2025-03-03
ETCD backend interpretes everything as UTF-8 string which could lead to key collisions (besides other things) when being used with ldap connector


ProbstDJakob:
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [x] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

master

### Storage Type

etcd

### Installation Type

Other (specify below)

### Expected Behavior

Either the ldap connector deserialises the values according to the schema or does treat the values as binary data.

### Actual Behavior

The etcd backend interprets everything as UTF-8 string which could lead to key collisions when being used with the ldap connector. Though the ldap library suggests that [`EntryAttribute.Values`](https://github.com/go-ldap/ldap/blob/b7e97bbe608464471a40dc5fececaf9c0acaacbc/v3/search.go#L345) returns an array of _decoded_ strings (with the raw values in `ByteValues`), the values are actually always the same and may contain arbitrary byte values. In our case we tried to use `objectSid` as `UserId` but it is not deserialised and thus contains binary data. The etcd storage backend then [serialises the value to json within `txnCreate`](https://github.com/dexidp/dex/blob/0f7554f001e3b5d1b75d5672442e87f573135211/storage/etcd/etcd.go#L505) which replaces all invalid UTF-8 sequences with the replacement character U+FFFD (`0xEFBFBD`). This could then lead to key collisions as the value is being used as `sub`. This may also apply to other connectors/storage backends.

### Configuration

```yaml

```

### Logs

```shell

```

Issue #4016, opened 2025-03-03
ETCD storage may produces key collisions


ProbstDJakob:
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/dexidp/dex/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://www.github.com/dexidp/dex/issues) for an issue that matches the one I want to file, without success.
- [x] I am not looking for support or already pursued the available [support channels](https://github.com/dexidp/dex/issues/new/choose) without success.

### Version

master

### Storage Type

etcd

### Installation Type

Other (specify below)

### Expected Behavior

Keys should be used as provided without altering them.

### Actual Behavior

The `userID` and `connID` are converted to lower case when being used as offline session key within [`keySession`](https://github.com/dexidp/dex/blob/0f7554f001e3b5d1b75d5672442e87f573135211/storage/etcd/etcd.go#L556-L558). I am not sure whether the `connID` is case sensitive or not, but at least the `userID` may be case sensitive and thus should not be converted to lower case as this could lead to collisions.

### Additional Information

I only looked into the etcd storage, but other storages may have the same problem.

### Configuration

```yaml

```

### Logs

```shell

```

nabokihms:
Not sure how to fix this without breaking anything somewhere...

ProbstDJakob:
I am not sure what happens when a client tires to refresh its token and the key could not be found in the storage, but if dex handles this case by logging out the client, a fix may not be that hard. One thing to consider is, whether old keys should be deleted, abandoned, or tried to still be resolved.

PR #4978, opened 2026-08-20
feature(ldap): Adds objectSid and objectUUID formatting when using Active Directory

SeanHood:
#### Overview

Adds support for formatting `objectSid` and `objectGUID` for use with the ldap connector and Active Directory.

Active Directory stores these two attributes as byte arrays. This change converts them to strings, allowing them to be used for any `*Attr` but intended for `idAttr`.

I believe that the Subject Claim within a JWT is intended to be a static non-changing value. In the context of AD `objectGUID` never changes and `objectSid` only changes in certain circumstances. Whereas a `DN` could change for a user either via a name change or if the object is moved to a new OU.

Prior to this change using either attribute would result in an error like so:
```
{"time":"2025-02-21T15:29:54.427305352Z","level":"ERROR","msg":"failed to marshal offline session ID","err":"string field contains invalid UTF-8","request_id":"6485c460-48ca-46bd-b1f1-30a4b9d14c6d"}
{"time":"2025-02-21T15:29:54.427340003Z","level":"ERROR","msg":"failed to create new access token","err":"failed to marshal offline session ID: string field contains invalid UTF-8","request_id":"6485c460-48ca-46bd-b1f1-30a4b9d14c6d"}
```
Closes: #3128

Implementation of `formatSidAttr` is inspired by: https://github.com/dotnet/runtime/blob/1f0e9eadd29b9c15dae9313234abf486f1c4ef96/src/libraries/System.Security.Principal.Windows/src/System/Security/Principal/SID.cs#L395

Test config:
```
connectors:
  - type: ldap
    id: ldap
    name: AD
    config:
      host: ad.example.com:636

      bindDN: uid=serviceaccount,cn=users,dc=example,dc=com
      bindPW: password

      userSearch:
        baseDN: OU=Users,DC=example,DC=com
        filter: "(objectClass=person)"
        username: sAMAccountName
        idAttr: objectSid
        emailAttr: mail
        nameAttr: cn
        preferredUsernameAttr: sAMAccountName

      groupSearch:
        baseDN: OU=Groups,DC=example,DC=com
        filter: "(objectClass=group)"
        userMatchers:
          - userAttr: DN
            groupAttr: "member:1.2.840.113556.1.4.1941:"
        nameAttr: cn
```

### DIFF
--- a/connector/ldap/ldap.go
+++ b/connector/ldap/ldap.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 
 	"github.com/go-ldap/ldap/v3"
+	"github.com/google/uuid"
 
 	"github.com/dexidp/dex/connector"
 )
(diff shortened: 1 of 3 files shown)


slrz:
How Dex handles non-UTF-8 data from LDAP attributes certainly leaves room for improvement, if not to say is totally broken (cf.  #4017). So I sympathize with this PR (and we apply something similar as a downstream patch).

But as you say, the OIDC subject is supposed to not change over time but this is what's going to happen for existing Dex users who chose objectSid as the idAttr.  One might argue that their usage was broken to begin with but still, this is a major breaking change and should be communicated as such in release notes.

If the log excerpts in your message really mean that using non-UTF8 attributes is no longer possible, then the breaking change already happened (but wasn't part of a release yet, I think),

slrz:
With regard to the actual code: are there cases where Dex would benefit from being aware of the structure and preferred text representation of NT SIDs and other binary attributes?  At least for the `sub` claim, [it is going to stuff it into a Protobuf encoder together with some extra data, Base64-encode the result and use *that* as the subject](https://github.com/dexidp/dex/blob/ab64ed778070e983cbb10cfc07ea4bb397d14312/server/tokens/claims.go#L139).

Might as well simplify the code and treat all binary strings the same, encoding them in some Base32 variant (or a plain hex string), always. Avoid any encoding where distinct inputs may yield outputs that differ in case only, [as the etcd backend  will  `strings.ToLower` them](https://github.com/dexidp/dex/blob/ab64ed778070e983cbb10cfc07ea4bb397d14312/storage/etcd/etcd.go#L698) (see #4016 for the issue).

Of course, if the project decides to do a breaking change, this should rather be fixed directly within the etcd backend.

PR #4987, opened 2026-08-27
Configure TLS Curve Preferences and Cipher Suites

### LINKED ISSUE #4913: allowedTLSCiphers flag can be made configurable to align for platform needs.
### Preflight Checklist

### Problem Description

right now, allowedTLSCiphers are not configurable https://github.com/dexidp/dex/blob/2c81c4457dfd87a112ec9912afefe57fd78f8877/cmd/dex/serve.go#L162 like tlsminversion and tlsmaxversion https://github.com/dexidp/dex/blob/2c81c4457dfd87a112ec9912afefe57fd78f8877/cmd/dex/serve.go#L180

### Proposed Solution

can we add support for making allowedTLSCiphers flag for web configurable like tlsminversion and tlsmaxversion.

### Alternatives Considered

_No response_

### Additional Information

_No response_

akhilnittala:
#### Overview

TLS Ciphersuites are hardcoded currently, to meet the platform side compliance ciphersuites needs to be configureable, so made changes for tls ciphersuites and curve preferences to be configurable for web.
Configures the tls ciphersuites and curvepreferences based on the parameters in config.yaml

Fixes: https://github.com/dexidp/dex/issues/4913


akhilnittala:
Hey @nabokihms @sagikazarmark could you please help in the review process? We need the PR in upstream. Thanks for any input!

cmd/dex/config.go
```diff
@@ -264,16 +266,43 @@ type PKCE struct {
...
+	"CurveP384":          tls.CurveP384,
+	"CurveP521":          tls.CurveP521,
+	"X25519":             tls.X25519,
+	"P-256":              tls.CurveP256,
+	"P-384":              tls.CurveP384,
+	"P-521":              tls.CurveP521,
+}
+
+// mapKeys returns the keys of a map as a slice.
+func mapKeys(m map[string]tls.CurveID) []string {
```
sagikazarmark: `slices.Sorted(maps.Keys(m))` would do the same from stdlib

cmd/dex/config.go
```diff
@@ -264,16 +266,43 @@ type PKCE struct {
...
+	Headers                    Headers        `json:"headers"`
+	TLSCert                    string         `json:"tlsCert"`
+	TLSKey                     string         `json:"tlsKey"`
+	TLSMinVersion              string         `json:"tlsMinVersion"`
+	TLSMaxVersion              string         `json:"tlsMaxVersion"`
+	AllowedTLSCiphers          []string       `json:"allowedTLSCiphers"`
+	AllowedOrigins             []string       `json:"allowedOrigins"`
+	AllowedHeaders             []string       `json:"allowedHeaders"`
+	ClientRemoteIP             ClientRemoteIP `json:"clientRemoteIP"`
+	AllowedTLSCurvePreferences []string       `json:"allowedTLSCurvePreferences"`
```
sagikazarmark: How about just `TLSCiphers` and `TLSCurvePreferences`? I would also move `TLSCurvePreferences` up to the TLS settings.

cmd/dex/serve.go
```diff
@@ -560,12 +560,29 @@ func runServe(options serveOptions) error {
 		if c.Web.TLSMaxVersion != "" {
 			tlsMaxVersion = allowedTLSVersions[c.Web.TLSMaxVersion]
 		}
+		cipherSuites := allowedTLSCiphers
```
sagikazarmark, line 566: I think this should be in Config.Validate. I can see the pattern above it, but that's probably historical.

cmd/dex/serve.go
```diff
@@ -640,6 +657,43 @@ func runServe(options serveOptions) error {
...
 
+// parseCipherSuites parses a list of cipher suite names and returns their corresponding IDs.
+func parseCipherSuites(names []string) ([]uint16, error) {
+	cipherMap := make(map[string]uint16)
+	for _, cs := range tls.CipherSuites() {
+		cipherMap[cs.Name] = cs.ID
+	}
+	for _, cs := range tls.InsecureCipherSuites() {
+		cipherMap[cs.Name] = cs.ID
+	}
```
sagikazarmark, line 662: This is quite a footgun IMO. Why did you add it?

cmd/dex/serve_test.go
```diff
@@ -27,3 +30,145 @@ func TestNewLogger(t *testing.T) {
...
+				}
+
+				return
+			}
+
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+
+			if !reflect.DeepEqual(got, tt.want) {
```
sagikazarmark: We usually use testify

cmd/dex/serve_test.go
```diff
@@ -27,3 +30,145 @@ func TestNewLogger(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := parseCurvePreferences(tt.input)
+
+			if tt.wantErr {
+				if err == nil {
+					t.Fatal("expected error, got nil")
+				}
+
+				if !strings.Contains(err.Error(), tt.errContains) {
```
sagikazarmark: We usually use testify

cmd/dex/config.go
```diff
@@ -264,16 +266,43 @@ type PKCE struct {
...
+}
+
+var allowedCurveNames = map[string]tls.CurveID{
```
sagikazarmark: Let's use one canonical format. For example: `CurveID.String()`. Do not parse different formats.

sagikazarmark (changes requested):
Thanks for the PR!

The approach looks good.

Blocking: SecP256r1MLKEM768/SecP384r1MLKEM1024 require Go 1.26.
Rebasing should resolve that.


I think it would also make sense to add X25519MLKEM768 (so users don't lose PQ key exchange when setting curve prefs).

sagikazarmark:
Also: maybe add an example to the config file?

cmd/dex/serve.go
akhilnittala: if someone wants to use the ciphers that is in the list of insecure cipher suites then the existing code will support it.

cmd/dex/serve.go
akhilnittala: done, added validations in config.Validate function.

cmd/dex/config.go
akhilnittala: done changed names.

cmd/dex/config.go
akhilnittala: done.

cmd/dex/config.go
akhilnittala: removed the entire function and replaced with slices.Sorted function in calling place.

akhilnittala:
> Also: maybe add an example to the config file?

added, thanks

akhilnittala:
@sagikazarmark addressed all your comments, could you please re-review it.

akhilnittala:
> Thanks for the PR!
> 
> The approach looks good.
> 
> Blocking: SecP256r1MLKEM768/SecP384r1MLKEM1024 require Go 1.26. Rebasing should resolve that.
> 
> I think it would also make sense to add X25519MLKEM768 (so users don't lose PQ key exchange when setting curve prefs).

done, added @sagikazarmark

cmd/dex/config.go
```diff
@@ -111,6 +112,21 @@ func (c Config) Validate() error {
...
+		c.Web.tlsCipherIDs = ciphers
+	}
+
+	if len(c.Web.TLSCurvePreferences) > 0 {
+		curves, err := parseCurvePreferences(c.Web.TLSCurvePreferences)
+		if err != nil {
+			return fmt.Errorf("invalid TLS curve preferences: %w", err)
+		}
+		c.Web.tlsCurveIDs = curves
+	}
```
sagikazarmark: Looking closer at this: `c` is a value receiver and Web is a value as well, so I think this will simply not set anything.

Can you please test it?

examples/config-dev.yaml
```diff
@@ -64,6 +64,20 @@ web:
...
+  # Cipher suites supported by the server.
+  # tlsCiphers:
+  # - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+  # - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+
+  # Curve preferences used during TLS key exchange.
+  # tlsCurvePreferences:
+  # - X25519
+  # - X25519MLKEM768
+  # - P256
```
sagikazarmark: I don't think this is valid

cmd/dex/serve.go
```diff
@@ -561,11 +563,20 @@ func runServe(options serveOptions) error {
...
+		}
+		curvePreferences := []tls.CurveID(nil)
+		if len(c.Web.tlsCurveIDs) > 0 {
+			curvePreferences = c.Web.tlsCurveIDs
+		}
 		baseTLSConfig := &tls.Config{
 			MinVersion:               uint16(tlsMinVersion),
 			MaxVersion:               uint16(tlsMaxVersion),
-			CipherSuites:             allowedTLSCiphers,
+			CipherSuites:             cipherSuites,
```
sagikazarmark: I think this can potentially cause some issues. According to the [docs](https://pkg.go.dev/crypto/tls#Config), `tls.Config.CipherSuites` allows setting ciphers for TLS 1.0-1.2:

```
	// CipherSuites is a list of enabled TLS 1.0–1.2 cipher suites. The order of
	// the list is ignored. Note that TLS 1.3 ciphersuites are not configurable.
```

[`tls.CipherSuites`](https://pkg.go.dev/crypto/tls#CipherSuites) on the other hand returns TLS 1.3 only ciphers, like `TLS_AES_256_GCM_SHA384`.

I suspect this would lead to handshake failures when TLS 1.2 is used.

We probably need to exclude ciphers that have `tls.CipherSuite.SupportedVersions` TLS 1.3 only (maybe even do a warning log so people know it isn't applied).

It should also be documented in the example config file.

sagikazarmark (changes requested):
Sorry @akhilnittala but I think I found a few more issues. Can you please take a look?

PR #4989, opened 2026-08-28
feat(api): add groups field to password requests and responses

### LINKED ISSUE #4972: gRPC API set groups on passwords
### Preflight Checklist

### Problem Description

storage.Password has a Groups field, but the gRPC API cannot set it. The Password message in api/v2/api.proto has only email, hash, username, user_id, and server/apiserver/passwords.go copies just those four into storage.Password. UpdatePasswordReq is narrower still, new_hash and new_username only.

Groups for local passwords were added in [#4456](https://github.com/dexidp/dex/pull/4456) and work through the config file. A user created over the API authenticates but has no groups, so any relying party that authorizes on the groups claim treats them as unprivileged. That leaves the config file as the only way to manage users with groups, which requires a restart to change.

### Proposed Solution

Add repeated string groups to the proto Password, copy it in CreatePassword, and add new_groups to UpdatePasswordReq. The storage field already exists.

### Alternatives Considered

_No response_

### Additional Information

Willing to contribute the PR if feature request is accepted.

CasLubbers:
#### Overview

Adds `groups` to the gRPC API's `Password` message and `new_groups` to `UpdatePasswordReq`, so local users managed over the API can carry groups. `storage.Password.Groups` already existed; only the API surface was missing.

Groups for local passwords were added in #4456 and work through the config file, but the gRPC API could not set them. A   
user created over the API authenticates with no groups, so any relying party that authorizes on the `groups` claim treats 
them as unprivileged. That left the config file as the only way to manage users with groups, which requires a restart to  dex to take effect. 
- `CreatePassword` copies `groups` into storage
- `UpdatePassword` accepts `new_groups` and counts it as a change on its own
- `ListPasswords` returns `groups`, so the field is not write-only

No storage changes and no migration, the column has existed since #4456.                                                 
                                                                                                                          
Closes #4972 

**Groups can be set and replaced, but not cleared.** Proto3 encodes an empty repeated field to the same zero bytes as an absent one, so `new_groups: []` arrives as nil and reads as "leave alone". This matches every repeated field already in the API, `UpdateClientReq` treats `redirect_uris`, `trusted_peers`, `allowed_connectors`, `sso_shared_with` and `post_logout_redirect_uris` the same way, and none of them can be cleared either. I followed the existing convention rather than making this one field special. If clearing is wanted, it seems better solved once for all six fields, a `FieldMask` on the request, or a wrapper message per field, and would need a new tag number, since changing tag 4 from `repeated string` to a message is wire-breaking.

**Other notes:**

- `api.pb.go` is regenerated with the pinned toolchain (protoc 29.3, protoc-gen-go 1.36.5); most of the diff is the       
descriptor blob re-wrapping. 
- `TestPassword` now covers groups end to end: set on create, read back via storage and `ListPasswords`, changed via      
update, and left untouched by an update that does not mention them. 
- Verified on a kind cluster with `storage.type: kubernetes` — groups set over gRPC reach the `groups` claim in the ID    
token, and a password created without groups omits the key in the CR.


server/apiserver/passwords.go
```diff
@@ -48,7 +49,7 @@ func (d dexAPI) UpdatePassword(ctx context.Context, req *api.UpdatePasswordReq)
 	if req.Email == "" {
 		return nil, errors.New("no email supplied")
 	}
-	if req.NewHash == nil && req.NewUsername == "" {
+	if req.NewHash == nil && req.NewUsername == "" && req.NewGroups == nil {
```
sagikazarmark, line 60: I' m not sure how proto works, but if you send an empty field over the wire, is it going to be an initialized slice on this side or nil?

Because if it's the latter, sending an empty list as a group will result in "nothing to update".

server/apiserver/passwords.go
```diff
@@ -67,6 +68,10 @@ func (d dexAPI) UpdatePassword(ctx context.Context, req *api.UpdatePasswordReq)
 			old.Username = req.NewUsername
 		}
 
+		if req.NewGroups != nil {
```
sagikazarmark: Should group names be validated? Eg. != "" or that they are unique? (We could use slices.Uniq for that, no need to validate)

server/apiserver/passwords.go
CasLubbers: Yes, it arrives as nil. Proto3 doesn't encode empty repeated fields, so [] and "not set" are identical on the wire. I tested this against a running dex before opening the PR:
- Sent on its own, new_groups: [] returns nothing to update
- Sent with another field, it's silently skipped
 
So groups can be replaced but not cleared, same as the repeated fields on UpdateClientReq. I kept that convention, but if you'd rather support clearing I can wrap it in a message (StringList new_groups) so presence is tracked. 

Added a comment in the code to make this explicit.

server/apiserver/passwords.go
CasLubbers: Makes sense, added both. Empty group names are now rejected, and groups are sorted and deduplicated on create and update.

There's no `slices.Uniq` in the stdlib, so I used `slices.Sort` + `slices.Compact` (Compact only removes adjacent duplicates, hence the sort). This does mean groups are stored sorted rather than in the order they were sent.

server/apiserver/passwords.go
sagikazarmark: Let's leave it this way for now. It's documented. If anyone complains, we can change it.

sagikazarmark (approved):
Thanks @CasLubbers

PR #4991, opened 2026-08-28
fix(server deadlock): resolve refresh connector data outside the rotation transaction

### LINKED ISSUE #5017: SQLite storage: a single refresh_token grant permanently wedges /token while /healthz stays green
### Summary

With SQLite storage, a **single** `grant_type=refresh_token` request deadlocks the token endpoint permanently: the request never completes, every later `/token` request (any grant type) hangs until the process restarts, and `/healthz` keeps returning 200 — so orchestrated deployments never recycle the wedged pod.

Present on `master` (verified at 7ace0e79); introduced by the per-grant `/token` handler refactor (2c81c4457d). Releases up to v2.45.1 are unaffected.

### Reproduction

Minimal config (SQLite + local password DB + public client, rotation is default):

```yaml
issuer: http://127.0.0.1:15556/dex
storage:
  type: sqlite3
  config:
    file: /tmp/dex-repro/dex.db
web:
  http: 127.0.0.1:15556
enablePasswordDB: true
oauth2:
  passwordConnector: local
staticClients:
  - id: test-client
    redirectURIs: ['http://127.0.0.1:15556/callback']
    name: 'test-client'
    public: true
staticPasswords:
  - email: "admin@example.com"
    hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"  # "password"
    username: "admin"
    name: "Admin"
    emailVerified: true
```

1. Password grant works fine:

```console
$ curl -s http://127.0.0.1:15556/dex/token -d grant_type=password -d username=admin@example.com \
    -d 'password=password' -d client_id=test-client --data-urlencode 'scope=openid offline_access'
{"access_token":"...","refresh_token":"..."}
```

2. The refresh grant never returns (10 s timeout, 0 bytes):

```console
$ curl -m 10 -w 'HTTP=%{http_code} time=%{time_total}s bytes=%{size_download}\n' \
    http://127.0.0.1:15556/dex/token -d grant_type=refresh_token \
    -d 'refresh_token=<rt>' -d client_id=test-client
HTTP=000 time=10.002080s bytes=0      # curl exit 28
```

3. While wedged, `/healthz` still passes **and a brand-new password grant also hangs**:

```console
$ curl -s http://127.0.0.1:15556/dex/healthz
Health check passed

$ curl -m 5 -w 'HTTP=%{http_code} time=%{time_total}s\n' http://127.0.0.1:15556/dex/token \
    -d grant_type=password -d username=admin@example.com -d 'password=password' ...
HTTP=000 time=5.006826s
```

### Root cause

`RefreshStore.Rotate` invokes the `freshIdentity` callback from inside the `UpdateRefreshToken` transaction. For a user without an offline session (every local-password user), that callback calls `refreshConnectorData` → `GetOfflineSessions`, a **db-level** query. The SQLite storage pins the pool to one connection (`SetMaxOpenConns(1)`), so the nested query waits forever for the connection the transaction already holds. Every later token request queues behind the exhausted pool. `/healthz` pings storage on a path that never contends with the rotation transaction.

Before the refactor the connector data was resolved during request verification, outside the transaction (v2.45.1 `server/refreshhandlers.go:207`); the refactor moved that read into the lazy closure.

### Fix

#4991 resolves the connector data before rotation starts (restores the pre-refactor ordering). With the fix, the same reproduction completes three consecutive rotating refreshes at ~6 ms each.

A goroutine dump taken at the deadlock is attached to #4991.

laoshanxi:
fix(server): resolve refresh connector data outside the rotation transaction

Rotate invokes the freshIdentity callback from inside UpdateRefreshToken's transaction, and the SQLite storage pools a single connection (SetMaxOpenConns(1)), so the GetOfflineSessions read issued by refreshConnectorData from that callback waits forever for the connection the transaction already holds. Every later token request then queues behind the exhausted pool: a single refresh grant wedges /token permanently while /healthz keeps passing.

Restore the pre-refactor behavior of resolving the connector data before rotation starts, gated on the same userIdent condition the callback used.

#### Overview

Moves the `refreshConnectorData` storage read out of the refresh-token rotation transaction. One-line behavior summary: with SQLite storage, a single `grant_type=refresh_token` request currently deadlocks the whole `/token` endpoint until process restart; after this change the same request (and any number of subsequent ones) succeed.

Reproduction (SQLite storage, local password connector, public client, rotation enabled):

```console
$ curl -s http://127.0.0.1:6062/auth/token \
    -d grant_type=password -d username=admin@example.com -d 'password=<pw>' -d client_id=test-client \
    --data-urlencode 'scope=openid offline_access'
{"access_token":"...","refresh_token":"...","id_token":"..."}     # fine

$ curl -m 8 http://127.0.0.1:6062/auth/token \
    -d grant_type=refresh_token -d 'refresh_token=<rt>' -d client_id=test-client
curl: (28) Operation timed out ... 0 bytes received               # never returns

$ curl -s http://127.0.0.1:6062/auth/healthz
{"storage":{"contiguousFailures":0,...}}                          # still "healthy"
```

That request never completes, and **every** later token request (any grant type) hangs until the process is restarted, while `/healthz` keeps returning 200 — in orchestrated environments the pod is never recycled automatically.

**Goroutine dump at the deadlock** (`SIGQUIT` right after the wedged refresh; irrelevant frames trimmed, paths shortened to `dex/` root):

```goroutine
goroutine 82 [select]:
runtime.gopark(...)
runtime.selectgo(...)
database/sql.(*DB).conn(...)                                    sql/sql.go:1369   <- waits for a free connection, forever
database/sql.(*DB).query(...)                                   sql/sql.go:1759
database/sql.(*DB).QueryContext.func1(...)                      sql/sql.go:1742
database/sql.(*DB).retry(...)                                   sql/sql.go:1576
database/sql.(*DB).QueryContext(...)                            sql/sql.go:1741
database/sql.(*DB).QueryRowContext(...)                         sql/sql.go:1842
database/sql.(*DB).QueryRow(...)                                sql/sql.go:1856
storage/sql.(*conn).QueryRow(...)                               storage/sql/sql.go:155      <- db-level query, NOT the tx
storage/sql.getOfflineSessions(...)                             storage/sql/crud.go:810
storage/sql.(*conn).GetOfflineSessions(...)                     storage/sql/crud.go:806
storage.(*staticClientsStorage).GetOfflineSessions(...)         (storage wrappers)
storage.(*staticPasswordsStorage).GetOfflineSessions(...)
storage.(*staticConnectorsStorage).GetOfflineSessions(...)
server.(*keyCacher).GetOfflineSessions(...)
server/grants.(*refresh).refreshConnectorData(...)              server/grants/refresh.go:247
server/grants.(*refresh).Authorize.func1(...)                   server/grants/refresh.go:115  <- the freshIdentity closure
server/tokens.(*RefreshStore).Rotate.func1(...)                 server/tokens/refresh_store.go:196
storage/sql.(*conn).UpdateRefreshToken.func1(...)               storage/sql/crud.go:349      <- INSIDE the tx callback
storage/sql.(*conn).ExecTx(...)                                 storage/sql/sql.go:170       <- tx holds the only connection
storage/sql.(*conn).UpdateRefreshToken(...)                     storage/sql/crud.go:344
server/tokens.(*RefreshStore).Rotate(...)                       server/tokens/refresh_store.go:211
server/grants.(*refresh).Authorize(...)                         server/grants/refresh.go:122
server/grants.(*Handler).dispatch(...)                          server/grants/grants.go:259
server/grants.(*Handler).handleToken(...)                       server/grants/grants.go:208
net/http.HandlerFunc.ServeHTTP(...) + prometheus/gorilla middleware frames
```

**Root cause analysis** — three code locations compose into a guaranteed single-request self-deadlock:

1. `RefreshStore.Rotate` runs the caller's `freshIdentity` callback from inside the `UpdateRefreshToken` transaction (`storage/sql.(*conn).ExecTx` opens the transaction and calls the updater; `refresh_store.go:196` invokes `freshIdentity(ctx)` there):

   ```go
   refreshTokenUpdater := func(old storage.RefreshToken) (storage.RefreshToken, error) {
       ...
       ident, idErr = freshIdentity(ctx)   // executes inside ExecTx
       ...
   }
   if err := rt.storage.UpdateRefreshToken(ctx, storageToken.ID, refreshTokenUpdater); err != nil {
   ```

2. For a user without an offline session — every local-password user — that closure reads storage at db level (`server/grants/refresh.go`):

   ```go
   freshIdentity := func(ctx context.Context) (connector.Identity, error) {
       if userIdent != nil {
           return tokens.IdentityFromClaims(userIdent.Claims), nil
       }
       connectorData, err := g.refreshConnectorData(ctx, refreshToken)  // -> GetOfflineSessions -> c.db.QueryRow
       ...
   }
   ```

3. The SQLite flavor pins the pool to exactly one connection (`storage/sql/sqlite.go`):

   ```go
   db.SetMaxOpenConns(1)
   ```

The transaction holds that one connection; the nested `QueryRow` waits for a second one that can never be released, so the callback — and with it the transaction, the HTTP handler, and every subsequent token request queued behind the exhausted pool — blocks forever. `/healthz` performs its storage ping through a path that never contends with the rotation transaction, which is why liveness probes stay green.

Before the per-grant handler refactor (`2c81c4457d`, "decompose the /token endpoint into per-grant handlers"), the equivalent read happened during request verification, *outside* the transaction (`v2.45.1 server/refreshhandlers.go:207`), so the rotation callback only touched already-captured data. The refactor moved that read into the lazy `freshIdentity` closure, introducing the deadlock; releases up to `v2.45.1` are unaffected.

**The fix** restores the pre-refactor ordering: resolve the connector data before `Rotate` starts, gated on the same `userIdent == nil` condition the closure used (session-backed identities never read it), and let the closure consume the pre-resolved value. Behavior is otherwise unchanged.

- Verified on `master` (SQLite, local password users, rotation enabled): where `master` wedges on the first refresh, this change survives 10 consecutive rotating refreshes plus further password grants, all sub-100ms.
- `refreshWithConnector` (an upstream IdP round-trip) still runs inside the transaction via the callback; it performs no storage access, so it cannot deadlock the pool, but if you'd rather keep the connection un-held during network I/O too, hoisting the entire identity refresh out of the rotation callback would be a larger behavioral change than this fix intends.
- More generally, with `MaxOpenConns(1)` any nested non-tx storage read inside an `ExecTx` callback is fatal on SQLite; it may be worth auditing the other `ExecTx` call sites for the same pattern in a follow-up.
- No new test included: the deadlock needs a real SQLite-backed server handling an HTTP refresh request; the existing unit suites don't exercise that combination. Happy to add one if there's a preferred harness for storage-level integration tests.

Fixes #5017.


laoshanxi:
Filed #5017 with a standalone reproduction (SQLite + local passwords + public client): the refresh grant wedges /token permanently while /healthz stays green, and with this patch the same sequence completes three rotating refreshes at ~6 ms each.

fxgameplan:
Confirming this affects OIDC connector setups too, not just local password auth — hit the identical deadlock in production (Dex fronting Authentik as an upstream OIDC provider, SQLite storage, refresh token rotation enabled). Our logs show the exact same refreshConnectorData/GetOfflineSessions call path in the goroutine trace above; /token wedged completely while /healthz stayed green, and a container restart was the only recovery. Confirms userIdent is nil for our connector type too, so this isn't limited to local password users as the issue title might suggest. Would be great to see this merged — happy to provide logs if useful.

laoshanxi:
> Confirming this affects OIDC connector setups too, not just local password auth — hit the identical deadlock in production (Dex fronting Authentik as an upstream OIDC provider, SQLite storage, refresh token rotation enabled). Our logs show the exact same refreshConnectorData/GetOfflineSessions call path in the goroutine trace above; /token wedged completely while /healthz stayed green, and a container restart was the only recovery. Confirms userIdent is nil for our connector type too, so this isn't limited to local password users as the issue title might suggest. Would be great to see this merged — happy to provide logs if useful.

so is the fix works for you?

fxgameplan:
Built and deployed your branch as a separate test instance (Dex fronting Authentik as an OIDC connector, same setup that hit the original deadlock). Manual refresh test came back fast (~90ms, no hang) so that looks promising. Letting it run naturally overnight to confirm it survives the real trigger condition (Dex's own internal refresh timer, ~24-25h in my case) rather than just a manual test. Will report back tomorrow.

fxgameplan:
Following up after letting this run for real, instead of just a quick manual test.  First connected the morning of the 17th, still connected right now (8am on the 19th, ~47 hours later). One refresh cycle happened at the 24hr mark (the 18th, ~13:39), came back clean and fast, no errors. Six signing key rotations have happened since with zero issues. This is well past the point where the old build would have deadlocked every time for us. Confident this fixes it on our end.

PR #4996, opened 2026-09-01
fix(keystone): separate public redirect host from internal API host

mridulgain:
## Summary
- `LoginURL()` was building the Shibboleth SSO redirect from `FederationConfig.Host`, the internal Keystone Service URL — broken when the public DU FQDN differs (KAAP-2257).
- Adds optional `PublicHost` (`publicKeystoneHost`), used only for the redirect, falling back to `Host` for backward compatibility. Server-to-server Keystone API calls still use `Host`.

## Test plan
- [x] `go test ./connector/keystone/...`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mridulgain:
Opened against the wrong repository by mistake (internal Platform9 fork change, not intended for upstream). Closing.

PR #5008, opened 2026-09-09
Add configurable ID token subject claim format and order

aduewel:
#### Overview

Add configurable formats and field ordering for Dex OAuth2/OIDC `sub` claims.

- Preserve the existing base64-encoded protobuf subject claim as the default.
- Add a URL-escaped plain subject format suitable for AWS IAM trust policies and wildcard matching.
- Add configurable `user-connector` and `connector-user` ordering.
- Preserve compatibility when parsing existing subjects across logout, introspection, auth flow, and refresh-token API paths.
- Document the new `oauth2.subjectClaim` and `oauth2.subjectClaimOrder` options and add coverage for configuration, encoding, parsing, and integration paths.

The plain format avoids protobuf length-prefix bytes in the claim, allowing AWS `StringLike` conditions to match meaningful subject components.

- Default behavior remains unchanged: base64 format with user ID before connector ID.
- Compatibility aliases for the previous terminology remain accepted.
- Validation: `go test ./...` passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,12 @@ ID Tokens contains standard claims assert which client app logged the user in, w
 }
 ```
 
+By default, Dex emits `sub` in a protobuf+base64url format. To use a
+wildcard-friendly string subject (for example in AWS IAM trust policies), set
+`oauth2.subjectClaim: plain` in Dex configuration.
+For plain subjects, you can also control order with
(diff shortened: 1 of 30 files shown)


aduewel:
Hi @nabokihms may I ask for a review here? Or is an issue mandatory for such an additional feature?

PR #5023, opened 2026-09-14
feat(openshift): Dynamically load clientSecret from a file without requiring a restart during secret rotation

anandf:
Assisted-By: Claude Code (Opus 4.6) for generating unit tests. Reviewed by Author.

#### Overview
When using Kubernetes `ServiceAccount` token as OAuth client to connect to an OpenShift cluster, the `ServiceAccount` token acts as the `clientSecret`.  While this works well for non-expiring legacy tokens, bound `ServiceAccount` tokens, in modern OpenShift clusters expire and are rotated periodically on disk, requiring Dex server restarts to pick up the updated token.

This PR adds support for specifying a file path where the ServiceAccount token is mounted (e.g., via projected volume mounts). Dex dynamically reads the latest token from disk during authentication, eliminating the need to restart Dex when tokens are rotated.

### What this PR does / why we need it
* Adds a new `clientSecretFile` field to the OpenShift connector configuration.
* Allows users to supply a token file path instead of a static `clientSecret`.
* When `clientSecretFile` is set, Dex reads the token from the file on demand. The token file is mounted as projected volume.
* **Backward Compatibility:** If `clientSecretFile` is empty or omitted, Dex defaults to the existing `clientSecret` behavior.

**Example Configuration:**
```yaml
connectors:
- type: openshift
  id: openshift
  name: OpenShift
  config:
    issuer: https://openshift.example.com
    clientID: system:serviceaccount:dex:dex-client
    clientSecretFile: /var/run/secrets/tokens/openshift-sa-token
```


anandf:
cc: @sabre1041 Please take a look, Thanks.

connector/openshift/openshift.go
```diff
@@ -24,13 +25,14 @@ const (
...
-	InsecureCA   bool     `json:"insecureCA"`
-	RootCA       string   `json:"rootCA"`
+	Issuer                  string   `json:"issuer"`
+	ClientID                string   `json:"clientID"`
+	ClientSecret            string   `json:"clientSecret"`
+	RedirectURI             string   `json:"redirectURI"`
+	Groups                  []string `json:"groups"`
+	InsecureCA              bool     `json:"insecureCA"`
+	RootCA                  string   `json:"rootCA"`
+	ServiceAccountTokenFile string   `json:"serviceAccountTokenFile"`
```
sabre1041:
```suggestion
	ClientSecretFile string   `json:"clientSecretFile"`
```

Given that the contents contains any token associated with a OpenShift client secret, I would recommend renaming this property to `clientSecretFile` as it is not limited to just containing a Kubernetes Service Account token. Other variables within this file should also be updated accordingly

sabre1041 (changes requested):
This is a really good enhancement @anandf. Added a comment related to the name of the property within he OpenShift connector. Tested and verified the functionality within an OpenShift cluster and the functionality remains the same -- with the added benefits provided in this enhancement

Keep in mind any change to the connector should require a modification to the [documentation/website repo](https://github.com/dexidp/website)

connector/openshift/openshift.go
anandf: I agree that the `clientSecretFile` is a better name to describe the function it provides and it need not necessarily be the SA token always. I have made the changes and also re-arranged the fields to keep `clientSecret` and `clientSecretFile` next to each other. PTAL.

anandf:
> Keep in mind any change to the connector should require a modification to the [documentation/website repo](https://github.com/dexidp/website)

Created this PR https://github.com/dexidp/website/pull/219 to update the documentation section of OpenShift.

sabre1041 (approved):
LGTM

Really solid integration

anandf:
@sagikazarmark Could you please help me to get this PR merged ? Thanks.

akhilnittala (approved):
LGTM

anandf:
@nabokihms Could you please help us with the pending checks and the merge of the PR. Thanks.

connector/openshift/openshift.go
```diff
@@ -261,6 +289,22 @@ func (c *openshiftConnector) user(ctx context.Context, client *http.Client) (u u
...
+func (c *openshiftConnector) currentOAuth2Config() (*oauth2.Config, error) {
+	if c.clientSecretFile == "" {
+		return c.oauth2Config, nil
+	}
+	clientSecretBytes, err := os.ReadFile(c.clientSecretFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read client secret file %q: %w", c.clientSecretFile, err)
+	}
+	cfg := *c.oauth2Config
+	cfg.ClientSecret = strings.TrimSpace(string(clientSecretBytes))
```
upramod: Small edge case here. If the secret file exists but is empty, or only has whitespace, we end up using an empty client secret and won't find out until the token request fails. Would it make sense to check the trimmed value here and return a clearer error? Just an extra check

Probably worth covering the empty file case in the rotation test too.

anandf: addressed your comments and added unit tests. PTAL. Thanks.

upramod: Checked the update. The empty and whitespace-only file cases are now rejected explicitly, including after rotation, and the tests cover both paths. Looks good to me.

anandf: Thanks @upramod. Could you also provide your approval via the `Submit Review` option ?

upramod (approved):

svghadi (approved):
LGTM.

